PR #15, opened 2016-01-20
Implement proper partition size detection.

kacf:
We need to go outside the comfort zone and do some low level OS
probing with syscall and unsafe pointers to achieve this.


kacf:
Something buggy with this PR. Opening a new one.

PR #30, opened 2016-02-15
MEN-56: Move logging framework to separate repositories.

kacf:
This way both server and client can use it.


kacf:
@maciejmrowiec, @pasinskim: Can one of you merge this, it's only decreasing coverage because I removed code that had higher coverage than the rest.

kacf:
Thanks!

PR #31, opened 2016-02-17
MEN-55: Add network update capability to Mender.

kacf:
It is used by specifying a URL to download the update from. Otherwise
it works exactly like the file updater.

### DIFF
--- a/rootfs.go
+++ b/rootfs.go
@@ -14,22 +14,94 @@
 package main
 
 import "fmt"
+import "net/http"
 import "io"
 import "os"
+import "strings"
+
+const minimumImageSize int64 = 4096
 
 func doRootfs(imageFile string) error {
+	var imageFd io.ReadCloser
+	var imageSize int64
+	var err error
+
+	if strings.HasPrefix(imageFile, "http:") ||
+		strings.HasPrefix(imageFile, "https:") {
+		// Network based update.
+		imageFd, imageSize, err = getHttpStream(imageFile)
+		if err != nil {
+			return err
+		}
+
+	} else {
+		// File based update.
+		imageFd, imageSize, err = getFileStream(imageFile)
+		if err != nil {
+			return err
+		}
+	}
+	defer imageFd.Close()
+
+	if err = writeToPartition(imageFd, imageSize); err != nil {
+		return err
+	}
+
+	if err = enableUpdatedPartition(); err != nil {
+		return fmt.Errorf("Unable to activate partition after update: "+
+			"%s", err.Error())
+	}
+
+	return nil
+}
+
+// Returns a byte stream which is a download of the given link, and also returns
+// the length of the file being downloaded.
+func getHttpStream(link string) (io.ReadCloser, int64, error) {
+	// Network based update.
+	resp, err := http.Get(link)
+	if err != nil {
+		return nil, 0, err
+	}
+
+	imageSize := resp.ContentLength
+	if imageSize < 0 {
+		return nil, 0, fmt.Errorf("Image size from '%s' is unknown. "+
+			"Will not continue with unknown image size.",
+			link)
+	} else if imageSize < minimumImageSize {
+		return nil, 0, fmt.Errorf("Less than 4KiB image update (%d "+
+			"bytes)? Something is wrong, aborting.",
+			imageSize)
+	}
+
+	return resp.Body, imageSize, nil
+}
+
+// Returns a byte stream of the fiven file, and also returns the size of the
+// file.
+func getFileStream(file string) (io.ReadCloser, int64, error) {
+	fd, err := os.Open(file)
+	if err != nil {
+		return nil, 0, fmt.Errorf("Not able to open image file: %s: %s\n",
+			file, err.Error())
+	}
+
+	imageInfo, err := fd.Stat()
+	if err != nil {
+		return nil, 0, fmt.Errorf("Unable to stat() file: %s: %s\n",
+			file, err.Error())
+	}
+
+	return fd, imageInfo.Size(), nil
+}
+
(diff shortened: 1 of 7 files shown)


rootfs.go
```diff
@@ -14,23 +14,76 @@
...
 
 import "fmt"
+import "net/http"
 import "io"
 import "os"
+import "strings"
 
 func doRootfs(imageFile string) error {
+	var imageFd io.Reader
+	var imageSize int64
```
pasinskim: Do we need to specify the size here? Shouldn't we use just int?

rootfs.go
```diff
@@ -14,23 +14,76 @@
...
+			return err
+		}
+		defer resp.Body.Close()
+
+		imageSize = resp.ContentLength
+		if imageSize < 0 {
+			return fmt.Errorf("Image size from '%s' is unknown. "+
+				"Will not continue with unknown image size.",
+				imageFile)
+		} else if imageSize < 4096 {
```
pasinskim: Maybe we should have some constant `min_image_size = 4096`?

rootfs.go
```diff
@@ -14,23 +14,76 @@
...
+		imageFd = resp.Body
+
+	} else {
+		// File based update.
+		fd, err := os.Open(imageFile)
+		if err != nil {
+			return fmt.Errorf("Not able to open image file: %s: %s\n",
+				imageFile, err.Error())
+		}
+		defer fd.Close()
```
pasinskim: I love this stuff in Go :+1:

rootfs_test.go
```diff
@@ -13,33 +13,66 @@
...
+
+const dummy string = "dummy_image.dat"
 
 func init() {
 	base_mount_device = "./dev/mmcblk0p"
 }
 
-func getModTime(t *testing.T, file string) time.Time {
-	info, err := os.Stat(file)
+// "Overlap" means that if one file is bigger, that part won't be checked.
```
pasinskim: Can we have better description of what this function is really doing?

rootfs_test.go
```diff
@@ -346,3 +382,89 @@ func TestPartitionsAPI(t *testing.T) {
 		mt.AssertTrue(t, err != nil)
 	}
 }
+
+func TestNetworkRootfs(t *testing.T) {
```
pasinskim: Would like to have some comment what is happening inside of this function.

rootfs_test.go
```diff
@@ -346,3 +382,89 @@ func TestPartitionsAPI(t *testing.T) {
...
+	prepareMockDevices(t)
+	defer cleanupMockDevices()
+
+	var server http.Server
+
+	server.Handler = http.FileServer(http.Dir("."))
+	addr := ":8081"
```
pasinskim: Some explanation why this is 8081?

rootfs.go
```diff
@@ -14,23 +14,76 @@
 package main
 
 import "fmt"
+import "net/http"
 import "io"
 import "os"
+import "strings"
 
 func doRootfs(imageFile string) error {
```
pasinskim: In general I would rather refactor this a bit and move file handling and network handling to separate functions. This will case the code to be more readable IMHO.

rootfs_test.go
```diff
@@ -346,3 +382,89 @@ func TestPartitionsAPI(t *testing.T) {
...
+			0,
+			0,
+			0,
+			0,
+			0,
+			0,
+			0}
+
+		runner = newRunner
+		err = doMain([]string{"-rootfs", "http://localhost:8081/" + dummy})
```
pasinskim: Should we have some const here `listener_port = 8081`?

rootfs_test.go
```diff
@@ -346,3 +382,89 @@ func TestPartitionsAPI(t *testing.T) {
...
+	addr := ":8081"
+	listen, err := net.Listen("tcp", addr)
+	mt.AssertNoError(t, err)
+
+	defer listen.Close()
+	go server.Serve(listen)
+
+	// Do this test twice, once with a valid update, and once with a too
+	// short/broken update.
+	for _, mode := range []int{0, os.O_TRUNC}[:] {
```
pasinskim: Maybe just for readability we can put this below in separate function like `ExecuteNetworkTest` or similar.

pasinskim:
@kacf looks really nice. In general I would like to see bit more comments and maybe some refactoring so that the functions are bit shorter. Other than those small details looks as ready to merge.

rootfs.go
kacf: 4K is quite arbitrary, it's just a safeguard against unreasonably small updates. This is just a safety net, not a guaranteed check.

But yes, I can use a constant.

kacf: Fixed!

rootfs_test.go
kacf: Done!

rootfs_test.go
kacf: Added a comment.

rootfs_test.go
kacf: I made it a constant and have added a comment.

rootfs.go
kacf: Done!

rootfs_test.go
kacf: Done!

kacf:
All issues fixed. Coverage has decreased 0.1%, but I propose merging anyway. It's just return statements for stream errors and stuff like that. It's very time consuming, and not very helpful, to write tests for those.

pasinskim:
Now I have the power to merge even if code coverage is lower than before :)
Agree that in this particular case it doesn't make that much sense to add artificial tests.

PR #32, opened 2016-02-24
JIRA-MEN-50: Client+Server TLS authenticated -bootstrap command. TODO finalise.

jimis:
DO NOT MERGE.

### DIFF
--- a/bootstrap.go
+++ b/bootstrap.go
@@ -1,0 +1,58 @@
+// Copyright 2016 Mender Software AS
+//
+//    Licensed under the Apache License, Version 2.0 (the "License");
+//    you may not use this file except in compliance with the License.
+//    You may obtain a copy of the License at
+//
+//        http://www.apache.org/licenses/LICENSE-2.0
+//
+//    Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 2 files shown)


bootstrap.go
```diff
@@ -0,0 +1,44 @@
...
+	fmt.Println("Sending HTTP GET to: ", serverURL)
+
+	response, err := httpClient.Get(serverURL)
+	if err != nil {
+		fmt.Println("HTTP GET failed:", err)
+		return nil // TODO
+	}
+
+	fmt.Println("Received headers:", response.Header)
+	respData, err := ioutil.ReadAll(response.Body)
```
kacf: I think you need to do `response.Body.Close()` afterwards.

kacf: A typical use case for `defer`. :-)

jimis: So no destructors for go. :-p Thanks, review much appreciated!

kacf: No unfortunately not, it's one of the things I still miss from C++. :-)

pasinskim:
This is already added to master after some refactoring. I will close this one.

PR #33, opened 2016-02-25
Add support for active partition detection through device ID.

kacf:
This is required for BeagleBone, where the `mount` output is useless
(it shows `/dev/root`, which isn't even a device). We still call the
`mount` command, but this is only a guide to go straight to the right
file. If the file is found to be incorrect, we iterate the `/dev`
directory until we find a device matching the device ID of the root
directory.

Also improve and restructure tests a bit. Give each test a separate
test function, and test more of the error cases for unexpected
partition numbers. And a test for the added functionality, of course.

MEN-30

### DIFF
--- a/partitions.go
+++ b/partitions.go
@@ -15,40 +15,115 @@
 
 package main
 
-import "strings"
 import "errors"
 import "fmt"
+import "io"
+import "os"
+import "path"
+import "strings"
+import "syscall"
 
 // Will change in testing.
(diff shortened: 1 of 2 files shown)


kacf:
@pasinskim: Here is the patch to make Beaglebone work.

Again the coverage fell, but I don't know how I could bring it up without serious effort. All the uncovered lines are either unlikely errors returned from the OS or `if` statements that could basically have been asserts (IOW, they never ever happen, but are there for safety).

PR #34, opened 2016-03-04
Add daemonization handling and bootstrapping.

pasinskim:


pasinskim:
@jimis this is not done yet, but you can see how this looks now. Fell free to review :)

bootstrap.go
```diff
@@ -0,0 +1,40 @@
...
+//TODO: this will be hardcoded for now but should be configurable in future
+const (
+	defaultCertFile   = "/data/certfile.crt"
+	defaultCertKey    = "/data/certkey.key"
+	defaultServerCert = "/data/server.crt"
+)
+
+func (c *Client) parseBootstrapResponse(response *http.Response) error {
+	// TODO: do something with the stuff received
+	if response.Status != "200 OK" {
```
maciejmrowiec:
```
reponse.StatusCode != http.StatusOK
```

bootstrap_test.go
```diff
@@ -0,0 +1,70 @@
...
+	authParams := authCmdLineArgsType{}
+	authParams.setDefaultKeysAndCerts("client.crt", "client.key", "server.crt")
+
+	_, authCreds := initClientAndServerAuthCreds(&authParams)
+
+	return Client{server, initClient(authCreds)}
+}
+
+func TestBootstrapSuccess(t *testing.T) {
+
```
maciejmrowiec: good candidate for table test

client.go
```diff
@@ -0,0 +1,151 @@
...
+//    See the License for the specific language governing permissions and
+//    limitations under the License.
+package main
+
+import "errors"
+import "github.com/mendersoftware/log"
+import "io/ioutil"
+import "net/http"
+import "crypto/tls"
+import "crypto/x509"
```
maciejmrowiec:
```
import (
    "crypto/tls"
    "crypto/x509"
   ....
)
```

client.go
```diff
@@ -0,0 +1,151 @@
...
+
+type authCmdLineArgsType struct {
+	// hostname or address to bootstrap to
+	bootstrapServer string
+	certFile        string
+	certKey         string
+	serverCert      string
+}
+
+func (cred *authCmdLineArgsType) setDefaultKeysAndCerts(clientCert string, clientKey string, serverCert string) {
```
maciejmrowiec:
```
setDefaultKeysAndCerts(clientCert string, clientKey string, serverCert string)
```

can be written as:
```
setDefaultKeysAndCerts(clientCert, clientKey, serverCert string)
```

client.go
```diff
@@ -0,0 +1,151 @@
...
+type authCredsType struct {
+	// Cert+privkey that authenticates this client
+	clientCert tls.Certificate
+	// Trusted server certificates
+	trustedCerts x509.CertPool
+}
+
+func (auth *authCredsType) initServerTrust(cred *authCmdLineArgsType) error {
+
+	trustedCerts := *x509.NewCertPool()
```
maciejmrowiec: why *?

```
trustedCerts := x509.NewCertPool()
CertPoolAppendCertsFromFile(trustedCerts, cred.serverCert)
....
auth.trustedCerts = *trustedCerts
```

client.go
```diff
@@ -0,0 +1,151 @@
...
+		return nil
+	}
+}
+
+func initClientAndServerAuthCreds(cred *authCmdLineArgsType) (error, authCredsType) {
+
+	var auth authCredsType
+
+	if err := auth.initServerTrust(cred); err != nil {
+		return err, (authCredsType{})
```
maciejmrowiec:
```
return err, (authCredsType{})
return err, authCredsType{}
```

client.go
```diff
@@ -0,0 +1,151 @@
...
+	}
+
+	return &http.Client{
+		Transport: &transport,
+	}
+}
+
+type httpReqType int
+
+const (
```
maciejmrowiec: have a look at
```
http.MethodGet
http. MethodPut
....
```

client.go
```diff
@@ -0,0 +1,151 @@
...
+	GET httpReqType = 1 + iota
+	PUT
+	POST
+)
+
+func (c *Client) sendRequest(reqType httpReqType, request string) (error, *http.Response) {
+
+	switch reqType {
+	case GET:
```
maciejmrowiec:
```
http.MethodGet
```

daemon.go
```diff
@@ -0,0 +1,71 @@
...
+	//TODO: check if starts with https://
+	config.server = server
+}
+
+func (config *daemonConfigType) setDeviceID() {
+	//TODO: get it from somewhere
+	config.deviceID = defaultDeviceID
+}
+
+func runAsDemon(config daemonConfigType, client *client) error {
```
jimis: typo: `s/Demon/Daemon/`

bootstrap.go
pasinskim: Actually response.Status is type string so this will not work.

client.go
pasinskim: Fixed!

bootstrap.go
maciejmrowiec: reponse.StatusCode is int

Status is more for print friendly

client.go
pasinskim: Fixed!

client.go
pasinskim: Fixed!

client.go
pasinskim: Fixed. Also returning authCredsType{}, err as it is advised that err is the last thing returned.

client.go
pasinskim: Fixed. Tricky part is that it is working from version 1.6.

bootstrap.go
```diff
@@ -0,0 +1,57 @@
...
+	return br.menderClient
+}
+
+func (br bootstrapRequester) formatRequest() clientRequestType {
+	return clientRequestType{br.reqType, br.request}
+}
+
+func (br bootstrapRequester) actOnResponse(response http.Response, respBody []byte) error {
+	// TODO: do something with the stuff received
+	if response.Status != "200 OK" {
```
maciejmrowiec: I would still use:
```
response.StatusCode != http. StatusOK
```

bootstrap_test.go
```diff
@@ -0,0 +1,78 @@
...
+	if err := doMain([]string{"-bootstrap", "127.0.0.1",
+		"-trusted-certs", "non-existing.crt"}); err == nil || err != errorAddingServerCertificateToPool {
+		t.Fatal("Can not override default certificate command line swhitch")
+	}
+}
+
+func TestBootstrapSuccess(t *testing.T) {
+	// Test server that always responds with 200 code, and specific payload
+	ts := httptest.NewTLSServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		w.WriteHeader(200)
```
maciejmrowiec: w.WriteHeader(http.StatusOK)

client.go
```diff
@@ -0,0 +1,188 @@
...
+	"crypto/x509"
+	"errors"
+	"io/ioutil"
+	"net/http"
+
+	"github.com/mendersoftware/log"
+)
+
+var (
+	errorLoadingClientCertificate = errors.New("Failed to load certificate and key")
```
maciejmrowiec: Naming covention is:

ErrSomeIssue

client.go
```diff
@@ -0,0 +1,188 @@
...
+
+type client struct {
+	BaseURL string
+	authCredsType
+	HTTPClient *http.Client
+}
+
+// Client initialization
+
+func initClient(args authCmdLineArgsType) (client, error) {
```
maciejmrowiec: Why by value? Extra copy.
Return pointer.

Constructor naming convention is:

NewMyType

client.go
```diff
@@ -0,0 +1,188 @@
...
+		Transport: &transport,
+	}
+
+	return httpsClient, nil
+}
+
+func (c *client) initServerTrust(args authCmdLineArgsType) error {
+
+	if args.serverCert == "" {
+		panic("certificate should be replaced with default one")
```
maciejmrowiec: maybe better to have some critical error instead of panic

client.go
```diff
@@ -0,0 +1,188 @@
...
+}
+
+// Client request sending and parsing
+
+type clientRequestType struct {
+	reqType string
+	request string
+}
+
+type clientWorker interface {
```
maciejmrowiec: to me this worker name suggests that it's supposed to be spawned as a process

client.go
```diff
@@ -0,0 +1,188 @@
...
+
+	response, data, err := client.sendRequest(request.reqType, request.request)
+	if err != nil {
+		return err
+	}
+
+	return req.actOnResponse(*response, data)
+}
+
+func (c *client) sendRequest(reqType string, request string) (*http.Response, []byte, error) {
```
maciejmrowiec: response has already body inside, no reason to return both ...

client.go
```diff
@@ -0,0 +1,188 @@
...
+
+		respData, err := ioutil.ReadAll(response.Body)
+		if err != nil {
+			return nil, nil, err
+		}
+		log.Debug("Received response body: ", string(respData))
+
+		return response, respData, nil
+	}
+	panic("unknown http request")
```
maciejmrowiec: prefer log.Fatal

daemon.go
```diff
@@ -0,0 +1,195 @@
...
+		log.Debug("No update available")
+
+		//TODO: check body to see if message is mallformed
+		return updateNoUpdateResponseType{}, nil
+
+	case updateResponseError:
+		//TODO: check body to see if message is mallformed
+		return updateErrorResponseType{}, nil
+
+	default:
```
maciejmrowiec: This is not entirely true, there can be authentication errors, timeouts, special server error codes. Anything from 3xx, 4xx, 5xx

daemon.go
```diff
@@ -0,0 +1,195 @@
...
+	"encoding/json"
+	"errors"
+	"io/ioutil"
+	"net/http"
+	"time"
+)
+import "github.com/mendersoftware/log"
+
+// create a channel so that we will be able to stop daemon
+var daemonQuit = make(chan bool)
```
maciejmrowiec: why global?

daemon.go
```diff
@@ -0,0 +1,195 @@
...
+		request:              config.server + "/" + config.deviceID + "/update",
+		menderClient:         client,
+		updateResponseParser: respParse,
+	}
+
+	for {
+		select {
+		case <-ticker.C:
+
+			log.Debug("Timer expired. Pulling server to check update.")
```
jimis: typo: `s/Pulling/Polling/`

daemon_test.go
```diff
@@ -0,0 +1,211 @@
...
+
+	if err := makeJobDone(updateRequester); err != nil {
+		t.Fatal(err)
+	}
+}
+
+func TestCheckPeriodicDaemonUpdate(t *testing.T) {
+
+	reqHandlingCnt := 0
+	pullInterval := time.Duration(100) * time.Millisecond
```
jimis: typo: `s/pull/poll`

daemon_test.go
```diff
@@ -0,0 +1,211 @@
...
+		w.WriteHeader(200)
+		w.Header().Set("Content-Type", "application/json")
+		fmt.Fprintln(w, correctUpdateResponse)
+		reqHandlingCnt++
+	}))
+	defer ts.Close()
+
+	client, _ := initClient(authCmdLineArgsType{ts.URL, "client.crt", "client.key", "server.crt"})
+	var config daemonConfigType
+	config.serverPullInterval = pullInterval
```
jimis: poll

daemon_test.go
```diff
@@ -0,0 +1,211 @@
...
+	config.server = ts.URL
+	config.deviceID = "fake_id"
+
+	go func() {
+		runAsDaemon(config, client, fakeParseUpdateResponse)
+	}()
+
+	timesPulled := 5
```
jimis: poll

daemon_test.go
```diff
@@ -0,0 +1,211 @@
...
+	}))
+	defer ts.Close()
+
+	client, _ := initClient(authCmdLineArgsType{ts.URL, "client.crt", "client.key", "server.crt"})
+	var config daemonConfigType
+	config.serverPullInterval = pullInterval
+	config.server = ts.URL
+	config.deviceID = "fake_id"
+
+	go func() {
```
maciejmrowiec: go runAsDaemon(config, client, fakeParseUpdateResponse)

main.go
```diff
@@ -121,14 +198,40 @@ func argsParse(args []string) (runOptionsType, error) {
...
+func startDaemon(args authCmdLineArgsType) error {
+	client, err := initClient(args)
+	if err != nil {
+		return err
 	}
 
-	runOptions.imageFile = *imageFile
-	runOptions.committing = *committing
+	//TODO: this is temporary only and should be replaced in future
+	server := getServerAddress()
```
maciejmrowiec: getServerAddress should set the default one

daemon.go
```diff
@@ -0,0 +1,195 @@
...
+
+	}
+	// ureachable
+}
+
+func runAsDaemon(config daemonConfigType, client client, respParse responseParserFunc) error {
+	// create channels for timer and stopping daemon
+	ticker := time.NewTicker(config.serverPullInterval)
+
+	updateRequester := updateRequester{
```
maciejmrowiec: this object could be param to this function, there are 3 params passed just to fill out this sturct, could be done outside

bootstrap.go
pasinskim: Makes sense. Fixing!

bootstrap_test.go
pasinskim: Fixing!

client.go
pasinskim: You mean it should be capital letter?

client.go
pasinskim: Fixing!

client.go
maciejmrowiec: I would use capital letter, in main import/export for method does not matter (you can't import main package to another project). otherwise you almost always want to export predefined errors in stand alone package so package user can compare errors

client.go
maciejmrowiec: -- golden rule: never return panic from stand alone package, if used you have to catch it inside of package

client.go
pasinskim: I will change to clientRequester (looks like 'er' ending is also naming convention for interfaces).

client.go
pasinskim: Maybe I am doing something wrong here, but Body is a type `Body io.ReadCloser` and you are supposed to close it once reading is done. It looks like when I am trying to do `defer Body.Close()` the Body content is not accessible outside `sendRequest()` function. That is why I am reading the body and storing it's content in separate place.

daemon.go
pasinskim: I think we can add it later. For now what we are interested in is if there is valid response we can handle or not. We are not having any failure recovery mechanism if we will receive anything else that we are expecting...

pasinskim: But it's a good thing that you've pointed this out.

client.go
maciejmrowiec: yes, but if you extract data from request, extract everything you need and let it go, or pass entire thing, now it's one foot here one foot there

daemon.go
pasinskim: I am using this one in tests so that I can kill the daemon. If there is a better way feel free to point me this out.

daemon_test.go
pasinskim: Nice. Didn't know about this syntax.

maciejmrowiec: you'd use 

```
z := "Time is"
go func() {
   fmt.Println(z, time.Now())
}()
```

if you want to put inline code that will be the routine or want to use closure

daemon.go
maciejmrowiec: sure, you can handle this as range 400-499 client errors 500-599 server errors

main.go
pasinskim: Good point!

daemon.go
pasinskim: Right. Good point.

PR #37, opened 2016-03-07
Update travis go to version 1.6

pasinskim:
Some functionalities we are using are introduced in go 1.6. Let's update Travis to use it.


pasinskim:
@maciejmrowiec we need this to have go 1.6 being used by Travis.

PR #39, opened 2016-03-31
Automated daemon additions

pasinskim:


pasinskim:
@kacf @maciejmrowiec can you take a look. Still some stuff I would like to test more and some daemon refactoring will be needed, but otherwise all functionality should be there.

client.go
```diff
@@ -21,74 +21,168 @@ import (
...
-	defaultServerCert       = "/data/server.crt"
-	minimumImageSize  int64 = 4096
+	minimumImageSize int64 = 4096 //kB
 )
 
 type RequestProcessingFunc func(response *http.Response) (interface{}, error)
 
 type Updater interface {
-	GetScheduledUpdate(RequestProcessingFunc, string) (interface{}, error)
+	GetScheduledUpdate(RequestProcessingFunc, string, string) (interface{}, error)
```
maciejmrowiec: could be nice for the interface to keep param names (if you would like to supply different implementation of the interface you don't have any hint about what the input is.)

client.go
```diff
@@ -21,74 +21,168 @@ import (
...
-	httpsClient.minImageSize = minimumImageSize
-	args.setDefaultKeysAndCerts(defaultCertFile, defaultCertKey, defaultServerCert)
+func NewUpdater(conf httpsClientConfig) (Updater, error) {
 
-	if err := httpsClient.initServerTrust(args); err != nil {
-		return nil
+	if conf == (httpsClientConfig{}) {
+		client := NewHttpClient()
+		if client == nil {
+			return nil, errors.New("Can not instantialte updater.")
```
maciejmrowiec: can be predefined in var at the top of the file

client.go
```diff
@@ -21,74 +21,168 @@ import (
...
+		}
+		return client, nil
+	}
+	client := NewHttpsClient(conf)
+	if client == nil {
+		return nil, errors.New("Can not instantialte updater.")
```
maciejmrowiec: can be predefined in var at the top of the file

client.go
```diff
@@ -21,74 +21,168 @@ import (
...
 
-func NewClient(args authCmdLineArgsType) *client {
-	var httpsClient client
-	httpsClient.minImageSize = minimumImageSize
-	args.setDefaultKeysAndCerts(defaultCertFile, defaultCertKey, defaultServerCert)
+func NewUpdater(conf httpsClientConfig) (Updater, error) {
 
-	if err := httpsClient.initServerTrust(args); err != nil {
-		return nil
+	if conf == (httpsClientConfig{}) {
```
maciejmrowiec: this compare is a bit risky, if it has any references then it will compare addresses (not values), even if it does not have it now, if added later it could potentially start misbehaving 

alternatife is reflext.DeepEqual but this is known to be expensive call so have to be used wisely

daemon.go
```diff
@@ -111,42 +115,45 @@ func (daemon *menderDaemon) Run() error {
...
+	log.Debug("Installing update to inactive partition.")
+	err = daemon.InstallUpdate(image, imageSize)
+	if err != nil {
+		log.Error("Can not install update: ", err)
+		return false
 	}
-	// we have no update
-	return false, nil
+
+	log.Info("Update instelled to inactive partition")
```
kacf: `s/instelled/installed/`

client.go
maciejmrowiec: Example: http://play.golang.org/p/mFtz7eRey_

client.go
```diff
@@ -137,6 +239,7 @@ type UpdateResponse struct {
 	Image struct {
 		URI      string
 		Checksum string
+		Yocto_ID string
```
maciejmrowiec: coding style: YoctoId

client.go
```diff
@@ -144,8 +247,9 @@ type UpdateResponse struct {
 
 func validateGetUpdate(update UpdateResponse) error {
 	// check if we have JSON data correctky decoded
-	if update.ID != "" && update.Image.ID != "" && update.Image.Checksum != "" && update.Image.URI != "" {
-		log.Info("Received correct request for getting image from: " + update.Image.URI)
+	if update.ID != "" && update.Image.ID != "" && update.Image.Checksum != "" &&
```
maciejmrowiec: checksum is optional attribute, may or may not be in payload

client.go
```diff
@@ -187,67 +291,3 @@ func processUpdateResponse(response *http.Response) (interface{}, error) {
 		return nil, errors.New("Invalid response received from server")
```
maciejmrowiec: Error can be predefined as global in var

maciejmrowiec: Error can be predefined as global in var

client.go
```diff
@@ -163,15 +267,15 @@ func processUpdateResponse(response *http.Response) (interface{}, error) {
 	case updateResponseHaveUpdate:
 		log.Debug("Have update available")
 
-		data := new(UpdateResponse)
-		if err := json.Unmarshal(respBody, data); err != nil {
+		data := UpdateResponse{}
+		if err := json.Unmarshal(respBody, &data); err != nil {
 			switch err.(type) {
 			case *json.SyntaxError:
 				return nil, errors.New("Error parsing data syntax")
```
maciejmrowiec: Error can be predefined as global in var

daemon.go
```diff
@@ -22,58 +21,41 @@ import (
...
 
-func NewDaemon(client Updater, device UInstallCommitRebooter) *menderDaemon {
-	daemon := menderDaemon{client, device, daemonConfigType{}, make(chan bool)}
+func NewDaemon(client Updater, device UInstallCommitRebooter,
+	mender Controler) *menderDaemon {
+
+	config := mender.GetDaemonConfig()
+
+	daemon := menderDaemon{
+		client,
```
maciejmrowiec: This is very much neat picking:
When initialising structure like this it's nice to use named params it's more readable and more resilient to changes to the original structure fields.

mender.go
```diff
@@ -0,0 +1,182 @@
...
+
+func (m mender) GetDaemonConfig() daemonConfig {
+	return daemonConfig{
+		time.Duration(m.config.PollIntervalSeconds) * time.Second,
+		m.config.ServerURL,
+		m.config.DeviceID,
+	}
+}
+
+func readCongigFile(config interface{}, fileName string) error {
```
kacf: `s/readCongigFile/readConfigFile/`

daemon.go
```diff
@@ -111,42 +115,45 @@ func (daemon *menderDaemon) Run() error {
...
+	}
 
-		log.Info("Update instelled to inactive partition")
-		if err := uinst.EnableUpdatedPartition(); err != nil {
-			return false, err
-		}
-		log.Debug("Inactive partition marked as first boot candidate.")
-		return true, nil
+	log.Debug("Installing update to inactive partition.")
+	err = daemon.InstallUpdate(image, imageSize)
```
maciejmrowiec: could be:

mender.go
```diff
@@ -0,0 +1,182 @@
...
+
+type Controler interface {
+	GetState() MenderState
+	GetCurrentImageID() string
+	GetDaemonConfig() daemonConfig
+	GetUpdaterConfig() httpsClientConfig
+}
+
+const (
+	defaultManifestFile = "/etc/build"
```
kacf: Perhaps this should be `/etc/yocto-build`, I think `/etc/build` is too generic.

client.go
pasinskim: Makes sense. Also I can see that go library also leaves names of parameters here.

client.go
pasinskim: This is something we need to change all over the places. I will leave this as is for now.

client.go
pasinskim: Yes. The thing I am interested in here is if there are any config parameters. If not I assume we are using http not https.

daemon.go
pasinskim: Fixing.

client.go
pasinskim: @maciejmrowiec will this work if server is sending yocto_id not yoctoid?

client.go
```diff
@@ -21,74 +21,168 @@ import (
...
+		return errorLoadingClientCertificate
+	}
+	c.clientCert = clientCert
+	return nil
+}
+
+func (c *httpClient) GetScheduledUpdate(process RequestProcessingFunc,
+	server string, deviceID string) (interface{}, error) {
+	// http client should be able to perform https requests
+	if strings.HasPrefix(server, "http") {
```
kacf: If your intention here is to check that any http protocol (http or https) is used, I think you should use two ORed checks with `http://` and `https://`. A bare URI can start with 'http' without having specified "http://" for example.

client.go
pasinskim: Nice to know :+1:

daemon.go
pasinskim: Agree. Will fix!

client.go
maciejmrowiec: sorry for that, didn't merge the update to the spec with this change

mender.go
pasinskim: Fixing.

daemon.go
pasinskim: Yup!

mender.go
pasinskim: Actually it is ```build_mender``` now.

client.go
pasinskim: That's a good point. Actually this is still bit tricky as we are not having the server and we are using 3rd party http or https services. I am sure this will be changed in future, but anyway I will change to check for either ```http://``` or ```https://```.

PR #44, opened 2016-04-06
Run ./check_signed_off.sh in before_script

GregorioDiStefano:
Signed-off-by: Gregorio Di Stefano <Gregorio.Di.Stefano@mender.io>

### DIFF
--- a/check_signed_off.sh
+++ b/check_signed_off.sh
@@ -1,0 +1,17 @@
+#!/bin/bash
+set -x
+
+echo "Checking range: ${TRAVIS_COMMIT_RANGE}"
+
+for i in $( (git rev-list --no-merges "$TRAVIS_COMMIT_RANGE") )
+do
+    COMMIT_MSG="$(git show -s --format=%B "$i")"
+    COMMIT_USER_EMAIL="$(git show -s --format="%an <%ae>" "$i")"
+    echo "$COMMIT_MSG" | grep -F "Signed-off-by: ${COMMIT_USER_EMAIL}" >/dev/null
+
+    if [ $? -ne 0 ]; then
+        echo >&2 "Commit ${i} is not signed off! Use --signoff with your commit."
+        exit 1
+    fi
+
+done
--- a/.travis.yml
+++ b/.travis.yml
@@ -74,6 +74,9 @@ before_script:
     # Test if the license strings are present in every .go file, with exceptions.
     - ./check_license.sh
 
+    # Test if all new commits are signed-off
+    - ./check_signed_off.sh
+
     # Perform static code analysys
     - go tool vet -unsafeptr=false `go list ./... | grep -v Godeps | sed -e "s,^,$GOPATH/src/,;s,.*:,,"`
 


GregorioDiStefano:
@kacf Please take a look

check_signed_off.sh
```diff
@@ -0,0 +1,21 @@
+#!/bin/bash
+
+echo "Checking range: ${TRAVIS_COMMIT_RANGE}"
+
+NEW_COMMITS=( $(git log  --format=format:%H "$TRAVIS_COMMIT_RANGE") )
```
kacf: Can be simplified by using `git rev-list`. Needs no arguments other than the range. :-)

check_signed_off.sh
```diff
@@ -0,0 +1,21 @@
...
+
+for i in "${NEW_COMMITS[@]}"
+do
+    PARENTS=$(git cat-file -p "$i" | grep -c "parent")
+    if [ "$PARENTS" -gt 1 ];
+     then
+        #if the commit is a merge, dont check the commit msg.
+        continue
+     else
```
kacf: This can also be simplified, just call `git rev-list --no-merges` above, and you will not get any merge commits in the output.

check_signed_off.sh
```diff
@@ -0,0 +1,21 @@
...
+        COMMIT_MSG=$(git show -s --format=%B "$i")
```
kacf: `git show -s` is nice! Didn't know about it, it's quite well hidden and I've been looking for that!

check_signed_off.sh
```diff
@@ -0,0 +1,21 @@
...
+        echo "$COMMIT_MSG" | grep '^Signed-off-by: ' >/dev/null || {
```
kacf: We need to be strict here and check that the *author* is signing it off, so the value here should be compared against `git show -s --format="%an <%ae>"` of the commit.

GregorioDiStefano:
I thought about this post-commit yesterday, should I eliminate the 'continue' statement as well?

kacf:
Nice stuff though!

kacf:
> I thought about this post-commit yesterday, should I eliminate the 'continue' statement as well?

The question of the `continue` statement becomes moot if you use `--no-merges`.

GregorioDiStefano:
@kacf

check_signed_off.sh
```diff
@@ -0,0 +1,17 @@
+#!/bin/bash
+set -x
+
+echo "Checking range: ${TRAVIS_COMMIT_RANGE}"
+
+for i in $( (git rev-list --no-merges "$TRAVIS_COMMIT_RANGE") )
+do
+    COMMIT_MSG="$(git show -s --format=%B "$i")"
+    COMMIT_USER_EMAIL="$(git show -s --format="%an <%ae>" "$i")"
+    echo "$COMMIT_MSG" | grep "^Signed-off-by: ${COMMIT_USER_EMAIL}" >/dev/null
```
kacf: One last nitpick (I promise it's the last one!): Emails can contain some characters that may have special meaning in regexes, `.` in particular. I would use fgrep instead of grep to grep for the exact string. Then you have to lose the `^` in the beginning, since it will not be a regex anymore, but I think this is acceptable. What do you think?

GregorioDiStefano: Haha, np. 
I changed the line by replace grep to grep -F, and and dumped the '^'

kacf:
Great, thanks!

PR #47, opened 2016-05-06
adding -version support

bboozzoo:
@kacf @mchalski @kjaskiewiczz @eysteins 

We still need to figure out how to populate version in non-travis builds. I'm willing to add a makefile with `go build -v -ldflags "-X main.Tag=$(git describe --always --tags --dirty)"` inside. For instance, with a bogus tag v1.0.0 at fddc44d that I've added locally, the log looks like this:

```
* 706c92b (HEAD -> bboozzoo/mender-version, github/bboozzoo/mender-version) main: add -version command line option
* 564fd93 version: import version helper from deviceadm
*   2425c19 (origin/master, origin/HEAD, github/master, master) Merge pull request #44 from GregorioDiStefano/testing_signoff_1
|\  
| * 7a8958b Run ./check_signed_off.sh in before_script
|/  
*   fddc44d (tag: v1.0.0) Merge pull request #42 from kacf/more_debug_output
|\  
| *   af0fa9b Merge branch 'more_debug_output'
```
Running `mender -version` will produce this output:

```
$ ./mender -version
v1.0.0-4-g706c92b
```

Dissecting, `v1.0.0` - tag name, `4` - number of patches from tag, `g706c92b` - HEAD is at rev 706c92b.

Plus, it should automatically work with Yocto.

It also looks like a good time to start tagging so that `git describes` produces some meaningful output.

@eysteins does the output above look ok to you?


version_test.go
```diff
@@ -0,0 +1,60 @@
...
+//    Unless required by applicable law or agreed to in writing, software
+//    distributed under the License is distributed on an "AS IS" BASIS,
+//    WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+//    See the License for the specific language governing permissions and
+//    limitations under the License.
+package main
+
+import (
+	"github.com/stretchr/testify/assert"
+	"testing"
```
kacf: This is nice. There is already `mendertesting.AssertTrue`, but it might be nicer to use something ready-made and more complete, like this.

kacf: Regarding `testify/assert`, that is.

main.go
```diff
@@ -224,12 +244,22 @@ func parseLogFlags(args logOptionsType) error {
 	return nil
 }
 
+func ShowVersion() {
+	v := fmt.Sprintf("%s\n", CreateVersionString())
+	os.Stderr.Write([]byte(v))
```
kacf: Should this perhaps be `Stdout` instead? It might be easier to use in scripts.

bboozzoo: On Fri, May 6, 2016 at 11:45 AM, Kristian Amlie <notifications@github.com>
wrote:

> In main.go
> <https://github.com/mendersoftware/mender/pull/47#discussion_r62308778>:
>
> > @@ -224,12 +244,22 @@ func parseLogFlags(args logOptionsType) error {
> >  	return nil
> >  }
> >
> > +func ShowVersion() {
> > +	v := fmt.Sprintf("%s\n", CreateVersionString())
> > +	os.Stderr.Write([]byte(v))
>
> Should this perhaps be Stdout instead? It might be easier to use in
> scripts.
>

stdout works too, I'll push another patch


-- 
Maciej Borzecki

kacf:
Nice! I think adding a makefile makes a lot of sense. People can still build using `go build`, but it's nice when you need a "proper" binary.

version_test.go
```diff
@@ -0,0 +1,60 @@
...
+//
+//    Unless required by applicable law or agreed to in writing, software
+//    distributed under the License is distributed on an "AS IS" BASIS,
+//    WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+//    See the License for the specific language governing permissions and
+//    limitations under the License.
+package main
+
+import (
+	"github.com/stretchr/testify/assert"
```
kacf: Oh, btw, new dependencies should be vendored, aka `godep save`.

(Using the new Go 1.6 vendoring would be even better, but this predates 1.6)

PR #49, opened 2016-05-10
removed dead code and unsed variables, updated .travis.yml

GregorioDiStefano:

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -58,6 +58,12 @@ before_install:
     # Merge all subpackage code coverage reports into a single report
     - go get -u github.com/modocache/gover
 
+    # Install varcheck
+    - go get -u github.com/opennota/check/cmd/varcheck
+
+    # Install deadcode
+    - go get -u github.com/mendersoftware/deadcode
+
 before_script:
     # Print build info that binary is compiled with.
     - echo $TRAVIS_COMMIT
@@ -79,6 +85,9 @@ before_script:
 
     # Perform static code analysys
     - go tool vet -unsafeptr=false `find . -name '*.go' |grep -v Godeps | grep -v _test.go`
+
+    - deadcode -ignore version.go:BuildNumber
+    - varcheck .
 
     # Fail builds when the cyclomatic complexity reaches 15 or more
     - gocyclo -over 15 `find . -iname '*.go' | grep -v 'Godeps' | grep -v '_test.go'`
--- a/client.go
+++ b/client.go
@@ -28,7 +28,6 @@ import (
 
 var (
 	errorLoadingClientCertificate      = errors.New("Failed to load certificate and key")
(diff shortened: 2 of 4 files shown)


kacf:
It found more dead code. :-)

bboozzoo:
@GregorioDiStefano can you split the commit in two? .travis.yml change is unrelated to removal of dead code.

GregorioDiStefano:
the .travis.yml update adds the actual checking for dead code

bboozzoo:
On Wed, May 11, 2016 at 10:51 AM, GregorioDiStefano <
notifications@github.com> wrote:

> the .travis.yml update adds the actual checking for dead code
>

Which is fine as it's not removing dead code. Not sure what's your
preferred approach to commits, I usually try to have 1 logical change in a
commit (though not always possible/practical). In this case I'd have 2
commits. First one removing dead code, followed by the second, enabling
travis checks.

If that's not the practice you follow, then it's also fine by me too :)

kacf:
I second @bboozzoo, the dead code removal should in the first commit, and then travis update in second commit (in that order, to avoid breaking the build in between).

GregorioDiStefano:
Thanks guys, that indeed makes sense. Will fix.

Meanwhile, I made a PR to deadcode: since we are using ldlfags to set the buildnumber, deadcode is detecting that variable as unused. I added the ability to ignore a filename:identifier by passing the -ignore flag. 

See here if you have any comments: https://github.com/tsenart/deadcode/pull/4

kacf:
> See here if you have any comments: tsenart/deadcode#4

Nice!

GregorioDiStefano:
can someone take a look at the changes in: https://github.com/GregorioDiStefano/deadcode
Considering that the author isn't interacting with any of the other pull requests, he probably wont look at this one either.

kacf:
It's a shame he doesn't respond.

bboozzoo:
Hmm @GregorioDiStefano are you sure the pull request is openend to the right repository? It seems that the original code comes from here: https://github.com/remyoudompheng/go-misc

GregorioDiStefano:
@bboozzoo Right, but it seems like he's been ignoring pull requests as well.

kacf:
Should you then point Travis to your repository instead, and add the exception for `BuildNumber`?

kacf:
You need to add the exception for `BuildNumber`, otherwise the test fails.

bboozzoo:
On Wed, May 18, 2016 at 1:25 PM, Kristian Amlie <notifications@github.com>
wrote:

> You need to add the exception for BuildNumber, otherwise the test fails.
>

How about we fork deadcode to mendersoftware, push necessary fixed to
master and grab if from there during the build?

-- 
Maciej Borzecki

GregorioDiStefano:
@bboozzoo done

PR #50, opened 2016-05-18
support/mender-device-identity: example platform integration script

bboozzoo:
The patch adds an example ntegration script for producing device
identity data. The script looks up a network interface with the lowest
ifindex, other than loopback interface, and grabs its MAC address. It is
assumed that /sys/class/net is available in the system.

The identity data is output to stdout and has a form of:
<key>=<value>

The script may produce some information to stderr, make sure that it's
accounted for by whoever is reading the output.

### DIFF
--- a/support/mender-device-identity
+++ b/support/mender-device-identity
@@ -1,0 +1,50 @@
+#!/bin/sh
+
+# Example script called by Mender agent to collect device identity
+# data. The script needs to be located at
+# $(bindir)/mender-device-identity path for the agent to find it.
+# The script shall exit with non-0 status on errors. In this case the
+# agent will discard any output the script may have produced.
+#
+# The script shall output identity data in <key>=<value> format, one
+# entry per line. Example
+#
+# $ ./mender-device-identity
+# mac=de:ad:ca:fe:00:01
+# cpuid=1112233
+#
+# The example script collects the MAC address of a network interface
+# with the lowest ifindex, other than the loopback device 'lo'. The
+# identity data is output in the following format:
+#
+# mac=00:01:02:03:04:05
+#
+
+set -ueo pipefail
+
+SCN=/sys/class/net
+min=65535
+ifdev=
+
+# find iface with lowest ifindex, except loopback
+for dev in $SCN/*; do
+    if [ $dev = "$SCN/lo" ]; then
+        continue
+    fi
+
+    idx=$(cat $dev/ifindex)
+    if [ $idx -lt $min ]; then
+        min=$idx
+        ifdev=$dev
+    fi
+done
+
+if [ -z "$ifdev" ]; then
+    echo "no suitable interfaces found" >&2
+    exit 1
+else
+    echo "using interface $ifdev" >&2
+    # grab MAC address
+    echo "mac=$(cat $ifdev/address)"
+
+fi


bboozzoo:
@kacf @pasinskim example script for https://tracker.mender.io/browse/MEN-381

kacf:
Looks good.

Should we perhaps name it something ala `device-identity-interface-mac`, in order to make it clear that this is just one of many possibilities? And should it go in directory `examples`, instead of support?

Just ideas, I'm fine as it is too.

bboozzoo:
On Wed, May 18, 2016 at 12:08 PM, Kristian Amlie <notifications@github.com>
wrote:

> Looks good.
>
> Should we perhaps name it something ala device-identity-interface-mac, in
> order to make it clear that this is just one of many possibilities? And
> should it go in directory examples, instead of support?
>
The name might be a bit too long :) I can put a comment at the top of the
script or a README in support directory.


As for examples, I'm not convinced. I'd rather go with support as it is
now. Later on we'll need to add a makefile to handle build and install
targets. There we may have something like:

```
install:
    ...
    install -t $(DESTDIR)$(bindir) support/mender-device-indentity
    ...
```

to give the right example of integration. What do you think?

bboozzoo:
@kacf added a description at the top of the script

support/mender-device-identity
```diff
@@ -0,0 +1,50 @@
+#!/bin/sh
+
+# Example script called by Mender agent to collect device identity
+# data. The script needs to be located at
+# $(bindir)/mender-device-indentity path for the agent to find it.
```
kacf: `s/indentity/identity/`

support/mender-device-identity
```diff
@@ -0,0 +1,50 @@
...
+# The script shall exit with non-0 status on errors. In this case the
+# agent will discard any output the script may have produced.
+#
+# The script shall output identity data in <key>=<value> format, one
+# entry per line. Example
+#
+# $ ./mender-device-identity
+# mac=de:ad:ca:fe:00:01
+# cpuid=1112233
```
kacf: Btw, will we pay any attention to order? We should perhaps keep it sorted in the client to make sure it doesn't mess up the hash.

kacf:
Looks good otherwise.

support/mender-device-identity
bboozzoo: On Wed, May 18, 2016 at 1:12 PM, Kristian Amlie <notifications@github.com>
wrote:

> In support/mender-device-identity
> <https://github.com/mendersoftware/mender/pull/50#discussion_r63684603>:
>
> > @@ -0,0 +1,50 @@
> > +#!/bin/sh
> > +
> > +# Example script called by Mender agent to collect device identity
> > +# data. The script needs to be located at
> > +# $(bindir)/mender-device-indentity path for the agent to find it.
> > +# The script shall exit with non-0 status on errors. In this case the
> > +# agent will discard any output the script may have produced.
> > +#
> > +# The script shall output identity data in <key>=<value> format, one
> > +# entry per line. Example
> > +#
> > +# $ ./mender-device-identity
> > +# mac=de:ad:ca:fe:00:01
> > +# cpuid=1112233
>
> Btw, will we pay any attention to order? We should perhaps keep it sorted
> in the client to make sure it doesn't mess up the hash.
>
Right, we'll sort it in the client

-- 
Maciej Borzecki

PR #51, opened 2016-05-20
documentation update - device identity

bboozzoo:
First patch wraps lines in README at 70/80 characters per line. The original wrapping was at 120 characters making it difficult to read and edit.

The second patch adds documentation for platform integration and device identity.

@kacf @pasinskim

### DIFF
--- a/README.md
+++ b/README.md
@@ -170,8 +170,50 @@ switched to roll-back mode and the partition with the update won't be
 used to boot. Instead, the active partition will be used as it has
 been proven to work correctly.
 
+7. Platform integration
+=======================
 
-7. Project roadmap
+Mender requires certain functionality to be provided by platform
+integration glue. This covers performing certain operations on the
+device or obtaining information about platform that Mender is running
+on.
+
+Platform integration is implemented with use of helper tools that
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -170,8 +170,50 @@ switched to roll-back mode and the partition with the update won't be
...
+
+Mender requires certain functionality to be provided by platform
+integration glue. This covers performing certain operations on the
+device or obtaining information about platform that Mender is running
+on.
+
+Platform integration is implemented with use of helper tools that
+Mender will call during runtime. The tools are named using `mender-`
+prefix (for example `mender-device-identity`), and need to be placed
+in `$PATH` (for instance in `/usr/bin/`).
```
kacf: I recall a discussion being had that we wanted to call `/usr/bin/mender-device-identity` directly, instead of relying on the path. At least for the moment, I think it's better not to leave it up to choice but to specify `/usr/bin` specifically. We can always relax the condition later if we find it necessary, but going the other way might be a bit more tricky if people already put it somewhere else.

kacf:
The rest looks good!

PR #55, opened 2016-05-25
mender.conf: add example mender config file

bboozzoo:
The config is based on mender.conf shipped by meta-mender Yocto
layer. The default server address is points to localhost:9080, so that
the the client will try to access the local services set up with docker.

### DIFF
--- a/main.go
+++ b/main.go
@@ -35,6 +35,7 @@ type logOptionsType struct {
 
 type runOptionsType struct {
 	version   *bool
+	config    *string
 	imageFile *string
 	commit    *bool
 	daemon    *bool
@@ -79,6 +80,9 @@ func argsParse(args []string) (runOptionsType, error) {
 
 	version := parsing.Bool("version", false, "Show mender agent version and exit.")
 
+	config := parsing.String("config", "/etc/mender/mender.conf",
+		"Configuration file location.")
+
 	commit := parsing.Bool("commit", false, "Commit current update.")
 
 	imageFile := parsing.String("rootfs", "",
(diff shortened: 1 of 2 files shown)


mender.conf
```diff
@@ -0,0 +1,12 @@
+{
+  "pollIntervalSeconds": 60,
+  "ServerURL": "localhost:9080",
+  "DeviceID": "ABCD-1234",
```
kacf: Shouldn't device ID be figured out using the identity script?

kacf:
Will you provide a meta-mender PR to copy that file into the image?

mender.conf
bboozzoo: On Wed, May 25, 2016 at 3:12 PM, Kristian Amlie <notifications@github.com>
wrote:

> In mender.conf
> <https://github.com/mendersoftware/mender/pull/55#discussion_r64569173>:
>
> > @@ -0,0 +1,12 @@
> > +{
> > +  "pollIntervalSeconds": 60,
> > +  "ServerURL": "localhost:9080",
> > +  "DeviceID": "ABCD-1234",
>
> Shouldn't device ID be figured out using the identity script?
>

The script is there, but it's not plugged in. We should update the config
once the integration is there.

-- 
Maciej Borzecki

kacf: Ok.

bboozzoo:
On Wed, May 25, 2016 at 3:13 PM, Kristian Amlie <notifications@github.com>
wrote:

> Will you provide a meta-mender PR to copy that file into the image?
>

IMO it's better to keep those separate. One provided here should be an
example to start with after someone builds the binary. The meta-mender one
is already rigged to work with the layer.

-- 
Maciej Borzecki

kacf:
Ah ok, I see. Perhaps we should give it an `.example` suffix then, just to be clear.

PR #56, opened 2016-05-26
Add very basic version of Mender client architecture overview.

pasinskim:


kacf:
Nice! But you're signing off with a different email address than you committed with. Both should be `@mender.io`.

kacf:
Fixed this for you here: #65

PR #58, opened 2016-05-27
bootstrap & cleanups

bboozzoo:
The series implements bootstrap handling as part of Mender controller. The old bootstrap procedure is unused, hence the code was removed from the tree.

The series introduces cleanups to state handling. The implicit state transitions as implemented before was replaced with explicit `TransitionState()` call. Daemon code was adjusted to work with changed controller.

@pasinskim @kacf @mchalski


mender.go
```diff
@@ -104,20 +132,61 @@ func (m mender) GetCurrentImageID() string {
...
+
+	newstate := MenderStateUnknown
+	var merr error
+
+	switch m.state {
+	case MenderStateInit:
+		if m.needsBootstrap() {
+			if err := m.doBootstrap(); err != nil {
+				newstate = MenderStateError
+				merr = err
```
kacf: Won't `MenderStateError` cause the daemon to exit? Failure to bootstrap would be quite common in bad reception circumstances, so that would be inappropriate.

mender.go
```diff
@@ -104,20 +132,61 @@ func (m mender) GetCurrentImageID() string {
...
+		} else {
+			if upg {
+				newstate = MenderStateRunningWithFreshUpdate
+			} else {
+				newstate = MenderStateWaitForUpdate
+			}
+		}
+	}
+
+	// recprd last errpr
```
kacf: `s/recprd last errpr/record last error/`

mender.go
```diff
@@ -104,20 +132,61 @@ func (m mender) GetCurrentImageID() string {
...
+				newstate = MenderStateBootstrapped
+			}
+		} else {
+			newstate = MenderStateBootstrapped
+		}
+	case MenderStateBootstrapped:
+		upg, err := m.hasUpgrade()
+		if err != nil {
+			newstate = MenderStateError
+			merr = err
```
kacf: Same here, daemon will exit (unless I have misunderstood the main loop).

mender.go
```diff
@@ -164,6 +244,52 @@ func (m mender) GetDaemonConfig() daemonConfig {
...
+			return err
+		}
+	}
+	return nil
+}
+
+func (m *mender) doBootstrap() error {
+
+	return nil
+}
```
kacf: I'm confused about the roles of `Bootstrap()` and `doBootstrap()`. Is one supposed to call the other? I assume that this is a stub right now that needs to be extended later?

mender.go
```diff
@@ -164,6 +244,52 @@ func (m mender) GetDaemonConfig() daemonConfig {
...
+				m.config.DeviceKey, err)
+			return err
+		}
+	}
+	return nil
+}
+
+func (m *mender) doBootstrap() error {
+
+	return nil
```
kacf: You should probably set `forceBootstrap = false` at the end of a successful bootstrap, either here or in `Bootstrap()`.

mender_test.go
```diff
@@ -24,7 +25,11 @@ func Test_updateState_readBootEnvError_returnsError(t *testing.T) {
 	fakeEnv := uBootEnv{&runner}
 	mender := NewMender(&fakeEnv)
 
-	if err := mender.updateState(); err == nil {
+	// pretend we're boostrapped
+	mender.state = MenderStateBootstrapped
+
+	if state := mender.TransitionState(); state != MenderStateError {
+		t.Logf("got state: %v", state)
 		t.FailNow()
```
kacf: A shortcut for this in go is:

```
-t.Logf("got state: %v", state)
-t.FailNow()
+t.Fatalf("got state: %v", state)
```

mender_test.go
```diff
@@ -179,48 +195,110 @@ func validateConfiguration(actual menderFileConfig) bool {
...
+	if err := mender.Bootstrap(); err != nil {
 		t.FailNow()
 	}
+
+	k := NewKeystore()
+	if err := k.Load("temp.key"); err != nil {
+		t.FailNow()
+	}
+
+	defer os.Remove("temp.key")
```
kacf: Shouldn't this be moved to right after the writing of the file?

mender.go
bboozzoo: `doBootstrap()` should probably go, not sure why I've left it around, makes no sense.

mender.go
bboozzoo: Yeah, I'll get rid of `doBoostrap()`.

mender.go
bboozzoo: Reception has no influence on bootstrap now. Though we may need to revise it later if any remote operation happens to be performed as part of bootstrap.

mender.go
bboozzoo: Yes, it will exit at this point. I think adding some sort of checking if error is transient may help here. Something in the line of:
```
func IsErrTransient(e error) bool {
....
}
```
and making sure to define sensible errors.

Anyway, we need to discuss the agent state machine as it's not that clear to me. The previous code would just enter `MenderStateUnknown` and continue working, even though the error was caused by a failure reading uboot environment (so it's not really transient). We ended up in an undefined state, cannot read uboot environment, but keep on poking the server for new updates, you see that there's something fishy around here?

Since it looks like a larger job, I propose to add the transient error check at this point, followed by opening a JIRA ticket to have the state machine defined in more detail.

@pasinskim what's your take on this?

daemon.go
```diff
@@ -59,12 +60,27 @@ func (daemon menderDaemon) StopDaemon() {
...
 	// figure out the state
-	switch daemon.GetState() {
-	case MenderFreshInstall:
-		//do nothing
-	case MenderRunningWithFreshUpdate:
-		daemon.CommitUpdate()
+	for {
+		switch daemon.TransitionState() {
+		case MenderStateRunningWithFreshUpdate:
+			daemon.CommitUpdate()
```
pasinskim: I know this was a part of refactoring, but `CommitUpdate()` in fact can return error and we should check for it. something to remember to add in future.

mender.go
```diff
@@ -26,50 +26,78 @@ import (
...
+//               v
+//             init
+//               |
+//               v
+//          bootstrapped
+//               |
+//       +-------+-------------+
+//       |                     |
+//       v                     v
+// fresh update         wait for update
```
pasinskim: I think this might be bit more complicated than that. What if you have the update but it is failed one (aka some checks are failing). Then we will need to reboot the device and use previously active partition as the one you will boot from.

bboozzoo: Well, yes, but IMO it's a part of another task. As I've indicated here https://github.com/mendersoftware/mender/pull/58#discussion_r65037067 the client state model is incomplete and needs working.

daemon.go
```diff
@@ -59,12 +60,27 @@ func (daemon menderDaemon) StopDaemon() {
 
 func (daemon *menderDaemon) Run() error {
 	// figure out the state
-	switch daemon.GetState() {
-	case MenderFreshInstall:
-		//do nothing
-	case MenderRunningWithFreshUpdate:
-		daemon.CommitUpdate()
+	for {
```
pasinskim: Not sure if we need a loop here. At least for now, all the state transitions are killing the daemon (which is another thing to discuss).

daemon.go
```diff
@@ -59,12 +60,27 @@ func (daemon menderDaemon) StopDaemon() {
...
+		switch daemon.TransitionState() {
+		case MenderStateRunningWithFreshUpdate:
+			daemon.CommitUpdate()
+		case MenderStateWaitForUpdate:
+			err := daemon.waitForUpdate()
```
pasinskim: I like moving this from main loop to separate function.

daemon.go
```diff
@@ -59,12 +60,27 @@ func (daemon menderDaemon) StopDaemon() {
...
+			if err != nil {
+				return err
+			}
+		case MenderStateError:
+			log.Errorf("entered error state due to: %s", daemon.LastError())
+			return daemon.LastError()
```
pasinskim: Maybe error-recovery is something we should discuss?

daemon.go
```diff
@@ -59,12 +60,27 @@ func (daemon menderDaemon) StopDaemon() {
...
-	case MenderRunningWithFreshUpdate:
-		daemon.CommitUpdate()
+	for {
+		switch daemon.TransitionState() {
+		case MenderStateRunningWithFreshUpdate:
+			daemon.CommitUpdate()
+		case MenderStateWaitForUpdate:
+			err := daemon.waitForUpdate()
+			if err != nil {
+				return err
```
pasinskim: If for some reason, you wont be able to fetch update, you will kill the daemon, which is not what I would expect.

bboozzoo: Well the problem is that fetch update, appply update and commit update should be separate states/state action. Having that you could easily decide about the next transition. Right now we can only guess based on error returned.

keystore.go
```diff
@@ -0,0 +1,129 @@
...
+}
+
+func (k *Keystore) Save(privPath string) error {
+	outf, err := os.OpenFile(privPath, os.O_CREATE|os.O_WRONLY, 0600)
+	if err != nil {
+		return err
+	}
+	defer outf.Close()
+
+	err = saveToPem(outf, k.private)
```
pasinskim: Should we check if we have `k.private` here?

keystore.go
```diff
@@ -0,0 +1,129 @@
...
+	k.private = key
+
+	return nil
+}
+
+func (k *Keystore) Private() *rsa.PrivateKey {
+	return k.private
+}
+
+func IsNoKeys(e error) bool {
```
pasinskim: Maybe `HaveNoKeys()`?

keystore_test.go
```diff
@@ -0,0 +1,61 @@
...
+	assert.Nil(t, k.Private())
+
+	// try using temp file, this time we should get unmarhall/load
+	// error
+	err = k.Load(tname)
+	assert.Error(t, err)
+	assert.False(t, IsNoKeys(err))
+	assert.Nil(t, k.Private())
+
+	// not changing random source, so this is no expected to fail
```
pasinskim: s/no/not

keystore_test.go
```diff
@@ -0,0 +1,61 @@
...
+	assert.NoError(t, k.Generate())
+
+	assert.NotNil(t, k.Private())
+
+	// try to save in at location that triggers an error
```
pasinskim: s/in at/in

keystore_test.go
```diff
@@ -0,0 +1,61 @@
...
+	assert.Error(t, k.Save("/foo"))
+
+	// try our temp location now
+	assert.NoError(t, k.Save(tname))
+
+	// we should be able to load a saved key
+	assert.NoError(t, k.Load(tname))
+
+	os.Remove(tname)
```
pasinskim: Shouldn't we have `defer os.Remove(tfile.Name()) just after checking that file was created?

mender.go
```diff
@@ -26,50 +26,78 @@ import (
...
+	MenderStateInit
+	// client is bootstrapped, i.e. ready to go
+	MenderStateBootstrapped
+	// update applied, waiting for commit
+	MenderStateRunningWithFreshUpdate
+	// wait for new update
+	MenderStateWaitForUpdate
+	// error occurred, call Controller.LastError() to obtain the
+	// error
+	MenderStateError
```
pasinskim: Nice cleanup here!

mender.go
```diff
@@ -104,20 +132,61 @@ func (m mender) GetCurrentImageID() string {
 	return imageID
 }
 
-func (m *mender) updateState() error {
+func (m *mender) changeState(state MenderState) {
+	log.Infof("Mender state: %v -> %v", m.state, state)
+	m.state = state
+}
+
+func (m *mender) hasUpgrade() (bool, error) {
```
pasinskim: For consistency I think it should be `hasUpdate()`, the name of boot variable is wrong as well, but this was there before...

mender.go
```diff
@@ -104,20 +132,61 @@ func (m mender) GetCurrentImageID() string {
...
-	return nil
+	return false, nil
+}
+
+func (m *mender) updateState() {
+
+	newstate := MenderStateUnknown
+	var merr error
+
+	switch m.state {
```
pasinskim: I am not sure if there is not some logic error here. Wont we always start with `MenderStateInit` (which is default state in constructor)? Then, it looks like `needsBootstrap()` will be always checked here, and we will end up with `MenderStateBootstrapped`.
If there are other places in the code relaying on the actual state check we will report wrong state as the on we are really in might be `MenderStateRunningWithFreshUpdate` or `MenderStateWaitForUpdate` which will be reported only after entering `updateState()` for the second time.

What is more `needsBootstrap()` might not work well. What if we've been already bootstrapped, then update and reset happened. We will end up running from inactive partition being not bootstrapped.

pasinskim: Please note that I might be wrong above, just want to make sure the logic is correct.

mender.go
pasinskim: I agree. Let's discuss it in some JIRA ticket and fix the state machine.

mender.go
pasinskim: Where we are actually going to bootstrap, and what bootstrapping really means? Should we send some data to server and verify it accepts us (at least check that TLS connection was successful)?

keystore.go
bboozzoo: fixed

mender.go
bboozzoo: Right now it means generating device keys.

mender.go
bboozzoo: fixed

mender_test.go
bboozzoo: Fixed.

How about I update some of the tests to use `github.com/stretchr/testify/assert`?

mender_test.go
bboozzoo: fixed

keystore.go
bboozzoo: Trying to stick to convention here: https://godoc.org/os#IsNotExist

keystore_test.go
bboozzoo: fixed

keystore_test.go
bboozzoo: fixed

keystore_test.go
bboozzoo: fixed

mender_test.go
kacf: Or that. Before I knew about that one I started using `mendertesting.AssertTrue`, but testify is probably more complete.

PR #59, opened 2016-05-28
fixed contribution link and added full ASL text

ThomasRyd:
Signed-off-by: ThomasRyd <thomas@mender.io>

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,13 +1,58 @@
-// Copyright 2016 Mender Software AS
-//
-//    Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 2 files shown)


kacf:
As stated previously, you cannot change the license text because the current license check expects the short text, so the build breaks. I'm already in the process of fixing this, I'm just prioritizing sprint tasks until tomorrow.

If you're worried that the full license text is not included, it *is* included in `COPYING`, which is the more traditional way in OSS. I will consolidate the two when I return to my ValueAdd.

kacf:
Superseded by https://github.com/mendersoftware/mender/pull/60.

PR #62, opened 2016-06-01
Add automatic license checking and update mendertesting repo.

kacf:
```
commit 3c3a4f68aa2699429c94d0f3a0993f36b0854e10
Author: Kristian Amlie <kristian.amlie@mender.io>
Date:   Wed Jun 1 08:52:31 2016

    Use full license text in LICENSE, and remove COPYING.
    
    Signed-off-by: Kristian Amlie <kristian.amlie@mender.io>

commit 40e0e4f7356d72e041b70a2a31675e53fbc90e35
Author: Kristian Amlie <kristian.amlie@mender.io>
Date:   Wed Jun 1 08:47:33 2016

    Add automatic license checking and update mendertesting repo.
    
    This means removing the Travis check, since it is now done implicitly
    by "go test".
    
    Signed-off-by: Kristian Amlie <kristian.amlie@mender.io>

commit 5b382e0c4291760e30b01001690568321e615e78
Author: Kristian Amlie <kristian.amlie@mender.io>
Date:   Tue May 10 16:35:08 2016

    Clean up junk after test.
    
    Signed-off-by: Kristian Amlie <kristian.amlie@mender.io>

commit b4caf3d5ed1a736956a653664321c2a4037ee964
Author: Kristian Amlie <kristian.amlie@mender.io>
Date:   Wed Jun 1 08:40:08 2016

    Switch from Godeps/_workspace/ to vendor/ (standard Go vendoring).
    
    Note that the default install action of Travis is to run Godep, but
    this messes with our vendored dependencies because it always pulls
    latest, therefore there are API incompatibilities. So define that
    action ourselves instead.
    
    Signed-off-by: Kristian Amlie <kristian.amlie@mender.io>
```

PR #63, opened 2016-06-01
build & version handling cleanups

bboozzoo:
Populate version information during build with output of `git describe --always --tags --dirty`. Then running `mender -version` produces this output:

```
$ ./mender -version
v1.0.0-44-ge7f4d6d
```

The series also introduces a Makefile. Although running `make` is not part of the usual Go workflow, it's necessary so that the same commands are ran during the build. It also provides a shorter way of invoking different go tools.

@kacf @pasinskim @eysteins @GregorioDiStefano 

change to `meta-mender` added in this PR: https://github.com/mendersoftware/meta-mender/pull/52


.travis.yml
```diff
@@ -50,19 +50,19 @@ before_install:
...
-    # Install varcheck
-    - go get -u github.com/opennota/check/cmd/varcheck
+    # Get tools used in Makefile
+    - make get-tools
 
-    # Install deadcode
-    - go get -u github.com/mendersoftware/deadcode
+install:
+    # Get dependencies
+    - go get -t ./...
```
kacf: This is going to conflict with https://github.com/mendersoftware/mender/pull/62, where we switch to Go 1.6 vendored dependencies (see the bottom commit). You'll have to keep getting each tool individually.

Makefile
```diff
@@ -0,0 +1,72 @@
...
+		echo "-- go getting $$t"; \
+		go get -u $$t; \
+	done
+
+test:
+	$(GO) test -v ./...
+
+check:
+	echo "-- checking license header"
+	./check_license.sh
```
kacf: Drop this, it has changed completely in https://github.com/mendersoftware/mender/pull/62, and is now covered by `go test`.

kacf:
+1 on the Makefile, we're going to need this for other configurable values as well, such as https://tracker.mender.io/browse/MEN-440.

The "check_signed_off: import improved version from deviceadm" commit you can drop completely, this will change radically in #62.

.travis.yml
kacf: I suppose you can run `make get-tools` instead?

kacf:
The rest looks good!

bboozzoo:
Ok, I"ll wait for #62 to get merged and rebase on top of that.

kacf:
I have merged #62! :-)

.travis.yml
bboozzoo: Not really. In my opinion, Travis buils are bad in the sense that we're not using dependencies vendored by godep (ones in `Godeps/_workspace`). This happens as a result of 2 issues. Firstly we were directly calling `go build ...` instead of `godep go build ..`. Secondly (a side-effect of the first one), Travis automatically runs `go get -t ./...` if there's no install script, hence external dependencies were downloaded into `GOPATH`, each directly from `master` branch. Since the builds were never run with `go build .. -v`, it's not really easy to spot which packages are used.

kacf: Indeed, I discovered this confusing behavior as well in #62. But now `go get ./...` is prevented, and everything is supposed to be picked up from `vendor`, aka Go 1.6 style.

kacf:
> check_signed_off: remove unused script

Of course! Doh!

Makefile
```diff
@@ -0,0 +1,71 @@
+GO ?= go
+GOFMT ?= gofmt
+V ?=
+PKGS = $(shell go list ./... | grep -v vendor)
+PKGFILES = $(shell find . \( -path ./vendor -o -path ./Godeps \) -prune \
+												-o -type f -name '*.go' -print)
```
kacf: Something weird with the tabs here? It's very far off to the right.

kacf:
Just a minor tab thing, otherwise looks good!

Makefile
bboozzoo: Yeah, switched to spacemacs recently, the 2 space tab stop is weird :)

kacf:
Great, thanks!

PR #67, opened 2016-06-07
MEN-440: Dynamic/configurable partition devices

kacf:
See individual commit messages for descriptions.

### DIFF
--- a/partitions.go
+++ b/partitions.go
@@ -28,12 +28,15 @@ var (
 	InvalidActivePartition         = errors.New("Invalid active partition")
 	RootPartitionDoesNotMatchMount = errors.New("Can not match active partition and any of mounted devices.")
 	ErrorNoMatchBootPartRootPart   = errors.New("No match between boot and root partitions.")
+	ErrorPartitionNumberNotSet     = errors.New("PartitionANumber and PartitionBNumber settings are not both set.")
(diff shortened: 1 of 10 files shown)


partitions.go
```diff
@@ -82,28 +85,53 @@ func (p *partitions) getPartitionSize(partition string) (int64, error) {
...
-	case "3":
-		p.inactive = mountPrefix + "2"
-		return p.inactive, nil
-	default:
+	// Check whether the active partition ends with the number of either the
+	// A or the B partition, and then assign the non-matching one to
+	// inactive.
+	var inactive string
+	var itsA, itsB bool
+	if active[len(active)-len(p.partitionANumber):] == p.partitionANumber {
```
bboozzoo: Seems like a guessing game. Can we instead look at `/sys/block/<device>/<device>*/partition` instead for the exact numbers?

kacf: Yes, good idea, that's much better. I'll make that switch.

kacf:
Alright, I have updated the branch, just adding one commit for easy review. I will squash it before merging though.

partitions.go
kacf: Because of other comments in the meta-mender PR, I decided to switch to full device names instead of numbers. Which means we can do direct string comparison, so this whole section is much simpler now.

bboozzoo:
Looks good to me, much better than before :)

PR #68, opened 2016-06-14
client state refactoring & basic authorization setup

bboozzoo:
While working on authorization implementation it turned out that the client code needs a refactoring, otherwise it would be hard to go through authorization steps in a clean manner.

The patch series introduces some major changes to the agent code, in rough detail these are:
* do away with inheritance (or struct/interface embeddeding) in favor of composition where possible; for instance `menderDaemon` does not need to export any interfaces, while `mender` should only export minimal set of interfaces
* introduce a single `Controler` interface
* split `Client` and `UpdateClient` to separate interfaces
* use a state pattern to implement client states, separates controler from actual state execution; operations on controler have been nicely moved to states
* introduction of `menderError` interface; this extends the `error` with extra methods for querying whether an error is fatal or not (used by error state handler)
 
Introuce some helpers/interfaces for implementation of authorization flow, these are:
* `IdentityDataGetter` for obtaining device identity information
* `AuthClient` actual client to the authorization service
* `FileSeqnum`, `Seqnum` number sequence generators with file storage (required for keeping track of auth request sequence numbers)
* `Store` a  generic storage wrapper

Since the authorization work is kind of separate from the rest, I can put that into a different branch and open a PR from that branch after we've merged the generic changes form this PR.

@pasinskim @kacf

### DIFF
--- a/LIC_FILES_CHKSUM.sha256
+++ b/LIC_FILES_CHKSUM.sha256
@@ -1,9 +1,12 @@
+# These are identical to Apache-2.0 license.
 bbb303820971c294a9a8e5eba5affcf1379036e877ea61c11cbf9400b2949483  LICENSE
 bbb303820971c294a9a8e5eba5affcf1379036e877ea61c11cbf9400b2949483  vendor/github.com/mendersoftware/mendertesting/LICENSE
 3591f687e2d6f49c83b1ec69577e8110afbde80be5ec81791bd86d2838ccd3de  vendor/github.com/mendersoftware/log/LICENSE
 bbb303820971c294a9a8e5eba5affcf1379036e877ea61c11cbf9400b2949483  vendor/github.com/mendersoftware/log/COPYING
 3591f687e2d6f49c83b1ec69577e8110afbde80be5ec81791bd86d2838ccd3de  vendor/github.com/mendersoftware/scopestack/LICENSE
 bbb303820971c294a9a8e5eba5affcf1379036e877ea61c11cbf9400b2949483  vendor/github.com/mendersoftware/scopestack/COPYING
+
+# These are identical to BSD 2 Clause license.
 192def682f6846c6a57e20cf044e224fa467948358163a18dd37bfe14578e6ab  vendor/github.com/pkg/errors/LICENSE
 
 # These are all identical to MIT license.
--- a/LIC_FILES_CHKSUM.sha256
+++ b/LIC_FILES_CHKSUM.sha256
@@ -4,6 +4,7 @@ bbb303820971c294a9a8e5eba5affcf1379036e877ea61c11cbf9400b2949483  vendor/github.
 bbb303820971c294a9a8e5eba5affcf1379036e877ea61c11cbf9400b2949483  vendor/github.com/mendersoftware/log/COPYING
 3591f687e2d6f49c83b1ec69577e8110afbde80be5ec81791bd86d2838ccd3de  vendor/github.com/mendersoftware/scopestack/LICENSE
 bbb303820971c294a9a8e5eba5affcf1379036e877ea61c11cbf9400b2949483  vendor/github.com/mendersoftware/scopestack/COPYING
+192def682f6846c6a57e20cf044e224fa467948358163a18dd37bfe14578e6ab  vendor/github.com/pkg/errors/LICENSE
 
 # These are all identical to MIT license.
 51a0c9ec7f8b7634181b8d4c03e5b5d204ac21d6e72f46c313973424664b2e6b  vendor/github.com/Sirupsen/logrus/LICENSE
--- a/client.go
+++ b/client.go
@@ -81,13 +81,6 @@ type httpsClientConfig struct {
 	isHttps    bool
 }
 
-type httpsClientAuthCreds struct {
-	// Cert+privkey that authenticates this client
-	clientCert tls.Certificate
-	// Trusted server certificates
-	trustedCerts x509.CertPool
-}
-
 func loadServerTrust(conf httpsClientConfig) (*x509.CertPool, error) {
 	if conf.serverCert == "" {
 		// TODO: this is for pre-production version only to simplify tests.
--- a/client.go
+++ b/client.go
@@ -16,14 +16,12 @@ package main
 import (
 	"crypto/tls"
 	"crypto/x509"
-	"encoding/json"
-	"errors"
-	"io"
 	"io/ioutil"
 	"net/http"
 	"strings"
 
 	"github.com/mendersoftware/log"
+	"github.com/pkg/errors"
 )
 
 var (
@@ -31,78 +29,47 @@ var (
 	errorAddingServerCertificateToPool = errors.New("Error adding trusted server certificate to pool.")
 )
 
-const (
-	minimumImageSize int64 = 4096 //kB
-)
-
 type RequestProcessingFunc func(response *http.Response) (interface{}, error)
 
-type Updater interface {
-	GetScheduledUpdate(processFunc RequestProcessingFunc, server string, deviceID string) (interface{}, error)
-	FetchUpdate(url string) (io.ReadCloser, int64, error)
-}
-
-// Client represents the http(s) client used for network communication.
-//
-type httpClient struct {
-	HTTPClient   *http.Client
-	minImageSize int64
-}
-
-type httpsClient struct {
-	httpClient
-	httpsClientAuthCreds
-}
-
 // Client initialization
-
-func NewUpdater(conf httpsClientConfig) (Updater, error) {
+func NewHttpClient(conf httpsClientConfig) (*http.Client, error) {
 
 	if conf == (httpsClientConfig{}) {
-		client := NewHttpClient()
-		if client == nil {
-			return nil, errors.New("Can not instantialte updater.")
-		}
-		return client, nil
+		return newHttpClient(), nil
 	}
-	client := NewHttpsClient(conf)
-	if client == nil {
-		return nil, errors.New("Can not instantialte updater.")
-	}
-	return client, nil
+
+	return newHttpsClient(conf)
 }
 
-func NewHttpClient() *httpClient {
-	var client httpClient
-	client.minImageSize = minimumImageSize
-	client.HTTPClient = &http.Client{}
-
-	return &client
+func newHttpClient() *http.Client {
+	return &http.Client{}
 }
 
-func NewHttpsClient(conf httpsClientConfig) *httpsClient {
-	var client httpsClient
-	client.httpClient = *NewHttpClient()
+func newHttpsClient(conf httpsClientConfig) (*http.Client, error) {
+	client := newHttpClient()
 
-	if err := client.initServerTrust(conf); err != nil {
-		log.Error("Can not initialize server trust: ", err)
-		return nil
+	trustedcerts, err := loadServerTrust(conf)
+	if err != nil {
+		return nil, errors.Wrapf(err, "cannot initialize server trust")
 	}
 
-	if err := client.initClientCert(conf); err != nil {
-		log.Error("Can not initialize client certificate: ", err)
-		return nil
+	clientcerts, err := loadClientCert(conf)
+	if err != nil {
+		return nil, errors.Wrapf(err, "can not load client certificate")
 	}
 
 	transport := http.Transport{
 		TLSClientConfig: &tls.Config{
-			RootCAs:      &client.trustedCerts,
-			Certificates: []tls.Certificate{client.clientCert},
+			RootCAs: trustedcerts,
 		},
 	}
 
-	client.HTTPClient.Transport = &transport
-	return &client
+	if clientcerts != nil {
+		transport.TLSClientConfig.Certificates = []tls.Certificate{*clientcerts}
+	}
+
+	client.Transport = &transport
+	return client, nil
 }
 
 // Client configuration
@@ -121,172 +88,46 @@ type httpsClientAuthCreds struct {
 	trustedCerts x509.CertPool
 }
 
-func (c *httpsClient) initServerTrust(conf httpsClientConfig) error {
+func loadServerTrust(conf httpsClientConfig) (*x509.CertPool, error) {
 	if conf.serverCert == "" {
 		// TODO: this is for pre-production version only to simplify tests.
 		// Make sure to remove in production version.
 		log.Warn("Server certificate not provided. Trusting all servers.")
-		return nil
+		return nil, nil
 	}
 
-	c.trustedCerts = *x509.NewCertPool()
+	certs := x509.NewCertPool()
 	// Read certificate file.
 	cacert, err := ioutil.ReadFile(conf.serverCert)
 	if err != nil {
-		return err
+		return nil, err
 	}
-	c.trustedCerts.AppendCertsFromPEM(cacert)
+	certs.AppendCertsFromPEM(cacert)
 
-	if len(c.trustedCerts.Subjects()) == 0 {
-		return errorAddingServerCertificateToPool
+	if len(certs.Subjects()) == 0 {
+		return nil, errorAddingServerCertificateToPool
 	}
-	return nil
+	return certs, nil
 }
 
-func (c *httpsClient) initClientCert(conf httpsClientConfig) error {
+func loadClientCert(conf httpsClientConfig) (*tls.Certificate, error) {
 	if conf.certFile == "" || conf.certKey == "" {
 		// TODO: this is for pre-production version only to simplify tests.
 		// Make sure to remove in production version.
 		log.Warn("No client key and certificate provided. Using system default.")
-		return nil
+		return nil, nil
 	}
 
 	clientCert, err := tls.LoadX509KeyPair(conf.certFile, conf.certKey)
 	if err != nil {
-		return errorLoadingClientCertificate
+		return nil, errorLoadingClientCertificate
 	}
-	c.clientCert = clientCert
-	return nil
+	return &clientCert, nil
 }
 
-func (c *httpClient) GetScheduledUpdate(process RequestProcessingFunc,
-	server string, deviceID string) (interface{}, error) {
-	// http client should be able to perform https requests
-	if strings.HasPrefix(server, "http://") || strings.HasPrefix(server, "https://") {
-		return c.getUpdateInfo(process, server, deviceID)
+func buildURL(server string) string {
+	if strings.HasPrefix(server, "https://") || strings.HasPrefix(server, "http://") {
+		return server
 	}
-	return c.getUpdateInfo(process, "http://"+server, deviceID)
-}
-
-func (c *httpsClient) GetScheduledUpdate(process RequestProcessingFunc,
-	server string, deviceID string) (interface{}, error) {
-	if strings.HasPrefix(server, "https://") {
-		return c.getUpdateInfo(process, server, deviceID)
-	}
-	return c.getUpdateInfo(process, "https://"+server, deviceID)
-}
-
-func (c *httpClient) getUpdateInfo(process RequestProcessingFunc, server string,
-	deviceID string) (interface{}, error) {
-	request := server + "/api/0.0.1/devices/" + deviceID + "/update"
-	r, err := c.makeAndSendRequest(http.MethodGet, request)
-	if err != nil {
-		log.Debug("Sending request error: ", err)
-		return nil, err
-	}
-
-	defer r.Body.Close()
-
-	return process(r)
-}
-
-// Returns a byte stream which is a download of the given link.
-func (c *httpClient) FetchUpdate(url string) (io.ReadCloser, int64, error) {
-	r, err := c.makeAndSendRequest(http.MethodGet, url)
-	if err != nil {
-		log.Error("Can not fetch update image: ", err)
-		return nil, -1, err
-	}
-
-	log.Debugf("Received fetch update response %v+", r)
-
-	if r.StatusCode != http.StatusOK {
-		log.Errorf("Error fetching shcheduled update info: code (%d)", r.StatusCode)
-		return nil, -1, errors.New("Error receiving scheduled update information.")
-	}
-
-	if r.ContentLength < 0 {
-		return nil, -1, errors.New("Will not continue with unknown image size.")
-	} else if r.ContentLength < c.minImageSize {
-		log.Errorf("Image smaller than expected. Expected: %d, received: %d", c.minImageSize, r.ContentLength)
-		return nil, -1, errors.New("Image size is smaller than expected. Aborting.")
-	}
-
-	return r.Body, r.ContentLength, nil
-}
-
-func (client *httpClient) makeAndSendRequest(method, url string) (*http.Response, error) {
-
-	res, err := http.NewRequest(method, url, nil)
-	if err != nil {
-		return nil, err
-	}
-
-	log.Debug("Sending HTTP [", method, "] request: ", url)
-	return client.HTTPClient.Do(res)
-}
-
-// possible API responses received for update request
-const (
-	updateResponseHaveUpdate = 200
-	updateResponseNoUpdates  = 204
-	updateResponseError      = 404
-)
-
-// have update for the client
-type UpdateResponse struct {
-	Image struct {
-		URI      string
-		Checksum string
-		YoctoID  string `json:"yocto_id"`
-		ID       string
-	}
-	ID string
-}
-
-func validateGetUpdate(update UpdateResponse) error {
-	// check if we have JSON data correctky decoded
-	if update.ID != "" && update.Image.ID != "" &&
-		update.Image.URI != "" && update.Image.YoctoID != "" {
-		log.Info("Correct request for getting image from: " + update.Image.URI)
-		return nil
-	}
-	return errors.New("Missing parameters in encoded JSON response")
-}
-
-func processUpdateResponse(response *http.Response) (interface{}, error) {
-	log.Debug("Received response:", response.Status)
-
-	respBody, err := ioutil.ReadAll(response.Body)
-	if err != nil {
-		return nil, err
-	}
-
-	switch response.StatusCode {
-	case updateResponseHaveUpdate:
-		log.Debug("Have update available")
-
-		data := UpdateResponse{}
-		if err := json.Unmarshal(respBody, &data); err != nil {
-			switch err.(type) {
-			case *json.SyntaxError:
-				return nil, errors.New("Error parsing data syntax")
-			}
-			return nil, errors.New("Error parsing data: " + err.Error())
-		}
-		if err := validateGetUpdate(data); err != nil {
-			return nil, err
-		}
-		return data, nil
-
-	case updateResponseNoUpdates:
-		log.Debug("No update available")
-		return nil, nil
-
-	case updateResponseError:
-		return nil, errors.New("Client not authorized to get update schedule.")
-
-	default:
-		return nil, errors.New("Invalid response received from server")
-	}
+	return "https://" + server
 }
--- a/client_test.go
+++ b/client_test.go
@@ -14,227 +14,39 @@
 package main
 
 import (
-	"errors"
-	"fmt"
-	"io"
-	"net/http"
-	"net/http/httptest"
-	"strings"
 	"testing"
+
+	"github.com/stretchr/testify/assert"
 )
 
-const correctUpdateResponse = `{
-"image": {
-"uri": "https://menderupdate.com",
-"checksum": "checksum",
-"yocto_id": "core-image-base",
-"id": "f81d4fae-7dec-11d0-a765-00a0c91e6bf6"
-},
-"id": "13876-123132-321123"
-}`
-
-const malformedUpdateResponse = `{
-"image": {
-"non_existing": "https://menderupdate.com",
-"checksum": "Hello, world!",
-},
-"bad_field": "13876-123132-321123"
-}`
-
-const brokenUpdateResponse = `{
-"image": {
-"uri": "https://menderupdate
-"checksum": "Hello, world!",
-"id": "f81d4fae-7dec-11d0-a765-00a0c91e6bf6"
-},
-"id": "13876-123132-321123"
-}`
-
-const missingFieldsUpdateResponse = `{
-"image": {
-"uri": "https://menderupdate.com",
-"id": "f81d4fae-7dec-11d0-a765-00a0c91e6bf6"
-},
-"id": "13876-123132-321123"
-}`
-
-var updateTest = []struct {
-	responseStatusCode    int
-	responseBody          []byte
-	shoulReturnError      bool
-	shouldCheckReturnCode bool
-	returnCode            int
-}{
-	{200, []byte(correctUpdateResponse), false, true, updateResponseHaveUpdate},
-	{204, []byte(""), false, true, updateResponseNoUpdates},
-	{404, []byte(`{
-	 "error": "Not found"
-	 }`), true, true, updateResponseError},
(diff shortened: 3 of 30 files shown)


client_auth.go
```diff
@@ -0,0 +1,122 @@
...
+	"bytes"
+	"encoding/json"
+	"io/ioutil"
+	"net/http"
+
+	"github.com/mendersoftware/log"
+	"github.com/pkg/errors"
+)
+
+type AuthGetter interface {
```
pasinskim: I think we need to agree on some convention how we are naming the interfaces. Here I think we should have `Requester` or `AuthRequester`.

client_auth.go
```diff
@@ -0,0 +1,122 @@
...
+	// returns device public key
+	Pubkey() (string, error)
+	// returns tenant token
+	Token() (string, error)
+	// returns identity data
+	IdentityData() (interface{}, error)
+}
+
+type AuthClient struct {
+	client *http.Client
```
pasinskim: I might remove this later on when I will go through the rest of the code, but for now I am wondering why we need `AuthClient` to be separate entity. Won't it be more intuitive if we will have `Updater` only? We can rename it as well, but essentially we need just one thing, and the fact that this is authenticated or not is just a state of the client. What is more, let's assume we will add reporting. Will we add new `Reporter` struct and add some interface and implementation for sending reports? Or maybe we will reuse the same client to do everything (send reports, ask for update and authenticate).
The only reason I can see to split this is if (or when) we will use different protocols/methods to send reports and to fetch update.

client_auth.go
```diff
@@ -0,0 +1,122 @@
...
+}
+
+func (u *AuthClient) Request(server string, datasrc AuthDataProvider) (interface{}, error) {
+
+	req, err := makeAuthRequest(server, datasrc)
+	if err != nil {
+		return nil, errors.Wrapf(err, "failed to build auth request")
+	}
+
+	log.Debugf("making authorization request to server %s with req: %s", server, req)
```
pasinskim: +1 for adding debug logs

client_auth.go
```diff
@@ -0,0 +1,122 @@
...
+	data, err := ioutil.ReadAll(rsp.Body)
+	if err != nil {
+		return nil, errors.Wrapf(err, "failed to receive response data")
+	}
+
+	// validate response
+	log.Debugf("received response data %v", data)
+
+	return data, nil
+}
```
pasinskim: Again, I am commenting as I go, but this looks pretty common no matter what kid of request we are building and sending. Maybe it makes sense to abstract this is one client as all the communication with the server will boil down to `makeRequest()`, `client.Do()` and `ReadAll()`.

client_auth.go
```diff
@@ -0,0 +1,122 @@
...
+
+type authReqData struct {
+	IdData      interface{} `json:"id_data"`
+	TenantToken string      `json:"tenant_token"`
+	Pubkey      string      `json:"pubkey"`
+}
+
+func makeAuthRequest(server string, datasrc AuthDataProvider) (*http.Request, error) {
+	url := buildURL(server) + "/v0.1/auth_requests"
```
pasinskim: I think correct API version is `0.0.1`. What is more, maybe it is worth to close it in some function similar to `buildURL()` or in fact it can be a part of `buildURL()`.

client_auth.go
```diff
@@ -0,0 +1,122 @@
...
+	enc := json.NewEncoder(databuf)
+
+	err = enc.Encode(&authd)
+	if err != nil {
+		return nil, errors.Wrapf(err, "failed to encode auth request")
+	}
+
+	req, err := http.NewRequest(http.MethodGet, url, databuf)
+	if err != nil {
+		return nil, err
```
pasinskim: I like the way of returning error strings with `errors.Wrap()`. Here we can receive some weird error which might be hard to figure out later on.

client_auth.go
```diff
@@ -0,0 +1,122 @@
...
+	}
```
pasinskim: Maybe some debug log `auth request created...`?

client_auth_test.go
```diff
@@ -0,0 +1,129 @@
...
+		MAC: "foo",
+	}
+	req, err = makeAuthRequest("mender.io", &testAuthDataProvider{
+		token: "token",
+		idata: &fakedata,
+	})
+	assert.NotNil(t, req)
+	assert.NoError(t, err)
+	assert.Equal(t, http.MethodGet, req.Method)
+	assert.Equal(t, "https://mender.io/v0.1/auth_requests", req.URL.String())
```
pasinskim: Version should be `0.0.1`.

client_auth_test.go
```diff
@@ -0,0 +1,129 @@
...
+	assert.NoError(t, err)
+	assert.Equal(t, "token", authreq.TenantToken)
+	assert.Equal(t, "", authreq.Pubkey)
+	assert.Equal(t, "foo", (authreq.IdData.(map[string]interface{})["mac"]))
+}
+
+func TestClientAuth(t *testing.T) {
+	// Test server that always responds with 200 code, and specific payload
+	ts := httptest.NewTLSServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		w.WriteHeader(200)
```
pasinskim: `w.WriteHeader(http.StatusOK)`

client_test.go
```diff
@@ -14,227 +14,39 @@
...
-	}
-}
-
-func Test_FetchUpdate_invalidRequest_UpdateFailing(t *testing.T) {
-	// Test server that always responds with 200 code, and specific payload
-	ts := httptest.NewTLSServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
-		w.WriteHeader(200)
-		w.Header().Set("Content-Type", "application/json")
+	assert.NotNil(t, cl)
+	// assert.NotNil(t, cl.Transport.TLSClientConfig)
```
pasinskim: Remove comment?

client_test.go
```diff
@@ -14,227 +14,39 @@
...
 
-		fmt.Fprint(w, "")
-	}))
-	defer ts.Close()
+	// no https config, we should obtain a httpClient
+	cl, err = NewHttpClient(httpsClientConfig{})
+	assert.NotNil(t, cl)
+	// assert.Nil(t, cl.Transport.TLSClientConfig)
```
pasinskim: remove comment

client.go
```diff
@@ -114,179 +81,46 @@ type httpsClientConfig struct {
...
-	case updateResponseError:
-		return nil, errors.New("Client not authorized to get update schedule.")
-
-	default:
-		return nil, errors.New("Invalid response received from server")
+func buildURL(server string) string {
+	if strings.HasPrefix(server, "https://") || strings.HasPrefix(server, "http://") {
+		return server
 	}
+	return "https://" + server
```
pasinskim: Just a note here as this is still bit messy without server supporting HTTPS and client not having authentication implemented. I am not sure if returning `https` by default is the correct way. Maybe we can see and modify later once the code will be more mature.

client_update.go
```diff
@@ -0,0 +1,183 @@
...
+	GetScheduledUpdate(server string, deviceID string) (interface{}, error)
+	FetchUpdate(url string) (io.ReadCloser, int64, error)
+}
+
+type UpdateClient struct {
+	client       *http.Client
+	minImageSize int64
+}
+
+func NewUpdateClient(conf httpsClientConfig) (*UpdateClient, error) {
```
pasinskim: As mentioned above maybe we should have one client only, and if this is HTTPS client we should make sure that the authentication requests are performed before update requests. Maybe just something for further refactoring.

client_update.go
```diff
@@ -0,0 +1,183 @@
...
+	}
+
+	up := UpdateClient{
+		client:       client,
+		minImageSize: minimumImageSize,
+	}
+	return &up, nil
+}
+
+func (u *UpdateClient) GetScheduledUpdate(server string, deviceID string) (interface{}, error) {
```
pasinskim: The idea of adding "processing function" as a parameter was to avoid tight coupling and rather use dependency injection. As you are simplifying this, which I think makes sense, maybe we can get rid of `getUpdateInfo()` parameter as well and rather have `GetScheduledUpdate()`  interface implementation for different clients instead.

client_update.go
```diff
@@ -0,0 +1,183 @@
...
+func makeUpdateCheckRequest(server, deviceID string) (*http.Request, error) {
+	url := server + "/api/0.0.1/devices/" + deviceID + "/update"
+	req, err := http.NewRequest(http.MethodGet, url, nil)
+	if err != nil {
+		return nil, err
+	}
+	return req, nil
+}
+
+func makeUpdateFetchRequest(server string) (*http.Request, error) {
```
pasinskim: `server` as a name of argument is a bit misleading here. In fact this is link for downloading image.

config.go
```diff
@@ -0,0 +1,82 @@
...
+}
+
+func (c menderConfig) GetUpdaterConfig() httpsClientConfig {
+	return httpsClientConfig{
+		c.HttpsClient.Certificate,
+		c.HttpsClient.Key,
+		c.ServerCertificate,
+		c.ClientProtocol == "https",
+	}
+}
```
pasinskim: +1 for removing this from controller

daemon.go
```diff
@@ -14,156 +14,53 @@
...
-		stopChannel:            make(chan bool),
+		mender:      mender,
+		stopChannel: make(chan bool),
 	}
 	return &daemon
 }
 
-func (daemon menderDaemon) StopDaemon() {
-	daemon.stopChannel <- true
+func (d menderDaemon) StopDaemon() {
```
pasinskim: Are we using this anywhere right now? I've seen you rather are setting error and checking this in main daemon loop. If so maybe we can get rid of it and remove `stopChannel` from `menderDaemon` as well.

bboozzoo: I think that the original intention was to be able to react to SIGINT/SIGTERM, but I don't see `StopDaemon()` being called anywhere.

daemon.go
```diff
@@ -14,156 +14,53 @@
...
 	// figure out the state
 	for {
-		switch daemon.TransitionState() {
-		case MenderStateRunningWithFreshUpdate:
-			daemon.CommitUpdate()
-		case MenderStateWaitForUpdate:
-			err := daemon.waitForUpdate()
-			if err != nil {
-				return err
+		state, cancelled := d.mender.GetState().Handle(d.mender)
```
pasinskim: +1 for nice refactoring of daemon loop

daemon_test.go
```diff
@@ -14,13 +14,13 @@
 package main
 
 import (
-	"errors"
-	"fmt"
+	// "errors"
```
pasinskim: remove comments

daemon_test.go
```diff
@@ -68,112 +67,138 @@ func fakeProcessUpdate(response *http.Response) (interface{}, error) {
...
+
+// 	timespolled := 5
+// 	time.Sleep(time.Duration(timespolled) * pollInterval)
+// 	daemon.StopDaemon()
+
+// 	if reqHandlingCnt < (timespolled - 1) {
+// 		assert.Fail(t, "Expected to receive at least %v requests - %v received",
+// 			timespolled-1, reqHandlingCnt)
+// 	}
+// }
```
pasinskim: We need to do something with those tests.

dirstore.go
```diff
@@ -0,0 +1,125 @@
...
+// Once writing to temp file is done, the caller should run Commit() method of
+// the WriteCloserCommitter interface.
+func (d DirStore) OpenWrite(name string) (WriteCloserCommitter, error) {
+	f, err := os.OpenFile(d.getTempPath(name), os.O_WRONLY|os.O_CREATE, 0600)
+	if err != nil {
+		log.Errorf("I/O write error for entry %v: %v", name, err)
+		return nil, err
+	}
+
+	wrc := &DirFile{
```
pasinskim: Why not `wc := ...` :)

dirstore_test.go
```diff
@@ -0,0 +1,96 @@
...
+	indata, err = ioutil.ReadAll(in)
+	in.Close()
+
+	assert.NoError(t, err)
+	assert.Equal(t, []byte(data), indata)
+
+	// check writer
+	out, err := d.OpenWrite("bar")
+	assert.NoError(t, err)
+	// there should be a temp path already, but no the actual target path
```
pasinskim: but no ~~the~~ actual

error.go
```diff
@@ -0,0 +1,67 @@
...
+//    See the License for the specific language governing permissions and
+//    limitations under the License.
+package main
+
+import (
+	"github.com/pkg/errors"
+)
+
+// mender specific error
+type menderError interface {
```
pasinskim: Having this, should we start to use in all over the places instead of "standard" error handling or this is related only to state transitions?

bboozzoo: IMO we could use it everywhere. Though I would not do the switch right now, rather introduce the changes gradually as we work on the code.

fileseqnum.go
```diff
@@ -0,0 +1,69 @@
...
+func NewFileSeqnum(name string, store Store) *FileSeqnum {
+	return &FileSeqnum{
+		name:  name,
+		store: store,
+	}
+}
+
+// Obtain next sequence number. In case of errors (read, write, parse etc.)
+// returned value is 0 and error is returned.
+func (fs *FileSeqnum) Get() (uint64, error) {
```
pasinskim: Maybe we can name it so that the name will indicate that in fact we are increasing the number every time we are calling this function. What about `GetAndIncrement()` or something?

pasinskim: What is more, this seems to be quite inefficient to read and write file every time we want to increment this number. In fact we might need to do this quite frequently. I don't have any better solution for now though.

bboozzoo: We'll only read/write when performing authorization, what should be performed only so often.

dirstore.go
```diff
@@ -0,0 +1,125 @@
...
+func (d DirStore) ReadAll(name string) ([]byte, error) {
+	in, err := d.OpenRead(name)
+	if err != nil {
+		return nil, err
+	}
+	defer in.Close()
+	return ioutil.ReadAll(in)
+}
+
+func (d DirStore) WriteAll(name string, data []byte) error {
```
pasinskim: I can notice, that we are calling this one only here: https://github.com/mendersoftware/mender/pull/68/files#diff-3010f87e14cb0d76f9687480fb1f78a6R63.
Honestly, I don't see what we are gaining there comparing to "ordinary" write. If we will crash in the middle of writing the file and we will end up with `()*%#&*_` sequence this is not worse than not updating the file and sending duplicate request.

pasinskim: Oh, wait. In fact we won't send a request so next time we will increase seq number as expected. Ignore my comment above.

identity_data.go
```diff
@@ -0,0 +1,102 @@
...
+	"bufio"
+	"bytes"
+	"strings"
+
+	// "github.com/mendersoftware/log"
+	"github.com/pkg/errors"
+)
+
+const (
+	identityDataHelper = "/usr/bin/mender-identity-data"
```
pasinskim: Maybe `identityDataReaderScript` or something?

bboozzoo: IMO helper is better as it's more generic. This needn't be a script, can be a binary too.

identity_data.go
```diff
@@ -0,0 +1,102 @@
...
+}
+
+// device identity data content
+type IdentityData struct {
+	// device MAC address
+	MAC string `json:"mac"`
+}
+
+func parseIdentityData(data []byte) (interface{}, error) {
+	idata := IdentityData{}
```
pasinskim: I've been expecting something else here. I thought that `identityDataHelper` will return already formated JSON. Mender role will be only to opaque this data and send to server. It will be server, which is validating content and the fields. Client can only check if JSON is correct, but that's it. We shouldn't tight to something specific here, like MAC.

keystore.go
```diff
@@ -68,18 +75,20 @@ func (k *Keystore) Save(privPath string) error {
...
+	outf, err := k.store.OpenWrite(privPath)
 	if err != nil {
 		return err
 	}
-	defer outf.Close()
 
 	err = saveToPem(outf, k.private)
 	if err != nil {
 		log.Errorf("failed to save key: %s", err)
 		return err
```
pasinskim: Won't we end up with resource leak here? Why we are not calling `outf.Close()`?

bboozzoo: Right, `outf` leaks if `saveToPem` failed.

main.go
```diff
@@ -279,19 +290,26 @@ func doMain(args []string) error {
...
 	case *runOptions.daemon:
-		controler := NewMender(env)
-		if err := controler.LoadConfig(*runOptions.config); err != nil {
-			return err
+		updater, err := NewUpdateClient(config.GetUpdaterConfig())
+		if err != nil {
+			return errors.New("Cannot initialize daemon. Error instantiating updater. Exiting.")
+		}
+
+		controler := NewMender(*config, MenderPieces{
```
pasinskim: I don't like `MenderPieces` name.

main.go
```diff
@@ -262,9 +267,15 @@ func doMain(args []string) error {
...
 
+	config, err := LoadConfig(*runOptions.config)
+	if err != nil {
+		return err
+	}
+
 	// in any case we will need to have a device
 	env := NewEnvironment(new(osCalls))
 	device := NewDevice(env, new(osCalls), "/dev/mmcblk0p")
+	store := NewDirStore(*runOptions.dataStore)
```
pasinskim: Can we move this under https://github.com/mendersoftware/mender/pull/68/files#diff-7ddfb3e035b42cd70649cc33393fe32cR292. You don't need this except when you are running as a daemon.

main.go
bboozzoo: Care to suggest an alternative?

mender.go
```diff
@@ -15,28 +15,30 @@ package main
...
+	UInstallCommitRebooter
+	Updater
+	StateRunner
 }
 
 const (
 	defaultManifestFile = "/etc/build_mender"
-	defaultKeyFile      = "/data/mender/mender-agent.pem"
+	defaultKeyFile      = "mender-agent.pem"
+	defaultDataStore    = "/var/lib/mender"
```
pasinskim: This will need to be followed up by some change in meta-mender repo, won't it?

mender.go
```diff
@@ -56,49 +58,72 @@ type MenderState int
...
 	MenderStateError
+	// exit state
+	MenderStateDone
 )
 
 type mender struct {
-	BootEnvReadWriter
-	state          MenderState
-	config         menderFileConfig
+	UInstallCommitRebooter
```
pasinskim: Anonymous `MenderPieces` can be a part of mender instead of repeating all those pieces again here.

main.go
pasinskim: Not sure if this makes sense now, but maybe split into two: `MenderServerIface` and `MenderHWIface`? Then HW will be everything related to interactions with device and server will be communication part?

mender.go
```diff
@@ -15,28 +15,30 @@ package main
...
 	"time"
 
 	"github.com/mendersoftware/log"
+	"github.com/pkg/errors"
 )
 
 type Controler interface {
-	Bootstrap() error
-	TransitionState() MenderState
+	Bootstrap() menderError
```
pasinskim: Not sure if `Bootstrap()` should be a part of `Controller` interface. It feels more like client part.

mender.go
```diff
@@ -15,28 +15,30 @@ package main
...
 	GetCurrentImageID() string
-	GetDaemonConfig() daemonConfig
-	GetUpdaterConfig() httpsClientConfig
-	LastError() error
+	GetUpdatePollInterval() time.Duration
+	HasUpgrade() (bool, menderError)
+	CheckUpdate() (*UpdateResponse, menderError)
```
pasinskim: There is some inconsistency here. I would rename `HasUpgrade` to `HasUpdate`.

pasinskim: Actually `HasUpgrade()` and `CheckUpdate()` also belongs more to the client than controller.

mender.go
bboozzoo: For now we have agreed on some common location in this ticket https://tracker.mender.io/browse/MEN-441
Ideally one can think of having this (or `$(prefix)` at least) configurable at built time, similarly to what we do with version information.

identity_data.go
bboozzoo: The example implementation as done in https://tracker.mender.io/browse/MEN-417 provides a key=value list of attributes.

client_auth.go
bboozzoo: Device authentication service API spec states that base path is `/v0.1`. In my opinion there's a larger problem with how we export the APIs outside of mender system in general. Let me write it down in an email.

state.go
```diff
@@ -0,0 +1,308 @@
...
+	"time"
+
+	"github.com/mendersoftware/log"
+	"github.com/pkg/errors"
+)
+
+type State interface {
+	// Perform state action, returns next state and boolean flag indicating if
+	// execution was cancelled or not
+	Handle(c Controler) (State, bool)
```
pasinskim: Maybe `Execute()`?

state.go
```diff
@@ -0,0 +1,308 @@
...
+	// Cancel state action, returns true if action was cancelled
+	Cancel() bool
+	// Return numeric state ID
+	Id() MenderState
```
pasinskim: I would probably change it to `State()` or `Get()`.

state.go
```diff
@@ -0,0 +1,308 @@
...
+	doneState = &FinalState{
+		BaseState{
+			id: MenderStateDone,
+		},
+	}
+)
+
+// Helper base state with some convenience methods
+type BaseState struct {
+	id MenderState
```
pasinskim: This is more internal thing, so it doesn't matter that much, but maybe we can have `state` here as well?

state.go
pasinskim: I am also not convinced with changing the state inside of the state handler. What do you think @kacf? If you both think this is better solution I am OK with it, but I would rather do this inside controller.

state.go
```diff
@@ -0,0 +1,308 @@
...
+	select {
+	case <-ticker.C:
+		log.Debugf("poll interval complete")
+		return updateCheckState, false
+	case <-u.cancel:
+		log.Infof("update wait canceled")
+	}
+
+	return u, true
+}
```
pasinskim: Definitely +1 for nice and clear solution with states and handlers.

state.go
```diff
@@ -0,0 +1,308 @@
...
+	}
+	return doneState, false
+}
+
+type FinalState struct {
+	BaseState
+}
+
+func (f *FinalState) Handle(c Controler) (State, bool) {
+	panic("reached final state")
```
pasinskim: I am not a fan of this one. I would rather return the statue up the stack and decide what to do elsewhere.

client_auth.go
pasinskim: Agree. Probably API documentation is broken as in all other places we are having `0.0.1`.

mender.go
bboozzoo: I disagree. It's an interface of the controller that can then be consumed by someone else (states in this case). If I move that to a client of some sort, then the code that drives the controller will have to interact with that client.

identity_data.go
pasinskim: Still I think it's wrong. Now, when you will need to change some identification data, you will have to change client code and recompile. Having the script returning JSON we will just need to verify if this is correct JSON and then send it to the server. Then it is the server validating the data and deciding if it is correct or not.

state.go
bboozzoo: `StateId()` maybe? It's already state everywhere, another `State()` method will make it a bit confusing.

state.go
bboozzoo: It's not really changing the state, it's only returning a new state. The state implements a strategy that works with the controller to do stuff. In fact it's a know design pattern - aka. state design pattern. Although, I did not implement a separate controller method that executes the state action (instead we just `GetState().Handle(..)` in daemon).

state.go
bboozzoo: That's the point. This is the final state, you should not execute as we're technically done. See what daemon code with it: https://github.com/mendersoftware/mender/pull/68/files/2139bf44a7ba4e67dd724ab7d5bd94e2803ef21c#diff-e7bcc6178f83a5063b3c27cc5a860780R55

identity_data.go
bboozzoo: Let's put it in backlog and address it later on.

mender.go
pasinskim: Doesn't it have to anyway? And in fact I see this bit differently. I think that controller should be main point where you are controlling the states and it should delegate the work to all needed sub-modules instead of `states` implementing controller interface.
The drawback of leaving this as is looks to be the fact that this interface will grow a lot once we will add new functionalities (like reporting, inventory and so on).
But this is just different view and something we should decide somewhere else (aka some meeting where we will go through architecture and discuss it further). I am OK with leaving this as is.

pasinskim:
Overall looks nice and clean. I like lot of the things you did with refactoring. The biggest things I don't agree with are:
- the way we are collecting, validating and sending identity data (we can have some follow up later on so I am OK with merging this as is)
- the architecture of `states` (again something which is not set in stone and possible to change later)
- we need to fix client tests which are commented now

The rest is just some typos, and names of the interfaces or functions.

client_test.go
bboozzoo: fixed

client_test.go
bboozzoo: fixed

client.go
bboozzoo: Ok, I'll leave it as it is now as we'll see how it works out.

client_auth.go
bboozzoo: renamed to `AuthRequester`

client_auth.go
bboozzoo: Actually both the update client and auth client have different interfaces, not sure if wrapping that up into a single client will help. However, what might help is sharing a single `http.Client`. This should allow us to benefit from pipelining as the client is supposed to keep the connections open.

client_auth.go
bboozzoo: Updated to send requests to `/api/0.0.1/authorization/auth_requests`. This will not work with current API gateway setup (actually the previous one didn't work either :) ).

client_auth.go
bboozzoo: fixed

client_auth_test.go
bboozzoo: fixed

client_auth_test.go
bboozzoo: fixed

client_update.go
bboozzoo: fixed

daemon_test.go
bboozzoo: fixed

daemon_test.go
bboozzoo: Fixed. Removed most of those, reworked the last one to check if actual mender state sequence was run.

dirstore.go
bboozzoo: I'll skip this one, we can rename it later on :)

dirstore_test.go
bboozzoo: rephrased the comment

keystore.go
bboozzoo: fixed

main.go
bboozzoo: fixed

mender.go
bboozzoo: Might be that we have different views on what Controller controls. I originally assumed that the controller will be driving the update check/download/install/apply/reboot steps, becoming a device/update controller of sorts.

For instance, I expect the reporting part to be separate from device controller, with only some loose coupling through a channel or callbacks. Similarly the inventory piece, it's kind of outside of the update cycle.

The central point where all of these come together would be the daemon code.

mender.go
bboozzoo: `UInstallCommitRebooter` and `Updater` are part of mender as they're also a part of `Controler` interface. `BootEnvReadWriter` is referenced in `env` field. We don't really keep any reference to store aside from passing it further to `Keystore`. I don't think that directly embedding `MenderPieces` provides any gain here aside from duplication as we'd have to explicitly initialize the embedded interfaces anyway.

state.go
bboozzoo: On second thought, `State.StateId()` is repeating `State` word too much. How about we just keep `Id()` or rename it to `GetId()` (though that could be equally well named `Id()` :) )

bboozzoo:
I'll reoder the patches, move authorization ones to `bboozzoo/bootstrap-states-with-auth` branch and then drop them so that `deadcode` checks don't fail.

bboozzoo:
Seems like I'll need to rebase as well.

bboozzoo:
```
--- FAIL: TestLicenses (1.19s)
	licenses.go:53: exit status 1
	licenses.go:54: LICENSE: OK
		vendor/github.com/mendersoftware/mendertesting/LICENSE: OK
		vendor/github.com/mendersoftware/log/LICENSE: OK
		vendor/github.com/mendersoftware/log/COPYING: OK
		vendor/github.com/mendersoftware/scopestack/LICENSE: OK
		vendor/github.com/mendersoftware/scopestack/COPYING: OK
		vendor/github.com/Sirupsen/logrus/LICENSE: OK
		vendor/github.com/davecgh/go-spew/LICENSE: OK
		vendor/github.com/stretchr/testify/LICENSE: OK
		vendor/github.com/stretchr/testify/LICENCE.txt: OK
		vendor/github.com/pmezard/go-difflib/LICENSE: OK
		vendor/golang.org/x/sys/LICENSE: OK
		sha256sum: WARNING: 2 lines are improperly formatted
		vendor/github.com/pkg/errors/LICENSE has missing or wrong entry in LIC_FILES_CHKSUM.sha256
```

What is that??

bboozzoo:
@kacf @pasinskim ping

LIC_FILES_CHKSUM.sha256
```diff
@@ -4,6 +4,7 @@ bbb303820971c294a9a8e5eba5affcf1379036e877ea61c11cbf9400b2949483  vendor/github.
 bbb303820971c294a9a8e5eba5affcf1379036e877ea61c11cbf9400b2949483  vendor/github.com/mendersoftware/log/COPYING
 3591f687e2d6f49c83b1ec69577e8110afbde80be5ec81791bd86d2838ccd3de  vendor/github.com/mendersoftware/scopestack/LICENSE
 bbb303820971c294a9a8e5eba5affcf1379036e877ea61c11cbf9400b2949483  vendor/github.com/mendersoftware/scopestack/COPYING
+192def682f6846c6a57e20cf044e224fa467948358163a18dd37bfe14578e6ab  vendor/github.com/pkg/errors/LICENSE
```
kacf: You can put this under a headline `# These are identical to BSD 2 Clause license.`, since that's what it is.

kacf:
Oh sorry, perhaps I wasn't clear: The BSD-2-Clause applies only to the new file you added. Everything in mendertesting is Apache-2.0. Feel free to add a header for that too though.

bboozzoo:
Licenses should be fixed now.

bboozzoo:
@pasinskim ping

pasinskim:
I think it looks nice. Assuming some follow-ups on the stuff mentioned here we can merge it. Please open appropriate JIRA tickets before merging, so that we will keep track of the things missing.

bboozzoo:
The authorization pieces were pulled from this series, so I've added tickets for issues that were part of this PR:

- https://tracker.mender.io/browse/MEN-473
- https://tracker.mender.io/browse/MEN-474

PR #69, opened 2016-06-20
MEN-341: Switch to "mender_boot_part" U-Boot environment variable.

kacf:
The intention is to avoid conflicts with similarly named variables
being used in various U-Boot implementations. It's also the variable
being used by the new generic U-Boot patch for Mender in meta-mender.


kacf:
This goes together with https://github.com/mendersoftware/meta-mender/pull/63.

bboozzoo:
+1

PR #70, opened 2016-06-20
Authorization workflow

bboozzoo:
The series of patches implemens a workflow for client authorization.

The first 5 patches are mostly cleanups/refactoring of things/helpers that were spotted along the way.

Device identity data is obtained using the helper script `support/mender-device-identity`. Right now the script outputs a per line `key=value` of identity data. This is parsed and collected into a single structure. @pasinskim suggested refactoring the script to just produce JSON output without the need for further parsing.

Seqnum is a sequence generator with storage backing. This is needed for authorization requests.

An authorization client is implemented by `AuthManager` interface, with `AuthDataMessenger` being a helper interface that wraps producing/consuming of authorization requests. Tried to make the interface useful for other transports than HTTP as well. 

@pasinskim @kacf

### DIFF
--- a/auth.go
+++ b/auth.go
@@ -1,0 +1,214 @@
+// Copyright 2016 Mender Software AS
+//
+//    Licensed under the Apache License, Version 2.0 (the "License");
+//    you may not use this file except in compliance with the License.
+//    You may obtain a copy of the License at
+//
+//        http://www.apache.org/licenses/LICENSE-2.0
+//
+//    Unless required by applicable law or agreed to in writing, software
+//    distributed under the License is distributed on an "AS IS" BASIS,
+//    WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+//    See the License for the specific language governing permissions and
+//    limitations under the License.
+package main
+
+import (
+	"bytes"
+	"encoding/json"
+	"os"
+
+	"github.com/mendersoftware/log"
+	"github.com/pkg/errors"
+)
+
+type AuthCode string
+
+type AuthReqData struct {
+	IdData      string `json:"id_data"`
+	TenantToken string `json:"tenant_token"`
+	Pubkey      string `json:"pubkey"`
+	SeqNumber   uint64 `json:"seq_no"`
+}
+
+type AuthRequest struct {
+	// request message data
+	Data []byte
+	// authorization code
+	Code AuthCode
+	// request signature
+	Signature []byte
+}
+
+// handler for authorization message data
+type AuthDataMessenger interface {
+	// build authorization request data, returns auth request or an error
+	MakeAuthRequest() (*AuthRequest, error)
+	// receive authoiriation data response, returns error if response is invalid
+	RecvAuthResponse([]byte) error
+}
+
+type AuthManager interface {
+	// returns true if authorization data is current and valid
+	IsAuthorized() bool
+	// returns device's authoirization code
+	AuthCode() (AuthCode, error)
+	// check if device key is available
+	HasKey() bool
+	// generate device key (will overwrite an already existing key)
+	GenerateKey() error
+
+	AuthDataMessenger
+}
+
+const (
+	authTokenName       = "authtoken"
+	authTenantTokenName = "authtentoken"
+	authSeqName         = "authseq"
+
+	noAuthCode = AuthCode("")
+)
+
+type MenderAuthManager struct {
+	store    Store
+	keyStore *Keystore
+	keyName  string
+	idSrc    IdentityDataGetter
+	seqNum   SeqnumGetter
+}
+
+func NewAuthManager(store Store, keyName string, idSrc IdentityDataGetter) AuthManager {
+	ks := NewKeystore(store)
+	if ks == nil {
+		return nil
+	}
+
+	mgr := &MenderAuthManager{
+		store:    store,
+		keyStore: ks,
+		keyName:  keyName,
+		idSrc:    idSrc,
+		seqNum:   NewFileSeqnum(authSeqName, store),
+	}
+
+	if err := ks.Load(keyName); err != nil && !IsNoKeys(err) {
+		log.Errorf("failed to load device keys from %v: %v", keyName, err)
+		return nil
+	}
+
+	return mgr
+}
+
+func (m *MenderAuthManager) IsAuthorized() bool {
+	adata, err := m.AuthCode()
+	if err != nil {
+		return false
+	}
+
+	if adata == noAuthCode {
+		return false
+	}
+
+	// TODO check if JWT is valid?
+
+	return true
+}
+
+func (m *MenderAuthManager) MakeAuthRequest() (*AuthRequest, error) {
+
+	var err error
+	authd := AuthReqData{}
+
+	idata, err := m.idSrc.Get()
+	if err != nil {
+		return nil, errors.Wrapf(err, "failed to obtain identity data")
+	}
+
+	authd.IdData = idata
+
+	// fill device public key
+	authd.Pubkey, err = m.keyStore.PublicPEM()
+	if err != nil {
+		return nil, errors.Wrapf(err, "failed to obtain device public key")
+	}
+
+	tentok, err := m.store.ReadAll(authTenantTokenName)
+	if err != nil {
+		return nil, errors.Wrapf(err, "failed to read tenant token")
+	}
+	log.Debugf("tenant token: %s", tentok)
+
+	// fill tenant token
+	authd.TenantToken = string(tentok)
+
+	// fetch sequence number
+	num, err := m.seqNum.Get()
+	if err != nil {
+		return nil, errors.Wrapf(err, "failed to obtain sequence number")
+	}
+	authd.SeqNumber = num
+
+	log.Debugf("authorization data: %v", authd)
+
+	databuf := &bytes.Buffer{}
+	enc := json.NewEncoder(databuf)
+
+	err = enc.Encode(&authd)
+	if err != nil {
+		return nil, errors.Wrapf(err, "failed to encode auth request")
+	}
+
+	reqdata := databuf.Bytes()
+
+	// generate signature
+	sig, err := m.keyStore.Sign(reqdata)
+	if err != nil {
+		return nil, errors.Wrapf(err, "failed to sign auth request")
+	}
+
+	return &AuthRequest{
+		Data:      reqdata,
+		Code:      AuthCode(tentok),
+		Signature: sig,
+	}, nil
+}
+
+func (m *MenderAuthManager) RecvAuthResponse(data []byte) error {
+	if len(data) == 0 {
+		return errors.New("empty auth response data")
+	}
+
+	if err := m.store.WriteAll(authTokenName, data); err != nil {
+		return errors.Wrapf(err, "failed to save auth token")
+	}
+	return nil
+}
+
+func (m *MenderAuthManager) AuthCode() (AuthCode, error) {
+	data, err := m.store.ReadAll(authTokenName)
+	if err != nil {
+		if os.IsNotExist(err) {
+			return noAuthCode, nil
+		}
+		return noAuthCode, errors.Wrapf(err, "failed to read auth token data")
+	}
+
+	return AuthCode(data), nil
+}
+
+func (m *MenderAuthManager) HasKey() bool {
+	return m.keyStore.Private() != nil
+}
+
+func (m *MenderAuthManager) GenerateKey() error {
+	if err := m.keyStore.Generate(); err != nil {
+		log.Errorf("failed to generate device key: %v", err)
+		return errors.Wrapf(err, "failed to generate device key")
+	}
+
+	if err := m.keyStore.Save(m.keyName); err != nil {
+		log.Errorf("failed to save keys to %s: %s", m.keyName, err)
+		return NewFatalError(err)
+	}
+	return nil
+}
--- a/auth_test.go
+++ b/auth_test.go
@@ -1,0 +1,134 @@
+// Copyright 2016 Mender Software AS
+//
+//    Licensed under the Apache License, Version 2.0 (the "License");
+//    you may not use this file except in compliance with the License.
+//    You may obtain a copy of the License at
+//
+//        http://www.apache.org/licenses/LICENSE-2.0
+//
+//    Unless required by applicable law or agreed to in writing, software
+//    distributed under the License is distributed on an "AS IS" BASIS,
+//    WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+//    See the License for the specific language governing permissions and
+//    limitations under the License.
+package main
+
+import (
+	"encoding/json"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+)
+
+func TestAuthManager(t *testing.T) {
+	ms := NewMemStore()
+
+	cmdr := newTestOSCalls("", 0)
+	am := NewAuthManager(ms, "devkey", IdentityDataRunner{
+		cmdr: &cmdr,
+	})
+	assert.IsType(t, &MenderAuthManager{}, am)
+
+	assert.False(t, am.HasKey())
+	assert.NoError(t, am.GenerateKey())
+	assert.True(t, am.HasKey())
+
+	assert.False(t, am.IsAuthorized())
+
+	code, err := am.AuthCode()
+	assert.Equal(t, noAuthCode, code)
+	assert.NoError(t, err)
+
+	ms.WriteAll(authTokenName, []byte("footoken"))
+	// disable store access
+	ms.Disable(true)
+	code, err = am.AuthCode()
+	assert.Error(t, err)
+	ms.Disable(false)
+
+	code, err = am.AuthCode()
+	assert.Equal(t, AuthCode("footoken"), code)
+	assert.NoError(t, err)
+}
+
+func TestAuthManagerRequest(t *testing.T) {
+	ms := NewMemStore()
+
+	var err error
+
+	badcmdr := newTestOSCalls("mac=foobar", -1)
+	am := NewAuthManager(ms, "devkey", IdentityDataRunner{
+		cmdr: &badcmdr,
+	})
+	_, err = am.MakeAuthRequest()
+	assert.Error(t, err, "should fail, cannot obtain identity data")
+	assert.Contains(t, err.Error(), "identity data")
+
+	cmdr := newTestOSCalls("mac=foobar", 0)
+	am = NewAuthManager(ms, "devkey", IdentityDataRunner{
+		cmdr: &cmdr,
+	})
+	_, err = am.MakeAuthRequest()
+	assert.Error(t, err, "should fail, no device keys are present")
+	assert.Contains(t, err.Error(), "device public key")
+
+	// generate key first
+	assert.NoError(t, am.GenerateKey())
+
+	_, err = am.MakeAuthRequest()
+	assert.Error(t, err, "should fail, no tenant token present")
+	assert.Contains(t, err.Error(), "tenant token")
+
+	// setup tenant token
+	ms.WriteAll(authTenantTokenName, []byte("tenant"))
+	// setup sequence number
+	ms.WriteAll(authSeqName, []byte("12"))
+
+	req, err := am.MakeAuthRequest()
+	assert.NoError(t, err)
+	assert.NotEmpty(t, req.Data)
+	assert.Equal(t, AuthCode("tenant"), req.Code)
+	assert.NotEmpty(t, req.Signature)
+
+	var ard AuthReqData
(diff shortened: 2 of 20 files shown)


auth.go
```diff
@@ -0,0 +1,218 @@
...
+	// authorization code
+	Code AuthCode
+	// request signature
+	Signature []byte
+}
+
+// handler for authorization message data
+type AuthDataMessenger interface {
+	// build authorization request data, returns auth request and authorization
+	// code for the request
```
pasinskim: Looks like it returns error, not authorization code?

auth.go
```diff
@@ -0,0 +1,218 @@
...
+	MakeAuthRequest() (*AuthRequest, error)
+	// receive authoiriation data response, returns error if response is invalid
+	RecvAuthResponse([]byte) error
+}
+
+type AuthManager interface {
+	// returns true if authorization data is current and valid
+	IsValid() bool
+	// return device's authoirization code
+	AuthCode() (AuthCode, error)
```
pasinskim: s/return/returns/

identity_data.go
```diff
@@ -0,0 +1,111 @@
...
+	for in.Scan() {
+		line := in.Text()
+
+		if len(line) == 0 {
+			continue
+		}
+
+		val := strings.Split(line, "=")
+
+		if len(val) != 2 {
```
kacf: Is there a chance that the identity data itself could contain `=`? I think it is a valid character in some keys. Maybe you should check `len(val) >= 2` instead?

kacf: Or even better, keep the existing check, but use `SplitN`, like this: `val := strings.SplitN(line, "=", 2)`. Then you'll always get the right number.

identity_data.go
```diff
@@ -0,0 +1,111 @@
...
+			return nil, errors.Errorf("incorrect line '%s'", line)
+		}
+
+		switch val[0] {
+		case "mac":
+			idata.MAC = val[1]
+		default:
+			return nil, errors.Errorf("unexpected token %s in line %s", val[0], line)
```
kacf: Maybe I'm misunderstanding something, but why does `mac` have a special meaning here? I thought we were just letting identity be a list of key/value pairs, why do we care if it's the MAC?

auth.go
```diff
@@ -0,0 +1,218 @@
...
+
+	mgr := &MenderAuthManager{
+		store:    store,
+		keystore: ks,
+		keyname:  keyname,
+		idsrc:    idsrc,
+		seqnum:   NewFileSeqnum(authSeqName, store),
+	}
+
+	if err := ks.Load(keyname); err != nil && IsNoKeys(err) == false {
```
pasinskim: Do we need `err != nil` here? I think `!IsNoKeys(err)` should do the job.

auth.go
```diff
@@ -0,0 +1,218 @@
...
+		log.Errorf("failed to load device keys: %s", err)
```
pasinskim: I would add `keyname` to the error message.

auth.go
```diff
@@ -0,0 +1,218 @@
...
+
+type MenderAuthManager struct {
+	store    Store
+	keystore *Keystore
+	keyname  string
+	idsrc    IdentityDataGetter
+	seqnum   SeqnumGetter
+}
+
+func NewAuthManager(store Store, keyname string, idsrc IdentityDataGetter) AuthManager {
```
pasinskim: s/keyname/keyName/

pasinskim: s/idsrc/idSrc/

identity_data.go
bboozzoo: Right, perhaps it should not matter at all, `map[string]string` should be enough.

auth.go
```diff
@@ -0,0 +1,218 @@
...
+	authTokenName       = "authtoken"
+	authTenantTokenName = "authtentoken"
+	authSeqName         = "authseq"
+
+	noAuthCode = AuthCode("")
+)
+
+type MenderAuthManager struct {
+	store    Store
+	keystore *Keystore
```
pasinskim: s/keystore/keyStore/

auth.go
```diff
@@ -0,0 +1,218 @@
...
+	keyname  string
```
pasinskim: s/keyname/keyName/

auth.go
```diff
@@ -0,0 +1,218 @@
...
+	idsrc    IdentityDataGetter
```
pasinskim: s/idsrc/idSrc/

auth.go
```diff
@@ -0,0 +1,218 @@
...
+	seqnum   SeqnumGetter
```
pasinskim: s/seqnum/seqNum/

auth.go
```diff
@@ -0,0 +1,218 @@
...
+	idata, err := m.idsrc.Get()
+	if err != nil {
+		return nil, errors.Wrapf(err, "failed to obtain identity data")
+	}
+
+	authd.IdData = idata
+
+	// fill device public key
+	authd.Pubkey = m.keystore.PublicPEM()
+	if err != nil {
```
pasinskim: I can not see `err` being used after calling `m.idsrc.Get()`, so probably here we are checking the old error value.

auth.go
```diff
@@ -0,0 +1,218 @@
...
+		return nil, errors.Wrapf(err, "failed to obtain device public key")
+	}
+
+	tentok, err := m.store.ReadAll(authTenantTokenName)
```
pasinskim: I am usually not complaining about this for local variables, but here I don't know why I always can see scout camp (and tent OK).

FYI
```
Finally, the convention in Go is to use MixedCaps or mixedCaps rather than underscores to write multiword names.
```

identity_data.go
bboozzoo: fixed

auth.go
bboozzoo: fixed

auth.go
bboozzoo: fixed

identity_data.go
bboozzoo: Fixed. Changed to `type IdentityData map[string]string`

auth.go
bboozzoo: fixed

auth.go
```diff
@@ -0,0 +1,218 @@
...
+	}
+
+	return AuthCode(data), nil
+}
+
+func (m *MenderAuthManager) HasKey() bool {
+	if m.keystore.Private() != nil {
+		return true
+	}
+	return false
```
pasinskim: Note: could be simplified to `return m.keystore.Private() != nil`.

auth.go
```diff
@@ -0,0 +1,218 @@
...
+}
+
+func (m *MenderAuthManager) GenerateKey() error {
+	if err := m.keystore.Generate(); err != nil {
+		log.Errorf("failed to generate device key: %v", err)
+		return errors.Wrapf(err, "failed to generate device key")
```
pasinskim: I think we need to agree on some convention here. In more places we are rather just returning error without logging anything. 
From the other hand maybe there is nothing wrong with logging the error twice.

auth.go
bboozzoo: fixed

auth.go
bboozzoo: fixed

auth_test.go
```diff
@@ -0,0 +1,91 @@
...
+	ms.ReadOnly(true)
+	err = am.RecvAuthResponse([]byte("fooresp"))
+	assert.Error(t, err)
+
+	ms.ReadOnly(false)
+	err = am.RecvAuthResponse([]byte("fooresp"))
+	tokdata, err := ms.ReadAll(authTokenName)
+	assert.NoError(t, err)
+	assert.Equal(t, []byte("fooresp"), tokdata)
+}
```
pasinskim: I would split it into multiple tests. This is quite a big test and could be hard to figure out what is broken once test will fail.

client_auth.go
```diff
@@ -0,0 +1,92 @@
...
+	Request(server string, datasrc AuthDataMessenger) ([]byte, error)
+}
+
+type AuthClient struct {
+	client *http.Client
+}
+
+func NewAuthClient(conf httpsClientConfig) (*AuthClient, error) {
+	client, err := NewHttpClient(conf)
+	if client == nil {
```
pasinskim: Should we check the `err` here as well, to be on a safe side?

client_auth.go
```diff
@@ -0,0 +1,92 @@
...
+		return nil, errors.Wrap(err, "failed to create auth client HTTP client")
+	}
+
+	au := AuthClient{
+		client: client,
+	}
+	return &au, nil
+}
+
+func (u *AuthClient) Request(server string, datasrc AuthDataMessenger) ([]byte, error) {
```
pasinskim: Note: `u` matches nothing. Maybe `ac` or something?

pasinskim: s/datasrc/dataSrc/

client_auth.go
```diff
@@ -0,0 +1,92 @@
...
+	defer rsp.Body.Close()
+
+	log.Debugf("receive response data")
+	data, err := ioutil.ReadAll(rsp.Body)
+	if err != nil {
+		return nil, errors.Wrapf(err, "failed to receive response data")
+	}
+
+	// validate response
+	log.Debugf("received response data %v", data)
```
pasinskim: The comment is bit misleading here I think, unless you mean to validate response manually by reading the logs...

client_auth.go
```diff
@@ -0,0 +1,92 @@
...
+
+	return data, nil
+}
+
+func makeAuthRequest(server string, datasrc AuthDataMessenger) (*http.Request, error) {
```
pasinskim: s/datasrc/dataSrc/

client_auth.go
```diff
@@ -0,0 +1,92 @@
...
+	url := buildURL(server) + "/api/0.0.1/authorization/auth_requests"
```
pasinskim: Perhaps this is something we shouldn't fix now, but I don't think that `/api/0.0.1` should be added "by-hand" here. Maybe we can have some small function similar to `buildUrl()`.

In fact, maybe `buildUrl()` can take more arguments, like `server` and `requestType` and return whole URL with version?

client_auth.go
```diff
@@ -0,0 +1,92 @@
...
+
+	var err error
```
pasinskim: I don't think you need this one here.

client_auth.go
bboozzoo: Bit excessive, the client will not support more than one version of the API. We could have for instance a common api prefix somewhere, as in
```
const apiPrefix = "/api/0.0.1"
```
and use that in functions that build URLs

client_auth.go
```diff
@@ -0,0 +1,92 @@
...
+
+	req, err := makeAuthRequest(server, datasrc)
+	if err != nil {
+		return nil, errors.Wrapf(err, "failed to build auth request")
+	}
+
+	log.Debugf("making authorization request to server %s with req: %s", server, req)
+	rsp, err := u.client.Do(req)
+	if err != nil {
+		return nil, errors.Wrapf(err, "failed to execute request")
```
pasinskim: Maybe `"failed to execute authorization request"`?

client_auth.go
```diff
@@ -0,0 +1,92 @@
...
+	}
+	defer rsp.Body.Close()
+
+	log.Debugf("receive response data")
+	data, err := ioutil.ReadAll(rsp.Body)
+	if err != nil {
+		return nil, errors.Wrapf(err, "failed to receive response data")
```
pasinskim: What about `"failed to receive authorization response data"`?

client_auth.go
```diff
@@ -0,0 +1,92 @@
...
+
+	req, err := datasrc.MakeAuthRequest()
+	if err != nil {
+		return nil, errors.Wrapf(err, "failed to obtain authorization message data")
+	}
+
+	dataio := bytes.NewBuffer(req.Data)
+	hreq, err := http.NewRequest(http.MethodGet, url, dataio)
+	if err != nil {
+		return nil, errors.Wrapf(err, "failed to create HTTP request")
```
pasinskim: `"failed to create authorization HTTP request"`?

auth.go
bboozzoo: Actually will not fix, `IsNoKeys(nil)` is also true, but I'm interested in a case when there was an error, but other than `ErrNoKeys`

config.go
```diff
@@ -72,7 +72,7 @@ func readConfigFile(config interface{}, fileName string) error {
 	return nil
 }
 
-func (c menderConfig) GetUpdaterConfig() httpsClientConfig {
+func (c menderConfig) GetHttpConfig() httpsClientConfig {
```
pasinskim: In fact we are returning `httpsClientConfig` here, some maybe at least `GetHttpsClientConfig()`?

client_auth.go
bboozzoo: fixed

client_auth.go
bboozzoo: fixed

client_auth.go
bboozzoo: fixed

client_auth.go
bboozzoo: Yeah, the comment was left behind by mistake.

client_auth.go
bboozzoo: fixed

config.go
bboozzoo: I would argue that `GetHttpConfig` is more natural. Changing it to `GetHttpsConfig` and returning a configuration without SSL related info would look awkward.

fileseqnum.go
```diff
@@ -0,0 +1,69 @@
...
+// Obtain next sequence number. In case of errors (read, write, parse etc.)
+// returned value is 0 and error is returned.
+func (fs *FileSeqnum) Get() (uint64, error) {
+
+	d, err := fs.store.ReadAll(fs.name)
+	if err != nil && !os.IsNotExist(err) {
+		return 0, errors.Wrapf(err, "seqnum data read failed")
+	}
+
+	newval := uint64(1)
```
pasinskim: Maybe `newval = SeqnumStartVal`?

client_auth.go
bboozzoo: fixed

client_auth.go
bboozzoo: fixed

fileseqnum.go
```diff
@@ -0,0 +1,69 @@
...
+
+	if !os.IsNotExist(err) {
+		v, err := strconv.ParseUint(string(d), 10, 64)
+		if err != nil {
+			return 0, errors.Wrapf(err, "seqnum data parse failed")
+		}
+
+		// check for overflow
+		if math.MaxUint64 == v {
+			newval = SeqnumStartVal
```
pasinskim: Let's just make sure to document this somewhere and that server will support this as well. What is more, this seems to be a bit tricky and vulnerable. Depending on how server-side will be implemented, if we will send `SeqnumStartVal` after any previously sent number it should be accepted since server couldn't know what client-side `math.MaxUint64`is.

Any thoughts?

bboozzoo: In fact we're doing this only to have a piece of message that is different in each time a request is sent. The server does not really care if the sequence starts from 1 or any other value, just that it's different each time and monotonically increasing.

auth.go
bboozzoo: Fixed, `PublicPEM()` should return an error (though IMO it's not really possible for this to happen).

auth.go
bboozzoo: Heh, makes my head go up and down :)

auth.go
bboozzoo: fixed

auth.go
bboozzoo: Ideally we should propagate the error up to a place that it has enough context to be meaningful. In practice, people either log everything or virtually nothing. Perhaps feedback from testing could be useful here.

main.go
```diff
@@ -258,6 +258,42 @@ func ShowVersion() {
 	os.Stdout.Write([]byte(v))
 }
 
+func doDaemon(config *menderConfig, dev *device, env *uBootEnv, opts *runOptionsType) error {
```
pasinskim: I would call this one `initializeDaemon()` and `return NewDaemon(controller)`. Then in `main()` we would have:

```
daemon, err := initializeDaemon()
if err != nil {
  return daemon.Run()
}
```

Seems more descriptive than bit cryptic `doDaemon()`. What do you think?

mender.go
```diff
@@ -90,39 +95,32 @@ type mender struct {
...
+	authreq        AuthRequester
+	authmgr        AuthManager
 }
 
 type MenderPieces struct {
 	updater Updater
 	device  UInstallCommitRebooter
 	env     BootEnvReadWriter
 	store   Store
+	authmgr AuthManager
```
pasinskim: s/authmgr/authMgr/

mender.go
```diff
@@ -90,39 +95,32 @@ type mender struct {
...
+	authreq AuthRequester
```
pasinskim: s/authreq/authReq/

auth.go
```diff
@@ -0,0 +1,218 @@
...
+
+	if err := ks.Load(keyname); err != nil && IsNoKeys(err) == false {
+		log.Errorf("failed to load device keys: %s", err)
+		return nil
+	}
+
+	return mgr
+}
+
+func (m *MenderAuthManager) IsValid() bool {
```
pasinskim: Maybe we can change to `IsAuthorized()` to have something more descriptive?

mender.go
```diff
@@ -196,18 +194,34 @@ func (m *mender) Bootstrap() menderError {
...
+		return NewTransientError(errors.Wrap(err, "authorization request failed"))
+	}
+
+	err = m.authmgr.RecvAuthResponse(rsp)
+	if err != nil {
+		return NewTransientError(errors.Wrap(err, "failed to parse authorization response"))
+	}
+
+	log.Info("successfuly received new authorization data")
+
```
pasinskim: Shouldn't we validate the response?

state.go
```diff
@@ -92,6 +100,37 @@ func (b *BaseState) Cancel() bool {
...
+
+func NewCancellableState(base BaseState) CancellableState {
+	return CancellableState{
+		base,
+		make(chan bool),
+	}
+}
+
+func (cs *CancellableState) StateAfterWait(next, same State, wait time.Duration) (State, bool) {
+	ticker := time.NewTicker(wait)
```
pasinskim: Quick search shows no calls to `ticker.Stop()` in the code. Is this expected?

state.go
```diff
@@ -92,6 +100,37 @@ func (b *BaseState) Cancel() bool {
...
 
+type CancellableState struct {
+	BaseState
+	cancel chan bool
+}
+
+func NewCancellableState(base BaseState) CancellableState {
+	return CancellableState{
+		base,
+		make(chan bool),
```
pasinskim: It seems that we are never closing this channel.

config.go
pasinskim: Makes sense.

fileseqnum.go
pasinskim: My point is not about what is the initial sequence number, but what is happening when we are having overflow. Since server does not know how big your bugger is it needs to accept `SeqnumStartVal` to be a valid sequence number after no matter what previous number was. This is leaving it vulnerable for all packets with `SeqnumStartVal` being used as seq number, doesn't it?

fileseqnum.go
bboozzoo: fixed

fileseqnum.go
bboozzoo: Seems like this needs to be brought up with the backend team, best if you talk directly with @mchalski @kjaskiewiczz @maciejmrowiec. Right now they expect a number and once the counter overflows it obviously wrap and start from `SeqnumStartVal`. It might be better though to use a random number, but that would obviously require server to keep track of the numbers it has seen to shomehow exclude replays. I'm quite sure, we won't resolve it here anyway.

main.go
bboozzoo: Fair enough. Renamed to `initDaemon`.

mender.go
bboozzoo: No, `authmgr` does that.

state.go
bboozzoo: fixed

state.go
bboozzoo: Added `CancellableState.Stop()`.

client_auth.go
bboozzoo: Added a helper `buildApiURL(server, url string) string` that the trick

fileseqnum.go
pasinskim: Maybe you are right, but submitting code which is broken by design to fix it later doesn't sound like a great idea either.

What is more, I think that random number might be even worse in this case...

pasinskim:
Looks nice, but we need some follow up on sequence numbers. JIRA task here: https://tracker.mender.io/browse/MEN-489

PR #71, opened 2016-06-22
Remove not needed lacks and stopChannel.

pasinskim:
I have noticed that we are not using stopChannel any more for stopping
daemon. This can be safely removed now.

What is more, looks like  mutexes are also not needed at this point
as Mender is one thread application so synchronization is not needed.

### DIFF
--- a/daemon.go
+++ b/daemon.go
@@ -13,39 +13,28 @@
 //    limitations under the License.
 package main
 
-import (
-	"sync"
-
-	"github.com/pkg/errors"
-)
+import "github.com/pkg/errors"
 
 // Config section
 
 type menderDaemon struct {
-	mender      Controller
-	stopChannel chan (bool)
-	stop        bool
-	stopLock    sync.Mutex
+	mender Controller
+	stop   bool
 }
 
 func NewDaemon(mender Controller) *menderDaemon {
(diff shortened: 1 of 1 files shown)


pasinskim:
@bboozzoo @kacf

kacf:
Commit is not signed off (dammit, why isn't the the license test catching this...). Otherwise looks good to me.

bboozzoo:
If we add support for SIGTERM via `signal.Notify()` (and that is likely to end up in a separate goroutine) we'll need the code you want to remove. Maybe not `chan(bool)` channel actually, but stop flag and mutex very much so.

bboozzoo:
On more thought in the topic of signals. The usual way to handle them is to block (most) signals at runtime and only allow specific signals at certain points in code. Not sure if Go is special in this matter, I guess not and the same set of rules will apply.

pasinskim:
IMHO, let's add this once we will really need it. Signals handling is one of the things which needs some attention, but as long as this is not there let's not try to make the things more complicated than have to be.

Any thoughts?

@kacf commit should be signed now.

kacf:
Agreed, we can always retrieve it from Git history later.

PR #74, opened 2016-06-28
bootstrap & authorize CLI switches

bboozzoo:
It may be reasonable to have separate command line switches for triggering bootstrap and authorization. The series renames the old `-bootstrap` switch to `-forcebootstrap` to make it more meaningful and introduces 2 new switches:
* `-bootstrap` - the client perform boostrap only and exit
* `-authorize` - the client performs a single authorization attempt and exits

Also extra cleanups for `runOptionsType`.

@kacf @pasinskim 

Edit: after review separate authorization and bootstrap switches were replaced with single `-bootstrap`.

### DIFF
--- a/main.go
+++ b/main.go
@@ -351,19 +351,13 @@ func doMain(args []string) error {
 	switch {
 
 	case *runOptions.imageFile != "":
-		if err := doRootfs(device, runOptions); err != nil {
-			return err
-		}
+		return doRootfs(device, runOptions)
 
 	case *runOptions.commit:
-		if err := device.CommitUpdate(); err != nil {
-			return err
-		}
+		return device.CommitUpdate()
 
 	case *runOptions.bootstrap:
-		if err := doBootstrapAuthorize(config, &runOptions); err != nil {
-			return err
-		}
+		return doBootstrapAuthorize(config, &runOptions)
 
 	case *runOptions.daemon:
 		d, err := initDaemon(config, device, env, &runOptions)
--- a/main.go
+++ b/main.go
@@ -39,6 +39,7 @@ type runOptionsType struct {
 	dataStore      *string
 	imageFile      *string
 	commit         *bool
+	bootstrap      *bool
 	daemon         *bool
 	bootstrapForce *bool
 	httpsClientConfig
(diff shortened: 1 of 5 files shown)


GregorioDiStefano:
@bboozzoo ignore the message from the cfengine-review-bot for now :)

pasinskim:
I don't think that splitting `bootstrap` and `authorize` options makes sense. This is adding some additional overhead and is making Mender less intuitive (first you need bootstrap and then authorize). What is more, what is the point of bootstrapping without authorization to the server? Those are integral operations, that we shouldn't split IMHO.

bboozzoo:
Should I wrap these two options into one, say `-authorize`? Are you sure that separate `-bootstrap` will not come in useful (it's not that you need to invoke it manually when doing authorization anyway)l?

pasinskim:
I would limit the number of possible options to minimum. Unless something is really needed we shouldn't add this IMHO. 
What is more, the command line and manual operations are bit blurry at the moment. Probably we should at some point decide what is needed and how command line should be organized. We've been thinking that commands-like syntax might be a better way (like git for example).
I would't spend time now to improve this though, as probably better option is to start user tests and get the feedback. Maybe something we should talk to @eysteins. 

So summing up, I would merge those two into one command. 
Before merging, please check if we need to change documentation.

bboozzoo:
What about the name of this option, is `-authorize` ok?

pasinskim:
I think that we've picked up `bootstrap` as this was the same in the other project of ours. Maybe `authorize` is more meaningful, but this is probably PO decision. I would stay with `bootstrap` for now and propose a name change in other PR. Then depending on the decision we would be able to either merge or reject it (the other one I mean). Does it make sense?

main.go
```diff
@@ -258,6 +262,37 @@ func ShowVersion() {
 	os.Stdout.Write([]byte(v))
 }
 
+func doBootstrapAuthorize(config *menderConfig, opts *runOptionsType) error {
+	store := NewDirStore(*opts.dataStore)
+
+	authreq, err := NewAuthClient(config.GetHttpConfig())
+	if err != nil {
+		return errors.New("cannot auth client")
```
kacf: Maybe include the original `err` in the error here, so you know what the reason is?

kacf:
Looks good otherwise!

FWIW, I think "authorize" is a better word actually, "bootstrap" feels like a legacy word to me. But I have no strong opinion, so I'm ok with that too.

kacf:
I'm happy with this, merge if you agree, @pasinskim!

PR #75, opened 2016-06-29
API client cleanup

bboozzoo:
Cleanup of API client handling. Until now both `UpdateClient` and `AuthClient` created an instance of `http.Client` for requests. This has been replaced by a single instance of `ApiClient` interface kept by mender controller. `ApiClient` is limited to a single call `Do(*http.Request) (*http.Response, error)` that allows for using vanilla `http.Client` instance or one created by `NewApiClient()` helper call.

@kacf @pasinskim

### DIFF
--- a/client.go
+++ b/client.go
@@ -33,10 +33,19 @@ var (
 	errorAddingServerCertificateToPool = errors.New("Error adding trusted server certificate to pool.")
 )
 
+// Mender API Client wrapper. A standard http.Client is compatible with this
+// interface and can be used without further configuration where ApiRequester is
+// expected. Instead of instantiating the client by yourself, one can also use a
+// wrapper call NewApiClient() that sets up TLS handling according to passed
+// configuration.
+type ApiRequester interface {
+	Do(req *http.Request) (*http.Response, error)
+}
+
 type RequestProcessingFunc func(response *http.Response) (interface{}, error)
 
 // Client initialization
-func NewHttpClient(conf httpsClientConfig) (*http.Client, error) {
+func NewApiClient(conf httpsClientConfig) (*http.Client, error) {
 
 	if conf == (httpsClientConfig{}) {
 		return newHttpClient(), nil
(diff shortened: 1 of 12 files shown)


client.go
```diff
@@ -33,24 +33,33 @@ var (
 	errorAddingServerCertificateToPool = errors.New("Error adding trusted server certificate to pool.")
 )
 
+// Mender API Client wrapper. A standard http.Client is compatible with this
+// interface and can be used without further configuration where ApiClient is
+// expected. Instead of instantiating the client by yourself, one can also use a
+// wrapper call NewApiClient() that sets up TLS handling according to passed
+// configuration.
+type ApiClient interface {
```
pasinskim: I would stick to the Go naming convention here with `er` suffix. we should have `doer interface`.

client.go
```diff
@@ -33,24 +33,33 @@ var (
...
-		return newHttpClient(), nil
+		return newApiHttpClient(), nil
 	}
 
-	return newHttpsClient(conf)
+	return newApiHttpsClient(conf)
 }
 
-func newHttpClient() *http.Client {
+func newApiHttpClient() *http.Client {
```
pasinskim: I think the old name was better here. In fact we are returning JUST http client.

client_auth.go
```diff
@@ -25,34 +25,27 @@ import (
...
-		client: client,
-	}
-	return &ac, nil
+func NewAuthClient() *AuthClient {
+	ac := AuthClient{}
+	return &ac
 }
 
-func (u *AuthClient) Request(server string, dataSrc AuthDataMessenger) ([]byte, error) {
+func (u *AuthClient) Request(client ApiClient, server string, dataSrc AuthDataMessenger) ([]byte, error) {
```
pasinskim: So here `AuthClient` actually contains `http.Client` field, but still while doing request we are injecting `ApiClient`. Why we are not doing this in `NewAuthClient`? It feels that this binding makes more sense in "constructor".

pasinskim: Forget about last comment!

bboozzoo: The embedded http.Client was removed

PR #77, opened 2016-06-30
Setup authorization data for API requests

bboozzoo:
Add authorization data to outgoing API requests. The series add an `ApiClient` helper with a convenience method `Request()` for creating request wrappers. Request wrapper `ApiRequest` implements `ApiRequester` interface and automatically append authorization data to the outgoing request if none is present.

Mender controller has been updated to cache authorization code so that we don't waste time/cycles for reloading it each time it's needed.

@pasinskim @kacf

### DIFF
--- a/auth.go
+++ b/auth.go
@@ -22,7 +22,7 @@ import (
 	"github.com/pkg/errors"
 )
 
-type AuthCode string
+type AuthToken string
 
 type AuthReqData struct {
 	IdData      string `json:"id_data"`
@@ -34,8 +34,8 @@ type AuthReqData struct {
 type AuthRequest struct {
 	// request message data
 	Data []byte
-	// authorization code
-	Code AuthCode
+	// authorization token
+	Token AuthToken
 	// request signature
 	Signature []byte
 }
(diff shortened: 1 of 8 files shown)


kacf:
Good stuff!

One question: Could we stick to calling it `authToken` throughout, instead of `authCode`? I find it's easier to understand the intention of the variable "token" than "code" (unless, *I'm* the one that's confused, and `authCode` is not the token :-) ).

bboozzoo:
I don't have any strong opinion here, `authToken` is fine. IDo you want to rename `AuthCode` type alias to `AuthToken` too?

kacf:
Yes, let's use the same terminology everywhere.

kacf:
> rename AuthToken to AuthCode

I think you're doing the opposite. :-)

bboozzoo:
On Fri, Jul 1, 2016 at 10:53 AM, Kristian Amlie <notifications@github.com>
wrote:

> rename AuthToken to AuthCode
>
> I think you're doing the opposite. :-)
>

oops, right ;)

PR #78, opened 2016-07-01
Status reporting

bboozzoo:
First take on status reporting.

Sending status over HTTP is captured in `StatusClient`, implemented in the first part of this series. Failure to send a status report results in an error that the caller is expected to handle in some way.

`Controller` interface got a new function `ReportStatusUpdate` that returns a `menderError` or nil.

The fun part is captured in state handling. Basically each 'update' state (commit, install, fetch, reboot) can transition to a new error state - `UpdateErrorState`. Also each of these states carries the update information. This is nice, because we don't need to keep it as part of Controller's internal state. Update information should be saved as soon as we decide to apply an update (ideally when update check returns valid update information) and restored right after (returned from?) `HasUpgrade()`.

@kacf @pasinskim

### DIFF
--- a/client_status.go
+++ b/client_status.go
@@ -1,0 +1,84 @@
+// Copyright 2016 Mender Software AS
+//
+//    Licensed under the Apache License, Version 2.0 (the "License");
+//    you may not use this file except in compliance with the License.
+//    You may obtain a copy of the License at
+//
+//        http://www.apache.org/licenses/LICENSE-2.0
+//
+//    Unless required by applicable law or agreed to in writing, software
+//    distributed under the License is distributed on an "AS IS" BASIS,
+//    WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+//    See the License for the specific language governing permissions and
+//    limitations under the License.
+package main
+
+import (
+	"bytes"
+	"encoding/json"
+	"fmt"
+	"net/http"
+
+	"github.com/mendersoftware/log"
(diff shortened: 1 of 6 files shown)


client_status.go
```diff
@@ -0,0 +1,87 @@
...
+		return errors.Wrapf(err, "failed to prepare status report request")
+	}
+
+	r, err := api.Do(req)
+	if err != nil {
+		log.Error("failed to report status: ", err)
+		return errors.Wrapf(err, "reporting status failed")
+	}
+
+	if r.StatusCode != http.StatusNoContent {
```
kacf: What is `StatusNoContent`? Is that the expected response?

client_status.go
```diff
@@ -0,0 +1,87 @@
...
+
+	out := &bytes.Buffer{}
+	enc := json.NewEncoder(out)
+	enc.Encode(&report)
+
+	req, err := http.NewRequest(http.MethodPut, url, out)
+	if err != nil {
+		return nil, err
+	}
+	return req, nil
```
kacf: I think you could shorten this to simply return `http.NewRequest` directly, which is (almost) equivalent (`req` isn't guaranteed to be `nil` in case of error, but this is unlikely to be significant).

client_status.go
bboozzoo: Yes, the API proposal lists 204 (No Content) as response.

kacf:
Looks good to me otherwise!

PR #79, opened 2016-07-05
Log upload

bboozzoo:
A series of patches that adds support for uploading logs in case of update failure. For now there's a single 'update failed' log being reported.

@kacf @pasinskim

### DIFF
--- a/state.go
+++ b/state.go
@@ -397,6 +397,16 @@ func NewUpdateErrorState(err menderError, update UpdateResponse) State {
 func (ue *UpdateErrorState) Handle(c Controller) (State, bool) {
 	// TODO error handling
 	c.ReportUpdateStatus(ue.update, statusFailure)
+
+	// TODO fetch logs from logger
+	c.UploadLog(ue.update, []LogEntry{
+		LogEntry{
+			Timestamp: time.Now().Format(time.RFC3339),
+			Level:     "error",
+			Message:   "update failed",
+		},
+	})
+
 	return initState, false
 }
 
--- a/client_log.go
+++ b/client_log.go
@@ -1,0 +1,82 @@
+// Copyright 2016 Mender Software AS
+//
+//    Licensed under the Apache License, Version 2.0 (the "License");
+//    you may not use this file except in compliance with the License.
+//    You may obtain a copy of the License at
+//
+//        http://www.apache.org/licenses/LICENSE-2.0
+//
+//    Unless required by applicable law or agreed to in writing, software
+//    distributed under the License is distributed on an "AS IS" BASIS,
+//    WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+//    See the License for the specific language governing permissions and
+//    limitations under the License.
+package main
+
+import (
+	"bytes"
+	"encoding/json"
+	"fmt"
+	"net/http"
+
+	"github.com/mendersoftware/log"
+	"github.com/pkg/errors"
+)
+
+type LogUploader interface {
+	Upload(api ApiRequester, server string, logs LogData) error
+}
+
+type LogEntry struct {
+	Timestamp string `json:"timestamp"`
+	Level     string `json:"level"`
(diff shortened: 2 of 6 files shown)


state.go
```diff
@@ -397,6 +397,16 @@ func NewUpdateErrorState(err menderError, update UpdateResponse) State {
...
 	c.ReportUpdateStatus(ue.update, statusFailure)
+
+	// TODO fetch logs from logger
+	c.UploadLog(ue.update, []LogEntry{
+		LogEntry{
+			Timestamp: time.Now().Format(time.RFC3339),
+			Level:     "error",
+			Message:   "update failed",
+		},
+	})
```
kacf: If we fail to upload the log, should we stay in the UpdateErrorState? I suppose it makes sense to stay there until we know that the server knows about our failure. What do you think?

bboozzoo: Yeah, might be the same scenario as with update status report. Say, if we fail to report update failure, we should just keep on trying until we succeed. There's also a question what happens when there's a reboot in between. And there's obviously possibility of rolling back. IMO we should go over the scenarios with @eysteins seems like we can't/should not really try to solve it here.

kacf: I think it's still a fairly technical discussion, we need to make sure we can always recover. So if we break it down:

* UpdateErrorState should never be reached in a daemon running with a non-committed update, because being in that situation means that there will be an immediate reboot and most likely rollback.
* After rolling back, and we are back in the old update, we enter UpdateErrorState, and stay there until we have successfully reported it.
* If there's a reboot while in UpdateErrorState (not necessarily because of us), we should stay in UpdateErrorState, IOW it should have no influence on the state. Only the report will change the state.

How to implement this technically is another matter, I'm just wondering if you agree with my logical rationale.

bboozzoo: > UpdateErrorState should never be reached in a daemon running with a non-committed update, because being in that situation means that there will be an immediate reboot and most likely rollback.

IMO any update state (fetch/install/commit) can transition to `UpdateErrorState`. What happens next should depend on where in the update pipeline the problem occurred. Fetch/install is harmless so we could go back to `UpdateCheck`. In contrast, commit failure could potentially trigger a rollback (something we still need to implement). Each update error may require additional reporting and upload of logs. That's also a reason why update information is captured in update related states as data (incl. `UpdateErrorState`). This is different from `ErrorState` that is suitable for other generic errors (no that we have many of those) and requires no special reporting.

I agree with your other points. There's only one concern, we need to preserve the state across reboots. Maybe that's something worth working on next.

kacf: You're right, this might deserve to be broken down even further:

* Fetch failure -> UpdateErrorState1
* Install failure -> UpdateErrorState2
* Boot update fails -> rollback & reboot -> UpdateErrorState3
* Boot update succeeds, but reporting update fails -> rollback & reboot -> UpdateErrorState4

The numbered `UpdateErrorStates` are arbitrary, just for discussion. The two first ones are as you say, harmless, nothing has really happened to the device yet. However, IMHO I think the states 2 and higher should be treated the same, since they are unlikely to be temporary, and should be reported in a mandatory fashion. The first `UpdateErrorState1` could however go back to `UpdateCheck`, since being unable to fetch is probably due to network outage. But I guess this one is debatable.

In any case, for such a "softer" type of error we could introduce a new state `UpdateSoftErrorState` or something similar, which unlike `UpdateErrorState`, can go straight back to `UpdateCheck`, even if the report fails.

And yes, we probably need to store some additional state for this.

bboozzoo: I've introduced `menderError` exactly for the purpose of distinguishing soft vs. hard errors (the error is considered transient unless `menderError.IsFatal()` returns true). This is also captured as part of error states data. IMO we can work on this, perhaps extend menderError with additional information.

Also, capturing the previous state information when transitioning to one of the error states might be useful in case we want to give special treatment to certain states that failed.

kacf:
@GregorioDiStefano: Actually I should have tried the PR analyzer with this one. Let me know if it gives trouble in testing.

PR #80, opened 2016-07-06
Document state transitions

bboozzoo:
State transitions in a diagram.

### DIFF
--- a/state.go
+++ b/state.go
@@ -21,6 +21,98 @@ import (
 	"github.com/pkg/errors"
 )
 
+// Each state implements Handle() - a state handler method that performs actions
+// on the Controller. The handler returns a new state, thus performing a state
+// transition. Each state can transition to an instance of ErrorState (or
+// UpdateErrorState for update related states). The handling of error states is
+// described further down.
+//
+// Regular state transitions:
+//
+//                             init
+//
+//                               |        (wait timeout expired)
+//                               |   +---------------------------------+
+//                               |   |                                 |
+//                               v   v                                 |
+//                                         (auth req. failed)
+//                          bootstrapped ----------------------> authorize wait
+//
+//                                |
+//                                |
+//                                |  (auth data avail.)
+//                                |
+//                                v
+//
+//                           authorized
+//
+//          (update needs     |   |
+//           commit)          |   |
+//         +------------------+   |
+//         |                      |
+//         v                      |          (wait timeout expired)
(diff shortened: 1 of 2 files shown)


state.go
```diff
@@ -21,6 +21,61 @@ import (
...
+//                                |  enabled)
+//                                |
+//                                v
+//
+//                              reboot
+//
+//                                |
+//                                v
+//
+//                              final (daemon exit)
```
kacf: Do we really exit after the reboot stage? IMHO init scripts should take care of this and kill us afterwards. This might be useful later if we add logic to detect failed reboots (forcing unclean reboot if a process is hanging, for example). Just a thought.

bboozzoo: Right now we do exit. The idea is that if we managed to call `reboot` without errors (no non-0 return code), there's nothing more we can do. The rest is ultimately handled by system init (systemd or whatever else).

kacf: Alright, fair enough.

kacf: Hmm, I think actually systemd might respawn the daemon, even if only for a split second before the reboot goes through. Are we prepared for that?

bboozzoo: We should be, but we are not at the moment. It should be possible to store the ID of new image for later comparison to detect if mender was started in the same image or a new one. Preferably, this should be a part of update state information we want to preserver across reboots.

FWIW we have `Restart=on-abort` in service definition in `meta-mender`, so unless we crash systemd will not restart the process after exiting.

kacf: Ah ok, should be good for the moment then.

state.go
```diff
@@ -21,6 +21,97 @@ import (
...
+//          |                       v                                  |
+//          |
+//          |                 update state -------------------> update error state
+//          |
+//          |                       |                                  |
+//          |                       |                                  |
+//          |                       v                                  |
+//          |                                                          |
+//          +-------------------> final (daemon exit) <----------------+
+//
```
kacf: I think we have to make a distinction in the state diagram, "running committed update" or not. Because it's not clear how we get to the final state from various other states (in fact maybe it should say reboot, because that's always how we get to the final state). Going to reboot from a "committed update" will never happen in the case of an error, so this should be reflected somehow. However, from a "non-committed update" it will usually happen.

bboozzoo: Pushed an updated diagram, should be clear enough now.

bboozzoo:
ping

PR #81, opened 2016-07-06
Logs

pasinskim:


pasinskim:
@kacf @bboozzoo please take a look. I will still work on this one. There are some integration parts needed. What is more, I will squash commits later. 
I've added PR to make it easier for commenting.

pasinskim:
FYI, Travis is failing because code is not used anywhere at the moment.

config_test.go
```diff
@@ -32,7 +32,8 @@ var testConfig = `{
   "PollIntervalSeconds": 60,
   "ServerURL": "mender.io",
   "DeviceID": "1234-ABCD",
-  "ServerCertificate": "/data/server.crt"
+  "ServerCertificate": "/data/server.crt",
+  "UpdateLogPath": "/data/logs/update.log"
```
kacf: I think this should be `/var/lib/mender` and `/var/lib/mender/log` (for the certificate as well). Yes, I know, it doesn't really matter in a test, but I think we should stick to what we're gonna use. Less confusing when reading it later.

config_test.go
```diff
@@ -47,7 +48,8 @@ var testConfigDevKey = `{
...
```
kacf: `/var/lib/mender`

config_test.go
```diff
@@ -99,6 +101,7 @@ func validateConfiguration(t *testing.T, actual *menderConfig) {
 		PollIntervalSeconds: 60,
 		ServerURL:           "mender.io",
 		ServerCertificate:   "/data/server.crt",
+		UpdateLogPath:       "/data/logs/update.log",
```
kacf: `/var/lib/mender`

deployment_logger.go
```diff
@@ -0,0 +1,248 @@
...
+	if dlm.logger == nil {
+		return ErrLoggerNotInitialized
+	}
+	_, err := dlm.logger.Write(log)
+	return err
+}
+
+func (dlm *DeploymentLogManager) Enable() error {
+	if dlm.loggingEnabled {
+		// TODO: maybe in fact we should return error
```
bboozzoo: Wouldn't be too useful

deployment_logger.go
```diff
@@ -0,0 +1,248 @@
...
+		return nil
+	}
+
+	// check if we have valid deploymend id
+	if dlm.deploymentID == "" {
```
bboozzoo: Then perhaps we should pass deployment ID as parameter to Enable(). Might be semantically nicer, as in 'enable logging with this deployment ID'.

deployment_logger.go
```diff
@@ -0,0 +1,248 @@
...
+	dlm.loggingEnabled = false
+	return nil
+}
+
+func (dlm DeploymentLogManager) getSortedLogFiles() ([]string, error) {
+
+	// list all the log files in log directory
+	fileExtension := filepath.Ext(dlm.logFileName)
+	logFiles, err :=
+		filepath.Glob(filepath.Join(dlm.logLocation, dlm.logFileName[0:len(dlm.logFileName)-len(fileExtension)]) + "*")
```
bboozzoo: Can you split it so that the whole line fits in the screen?

main.go
```diff
@@ -297,11 +299,19 @@ func initDaemon(config *menderConfig, dev *device, env *uBootEnv,
...
+		deploymentLogsLocation = defaultDeploymentLogLocation
+	}
+	logManager := NewDeploymentLogManager(deploymentLogsLocation)
+
 	controller, err := NewMender(*config, MenderPieces{
 		device:  dev,
 		env:     env,
 		store:   store,
 		authMgr: authmgr,
+		logMgr:  logManager,
```
bboozzoo: Don't think log manager should be part of mender. IMO it's outside of controller's area of 'control'

mender.go
```diff
@@ -318,6 +323,27 @@ func (m *mender) UploadLog(update UpdateResponse, logs []LogEntry) menderError {
 	return nil
 }
 
+func (m *mender) StartLogWriting(deploymentID string) error {
+	// check if there is id from last deployment (we coudl do reboot)
+	if deploymentID == "" {
```
bboozzoo: So all of this could actually be done in `UpdateFetchState` and passed along as some sort of update context between the update states. Disable logs right after leaving reboot, reopen when entering update commit state.

deployment_logger.go
```diff
@@ -0,0 +1,248 @@
...
+	}
+
+	// rename log files; only those not removed
+	for i := range logFiles {
+		os.Rename(logFiles[i], dlm.rotateLogFileName(logFiles[i]))
+	}
+}
+
+// we need to check what was last deployment ID logged in case of reboot or power-loss
+func (dlm DeploymentLogManager) GetDeploymentIDFromLogFile() (string, error) {
```
bboozzoo: IMO this is not part of deployment logging. Update state information should be preserved elsewhere, perhaps we need to add state context passed between update states.

What I'm missing is a way of obtaining the logs for a particular deployment ID. This is something we need in UpdateError state where we should load logs of the deployment that failed..

deployment_logger.go
```diff
@@ -0,0 +1,254 @@
...
+	maxLogFiles int
+	// it is easy to add logging hook, but not so much remove it;
+	// we need a mechanism for emabling and disabling logging
+	loggingEnabled bool
+}
+
+func NewDeploymentLogManager(logDirLocation string) *DeploymentLogManager {
+	return &DeploymentLogManager{
+		logLocation: filepath.Dir(logDirLocation),
+		logFileName: filepath.Base(logDirLocation),
```
bboozzoo: Maybe the user should not be able to change the log file name, because it's an internal part of the client. Instead I'd try to enforce a naming scheme that makes it easy to extract deployment ID, for instance: `"%04d.<deployment-id>"`. Log ID would come first, then deployment ID (maybe with some escaping so that it can be used as part of a filename).

Also, log location should perhaps be relative to the global data path location, that defaults to `/var/lib/mender`, but can be overridden by passing `-data` in command line.

deployment_logger.go
pasinskim: I was thinking that when you are enabling logging twice there must be some error in the logic (kind of when you are trying to lock mutex which is already locked). Maybe we could somehow notify about those kind of issues.

deployment_logger.go
pasinskim: From the other hand we will need to call `Enable()` at least from two places: 
- when we are starting new update (here we will know what id is as we will receive it from server)
- when we are committing new update (at this point we will have to read it from somewhere; most likely some persistent location as at this point we will be after reboot)
- after some error happened (like we started update; then there was power-loss; and after reboot we need to report back to server that update was broken because ...)

In all cases, but first one, we will need to get deployment id from somewhere.

main.go
pasinskim: I am open for suggestions here.

deployment_logger.go
bboozzoo: Logging a warning may be enough here. I don't really see that as an error, nothing like the mutex example that you gave.

mender.go
pasinskim: I think I will need some more hints how you see this.

deployment_logger.go
bboozzoo: Well yes, but I'm slightly uncomfortable on depending on some part of state being initialized and returning an error if it's not. Restoring in-progress update information is not part of logging anyway, so I think it's safe to assume that whoever calls that method has all the necessary information.

deployment_logger.go
pasinskim: I disagree, I think. Or I might not fully understand what you are referring to. 
Deployment id is not part of the state. It is just identifier of current (last) update. We need to store it somewhere, as we will log it after device will be rebooted.
I didn't consider obtaining logs for particular deployment id important. Simply, I assume last log file is containing information about last deployment (aka file which is not rotated yet). If you want to get info about last deployment you simply need to read ALWAYS the same file. 
Probably something to discuss, if we need to find log file for given deployment id.

deployment_logger.go
pasinskim: I agree with the first part. We shouldn't allow user to change this. I was more thinking, that this should be configurable as we might need to change this ourselves.

As for the name scheme, I've been thinking about this as well. There are obviously some pros and cons. 
With current solution, you always get not rotated file if you are interested in last update logs. The file name is always the same for last update. 
Deployment id is part of each log message, which is adding some overhead, but from the other hand server can store each log line independently. 

I guess this is more of API design, and server processing side, but in this case (after seeing API implementation) I think we can remove deployment id from log file. 

Let's discuss naming convention separately.

config_test.go
```diff
@@ -32,7 +32,8 @@ var testConfig = `{
   "PollIntervalSeconds": 60,
   "ServerURL": "mender.io",
   "DeviceID": "1234-ABCD",
-  "ServerCertificate": "/data/server.crt"
+  "ServerCertificate": "/var/lib/mender/server.crt",
```
bboozzoo: Actually the certificate is somewhere in ${sysconfdir}/mender in the image. Though I guess it does not really matter which path we end up using in tests.

deployment_logger.go
```diff
@@ -0,0 +1,179 @@
...
+	dlm.deploymentID = deploymentID
+
+	// we might have new deployment so might need to rotate files
+	dlm.Rotate()
+
+	// instanciate logger
+	logFileName := fmt.Sprintf(logFileNameScheme, 1, deploymentID)
+	dlm.logger = NewFileLogger(filepath.Join(dlm.logLocation, logFileName))
+	if dlm.logger == nil {
+		return ErrLoggerNotInitialized
```
bboozzoo: try this:
`errors.Wrapf(err, ErrLoggerNotInitialized)`

This should give some context why we failed to initialize the logger.

main.go
```diff
@@ -334,6 +341,9 @@ func doMain(args []string) error {
 
 	env := NewEnvironment(new(osCalls))
 	device := NewDevice(env, new(osCalls), config.GetDeviceConfig())
+
+	DeploymentLogger = NewDeploymentLogManager("")
```
bboozzoo: NewDeploymentLogManager(*runOptions.dataStore)

main.go
```diff
@@ -56,6 +57,8 @@ var (
 
 var defaultConfFile string = "/etc/mender/mender.conf"
 
+var DeploymentLogger *DeploymentLogManager
```
bboozzoo: Might be a bit better to have an instance rather than a pointer here.

deployment_logger.go
```diff
@@ -0,0 +1,179 @@
...
+type FileLogger struct {
+	logFileName string
+	logFile     io.WriteCloser
+}
+
+// NewFileLogger creates instance of file logger; it is initialized
+// just before logging is started
+func NewFileLogger(name string) *FileLogger {
+	// open log file
+	logFile, err := os.OpenFile(name, os.O_RDWR|os.O_CREATE|os.O_APPEND, 0666)
```
bboozzoo: One thing I'm not sure of, but maybe it makes more sense to open & close the file in `Write()` rather than doing it here and dragging `io.WriteCloser` around.

config_test.go
pasinskim: Yup. This does not matter in tests, but it's changed to avoid confusions, where data should be stored.

deployment_logger.go
pasinskim: Fixing!

deployment_logger.go
bboozzoo: `0666` mask is not the best idea here, I'd suggest using 0600 instead.

main.go
pasinskim: Right, thanks!

deployment_logger.go
pasinskim: Or actually not. There is no context to be added here.

bboozzoo: Maybe NewFileLogger() should return (*FileLogger, error). Kind of does IO operations, so returning the actual error might make sense here. On the other hand, if we opened/closed the file in `FileLogger.Write()`, there is no need to return any errors here.

main.go
pasinskim: Matter of taste I think. It might be easier with pointer to check if it is nil or not. 
What is more, in our log library `Log` is also a pointer.

deployment_logger.go
pasinskim: Fixing mask.

pasinskim: I did this originally (opening and closing in `write()`). I am not sure either which method is better. It's kind of performance vs reliability. 
Maybe we can leave as is and change if we will see some issues with that?

bboozzoo:
+1, squash some commits if possible and we can merge

PR #82, opened 2016-07-07
Preserve state data across reboots

bboozzoo:
A proposal of a method for preserving state information across reboots. Update related actions such as reporting of log upload require information on the update that is applied. Since update will normally involve a reboot at some point, we need to capture and preserve update information across reboots, so that it's possible to report update progress after the system was restarted.

The series adds a new entry called `state` to client data store. The `state` file contains JSON serialized `StateData` structure, that captures both the update response and the last state ID. State data is stored when entering a state handler. Loading happens only when entering `UpdateCommitState`. The state entry is removed after an update has been committed.

There are still some open questions about what to do where a reboot happens mid update. For instance, when update is being installed. I expect that the state ID field will come in useful here. Since state data also captures the download URL, we should be able to resume from say, `UpdateFetchState` when needed. Anyways, I expect this to be a starting point for a longer discussion.

### DIFF
--- a/daemon.go
+++ b/daemon.go
@@ -20,12 +20,16 @@ import "github.com/pkg/errors"
 type menderDaemon struct {
 	mender Controller
 	stop   bool
+	sctx   StateContext
 }
 
-func NewDaemon(mender Controller) *menderDaemon {
+func NewDaemon(mender Controller, store Store) *menderDaemon {
 
 	daemon := menderDaemon{
 		mender: mender,
+		sctx: StateContext{
+			store: store,
+		},
 	}
 	return &daemon
 }
@@ -41,7 +45,7 @@ func (d *menderDaemon) shouldStop() bool {
 func (d *menderDaemon) Run() error {
 	// figure out the state
 	for {
-		state, cancelled := d.mender.GetState().Handle(d.mender)
+		state, cancelled := d.mender.RunState(&d.sctx)
 		if state.Id() == MenderStateError {
 			es, ok := state.(*ErrorState)
 			if ok {
--- a/mender.go
+++ b/mender.go
@@ -313,3 +313,7 @@ func (m *mender) SetState(s State) {
 func (m *mender) GetState() State {
 	return m.state
 }
+
+func (m *mender) RunState(ctx *StateContext) (State, bool) {
+	return m.state.Handle(ctx, m)
+}
--- a/state.go
+++ b/state.go
@@ -14,7 +14,9 @@
 package main
 
 import (
+	"encoding/json"
 	"io"
+	"os"
 	"time"
 
 	"github.com/mendersoftware/log"
@@ -113,10 +115,16 @@ import (
 //                           (daemon exit)
 //
 
+// state context carrying over data that may be used by all state handlers
+type StateContext struct {
+	// data store access
+	store Store
+}
+
 type State interface {
 	// Perform state action, returns next state and boolean flag indicating if
 	// execution was cancelled or not
-	Handle(c Controller) (State, bool)
+	Handle(ctx *StateContext, c Controller) (State, bool)
 	// Cancel state action, returns true if action was cancelled
 	Cancel() bool
 	// Return numeric state ID
@@ -128,8 +136,22 @@ type StateRunner interface {
 	SetState(s State)
 	// Obtain runner's state
 	GetState() State
-	// TODO generic state run action
+	// Run the currently set state with this context
+	RunState(ctx *StateContext) (State, bool)
 }
+
+// state information that can be used for restring state from storage
+type StateData struct {
+	// update reponse data for the update that was in progress
+	UpdateInfo UpdateResponse
+	// id of the last state to execute
+	Id MenderState
+}
+
+const (
+	// name of file where state data is stored across reboots
+	stateDataFileName = "state"
+)
 
 var (
 	initState = &InitState{
@@ -220,7 +242,7 @@ type InitState struct {
 	BaseState
 }
 
-func (i *InitState) Handle(c Controller) (State, bool) {
+func (i *InitState) Handle(ctx *StateContext, c Controller) (State, bool) {
 	log.Debugf("handle init state")
 	if err := c.Bootstrap(); err != nil {
 		log.Errorf("bootstrap failed: %s", err)
@@ -233,7 +255,7 @@ type BootstrappedState struct {
 	BaseState
 }
 
-func (b *BootstrappedState) Handle(c Controller) (State, bool) {
+func (b *BootstrappedState) Handle(ctx *StateContext, c Controller) (State, bool) {
 	log.Debugf("handle bootstrapped state")
 	if err := c.Authorize(); err != nil {
 		log.Errorf("authorize failed: %v", err)
@@ -261,7 +283,7 @@ func NewUpdateCommitState(update UpdateResponse) State {
 	}
 }
 
-func (uc *UpdateCommitState) Handle(c Controller) (State, bool) {
+func (uc *UpdateCommitState) Handle(ctx *StateContext, c Controller) (State, bool) {
 	log.Debugf("handle update commit state")
 	err := c.CommitUpdate()
 	if err != nil {
@@ -278,6 +300,9 @@ func (uc *UpdateCommitState) Handle(c Controller) (State, bool) {
 		// return NewUpdateErrorState(merr, uc.update), false
 	}
 
+	// cleanup state data
+	RemoveStateData(ctx.store)
+
 	// done?
 	return updateCheckWaitState, false
 }
@@ -286,7 +311,7 @@ type UpdateCheckState struct {
 	BaseState
 }
 
-func (u *UpdateCheckState) Handle(c Controller) (State, bool) {
+func (u *UpdateCheckState) Handle(ctx *StateContext, c Controller) (State, bool) {
 	log.Debugf("handle update check state")
 	update, err := c.CheckUpdate()
 	if err != nil {
@@ -296,8 +321,6 @@ func (u *UpdateCheckState) Handle(c Controller) (State, bool) {
 	}
 
 	if update != nil {
-		// TODO: save update information state
-
 		// custom state data?
 		return NewUpdateFetchState(*update), false
 	}
@@ -319,7 +342,15 @@ func NewUpdateFetchState(update UpdateResponse) State {
 	}
 }
 
-func (u *UpdateFetchState) Handle(c Controller) (State, bool) {
+func (u *UpdateFetchState) Handle(ctx *StateContext, c Controller) (State, bool) {
+	if err := StoreStateData(ctx.store, StateData{
+		Id:         u.Id(),
+		UpdateInfo: u.update,
+	}); err != nil {
+		log.Errorf("failed to store state data in fetch state: %v", err)
+		return NewUpdateErrorState(NewTransientError(err), u.update), false
+	}
+
 	// report downloading, don't care about errors
 	c.ReportUpdateStatus(u.update, statusDownloading)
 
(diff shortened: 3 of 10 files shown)


state.go
```diff
@@ -113,10 +115,16 @@ import (
 //                           (daemon exit)
 //
 
+type StateContext struct {
+	ctrl   Controller
+	update *UpdateResponse
```
pasinskim: I would rename it to something more meaningful. Maybe `lastUpdateData`?

state.go
```diff
@@ -113,10 +115,16 @@ import (
 //                           (daemon exit)
 //
 
+type StateContext struct {
+	ctrl   Controller
```
pasinskim: Why do we need a Controller here?

state.go
```diff
@@ -129,8 +137,19 @@ type StateRunner interface {
 	// Obtain runner's state
 	GetState() State
 	// TODO generic state run action
+	RunState(ctx *StateContext) (State, bool)
```
pasinskim: I don't get the idea. How is `RunState()` different from `Handle()` and why we need both?

daemon.go
```diff
@@ -20,12 +20,17 @@ import "github.com/pkg/errors"
...
 	stop   bool
+	sctx   StateContext
 }
 
-func NewDaemon(mender Controller) *menderDaemon {
+func NewDaemon(mender Controller, store Store) *menderDaemon {
 
 	daemon := menderDaemon{
 		mender: mender,
+		sctx: StateContext{
```
pasinskim: s/sctx/sCtx/

mender.go
```diff
@@ -313,3 +313,7 @@ func (m *mender) SetState(s State) {
 func (m *mender) GetState() State {
 	return m.state
 }
+
+func (m *mender) RunState(ctx *StateContext) (State, bool) {
```
pasinskim: Maybe `HandleState`?

state.go
```diff
@@ -129,8 +137,19 @@ type StateRunner interface {
...
+}
+
+type StateData struct {
+	Update UpdateResponse
```
pasinskim: I would work a bit on changing the names. Maybe `LastUpdateInfo` and `State`?

state.go
```diff
@@ -129,8 +137,19 @@ type StateRunner interface {
...
+	Id     MenderState
 }
 
+const (
+	// name of file where state data is stored across reboots
+	stateDataName = "state"
```
pasinskim: `stateDataFileName`?

state.go
```diff
@@ -278,6 +297,9 @@ func (uc *UpdateCommitState) Handle(c Controller) (State, bool) {
 		// return NewUpdateErrorState(merr, uc.update), false
 	}
 
+	// cleanup state data
+	RemoveStateData(ctx.store)
```
pasinskim: Are we sending some status back to server? I think it is important to make sure that communication with server happens before we are removing state data.

state.go
```diff
@@ -422,17 +452,26 @@ type AuthorizedState struct {
 	BaseState
 }
 
-func (a *AuthorizedState) Handle(c Controller) (State, bool) {
-	// TODO HasUpgrade should return update information
-	has, err := c.HasUpgrade()
+func (a *AuthorizedState) Handle(ctx *StateContext, c Controller) (State, bool) {
+	sd, err := LoadStateData(ctx.store)
 	if err != nil {
```
pasinskim: Will it be possible to recover wen we loose state file for some reason?

state.go
```diff
@@ -530,6 +574,27 @@ type FinalState struct {
 	BaseState
 }
 
-func (f *FinalState) Handle(c Controller) (State, bool) {
+func (f *FinalState) Handle(ctx *StateContext, c Controller) (State, bool) {
 	panic("reached final state")
 }
+
+func StoreStateData(store Store, sd StateData) error {
```
pasinskim: TODO: as decided today we will evaluate storing this in some DB, which will be a part of another task.

store_test.go
```diff
@@ -111,8 +111,9 @@ func (ms *MemStore) Commit(name string, data []byte) error {
 	return nil
 }
 
-func (ms *MemStore) Remove(name string) {
+func (ms *MemStore) Remove(name string) error {
```
pasinskim: Doesn't look like we need to return anything here.

daemon.go
```diff
@@ -41,7 +46,7 @@ func (d *menderDaemon) shouldStop() bool {
 func (d *menderDaemon) Run() error {
 	// figure out the state
 	for {
-		state, cancelled := d.mender.GetState().Handle(d.mender)
+		state, cancelled := d.mender.RunState(&d.sctx)
```
pasinskim: This looks a bit fishy. In real OO language I would write we are calling a method on an object which we are passing as an argument to itself (not explicitly, but `mender` is part of `sctx`).

pasinskim:
I like the idea in general, but state is becoming bit messy now. We have `State` and `StateRunner` interfaces, some global variables (`initState`, `bootstrappedState`, ...), then lot of individual state structures (`BaseState`, `InitState`, `BootstrappedState`) and now extra `StateContext` and `StateData`. 
I think we should consider some refactoring of this code.

state.go
bboozzoo: Right, we don't. Must have left it behind.

state.go
bboozzoo: Kind of a helper that should have been here from the start. So instead of doing `StateRunner.GetState().Handle(ctx, ctrl)` we just call `StateRunner.RunState(ctx)`. Maybe it should take `Controller` as parameter too.

state.go
bboozzoo: `State` name is already taken.

Also `LastUpdateInfo` might be a bit confusing. When is the update cosidered to be last? For instance in commit state, update info is technically not the last update but for the current one.

daemon.go
bboozzoo: This is becoming a bit extreme. Let's agree that I skip this one.

state.go
bboozzoo: ok

state.go
bboozzoo: Status is send just above this line, see `c.ReportUpdateStatus(...)`, but that fails as the backend is not ready to receive the reports yet. Effectively, we always take the success path here, but this will need an update once backend is ready.

state.go
bboozzoo: No, I don't think so. If there's no update information (we get ErrNotExist) then we should treat it as if no update was in progress. 

I still need to complete the logic here. Looks like we'll need to involve image ID to figure out if we're really running the new image or not. I think that the logic may look like this, but I'd like to know your opinions @pasinskim @kacf 

```
LoadStateData()
if err != ErrNotExist 
    can't do anythng about it && raise error?
if last state in (MenderStateReboot, MenderStateInstall)
    if HasUpgrade()
        if currentImageID == UpdatedImageID() 
            update complete && report success
        else
            we either crashed and were restarted or update image
            is the same as previous??
    else
        so update was being installed (or was installed) but upgrade_avaialble was not flipped, 
        hence we're running the old image, start from scratch
else
    ignore previous update, start from scratch

RemoveStateData()
```

store_test.go
bboozzoo: `Store` interface defines `Remove(string) error`.

bboozzoo:
The naming may not be the most fortunate. Let's go one by one:

- `State` (`StateHandler`?) is something that implements the state logic. The handler `State.Handle()` takes a `StateContext` and `Controller` as parameters. The state primarly acts on controller.

- `StateRunner` something that has the state, for instance controller in our case, but equally well that may be a different entity. Basically you set a state of the runner to somethng, and then tell it to run the state (hence `RunState()`) on itself.

- `StateContext`, provides asccess to functionality that state handlers may need (store for instance)

- `StateData` is something that we save to file and restore

- `initState`, `bootstrappedState` are are declared as variables as these are always the same (i.e. these states do not keep any extra data that changes at runtime)

- update states `UpdateFetchState`, `UpdateInstallState` drag update information as their data, so they need to be created dynamically via `NewUpdate..State()`

- for clarity each state has its own struct to keep its private data

state.go
pasinskim: So maybe in that case `Handle()` should be `handle()` and we can use it as a local non-exported helper function?

daemon.go
pasinskim: Yup. Let's leave this as is. Just for you to know that I saw it :)

state.go
pasinskim: So maybe simply `UpdateInfo`?

state.go
pasinskim: Roger that!

state.go
bboozzoo: ok

state.go
bboozzoo: Removed that along with controller. Neither is needed here.

state.go
pasinskim: Couple of things here. First I think it might have sense to consider update as failed if the state can not be retrieved. 
As for the logic:
```
lastKnownState, imageID := LoadStateData()
if err != ErrNotExist
  if HasUpgrade()
    // probably something bad happened during update
    send broken update to server // might be tricky as we don't have update id
  else
    log error
    start from scratch?
// we have installed update correctly (can we enter this state if not?)
swith lastKnownState 
  case MenderStateReboot
    if HasUpgrade()
      if curentInamgeID == imageID
        update complete && report success
      else
        send broken update to server with imageID
    else
      last state was reboot but no update; is this valid?

  case MenderStateAuthorized
    do nothing; we simply rebooted

  default
    // update was in progress, but something went wrong
    report broken update to server with imageID
    start from scratch

RemoveStateData()
```

store_test.go
pasinskim: Got it!

pasinskim:
Let's leave the names as is right now. I will think about making those more intuitive. Let's improve and discuss in separate PR.

state.go
```diff
@@ -422,19 +461,66 @@ type AuthorizedState struct {
...
+
 	if err != nil {
-		log.Errorf("has upgrade check failed: %s", err)
-		// we may or may now have an upddate ready
-		return NewErrorState(err), false
+		log.Errorf("failed to restore update information: %v", err)
+		me := NewFatalError(errors.Wrapf(err, "failed to restore update information"))
+
+		// report update error with unknown deployment ID
+		// TODO: fill current image ID?
```
pasinskim: Not really. This might be misleading. Better report that we don't know which id failed. Maybe server will be able to figure out what was last update id.

state.go
```diff
@@ -422,19 +461,66 @@ type AuthorizedState struct {
...
+		log.Errorf("has upgrade check failed: %v", haserr)
+		me := NewFatalError(errors.Wrapf(err, "failed to perform 'has upgrade' check"))
+		return NewUpdateErrorState(me, sd.UpdateInfo), false
+	}
+
 	if has {
-		// TODO restore update information
-		return NewUpdateCommitState(UpdateResponse{}), false
+		if sd.UpdateInfo.Image.YoctoID == c.GetCurrentImageID() {
+			log.Infof("successfully running in new image %v", c.GetCurrentImageID())
```
pasinskim: `successfully running with the new image %v`?

state.go
```diff
@@ -422,19 +461,66 @@ type AuthorizedState struct {
...
+			// update info and has upgrade flag is there, we're running the new
```
pasinskim: s/is there/are there/

state.go
```diff
@@ -422,19 +461,66 @@ type AuthorizedState struct {
...
+			// update, everything looks good, proceed with committing
+			return NewUpdateCommitState(sd.UpdateInfo), false
+		} else {
+			// seems like we're running in a different image than expected from update
+			// information, best report an error
+			log.Errorf("running in image %v, expected updated image %v",
```
pasinskim: `running with image %v, expected updated image %v`

state.go
```diff
@@ -422,19 +461,66 @@ type AuthorizedState struct {
...
+				c.GetCurrentImageID(), sd.UpdateInfo.Image.YoctoID)
+			me := NewFatalError(errors.Errorf("restarted with old image %v, "+
+				"expected updated image %v",
+				c.GetCurrentImageID(), sd.UpdateInfo.Image.YoctoID))
+			return NewUpdateErrorState(me, sd.UpdateInfo), false
+		}
 	}
 
+	// we have upgrade info but has flag is not set
+	log.Infof("update info for deployment %v present, but has flag is false",
```
pasinskim: Maybe `(...) present, but no update is installed`?

state.go
```diff
@@ -516,7 +605,16 @@ func NewRebootState(update UpdateResponse) State {
 	}
 }
 
-func (e *RebootState) Handle(c Controller) (State, bool) {
+func (e *RebootState) Handle(ctx *StateContext, c Controller) (State, bool) {
+	if err := StoreStateData(ctx.store, StateData{
+		Id:         e.Id(),
+		UpdateInfo: e.update,
+	}); err != nil {
+		// TODO: too late to do anything now, update is installed and enabled
```
pasinskim: Shouldn't we simply continue reboot?

bboozzoo: Hmm, maybe you're right, we've already reached reboot, so it's logical that install has completed with success and state data was written there.

PR #84, opened 2016-07-11
Mandatory update status reporting

bboozzoo:
@kacf @pasinskim

### DIFF
--- a/state.go
+++ b/state.go
@@ -146,6 +146,8 @@ type StateData struct {
 	UpdateInfo UpdateResponse
 	// id of the last state to execute
 	Id MenderState
+	// update status
+	UpdateStatus string
 }
 
 const (
@@ -299,23 +301,11 @@ func (uc *UpdateCommitState) Handle(ctx *StateContext, c Controller) (State, boo
 	err := c.CommitUpdate()
 	if err != nil {
 		log.Errorf("update commit failed: %s", err)
-		return NewErrorState(NewFatalError(err)), false
+		return NewUpdateErrorState(NewFatalError(err), uc.update), false
 	}
-
-	if merr := c.ReportUpdateStatus(uc.update, statusSuccess); merr != nil {
-		log.Errorf("failed to report success status: %v", err)
-		// TODO: until backend has implemented status reporting, this error cannot
-		// result in update being aborted. Once required API endpoint is available
-		// commend the line below and remove this comment.
-
-		// return NewUpdateErrorState(merr, uc.update), false
-	}
-
-	// cleanup state data
-	RemoveStateData(ctx.store)
 
 	// done?
-	return updateCheckWaitState, false
+	return NewUpdateStatusReportState(uc.update, statusSuccess), false
 }
 
 type UpdateCheckState struct {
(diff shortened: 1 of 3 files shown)


state.go
```diff
@@ -584,19 +585,75 @@ func NewUpdateErrorState(err menderError, update UpdateResponse) State {
...
+
+func (usr *UpdateStatusReportState) Handle(ctx *StateContext, c Controller) (State, bool) {
+	if err := StoreStateData(ctx.store, StateData{
+		Id:           usr.Id(),
+		UpdateInfo:   usr.update,
+		UpdateStatus: usr.status,
+	}); err != nil {
+		log.Errorf("failed to store state data in update status report state: %v",
+			err)
+		// TODO: what now?
```
kacf: Should be the same as failure to report I think, rollback and reboot.

state.go
```diff
@@ -584,19 +585,75 @@ func NewUpdateErrorState(err menderError, update UpdateResponse) State {
...
+						Timestamp: time.Now().Format(time.RFC3339),
+						Level:     "error",
+						Message:   "update failed",
+					},
+				})
+			}
+		}
+
+		log.Debug("reporting complete")
+		break
```
kacf: Won't this `break` always be hit? It looks like you're trying to do some repeated attempts to upload the log, but I cannot see that the loop can ever be executed more than once, not even with the currently commented code.

state_test.go
```diff
@@ -195,13 +198,65 @@ func TestStateUpdateError(t *testing.T) {
 		store: ms,
 	}
 	sc := &stateTestController{}
-	es = NewUpdateErrorState(fooerr, UpdateResponse{})
-	es.Handle(&ctx, sc)
+	es = NewUpdateErrorState(fooerr, update)
+	s, _ := es.Handle(&ctx, sc)
+	assert.IsType(t, &UpdateStatusReportState{}, s)
+	// verify that update status report state data is correct
+	usr, _ := s.(*UpdateStatusReportState)
```
kacf: Minor: Maybe write without `_` to catch the error that it is the wrong type?

bboozzoo: There's an assert just above, `assert.IsType(t, &UpdateStatusReportState{}, s)`

state.go
bboozzoo: Makes sense. Since we don't support rollback at the moment, I'll update the comment to reflect that.

state_test.go
kacf: Right, doh!

state.go
```diff
@@ -573,19 +585,75 @@ func NewUpdateErrorState(err menderError, update UpdateResponse) State {
...
-		LogEntry{
-			Timestamp: time.Now().Format(time.RFC3339),
-			Level:     "error",
-			Message:   "update failed",
-		},
-	})
+	return NewUpdateStatusReportState(ue.update, statusFailure), false
+}
+
+// Wrapper for mandatory update state reporting, the state handler will attempt
```
pasinskim: ` Wrapper for mandatory update status reporting. The state handler will attempt`

pasinskim:
@bboozzoo overall looks OK. Just fix minor issues if there are any outstanding. What is more, make sure to comment out the appropriate code so that client won't break if we can not send logs and status to the server. 
You can add in comments reference to https://tracker.mender.io/browse/MEN-536 (sending status) and possibly https://tracker.mender.io/browse/MEN-437 (sending logs).

PR #85, opened 2016-07-12
Make sure to enable and disable deployment logging when needed.

pasinskim:

### DIFF
--- a/client_log.go
+++ b/client_log.go
@@ -15,7 +15,6 @@ package main
 
 import (
 	"bytes"
-	"encoding/json"
 	"fmt"
 	"net/http"
 
@@ -27,16 +26,10 @@ type LogUploader interface {
 	Upload(api ApiRequester, server string, logs LogData) error
 }
 
-type LogEntry struct {
-	Timestamp string `json:"time"`
-	Level     string `json:"level"`
-	Message   string `json:"msg"`
-}
-
 type LogData struct {
 	deviceID     string
 	deploymentID string
-	Messages     []LogEntry `json:"messages"`
+	Messages     []byte `json:"messages"`
 }
 
 type LogUploadClient struct {
@@ -74,9 +67,5 @@ func makeLogUploadRequest(server string, logs *LogData) (*http.Request, error) {
 		logs.deviceID, logs.deploymentID)
 	url := buildApiURL(server, path)
 
-	out := &bytes.Buffer{}
-	enc := json.NewEncoder(out)
-	enc.Encode(&logs)
-
-	return http.NewRequest(http.MethodPut, url, out)
+	return http.NewRequest(http.MethodPut, url, bytes.NewReader(logs.Messages))
 }
--- a/client_log.go
+++ b/client_log.go
@@ -28,9 +28,9 @@ type LogUploader interface {
 }
 
 type LogEntry struct {
-	Timestamp string `json:"timestamp"`
+	Timestamp string `json:"time"`
 	Level     string `json:"level"`
-	Message   string `json:"message"`
+	Message   string `json:"msg"`
 }
 
 type LogData struct {
--- a/daemon_test.go
+++ b/daemon_test.go
@@ -121,6 +121,7 @@ func TestDaemonRun(t *testing.T) {
 		0,
 	}
 	daemon := NewDaemon(dtc, NewMemStore())
+	DeploymentLogger = NewDeploymentLogManager(".")
 
 	go daemon.Run()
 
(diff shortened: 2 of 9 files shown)


deployment_logger.go
```diff
@@ -166,14 +166,86 @@ func (dlm DeploymentLogManager) Rotate() {
...
+			return file, nil
+		}
+	}
+	return "", nil
+}
+
+// GetLogs is returnig logs as a JSON string. Function is having the same
+// signature as json.Marshal() ([]byte, error)
+func (dlm DeploymentLogManager) GetLogs(deploymentID string) ([]byte, error) {
+	logFileName, err := dlm.findLogsForSpecificID(deploymentID)
```
bboozzoo: `dlm.findLogsForSpecificID()` can return `"", nil`, I don't see `""` being handled. Maybe return `os.ErrNotExist` from `findLogsForSpecificID()`?

client_log.go
```diff
@@ -27,16 +26,10 @@ type LogUploader interface {
...
-	Timestamp string `json:"timestamp"`
-	Level     string `json:"level"`
-	Message   string `json:"message"`
-}
-
 type LogData struct {
 	deviceID     string
 	deploymentID string
-	Messages     []LogEntry `json:"messages"`
+	Messages     []byte `json:"messages"`
```
bboozzoo: Any reason why we replace a structured log entry that fits what API expects with raw bytes sequence?

daemon_test.go
```diff
@@ -121,6 +121,7 @@ func TestDaemonRun(t *testing.T) {
 		0,
 	}
 	daemon := NewDaemon(dtc, NewMemStore())
+	DeploymentLogger = NewDeploymentLogManager(".")
```
bboozzoo: Try using a `ioutil.TempDir()` instead of current directory.

state.go
```diff
@@ -643,22 +666,24 @@ func (usr *UpdateStatusReportState) Handle(ctx *StateContext, c Controller) (Sta
...
-						Message:   "update failed",
-					},
-				})
+
+				logs, err := DeploymentLogger.GetLogs(usr.update.ID)
+				if err != nil {
+					log.Errorf("Failed to get deployment logs for deployment [%v]: %v",
+						usr.update.ID, err)
+				} else {
+					c.UploadLog(usr.update, logs)
```
bboozzoo: Either handle `UploadLog()` errors or leave a TODO note.

deployment_logger.go
```diff
@@ -166,14 +166,86 @@ func (dlm DeploymentLogManager) Rotate() {
...
 
+	// do we have some log files already
+	if len(logFiles) == 0 {
+		return
+	}
+
 	// check if we need to delete oldest file
 	for len(logFiles) > dlm.maxLogFiles {
 		os.Remove(logFiles[0])
 		logFiles = append(logFiles[1:])
```
bboozzoo: Can we replace this with just `logFiles = logFiles[1:]` as well?

client_log.go
```diff
@@ -27,16 +26,10 @@ type LogUploader interface {
 	Upload(api ApiRequester, server string, logs LogData) error
 }
 
-type LogEntry struct {
-	Timestamp string `json:"timestamp"`
```
bboozzoo: FYI, the API will try to parse Timestamp with default format for timestamps in JSON, that happens to be RFC3339.

deployment_logger.go
pasinskim: Thanks. Will fix it!

client_log.go
pasinskim: As agreed logs are written as a JSON entries to the log file. There is no need to marshal those and convert to `LogEntry` struct as the raw bytes returned from file are already JSON.

state.go
pasinskim: Actually, maybe you have some idea how to handle this error? We are already in the error state and if we can not upload logs there is anyway nothing we can do.

client_log.go
pasinskim: Yup. Will make sure this is the format that is written to log file.

deployment_logger.go
pasinskim: Will fix!

state.go
bboozzoo: Might need to retry the upload a couple of times if it fails. I'm not sure how important it is to have these logs.

Actually, another thought, what if succeed in reporting the status but fail to upload the log, or why is this process split in two separate messages? Maybe it'd be better to report status _and_ upload logs at the same time, in a single message.

daemon_test.go
pasinskim: Thanks. Fixing!

PR #86, opened 2016-07-12
Allow for skipping SSL certificate validation

bboozzoo:
It might be useful to skip SSL certificate validation when troubleshooting integration issues.

@kacf @pasinskim

### DIFF
--- a/config.go
+++ b/config.go
@@ -28,6 +28,7 @@ type menderConfig struct {
 	HttpsClient    struct {
 		Certificate string
 		Key         string
+		SkipVerify  bool
 	}
 	RootfsPartA         string
 	RootfsPartB         string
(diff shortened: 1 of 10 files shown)


main.go
```diff
@@ -132,6 +133,7 @@ func argsParse(args []string) (runOptionsType, error) {
 			*certKey,
 			*serverCert,
 			false,
+			*skipVerify,
```
kacf: I suggest using labels here, there are quite a few entries, including two bools right next to each other. If we somehow mess up the order later on, the compiler won't warn you as is.

config.go
```diff
@@ -79,6 +80,7 @@ func (c menderConfig) GetHttpConfig() httpsClientConfig {
 		c.HttpsClient.Key,
 		c.ServerCertificate,
 		c.ClientProtocol == "https",
+		c.HttpsClient.SkipVerify,
```
kacf: I suggest labels here as well.

kacf:
Otherwise +1 from me!

main.go
bboozzoo: Makes sense, will do that.

bboozzoo: fixed

config.go
bboozzoo: fixed

kacf:
trigger build

PR #87, opened 2016-07-14
Deployments API changes & Device ID

bboozzoo:
Small patch series that updates the API endpoints of deployments service. 

Also, Device ID is no longer needed as we no longer need to explicitly pass it in any outgoing request. The series addresses this by removing `DeviceID` from configuration.

@kacf


kacf:
trigger build

PR #90, opened 2016-08-08
Set debug as default log level for Mender client.

pasinskim:
Having log level set to debug we will make sure all log messages will
be available for the journal. Then we can configure journal which
level messages will be filtered out and which ones will be stored in
the system.

### DIFF
--- a/main.go
+++ b/main.go
@@ -192,7 +192,7 @@ func addLogFlags(f *flag.FlagSet) logOptionsType {
 		"'debug', 'info', 'warning', 'error', 'fatal' or 'panic'. "+
 		"Earlier log levels will also log the subsequent levels (so "+
 		"'debug' will log everything). The default log level is "+
-		"'warning'.")
+		"'info'.")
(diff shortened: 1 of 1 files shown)


pasinskim:
@bboozzoo @kacf @GregorioDiStefano what do you think about this one? The reason is some messages are filtered out by default and those are not accessible from `journalctl`. Having all messages being logged by default we can filter those using `journalctl` filtering options. We can also configure journal to set which log level messages we want to keep.

kacf:
I think debug is too high. Yes, you can use journalctl to filter after the fact, but the default behavior will be extremely noisy. If Mender is going to be a well behaved daemon I think it should be more quiet, logging only errors and info messages that are significant. Other daemons don't start out in debug mode, do they?

bboozzoo:
Debug or no debug, either way is fine with me.

pasinskim:
@kacf I will do info default level then.

PR #92, opened 2016-08-08
Finalize sending deployment logs and status to server.

pasinskim:
Use `maxReportSendingTries` to make sure that we are trying to send
status and logs finite number of times.
If we are not able to send status or logs we are marking update as
failed and trying to rollback (rollback needs to be implemented).

### DIFF
--- a/state.go
+++ b/state.go
@@ -612,8 +612,9 @@ func (ue *UpdateErrorState) Handle(ctx *StateContext, c Controller) (State, bool
 // update is deemed as failed.
 type UpdateStatusReportState struct {
 	CancellableState
-	update UpdateResponse
-	status string
+	update             UpdateResponse
+	status             string
+	triesSendingReport int
 }
 
 func NewUpdateStatusReportState(update UpdateResponse, status string) State {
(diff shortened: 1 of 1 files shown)


pasinskim:
@bboozzoo what do you think? We are right now trying to send status only finite number of times (hard-coded to `5`). The same is for logs. 
If this is failing, we are considering update as broken and need to roll-back (needs to be implemented).

kacf:
Looks good to me, but @bboozzoo should take a look too.

state.go
```diff
@@ -612,8 +612,9 @@ func (ue *UpdateErrorState) Handle(ctx *StateContext, c Controller) (State, bool
 // update is deemed as failed.
 type UpdateStatusReportState struct {
 	CancellableState
-	update UpdateResponse
-	status string
+	update             UpdateResponse
+	status             string
+	triesSendingReport int
```
bboozzoo: Not sure if there's any use in keeping retry counter in the state.

bboozzoo:
Other than the comment above, the rest looks sane. Have you tested it in a vm?

state.go
pasinskim: The reason is this line: https://github.com/mendersoftware/mender/pull/92/files#diff-a987904f2c16109d2be403c8d8d7f93eR703. If the waiting loop will be interrupted we won't reset the counter.

bboozzoo: I think we don't really care in this case, do we?

PR #94, opened 2016-08-09
Make sure that deployment logs are formatted correctly.

pasinskim:
At the moment server endpoint requires JSON log format as follows:
{"level":"info", "message":"log", "timestamp":"2016-08-09T12:03:53Z"}

We are adding new custom log formatter to make sure that logs are
formatted as expected by the server.


bboozzoo:
Seems to work :+1:

PR #95, opened 2016-08-10
close http.Response Body

bboozzoo:
make sure that we close Body of each http.Response

@kacf @pasinskim @GregorioDiStefano

### DIFF
--- a/client_update.go
+++ b/client_update.go
@@ -81,13 +81,16 @@ func (u *UpdateClient) FetchUpdate(api ApiRequester, url string) (io.ReadCloser,
 	log.Debugf("Received fetch update response %v+", r)
 
 	if r.StatusCode != http.StatusOK {
+		r.Body.Close()
 		log.Errorf("Error fetching shcheduled update info: code (%d)", r.StatusCode)
 		return nil, -1, errors.New("Error receiving scheduled update information.")
 	}
 
 	if r.ContentLength < 0 {
+		r.Body.Close()
 		return nil, -1, errors.New("Will not continue with unknown image size.")
 	} else if r.ContentLength < u.minImageSize {
+		r.Body.Close()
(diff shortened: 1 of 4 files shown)


client_update.go
```diff
@@ -86,8 +86,10 @@ func (u *UpdateClient) FetchUpdate(api ApiRequester, url string) (io.ReadCloser,
 	}
 
 	if r.ContentLength < 0 {
+		defer r.Body.Close()
```
pasinskim: What about closing this couple of lines earlier - here: https://github.com/mendersoftware/mender/pull/95/files#diff-c1a9cef21663a7f28d8d2eb327d47bffR80

bboozzoo: Uhh right, we should also close it there.

bboozzoo: fixed

pasinskim: Why we can not have just one `defer r.Body.Close()` after (or before) `log.Debugf("Received fetch update response %v+", r)`?

bboozzoo: Because we still want to receive the update ;) otherwise io stream will get closed when fetch update returns.

pasinskim: Crap. You are right. I did not follow the logic. In this case I think you can skip `defer` and close the stream explicitly. What is more, you are right that we need to close the stream in `func (u *UpdateInstallState) Handle(ctx *StateContext, c Controller) (State, bool)`.
Can you add something like below:

```
func (u *UpdateInstallState) Handle(ctx *StateContext, c Controller) (State, bool) {

	defer u.imagein.Close()

	// start deployment logging
	DeploymentLogger.Enable(u.update.ID)

```

bboozzoo: Right, fixed!

pasinskim:
Looks good now.

PR #96, opened 2016-08-11
Check if there is enough disk space for logs before starting update.

pasinskim:
If there is not enough space for storing deployment logs we need
to invalidate the update.

### DIFF
--- a/deployment_logger.go
+++ b/deployment_logger.go
@@ -107,9 +107,7 @@ func (dlm *DeploymentLogManager) haveEnoughSpaceForStoringLogs() bool {
 	// Available blocks * size per block = available space in bytes
 	availableSpace := stat.Bavail * uint64(stat.Bsize)
 	if availableSpace < dlm.minLogSizeBytes {
-		fmt.Printf("Not enough space for storing logs: %v (required: %v)",
-			availableSpace, dlm.minLogSizeBytes)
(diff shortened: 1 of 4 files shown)


bboozzoo:
Looks good to me.

deployment_logger.go
```diff
@@ -94,11 +99,32 @@ func (dlm DeploymentLogManager) WriteLog(log []byte) error {
...
+func (dlm *DeploymentLogManager) haveEnoughSpaceForStoringLogs() bool {
+	var stat syscall.Statfs_t
+	syscall.Statfs(dlm.logLocation, &stat)
+
+	// Available blocks * size per block = available space in bytes
+	availableSpace := stat.Bavail * uint64(stat.Bsize)
+	if availableSpace < dlm.minLogSizeBytes {
+		fmt.Printf("Not enough space for storing logs: %v (required: %v)",
+			availableSpace, dlm.minLogSizeBytes)
+		fmt.Printf("Directory: %v", dlm.logLocation)
```
kacf: You should use `log.Errorf()`, no?

kacf:
Otherwise looks good.

deployment_logger.go
pasinskim: Actually I shouldn't log at all. No space for logs anyway. Will just return error.

PR #97, opened 2016-08-11
Make sure to report failed update after rebooting with broken image.

pasinskim:
If we are installing broken image device is rebooted reaching
`we have upgrade info but has flag is not set` point.
Here we are sending no information to the server that last update was
broken and we are entering infinite loop of fetching and installing
broken update.
Right now we will try to report broken update so that server will be able
to remove it from the queue.

### DIFF
--- a/state.go
+++ b/state.go
@@ -548,14 +548,10 @@ func (a *AuthorizedState) Handle(ctx *StateContext, c Controller) (State, bool) 
 	}
 
 	// we have upgrade info but has flag is not set
+	// most probably last update failed
 	log.Infof("update info for deployment %v present, but update flag is not set",
 		sd.UpdateInfo.ID)
-
-	// starting from scratch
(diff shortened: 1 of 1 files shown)


bboozzoo:
IIRC we reboot only if we were able to set upgrade_avalable flag. So, at least in theory, this scenario should not happen right?

pasinskim:
@bboozzoo see https://tracker.mender.io/browse/MEN-589
It looks like we did reboot, then failed to start new image (as it was broken), and implicitly rollback to previous one. At this point we are not sending failed update status to server and server we are receiving the same (broken) image. Then reboot, broken update, ...

Does it make sense?

bboozzoo:
Discussed offline, things are clear now. IMO it's ok to merge it.

pasinskim:
@bboozzoo great. Thanks!

PR #100, opened 2016-08-22
Improve Mender state logic

pasinskim:
Add new update verification and rollback states.
Improve logging and fix default log level.
Refactor log and status sending. Fix some infinite loops while storing deployment logs.
Fix sending empty logs to server.
Update tests and documentation.

### DIFF
--- a/state.go
+++ b/state.go
@@ -877,7 +877,13 @@ func NewRollbackState(update UpdateResponse) State {
 }
 
 func (rs *RollbackState) Handle(ctx *StateContext, c Controller) (State, bool) {
-	// TODO: swap active partition
+	// swap active and inactive partitions
+	if err := c.Rollback(); err != nil {
+		log.Errorf("swapping active and inactive partitions failed: %s", err)
+		// TODO: what can we do here
+		return NewErrorState(NewFatalError(err)), false
+	}
+
 	if err := c.Reboot(); err != nil {
 		log.Errorf("error rebooting device: %v", err)
 		return NewErrorState(NewFatalError(err)), false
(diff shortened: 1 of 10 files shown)


pasinskim:
@kacf @GregorioDiStefano please take a look. There is quite a lot of changes, but I have verified manually that everything is working. 
I will add missing tests soon.

state.go
```diff
@@ -585,33 +583,29 @@ func (a *AuthorizedState) Handle(ctx *StateContext, c Controller) (State, bool)
 
 		// update prosess was initialized but stopped in the middle
 	case MenderStateUpdateFetch:
+		fallthrough
 	case MenderStateUpdateInstall:
```
kacf: You can use multiple values in one case block, like this:

```
case MenderStateUpdateFetch, MenderStateUpdateInstall:
```

state.go
```diff
@@ -307,7 +307,7 @@ func (uv *UpdateVerifyState) Handle(ctx *StateContext, c Controller) (State, boo
 		return NewUpdateErrorState(NewTransientError(err), uv.update), false
 	}
 
-	log.Debugf("handle update verify state")
+	log.Info("handle update verify state")
```
kacf: Did you mean to include this? It looks quite debug-like to me.

kacf:
It's hard to review all of it in detail, but looks good on the surface.

state.go
pasinskim: Yup. Will change that.

state.go
pasinskim: This is a small work-around for now if we are having no logs at all. I have added other logs to be stored in file, so can change the log level here.

pasinskim:
Verified with end to end tests. Merging now!

PR #102, opened 2016-09-01
Simplify copying update stream to inactive partition.

pasinskim:

### DIFF
--- a/device.go
+++ b/device.go
@@ -177,26 +177,10 @@ func writeToPartition(image io.Reader, imageSize int64, partition string) error 
 	}
 	defer partFd.Close()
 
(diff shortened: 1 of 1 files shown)


device.go
```diff
@@ -177,26 +177,10 @@ func writeToPartition(image io.Reader, imageSize int64, partition string) error
...
-				return fmt.Errorf("Error while writing to partition: %s: %s",
-					partition, writeErr.Error())
-			}
-		}
-
-		if readErr == io.EOF {
-			break
-		}
+	if _, err = io.Copy(partFd, image); err != nil {
+		return err
```
kacf: I should have known Go would have had something!

PR #103, opened 2016-09-02
Reduce binary size by removing debug symbols

GregorioDiStefano:
Shrinks the mender binary by 30%, but debug symbols are lost.

@kacf @bboozzoo


kacf:
+1, most local development will use `go build` which is not affected, so this shouldn't impact development, but will make the image smaller.

PR #106, opened 2016-09-05
Warn update will not be performed if update image id same as current

GregorioDiStefano:
@maciejmrowiec @bboozzoo

### DIFF
--- a/mender.go
+++ b/mender.go
@@ -286,6 +286,7 @@ func (m *mender) CheckUpdate() (*UpdateResponse, menderError) {
 	log.Debugf("received update response: %v", update)
 
 	if update.Image.YoctoID == currentImageID {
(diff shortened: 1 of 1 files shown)


kacf:
Makes sense to me. Normally this shouldn't happen, because the server should already know which version the client is running, but if it does happen it's good to have a message.

I'll give the people you mentioned a chance to look, but let me know if you want me to merge.

GregorioDiStefano:
some issues that still exist:

1) using the REST api, there is no feedback regarding the server already knowing that the client is already running this image id

2) once you deploy an image with the same image id, the status (deployments/<deployment id>/statistics) of your deployment is never updated: if you are only using the rest api, there is no way of knowing the reason why your deployment isnt being deployed.

pasinskim:
I'll merge this, but as we've agreed in this case we should return update successful to the server.

PR #108, opened 2016-09-09
inventory reporting & improvements

bboozzoo:
Introduce inventory support. `Controller` interface is extended with `InventoryRefresh()` method. New state `InventoryUpdateState` is added to the state machine, and is executed right after `UpdateCheckState` in case no update is present. Additionally, the state will be run when transitioning from `AuthorizedState`, so that we get a refresh of inventory data if Mender client was restarted.

A couple of extra commits introduce (incomplete) support for 'local' builds. Something I wanted to explore before, but had it only as a couple of stashed patches. The idea is that it would be nice to be able to build and run mender locally as part of one's development cycle. Right now we can build the client, but running it will bail out early in `HasUpgrade()`. Even if one adds `return false, nil` over there, the client will fail once again when attempting to run identity data script. Now, with inventory support, the inventory scripts will not be run because we just assume a specific prefix in the code. 

When building for local environment, the client should use paths that lead it back to the source tree, also any code that would normally reference u-boot or fiddle with partitions, should interact with local files instead.

If agent was written in C we could use a couple of defines/headers to work around that, however in Go the only reasonable method I found is build tags, hence the series introduces a 'local' build tag (triggered by `make LOCAL=1`) that for now only influences how system paths are found. So, with this PR, we only need this small patch

```
diff --git a/mender.go b/mender.go
index 28a3d41..ffc9c3c 100644
--- a/mender.go
+++ b/mender.go
@@ -197,6 +197,7 @@ func (m mender) GetCurrentImageID() string {
 }
 
 func (m *mender) HasUpgrade() (bool, menderError) {
+       return false, nil
        env, err := m.env.ReadEnv("upgrade_available")
        if err != nil {
                return false, NewFatalError(err)
```
to be able to build & run the client locally, and get it through authorization/update check/inventory update steps. It should be quite easy to add `device_local.go` and implement environment & image writing on local files. 

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -24,8 +24,17 @@ ifeq ($(V),1)
 BUILDV = -v
 endif
 
+TAGS =
+ifeq ($(LOCAL),1)
+TAGS += local
+endif
+
+ifneq ($(TAGS),)
+BUILDTAGS = -tags '$(TAGS)'
+endif
+
 build:
-	$(GO) build $(GO_LDFLAGS) $(BUILDV)
+	$(GO) build $(GO_LDFLAGS) $(BUILDV) $(BUILDTAGS)
 
 install:
 	$(GO) install $(GO_LDFLAGS) $(BUILDV)
--- a/Makefile
+++ b/Makefile
@@ -37,7 +37,7 @@ build:
 	$(GO) build $(GO_LDFLAGS) $(BUILDV) $(BUILDTAGS)
 
 install:
-	$(GO) install $(GO_LDFLAGS) $(BUILDV)
+	$(GO) install $(GO_LDFLAGS) $(BUILDV) $(BUILDTAGS)
 
 clean:
 	$(GO) clean
--- a/client_inventory.go
+++ b/client_inventory.go
@@ -1,0 +1,74 @@
+// Copyright 2016 Mender Software AS
+//
+//    Licensed under the Apache License, Version 2.0 (the "License");
+//    you may not use this file except in compliance with the License.
+//    You may obtain a copy of the License at
+//
+//        http://www.apache.org/licenses/LICENSE-2.0
+//
+//    Unless required by applicable law or agreed to in writing, software
+//    distributed under the License is distributed on an "AS IS" BASIS,
+//    WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+//    See the License for the specific language governing permissions and
+//    limitations under the License.
+package main
+
+import (
+	"bytes"
+	"encoding/json"
+	"net/http"
+
+	"github.com/mendersoftware/log"
+	"github.com/pkg/errors"
+)
+
+type InventoryReporter interface {
+	Submit(api ApiRequester, server string, data interface{}) error
+}
+
+type InventoryClient struct {
+}
+
+func NewInventoryClient() InventoryReporter {
+	return &InventoryClient{}
+}
+
+// Report status information to the backend
+func (i *InventoryClient) Submit(api ApiRequester, url string, data interface{}) error {
+	req, err := makeInventorySubmitRequest(url, data)
+	if err != nil {
+		return errors.Wrapf(err, "failed to prepare inventory submit request")
+	}
+
+	r, err := api.Do(req)
+	if err != nil {
+		log.Error("failed to submit inventory data: ", err)
+		return errors.Wrapf(err, "inventory submit failed")
+	}
+
+	defer r.Body.Close()
+
+	if r.StatusCode != http.StatusOK {
+		log.Errorf("got unexpected HTTP status when submitting to inventory: %v", r.StatusCode)
+		return errors.Errorf("inventory submit failed, bad status %v", r.StatusCode)
+	}
+	log.Debugf("inventory update sent, response %v", r)
+
+	return nil
+}
+
+func makeInventorySubmitRequest(server string, data interface{}) (*http.Request, error) {
+	url := buildApiURL(server, "/inventory/device/attributes")
+
+	out := &bytes.Buffer{}
+	enc := json.NewEncoder(out)
+	enc.Encode(&data)
+
+	hreq, err := http.NewRequest(http.MethodPatch, url, out)
+	if err != nil {
+		return nil, errors.Wrapf(err, "failed to create inventory HTTP request")
+	}
+
+	hreq.Header.Add("Content-Type", "application/json")
+	return hreq, nil
+}
--- a/client_inventory.go
+++ b/client_inventory.go
@@ -22,14 +22,14 @@ import (
 	"github.com/pkg/errors"
 )
 
-type InventoryReporter interface {
+type InventorySubmitter interface {
 	Submit(api ApiRequester, server string, data interface{}) error
 }
 
 type InventoryClient struct {
 }
 
-func NewInventoryClient() InventoryReporter {
+func NewInventoryClient() InventorySubmitter {
 	return &InventoryClient{}
 }
 
--- a/inventory.go
+++ b/inventory.go
@@ -1,0 +1,21 @@
+// Copyright 2016 Mender Software AS
+//
+//    Licensed under the Apache License, Version 2.0 (the "License");
+//    you may not use this file except in compliance with the License.
+//    You may obtain a copy of the License at
+//
+//        http://www.apache.org/licenses/LICENSE-2.0
+//
+//    Unless required by applicable law or agreed to in writing, software
+//    distributed under the License is distributed on an "AS IS" BASIS,
+//    WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+//    See the License for the specific language governing permissions and
+//    limitations under the License.
+package main
+
+type InventoryAttribute struct {
+	Name  string      `json:"name"`
+	Value interface{} `json:"value"`
+}
+
+type InventoryData []InventoryAttribute
--- a/inventory_data.go
+++ b/inventory_data.go
@@ -16,6 +16,7 @@ package main
 import (
 	"bufio"
 	"bytes"
+	"io"
 	"io/ioutil"
 	"os"
 	"path"
@@ -66,89 +67,92 @@ func listRunnable(dpath string) ([]string, error) {
 	return runnable, nil
 }
 
-type tempInventoryAttribute []string
-
-func (tia tempInventoryAttribute) ToInventoryAttribute(name string) InventoryAttribute {
-	if len(tia) > 1 {
-		return InventoryAttribute{Name: name, Value: []string(tia)}
-	} else if len(tia) == 1 {
-		return InventoryAttribute{Name: name, Value: tia[0]}
-	}
-	return InventoryAttribute{Name: name, Value: ""}
-}
-
-type tempInventoryData map[string]tempInventoryAttribute
-
-func (tid tempInventoryData) Add(name string, values ...string) {
-	_, has := tid[name]
-	if has {
-		tid[name] = append(tid[name], values...)
-	} else {
-		tid[name] = values
-	}
-}
-
-func (tid tempInventoryData) Append(idata tempInventoryData) {
-	for k, v := range idata {
-		tid.Add(k, []string(v)...)
-	}
-}
-
-func (tid tempInventoryData) ToInventoryData() InventoryData {
-	data := make(InventoryData, 0, len(tid))
-	for k, v := range tid {
-		data = append(data, v.ToInventoryAttribute(k))
-	}
-	return data
-}
-
 func (id *InventoryDataRunner) Get() (InventoryData, error) {
 	tools, err := listRunnable(id.dir)
 	if err != nil {
 		return nil, errors.Wrapf(err, "failed to list tools for inventory data")
 	}
 
-	idata := tempInventoryData{}
+	idec := NewInventoryDataDecoder()
 	for _, t := range tools {
-		data, err := id.cmd.Command(t).Output()
+		cmd := id.cmd.Command(t)
+		out, err := cmd.StdoutPipe()
 		if err != nil {
+			log.Errorf("failed to open stdout for inventory tool %s: %v", t, err)
+			continue
+		}
+
+		if err := cmd.Start(); err != nil {
 			log.Errorf("inventory tool %s failed with status: %v", t, err)
 			continue
 		}
 
-		td, err := parseInventoryData(data)
-		if err != nil {
+		if _, err := io.Copy(idec, out); err != nil {
 			log.Warnf("inventory tool %s returned unparsable output: %v", t, err)
 			continue
 		}
-		idata.Append(td)
+
+		if err := cmd.Wait(); err != nil {
+			log.Warnf("inventory tool %s wait failed: %v", t, err)
+		}
 	}
-	return idata.ToInventoryData(), nil
+	return idec.GetInventoryData(), nil
 }
 
-func parseInventoryData(data []byte) (tempInventoryData, error) {
-	td := tempInventoryData{}
+type InventoryDataDecoder struct {
+	data map[string]InventoryAttribute
+}
 
-	in := bufio.NewScanner(bytes.NewBuffer(data))
-	for in.Scan() {
-		line := in.Text()
+func NewInventoryDataDecoder() *InventoryDataDecoder {
+	return &InventoryDataDecoder{
+		make(map[string]InventoryAttribute),
+	}
+}
 
-		if len(line) == 0 {
+func (id *InventoryDataDecoder) GetInventoryData() InventoryData {
+	idata := make(InventoryData, 0, len(id.data))
+	for _, v := range id.data {
+		idata = append(idata, v)
+	}
+	return idata
+}
+
+func (id *InventoryDataDecoder) Write(p []byte) (n int, err error) {
+	r := bufio.NewScanner(bytes.NewBuffer(p))
+
+	for {
+		if !r.Scan() {
+			if r.Err() != nil {
+				return 0, r.Err()
+			} else {
+				return len(p), nil
+			}
+		}
+		ia, err := readAttr(r.Text())
+		if err != nil {
+			return 0, err
+		}
+
+		if data, ok := id.data[ia.Name]; ok {
+			switch data.Value.(type) {
+			case string:
+				newVal := []string{data.Value.(string), ia.Value.(string)}
+				id.data[ia.Name] = InventoryAttribute{ia.Name, newVal}
+			case []string:
+				newVal := append(data.Value.([]string), ia.Value.(string))
+				id.data[ia.Name] = InventoryAttribute{ia.Name, newVal}
+			}
 			continue
+		} else {
+			id.data[ia.Name] = ia
 		}
-
-		val := strings.SplitN(line, "=", 2)
-
-		if len(val) < 2 {
-			return nil, errors.Errorf("incorrect line '%s'", line)
-		}
-
-		td.Add(val[0], val[1])
 	}
+}
 
-	if len(td) == 0 {
-		return nil, errors.Errorf("obtained no output")
+func readAttr(p string) (InventoryAttribute, error) {
+	val := strings.SplitN(p, "=", 2)
+	if len(val) < 2 {
+		return InventoryAttribute{}, errors.Errorf("incorrect line '%s'", p)
 	}
-
-	return td, nil
+	return InventoryAttribute{val[0], val[1]}, nil
 }
--- a/inventory_data.go
+++ b/inventory_data.go
@@ -1,0 +1,162 @@
+// Copyright 2016 Mender Software AS
+//
+//    Licensed under the Apache License, Version 2.0 (the "License");
+//    you may not use this file except in compliance with the License.
+//    You may obtain a copy of the License at
+//
+//        http://www.apache.org/licenses/LICENSE-2.0
+//
+//    Unless required by applicable law or agreed to in writing, software
+//    distributed under the License is distributed on an "AS IS" BASIS,
+//    WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+//    See the License for the specific language governing permissions and
+//    limitations under the License.
+package main
+
+import (
+	"bufio"
+	"bytes"
+	"os"
+	"path"
+	"strings"
+	"syscall"
+
+	"github.com/mendersoftware/log"
+	"github.com/pkg/errors"
+)
+
+const (
+	inventoryToolPrefix = "mender-inventory-"
+)
+
+type InventoryDataGetter interface {
+	Get() (InventoryData, error)
+}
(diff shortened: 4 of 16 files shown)


mender.go
```diff
@@ -43,7 +45,10 @@ type Controller interface {
 const (
 	defaultManifestFile = "/etc/mender/build_mender"
```
kacf: Did you forget to update this one?

bboozzoo: On Fri, Sep 9, 2016 at 8:42 AM, Kristian Amlie <notifications@github.com>
wrote:

> In mender.go
> <https://github.com/mendersoftware/mender/pull/108#discussion_r78135057>:
>
> > @@ -43,7 +45,10 @@ type Controller interface {
> >  const (
> >  	defaultManifestFile = "/etc/mender/build_mender"
>
> Did you forget to update this one?
>
Yup, looks like it.

paths_local.go
```diff
@@ -0,0 +1,43 @@
...
+func getStateDirPath() string {
+	return getRunningBinaryPath()
+}
+
+func getConfDirPath() string {
+	return getRunningBinaryPath()
+}
+
+func getBinDirPath() string {
+	return path.Join(getRunningBinaryPath(), "support")
```
kacf: Is it right to append "support" to the end here? Won't the binary end up in the mender root directory when you run `go build`?

Makefile
```diff
@@ -24,8 +24,17 @@ ifeq ($(V),1)
...
+ifneq ($(TAGS),)
+BUILDTAGS = -tags '$(TAGS)'
+endif
+
 build:
-	$(GO) build $(GO_LDFLAGS) $(BUILDV)
+	$(GO) build $(GO_LDFLAGS) $(BUILDV) $(BUILDTAGS)
 
 install:
 	$(GO) install $(GO_LDFLAGS) $(BUILDV)
```
kacf: Should add `$(BUILDTAGS) here as well.

paths_local.go
bboozzoo: On Fri, Sep 9, 2016 at 8:47 AM, Kristian Amlie <notifications@github.com>
wrote:

> In paths_local.go
> <https://github.com/mendersoftware/mender/pull/108#discussion_r78135451>:
>
> > +}
> > +
> > +func getDataDirPath() string {
> > +	return path.Join(getRunningBinaryPath(), "support")
> > +}
> > +
> > +func getStateDirPath() string {
> > +	return getRunningBinaryPath()
> > +}
> > +
> > +func getConfDirPath() string {
> > +	return getRunningBinaryPath()
> > +}
> > +
> > +func getBinDirPath() string {
> > +	return path.Join(getRunningBinaryPath(), "support")
>
> Is it right to append "support" to the end here? Won't the binary end up
> in the mender root directory when you run go build?
>
It will. We also keep sample scripts/tools in $(srcdir)/support, so that's
where `getBinDirPath()` will point to. I run mender right from the source
directory, not sure what others do. Feel free to suggest alternatives.
Perhaps `os.Getwd()`?

kacf: Right, I see the problem... Maybe, now that we've introduced `/usr/share/mender`, that it would be appropriate to put the identity script somewhere in there as well. After all, the identity script is never interesting as a standalone command, only when executed by Mender, hence it makes sense to put in a place local to Mender. What do you think?

kacf:
Nice solution to the "local" problem. But I'm wondering whether the condition should be inverted, that the default should be "local", and Yocto will override it in its build. Otherwise `go build` will never do the right thing, I think? What do you think?

Makefile
bboozzoo: fixed

paths_local.go
bboozzoo: We could, it's a simple change. Though, it may require some work on meta-mender and documentation, so I'm not sure it's worth the effort.

bboozzoo:
> Nice solution to the "local" problem. But I'm wondering whether the condition should be inverted, that the default should be "local", and Yocto will override it in its build. Otherwise go build will never do the right thing, I think? What do you think?

IMO we should build the 'board' version of code by default. The 'local' variant is more like a hack for development purpose only. It'd be useful to mention it in HACKING.md, but it's not really useful unless someone is working on the code. @pasinskim thoughts?

kacf:
Perhaps you're right. This is particularly important is somebody is testing it out *outside* Yocto (and there are some potentially interested candidates for this). But yes, mentioning it in `HACKING.md` sounds like a good idea.

paths_local.go
kacf: It seems it's not mentioned in the documentation (which is a bit surprising...), so we're off the hook there. I propose we make the changes together with https://github.com/mendersoftware/meta-mender/pull/129. Better now than later.

bboozzoo:
Depends on https://github.com/mendersoftware/meta-mender/pull/129

kacf:
+1

client_inventory.go
```diff
@@ -0,0 +1,74 @@
...
+import (
+	"bytes"
+	"encoding/json"
+	"net/http"
+
+	"github.com/mendersoftware/log"
+	"github.com/pkg/errors"
+)
+
+type InventoryReporter interface {
```
pasinskim: Do we need an interface here?

client_inventory.go
```diff
@@ -0,0 +1,74 @@
...
+	Submit(api ApiRequester, server string, data interface{}) error
```
pasinskim: For naming convention, if there is just one method, we should name the interface `submiter`.

client_inventory.go
pasinskim: In fact this is already messy with the client interfaces. We have a `StatusReporter` in `client_status.go`, `LogUploader` in `client_log.go`, `AuthRequester` in `client_auth.go` and so on. In fact all are doing the same thing: they are formating request are sending it to the server. 
The idea of interfaces should be to have one (maybe `Requester`) and use it interchangeably in the client code. Something to think of, but this will be rather bigger refactoring.

bboozzoo: I don't agree. Each client takes a different data for submitting. If we implement what you suggest, the problem will be moved elsewhere as we'll need to have a bunch of formatters that format the data in a way suitable for submission.

> have one (maybe Requester) and use it interchangeably in the client code. 

it's right here:
```
// Mender API Client wrapper. A standard http.Client is compatible with this
// interface and can be used without further configuration where ApiRequester is
// expected. Instead of instantiating the client by yourself, one can also use a
// wrapper call NewApiClient() that sets up TLS handling according to passed
// configuration.
type ApiRequester interface {
	Do(req *http.Request) (*http.Response, error)
}
```
https://github.com/mendersoftware/mender/blob/master/client.go#L37

client_inventory.go
```diff
@@ -0,0 +1,74 @@
...
+
+type InventoryClient struct {
+}
+
+func NewInventoryClient() InventoryReporter {
+	return &InventoryClient{}
+}
+
+// Report status information to the backend
+func (i *InventoryClient) Submit(api ApiRequester, url string, data interface{}) error {
```
pasinskim: Look at for example `Upload` method in `client_log.go`. This is exact same code with the difference of how we are making request. Maybe we should unify this somehow instead of copy-paste. 
What I am thinking is:

```
type MakeRequest func(url string, data interface{}) (*http.Request, error)

type Requester interface {
	Request(api ApiRequester, server string, data interface{}, makeFunc MakeRequest)
}
```
We can also skip `data interface{}` above and assume `data` will be a part of receiver on which we will call `MakeRequest()` function.

client_inventory.go
```diff
@@ -0,0 +1,74 @@
...
+	if r.StatusCode != http.StatusOK {
+		log.Errorf("got unexpected HTTP status when submitting to inventory: %v", r.StatusCode)
+		return errors.Errorf("inventory submit failed, bad status %v", r.StatusCode)
+	}
+	log.Debugf("inventory update sent, response %v", r)
+
+	return nil
+}
+
+func makeInventorySubmitRequest(server string, data interface{}) (*http.Request, error) {
```
pasinskim: Some room for unification here as well. This is exact same code as in `client_status.go`.

inventory_data.go
```diff
@@ -0,0 +1,162 @@
...
+
+const (
+	inventoryToolPrefix = "mender-inventory-"
+)
+
+type InventoryDataGetter interface {
+	Get() (InventoryData, error)
+}
+
+func NewInventoryDataGetter(scriptsDir string) InventoryDataGetter {
```
pasinskim: I might be wrong here, but shouldn't we name it `NewInventoryDataRunner`? This indeed returns `InventoryDataGetter` interface which `InventoryDataRunner` implements, but in fact we are not creating interface itself.

client_inventory.go
bboozzoo: I do not see how that improves things. Instead of 4-5 client structs, we'll have 4-5 `makeFunc`s. Plus, you need to account for interpretation of return statuses. For instance log & status expect 204, inventory expects 200, update expects 204. You'd have to wrap that, again, in some endpoint specific code that would be passed as (5th?) argument to `Request()`. Not really useful, don't you agree?

Besides, each client should be so simple that we could easily drop instantiaing update/log/status clients in `Mender` constructor, and just create them on demand, even using a struct initializer, like you're supposed to do it using the 'Go-way'.

inventory_data.go
```diff
@@ -0,0 +1,162 @@
...
+
+	"github.com/mendersoftware/log"
+	"github.com/pkg/errors"
+)
+
+const (
+	inventoryToolPrefix = "mender-inventory-"
+)
+
+type InventoryDataGetter interface {
```
pasinskim: Again, do we need interface here?

inventory_data.go
```diff
@@ -0,0 +1,162 @@
...
+	dp, err := os.Open(dpath)
+	if err != nil {
+		return nil, errors.Wrapf(err, "failed to open %s", dpath)
+	}
+
+	finfos, err := dp.Readdir(0)
+	if err != nil {
+		// don't care about any FileInfo that were read up to this point
+		return nil, errors.Wrapf(err, "failed to readdir")
+	}
```
pasinskim: Use `ReadDir(dpath)` instead.

bboozzoo: Where `ReadDir()` comes from? I can only see `Readdir()`, https://godoc.org/os#File.Readdir

client_inventory.go
bboozzoo: Renamed to `InventorySubmitter`

inventory_data.go
bboozzoo: removed

inventory_data.go
bboozzoo: renamed to `NewInventoryDataRunner()`

inventory.go
```diff
@@ -0,0 +1,21 @@
...
+//    Unless required by applicable law or agreed to in writing, software
+//    distributed under the License is distributed on an "AS IS" BASIS,
+//    WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+//    See the License for the specific language governing permissions and
+//    limitations under the License.
+package main
+
+type InventoryAttribute struct {
+	Name  string      `json:"name"`
+	Value interface{} `json:"value"`
```
pasinskim: Why do we need `interface{}` here? We are scanning text output and storing everything as strings, right?

inventory.go
```diff
@@ -0,0 +1,21 @@
...
+}
+
+type InventoryData []InventoryAttribute
```
pasinskim: I think having here a `type InventoryData map[string][]InventoryAttribute` or even `type InventoryData map[string][]string` would simplify implementation a lot. Then whole reading part could be done as: 

```
type InventoryData map[string][]interface{}

func (id InventoryData) Write(p []byte) (n int, err error) {
	r := bufio.NewReader(bytes.NewBuffer(p))
	for {
		data, err := r.ReadBytes('\n')
		if err == io.EOF {
			return len(p), nil
		} else if err != nil {
			return 0, err
		}
		ia, _ := readAttr(data[:len(data)-1])
		id[ia.Name] = append(id[ia.Name], ia.Value)
	}
}

func readAttr(p []byte) (*InventoryAttribute, error) {
	val := strings.SplitN(string(p), "=", 2)
	// TOOD: check for errors
	return &InventoryAttribute{val[0], val[1]}, nil
}
```

Having this you can read data using something like:
```
mr := io.MultiReader(listRunnable) //os.File implements io.Reader
io.Copy(new(InventoryData), mr) //now InventoryData implements io.Writer
```

inventory.go
bboozzoo: See inventory spec. The value can either be a single string, or a list of strings. IIRC you've come up with  this design. Multiple values with same key were supposed to be merged into a single list https://tracker.mender.io/browse/MEN-359?focusedCommentId=74308&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-74308

inventory_data.go
```diff
@@ -0,0 +1,162 @@
...
+type InventoryDataRunner struct {
+	dir string
+	cmd Commander
+}
+
+func listRunnable(dpath string) ([]string, error) {
+	dp, err := os.Open(dpath)
+	if err != nil {
+		return nil, errors.Wrapf(err, "failed to open %s", dpath)
+	}
```
pasinskim: Should we have `defer dp.Close()` here?

inventory_data.go
```diff
@@ -0,0 +1,162 @@
...
+		runnable = append(runnable, path.Join(dpath, finfo.Name()))
+	}
+
+	return runnable, nil
+}
+
+type tempInventoryAttribute []string
+
+func (tia tempInventoryAttribute) ToInventoryAttribute(name string) InventoryAttribute {
+	if len(tia) > 1 {
```
pasinskim: Do we need to export `ToInventoryAttribute`?

inventory_data.go
```diff
@@ -0,0 +1,162 @@
...
+		return InventoryAttribute{Name: name, Value: []string(tia)}
+	} else if len(tia) == 1 {
+		return InventoryAttribute{Name: name, Value: tia[0]}
+	}
+	return InventoryAttribute{Name: name, Value: ""}
+}
+
+type tempInventoryData map[string]tempInventoryAttribute
+
+func (tid tempInventoryData) Add(name string, values ...string) {
```
pasinskim: Same here. If we are not intended to export `Add` let's just use `add`.

inventory_data.go
```diff
@@ -0,0 +1,162 @@
...
+	_, has := tid[name]
+	if has {
+		tid[name] = append(tid[name], values...)
+	} else {
+		tid[name] = values
+	}
+}
+
+func (tid tempInventoryData) Append(idata tempInventoryData) {
```
pasinskim: Are we exporting?

inventory_data.go
```diff
@@ -0,0 +1,162 @@
...
+	for k, v := range idata {
+		tid.Add(k, []string(v)...)
+	}
+}
+
+func (tid tempInventoryData) ToInventoryData() InventoryData {
```
pasinskim: Same here...

inventory.go
bboozzoo: `InventoryAttribute` is passed directly to JSON encoder and sent to inventory service. With your proposal each inventory value will be a `[]string`. Is that desired?

 @maciejmrowiec @marekswiecznik  what should device attributes look like  here?

inventory_data.go
bboozzoo: `tempInventoryData` is clearly not exported

state.go
```diff
@@ -188,6 +188,12 @@ var (
 		},
 	}
 
+	inventoryUpdateState = &InventoryUpdateState{
+		BaseState{
+			id: MenderStateInventoryUpdate,
+		},
+	}
+
```
pasinskim: Please update also the state machine schematic at the very top of the file. It would be nice to have it up to date.

inventory_data.go
pasinskim: See here: https://golang.org/pkg/io/ioutil/#ReadDir. It in fact wraps `Readdir()`.

client_inventory.go
pasinskim: I do not agree actually. This is about consistency more. Receiver on which behalf you will call `Request()` or something else might (and will) be specific for each client. This is more about having some convention here as code is very similar.

client_inventory.go
pasinskim: I've been more thinking that we are repeating the same or similar code in couple of places already. Also the naming is different everywhere. Even though all clients are doing more or less the same things one is named `LogUploader`, other one `AuthRequester` but it boils down to: get data, endpoint url and send request. I've been thinking about abstracting this a bit more.

inventory.go
pasinskim: Yup. But isn't single string also `[]string`? At the end it always is slice of strings with 0 or more elements (whether we should support 0 is though different discussion).

bboozzoo: `[]string{"foo"} ` becomes `["foo"]` in JSON, while `string(foo)` encodes to `"foo"`. Given the outcome of yours and @eysteins discussion, multiple entries are wrapped into a list. On the other side of the system, the API expects attributes to come in this format `{"name": "foo", "value": <valid JSON>}`. Attributes appearing multiple times will just overwrite it's previous value. Eventually, the client can either send everything as a list, or try to preserve the 'format' you came up with.

I have no opinion here, best if you discuss that with @eysteins @maciejmrowiec and other interested parties. It would be nice if you decide on something in soon enough for me to update the PR, otherwise we'll have another sprint with incomplete tasks.

inventory_data.go
bboozzoo: `ToInventoryAttribute` just follows the convention of naming methods that are of some types 'public' API. Also keep in mind that all of this is part of 'main' package, hence it's not really exported. Renaming the method to `toInventoryAttribute` will not hide it from other files either.

inventory.go
bboozzoo: Took another look at your proposal, I'm afraid it does not relate to what we want to do.
1. we're not reading files, but running external tools and fetching their output
2. we'd have to fork multiple `mender-inventory-*` tool instances and collect output from pipe of each - this complicates things via somewhat flaky pipe handling
3. InventoryData is passed directly to JSON encoder (in your case, we'd still need an intermediate type to conform to backend API, basically implement the whole 'conversion' just like it's done for `tempInventoryData`
4. it's not clear what you want to do with same key but produced by different tools

Thanks for reminding me about `strings.SplitN()` though. IMO, we should use it in identity data handling too.

state.go
```diff
@@ -569,7 +575,7 @@ func (a *AuthorizedState) Handle(ctx *StateContext, c Controller) (State, bool)
 	// handle easy case first, no update info present, means no update in progress
 	if err != nil && os.IsNotExist(err) {
 		log.Debug("no update in progress, proceed")
-		return updateCheckWaitState, false
+		return inventoryUpdateState, false
```
bboozzoo: State transitions go like this:
* `init -> bootstrap -> authorized --(no update)--> inventory update -> update check wait`
* and after a complete update cycle
`init -> boostrap -> authorized -- (with update) --> <apply, commit, etc.> (clear update) -> update status report -> init` (see above)
* or with errors
`init -> boostrap -> authorized -- <some update states> -> error state (clear update) -> init` (see above)

I believe I asked for input from you and @kacf when starting this task, the recommendation was to add another state into existing SM. I'm also not sure if it's ok to send inventory reports while applying an update. We would need extra synchronization to prevent this from happening if inventory was to run in parallel.

inventory.go
pasinskim: I see this now. I thought JSON marshaler is bit smarter and detects that there is a single string inside slice. From the other hand this would get things even worse. Anyway, let's stick to what was agreed.

inventory.go
pasinskim: Threat my proposal as example here. I see why we might need structure, so then the example will be more like:
```
type InventoryAttribute struct {
	Name  string      `json:"name"`
	Value interface{} `json:"value"`
}

type InventoryData struct {
	data map[string]InventoryAttribute
}

func (id *InventoryData) Write(p []byte) (n int, err error) {
	r := bufio.NewReader(bytes.NewBuffer(p))
	for {
		data, err := r.ReadBytes('\n')
		if err == io.EOF {
			return len(p), nil
		} else if err != nil {
			return 0, err
		}
		ia, _ := readAttr(data[:len(data)-1])
		if data, ok := id.data[ia.Name]; ok {
			switch data.Value.(type) {
			case string:
				newVal := []string{data.Value.(string), ia.Value.(string)}
				id.data[ia.Name] = InventoryAttribute{ia.Name, newVal}
			case []string:
				newVal := append(data.Value.([]string), ia.Value.(string))
				id.data[ia.Name] = InventoryAttribute{ia.Name, newVal}
			}
			continue
		}
		id.data[ia.Name] = ia
	}
}

func readAttr(p []byte) (InventoryAttribute, error) {
	val := strings.SplitN(string(p), "=", 2)
	//TODO: error handling
	return InventoryAttribute{val[0], val[1]}, nil
}
```

It should generate exact same JSON as your proposal. 
Next, I think there is slight misunderstanding of how to use `MultiReader` and `io.Copy`. As an argument to `MultiReader` we can take `StdoutPipe` from `exec` which should be fine to handle or we can use `bytes.NewBuffer()` which implements `io.Reader` interface.
We can of course also take `Output` as you did and call `func (id *InventoryData) Write(p []byte) (n int, err error)` directly for each script.
As for the last, I guess it is the same issue with existing design too. Since we are not distinguishing which script is producing which output IMHO we should threat those equally (no difference if there are multiple same values from one script or from multiple).

To sum up, I come up with some different proposal as I have seen existing one not very intuitive. I agree, though that you can say the same thing to the one I've proposed...

inventory_data.go
pasinskim: Agree. But this is actually something we should fix as well. Probably we should move some parts of the code into separate packages as it gets messy with what we are having right now.

state.go
pasinskim: Agree for now. What I've been thinking (and it is actually my bad that I am late to the party) is that in future maybe we will need some configuration option, how frequently you will need to update inventory (same as how often do you want to ask for new updates).
Then we will have to refactor this as there is no chance we will be able to do so (interval handling for inventory sending).

inventory_data.go
```diff
@@ -0,0 +1,158 @@
...
+			continue
+		} else {
+			id.data[ia.Name] = ia
+		}
+	}
+}
+
+func readAttr(p string) (InventoryAttribute, error) {
+	val := strings.SplitN(p, "=", 2)
+	if len(val) < 2 {
```
pasinskim: Shouldn't we have `if len(val) != 2` here?

bboozzoo: Can't have more than 2 because of `SplitN`, can't have less than 0, so `!= 2` will have the same effect as `< 2`. I can change it to `!= 2` if that's needed.

pasinskim: Right, I haven't noticed that there is `SplitN` used.

PR #111, opened 2016-09-14
client: enable HTTP/2 for client requests

bboozzoo:
Enable HTTP/2 support. Falls back to HTTP/1.1 if server lacks proper
support.


pasinskim:
Looks OK to me.

PR #113, opened 2016-09-19
MEN-626 Return success in case of installing image with existing id.

pasinskim:
If we are installing image which id is the same as currently running
we are reporting update success to the server.


pasinskim:
@GregorioDiStefano @bboozzoo this is a workaround for the case where we are trying to install image with the same id as we are currently running.
Instead of silently skip installing it we are reporting success to the server.

PR #117, opened 2016-09-20
Add required `device_type` to inventory data.

pasinskim:

### DIFF
--- a/inventory_data.go
+++ b/inventory_data.go
@@ -68,6 +68,12 @@ func listRunnable(dpath string) ([]string, error) {
 }
 
 func (id *InventoryDataRunner) AddDeviceType(device string, inv InventoryData) error {
+	if inv == nil {
+		inv = make(InventoryData, 0, 1)
+		inv = append(inv, InventoryAttribute{Name: "device_type", Value: device})
+		return nil
+	}
+
 	haveDevType := false
 	for _, attr := range inv {
 		if strings.Compare(attr.Name, "device_type") == 0 {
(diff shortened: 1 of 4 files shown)


mender.go
```diff
@@ -165,35 +165,41 @@ func NewMender(config menderConfig, pieces MenderPieces) (*mender, error) {
...
 	}
 	if err := scanner.Err(); err != nil {
 		log.Error(err)
 	}
-	return imageID
+	return ""
+}
+
+func (m mender) GetCurrentImageID() string {
+	return m.getManifestData("IMAGE_ID")
```
bboozzoo: are we interested in reporting IMAGE_ID too?

inventory_data.go
```diff
@@ -67,6 +67,20 @@ func listRunnable(dpath string) ([]string, error) {
 	return runnable, nil
 }
 
+func (id *InventoryDataRunner) AddDeviceType(device string, inv InventoryData) error {
```
bboozzoo: how about adding a separate method to InventoryData that adds custom keys, say:
```
func (i *InventoryData) AddAttributes(map[string]interface{}) {
    ...
}
```

mender.go
pasinskim: Since we are having it already maybe we can report it as well.

inventory_data.go
pasinskim: This is a bit tricky. I wanted to add a device type only it this is not there already. Maybe instead we should replace existing one with the arbitrary one you are adding here.

PR #120, opened 2016-09-21
Make sure to remove authentication token if device is unauthorized.

pasinskim:


bboozzoo:
Add tests for mender?

pasinskim:
@bboozzoo I've added test checking that the token is removed when client is receiving `StatusUnauthorized`.

PR #121, opened 2016-09-21
state: use time based limits for update state report attempts

bboozzoo:
Retry update status report attempts for 5 minutes, after that return an
error and assume that reporting failed.

### DIFF
--- a/state.go
+++ b/state.go
@@ -735,10 +735,16 @@ func sendStatus(update UpdateResponse, status string, c Controller) menderError 
 	return c.ReportUpdateStatus(update, status)
 }
 
-const maxReportSendingTries = 5
+var maxReportSendingTime = 5 * time.Minute
 
 func (usr *UpdateStatusReportState) trySend(send SendData, c Controller) (error, bool) {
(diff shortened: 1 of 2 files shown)


kacf:
You are using time based values to calculate the maximum number of tries, but actual time is not used while reporting. If the reason you cannot report requires sockets to time out, each attempt will be time consuming, and this could go on for a while, if I understand it correctly. Would it be better to use actual time to decide when to stop?

(btw, I still think this is better than what we had before, just wondering if we should pursue it or not)

bboozzoo:
I was thinking more in the lines of _do that many polls to cover at least 5minutes, given current poll interval_. This way we'll do at most _n_ polls. Calculating actual time taken by the whole process may be a bit troublesome. I figured that if a server takes a while to respond (default `http.Client{}` has a 0 timeout) or there's a bad network, we may end up trying only a couple of times before our time slot expires.

BTW. about timeouts, we really should set one for the client

kacf:
Alright, makes sense. I'll merge this then.

PR #126, opened 2016-09-28
Artifacts support & local client development

bboozzoo:
Here it comes. Bulk of the changes is update artifact support. 

Included also some extra changes that enhance local client development. Usage examples and a brief intro is provided in HACKING.md.


device.go
```diff
@@ -92,21 +81,35 @@ func (d *device) InstallUpdate(image io.ReadCloser, size int64) error {
...
-	partitionSize, err := d.getPartitionSize(inactivePartition)
-	if err != nil {
+	if bsz, err := b.Size(); err != nil {
+		log.Errorf("failed to read size of block device %s: %v",
+			inactivePartition, err)
 		return err
+	} else if bsz < uint64(size) {
+		log.Errorf("update (%v bytes) is larger than the size of device %s (%v bytes)",
+			size, inactivePartition, bsz)
+		return syscall.ENOSPC
```
pasinskim: Hope we are not importing `syscall` just to return this error.

device.go
```diff
@@ -92,21 +81,35 @@ func (d *device) InstallUpdate(image io.ReadCloser, size int64) error {
...
+	w, err := io.Copy(b, image)
+	if err != nil {
+		log.Errorf("failed to write image data to device %v: %v",
+			inactivePartition, err)
 	}
 
-	if size <= partitionSize {
-		if err := writeToPartition(image, size, inactivePartition); err != nil {
-			return err
+	log.Infof("written %v/%v bytes of update to device %v",
```
pasinskim: s/written/wrote/

device.go
```diff
@@ -92,21 +81,35 @@ func (d *device) InstallUpdate(image io.ReadCloser, size int64) error {
...
+		w, size, inactivePartition)
+
+	if cerr := b.Close(); cerr != nil {
```
pasinskim: Should this be a `defer` function? What will happen if there is a `io.Copy()` error?

dirstore.go
```diff
@@ -72,7 +72,7 @@ func (d DirStore) WriteAll(name string, data []byte) error {
 func (d DirStore) OpenRead(name string) (io.ReadCloser, error) {
 	f, err := os.Open(path.Join(d.basepath, name))
 	if err != nil {
-		log.Errorf("I/O read error for entry %v: %v", name, err)
+		log.Debugf("I/O read error for entry %v: %v", name, err)
```
pasinskim: Why this is a debug now?

rootfs.go
```diff
@@ -74,3 +76,20 @@ func doRootfs(device UInstaller, args runOptionsType) error {
 
 	return device.EnableUpdatedPartition()
 }
+
+// Returns a byte stream of the fiven file, and also returns the size of the
```
pasinskim: s/fiven/given/

pasinskim: Also `FetchUpdateFromFile returns ...`

utils/limited_writer.go
```diff
@@ -0,0 +1,51 @@
...
+	W io.Writer // underlying writer
+	N uint64    // number of bytes remaining
+}
+
+func (lw *LimitedWriter) Write(p []byte) (int, error) {
+	if lw.W == nil {
+		return 0, syscall.EBADF
+	}
+	var selferr error
+	towrite := p
```
pasinskim: `toWrite`

device_local.go
```diff
@@ -0,0 +1,87 @@
...
+	device := device{env, sc, partitions}
+	return &device
+}
+
+func (d *device) Reboot() error {
+	log.Infof("reboot")
+	return nil
+}
+
+func (d *device) Rollback() error {
```
kacf: `Reboot()` I get, but why is for example `Rollback` and `InstallUpdate` necessary? Can't you simply modify the configuration file containing partitions to point to files instead?

device_local.go
```diff
@@ -43,7 +47,28 @@ func NewDevice(env BootEnvReadWriter, sc StatCommander, config deviceConfig) *de
 		active:            "",
 		inactive:          "",
 	}
+
+	if partitions.rootfsPartA == "" {
+		partitions.rootfsPartA = "mmcblk0p1"
+	}
+	if partitions.rootfsPartB == "" {
+		partitions.rootfsPartB = "mmcblk0p2"
+	}
```
kacf: Same here. I think we should use the configuration file for this.

device.go
kacf: There doesn't seem to be any more standard way of returning this type of error. At least I can't find it.

HACKING.md
```diff
@@ -50,3 +52,106 @@ Overview of Mender client architecture.
...
+│           ├── mender-inventory-hostinfo
+│           └── mender-inventory-network
+└── var
+    └── lib
+        └── mender                <-- client state directory
+            ├── authseq           <-- authorization sequence
+            ├── authtentoken      <-- tenan token
+            ├── authtoken         <-- API authorization token
+            ├── deployments.0001.31e6216f-1180-44c8-8e3a-2950e557bb46.log  <-- logs from ..
+            ├── deployments.0002.adcc2dc4-6f98-43e4-92a9-dfbae40bc059.log  <--   failed deployments
```
bboozzoo: fixed

HACKING.md
```diff
@@ -50,3 +52,106 @@ Overview of Mender client architecture.
 +------------------+       +------------------------+        +--------------------+
 
 ```
+
+## Hacking on Mender client locally
+
+Build mender with support for local development:
```
bboozzoo: I'm not sure about this. It seems sufficiently readable for me. Your proposal just makes it more verbose, something you want to avoid when providing instructions.

device.go
```diff
@@ -152,3 +152,16 @@ func (d *device) CommitUpdate() error {
 	// For now set only appropriate boot flags
 	return d.WriteEnv(BootVars{"upgrade_available": "0"})
 }
+
+func (d *device) HasUpdate() (bool, error) {
```
kacf: Am I right that this goes together with the commit "mender: all BootEnv handling moved to device"? Will that commit even compile without this? In general I think such functionality moves should either be grouped in the same commit or ordered in a way so that the code always compiles.

(Github doesn't always order things correctly though, so apologies if I'm wrong)

device_local.go
```diff
@@ -126,6 +126,19 @@ func (d *device) CommitUpdate() error {
 	return d.WriteEnv(BootVars{"upgrade_available": "0"})
 }
 
+func (d *device) HasUpdate() (bool, error) {
```
kacf: Same here, regarding compilability.

LIC_FILES_CHKSUM.sha256
```diff
@@ -19,7 +19,3 @@ bbb303820971c294a9a8e5eba5affcf1379036e877ea61c11cbf9400b2949483  vendor/github.
 2eb550be6801c1ea434feba53bf6d12e7c71c90253e0a9de4a4f46cf88b56477  vendor/github.com/pmezard/go-difflib/LICENSE
 2d36597f7117c38b006835ae7f537487207d8ec407aa9d9980794b2030cbc067  vendor/golang.org/x/sys/LICENSE
 2d36597f7117c38b006835ae7f537487207d8ec407aa9d9980794b2030cbc067  vendor/golang.org/x/net/LICENSE
-0634b008cee55ca01f0888d2f5aba2d34e66c3f52c31a4e16a5d5d33d0c2a03e  vendor/github.com/bmatsuo/lmdb-go/LICENSE.md
```
kacf: Please add a comment in the commit message why we are removing it.

block_device.go
```diff
@@ -0,0 +1,100 @@
...
+}
+
+// Write data `p` to underlying block device. Will automatically open device for
+// write. Otherwise, behaves like io.Writer.
+func (bd *BlockDevice) Write(p []byte) (int, error) {
+	if bd.out == nil {
+		log.Infof("opening device %s for writing", bd.Path)
+		out, err := os.OpenFile(bd.Path, os.O_WRONLY, 0)
+		if err != nil {
+			return 0, err
```
bboozzoo: It's a low level interface, I'd rather leave the error unwrapped here. If open fails you'll get an error like `open /dev/foobar: permission denied` (just an example).

block_device.go
```diff
@@ -0,0 +1,100 @@
...
+		}
+
+		size, err := BlockDeviceSize(out)
+		if err != nil {
+			log.Errorf("failed to read block device size: %v", err)
+			defer out.Close()
+			return 0, err
+		}
+		log.Infof("   partition %s size: %v", bd.Path, size)
```
bboozzoo: fixed

block_device.go
```diff
@@ -0,0 +1,100 @@
...
+		log.Infof("opening device %s for writing", bd.Path)
+		out, err := os.OpenFile(bd.Path, os.O_WRONLY, 0)
+		if err != nil {
+			return 0, err
+		}
+
+		size, err := BlockDeviceSize(out)
+		if err != nil {
+			log.Errorf("failed to read block device size: %v", err)
+			defer out.Close()
```
bboozzoo: removed defer

block_device.go
```diff
@@ -0,0 +1,100 @@
...
+func (bd *BlockDevice) Close() error {
+	if bd.out != nil {
+		if err := bd.out.Sync(); err != nil {
+			log.Errorf("failed to fsync partition %s: %v", bd.Path, err)
+			return err
+		}
+		if err := bd.out.Close(); err != nil {
+			log.Errorf("failed to close partition %s: %v", bd.Path, err)
+			return err
+		}
```
bboozzoo: fixed

bboozzoo: FYI, `Close()` errors are just logged now, there's no error returned.

block_device.go
```diff
@@ -0,0 +1,100 @@
...
+// Get the size of underlying block device. Automatically opens a new fd in
+// O_RDONLY mode.
+func (bd *BlockDevice) Size() (uint64, error) {
+	out, err := os.OpenFile(bd.Path, os.O_RDONLY, 0)
+	if err != nil {
+		return 0, err
+	}
+	defer out.Close()
+
+	size, err := BlockDeviceSize(out)
```
bboozzoo: fixed

block_device.go
kacf: I think I agree with @bboozzoo, this is too low level to add additional information. The upper code layers should do that.

HACKING.md
```diff
@@ -50,3 +52,106 @@ Overview of Mender client architecture.
...
+```
+
+Provide a dummy tenant token:
+
+```
+$ echo dummy > $PWD/prefix/var/lib/mender/authtentoken
+```
+
+Use example config file as template (edit as needed, ex. point the client to
+http://docker.mender.io:9080):
```
bboozzoo: 9080 is still there in docker-compose setup

block_device.go
```diff
@@ -0,0 +1,100 @@
...
+
+import (
+	"os"
+
+	"github.com/mendersoftware/log"
+	"github.com/mendersoftware/mender/utils"
+)
+
+var (
+	BlockDeviceSize BlockDeviceGetSizeFunc = getBlockDeviceSize
```
bboozzoo: You mean `BlockDeviceGetSizeOf` or `BlockDeviceSizeOf` ?

bboozzoo: Renamed to `BlockDeviceGetSizeOf`

device.go
bboozzoo: @pasinskim we are

block_device.go
```diff
@@ -0,0 +1,100 @@
...
+)
+
+// Helper for obtaining the size of a block device.
+type BlockDeviceGetSizeFunc func(file *os.File) (uint64, error)
```
bboozzoo: fixed

block_device.go
```diff
@@ -0,0 +1,100 @@
...
+
+// Wrapper for a block device, implements io.Writer and io.Closer interfaces.
+type BlockDevice struct {
+	Path string   // device path, ex. /dev/mmcblk0p1
+	out  *os.File // os.File for writing
+	size uint64
```
bboozzoo: fixed, also removed `size`

dirstore.go
bboozzoo: First, it would always appear when starting mender, makes one wonder why there's an error reported right away. Second, the error is returned, so it's better we allow the caller to log the error as needed.

rootfs.go
bboozzoo: fixed

utils/limited_writer.go
bboozzoo: fixed

device_local.go
bboozzoo: Can you elaborate? The configuration is read elsewhere and the fields with be populated with what was read at that point. This is just a fallback in case partitions were not provided in the config file.

device_local.go
bboozzoo: I'm not sure I understand your concern here, can you elaborate?

LIC_FILES_CHKSUM.sha256
bboozzoo: ok

HACKING.md
```diff
@@ -50,3 +52,106 @@ Overview of Mender client architecture.
...
+
+```
+$ cp mender.conf.example $PWD/prefix/etc/mender/mender.conf
+```
+
+Use example device identity script:
```
bboozzoo: Mentioning authentication without explaining how it works in the system is pointless. For all the reader knows, the script provides device identity.

device.go
bboozzoo: Reordered and squashed a number of commits (possibly blowing up github review in the process).

bboozzoo:
I understand you guys feel uneasy about local development support, but I cannot really see how it's possible to play with the code without getting a Yocto build running and using only qemu (BBB is still a no go?). I would not call that a quick and easy development path.

If that's a large problem I can skip most of the patches adding  a local variant and keep those in my local fork. It would be nice though, if the patch shuffling interfaces and structs around was merged. It's actually an enabler for the rest of the 'local' stuff.

device_local.go
kacf: I'm thinking about `mender.conf`. It contains the fields `RootfsPartA` and `RootfsPartB`. Can't we just use this configuration file, and point to, say `/home/bboozzoo/go/mender/prefix/dev/mmcblkp1`? If they point to this location, no special handling for local mode should be necessary, or?

kacf:
> I understand you guys feel uneasy about local development support, but I cannot really see how it's possible to play with the code without getting a Yocto build running and using only qemu (BBB is still a no go?). I would not call that a quick and easy development path.

I understand the need for it, and I'm even (somewhat) in favor, but I do want to keep it minimal.

@pasinskim, what do you do in your day-to-day development? Do you always use Yocto?

(I haven't touched the client in a long time myself, so I can only offer a generic viewpoint)

device_local.go
bboozzoo: How does that relate to `Rollback()`?

bboozzoo:
@pasinskim noticed your comment about `mender -rootfs`. Is it possible for anyone to create update artifacts right now? I have not updated that part. `doRootfs()` uses `device.InstallUpdate()` so it's  expecting a regular rootfs image instead of an artifact.

pasinskim:
@bboozzoo I guess this has become somewhat gray area right now. I don't know how this will evolve, but for now I think this is OK to use regular file to perform manual update. 
As for the local development, I agree this is bit tricky at the moment. I would really like to have PR analyzer soon to solve at least part of this problem (it will automatically build Yocto and create docker container). 
With the new different code path we will need some discipline and I am afraid that local development path will diverge from main production path.
It will also require extra effort to add some features to both local and production paths.

But since I'm not alone here and you guys think this is needed we can vote. But let's be aware of all drawbacks. 
For the reference my vote is *NO*.

device_local.go
kacf: My point is that you have dedicated `_local` versions of `Rollback`, `InstallUpdate`, `EnableUpdatedPartition`, `CommitUpdate` and `HasUpdate`, all of which seem to be unneeded if you point the `RootfsPartA` and `RootfsPartB` to local files using the configuration file. And assuming that `BootEnv` is behaving as expected when using the `_local` version.

kacf:
My vote is a very weak yes, but I guess that tips it in favor. We should be constantly looking for ways to minimize and even remove this "local" code though. Perhaps a better solution will present itself later.

bboozzoo:
Pushed updated tree with local support removed. There's still some things remaining, like paths_local.go and build tags. These can be addressed later.

dirstore.go
bboozzoo: Comment regarding what?

device.go
bboozzoo: I don't understand, it's already part of the binary, so we're not adding anything new here.

device.go
bboozzoo: fixed

device.go
kacf: Yeah, I don't think this is a problem. `syscall` is part of the binary regardless (it's used by the Go runtime code itself). Besides I think this is the standard way to return errors that aren't listed specifically in the `errors` package (and there are not many of those).

PR #128, opened 2016-10-04
Stable

pasinskim:


pasinskim:
It builds locally so I will try to merge to stable. If this will fail I will revert this one and try to fix.

PR #135, opened 2016-10-20
LMDB store backend

bboozzoo:
A set of changes that introduce LMDB as one of the stores (`DBStore` in our case). 

First patch restores lmdb-go dependency that was reverted in master.

The second patch extends `Store` with `Cleanup()` operation. While other stores do not need any special cleanup, we need to close `lmdb.Env`. Third and fourth patches setup calling of `Cleanup()`.

5th patch adds LMDB data store, with 2 more patches on top that add extra checking and enable `DBStore` by default.

6th patch is a small refactoring of Keystore. We have an opportunity to do some cleanups here, mainly Keystore is always used in the context of a single key. Thus it makes no sense to pass key (file) name in each `Save`/`Load` call.

The last patch refactors AuthManager to support injection of Keystore and tenant token. There's a small problem with how we access tenant token and device key. Right now the token and key are expected to be stored in file system. One option would be to load both into DB. Equally well we can just cleanup the AuthManager API a little bit and allow for injection of token and keystore, and while doing that, wrap the whole set of parameters into a single `AuthManagerConfig` struct.

@kacf @pasinskim

### DIFF
--- a/auth.go
+++ b/auth.go
@@ -17,6 +17,7 @@ import (
 	"bytes"
 	"encoding/json"
 	"os"
+	"strings"
 
 	"github.com/mendersoftware/log"
 	"github.com/pkg/errors"
@@ -64,37 +65,44 @@ type AuthManager interface {
 }
 
 const (
-	authTokenName       = "authtoken"
-	authTenantTokenName = "authtentoken"
-	authSeqName         = "authseq"
+	authTokenName = "authtoken"
+	authSeqName   = "authseq"
 
 	noAuthToken = AuthToken("")
 )
 
 type MenderAuthManager struct {
-	store    Store
-	keyStore *Keystore
-	keyName  string
-	idSrc    IdentityDataGetter
-	seqNum   SeqnumGetter
+	store       Store
+	keyStore    *Keystore
+	idSrc       IdentityDataGetter
+	seqNum      SeqnumGetter
+	tenantToken AuthToken
 }
 
-func NewAuthManager(store Store, keyName string, idSrc IdentityDataGetter) AuthManager {
-	ks := NewKeystore(store)
-	if ks == nil {
+type AuthManagerConfig struct {
+	AuthDataStore  Store              // authorization data store
+	KeyStore       *Keystore          // key storage
+	IdentitySource IdentityDataGetter // provider of identity data
+	TenantToken    []byte             // tenant token
+}
+
+func NewAuthManager(conf AuthManagerConfig) AuthManager {
+
+	if conf.KeyStore == nil || conf.IdentitySource == nil ||
+		conf.AuthDataStore == nil {
 		return nil
 	}
 
 	mgr := &MenderAuthManager{
-		store:    store,
-		keyStore: ks,
-		keyName:  keyName,
-		idSrc:    idSrc,
-		seqNum:   NewFileSeqnum(authSeqName, store),
+		store:       conf.AuthDataStore,
+		keyStore:    conf.KeyStore,
+		idSrc:       conf.IdentitySource,
+		seqNum:      NewFileSeqnum(authSeqName, conf.AuthDataStore),
+		tenantToken: AuthToken(conf.TenantToken),
 	}
 
-	if err := ks.Load(keyName); err != nil && !IsNoKeys(err) {
-		log.Errorf("failed to load device keys from %v: %v", keyName, err)
+	if err := mgr.keyStore.Load(); err != nil && !IsNoKeys(err) {
+		log.Errorf("failed to load device keys: %v", err)
 		return nil
 	}
 
@@ -134,12 +142,7 @@ func (m *MenderAuthManager) MakeAuthRequest() (*AuthRequest, error) {
 		return nil, errors.Wrapf(err, "failed to obtain device public key")
 	}
 
-	tentok, err := m.store.ReadAll(authTenantTokenName)
-	if err != nil {
-		return nil, errors.Wrapf(err, "failed to read tenant token")
-	}
-
-	tentok = bytes.TrimSpace(tentok)
+	tentok := strings.TrimSpace(string(m.tenantToken))
 
 	log.Debugf("tenant token: %s", tentok)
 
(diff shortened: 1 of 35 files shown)


dbstore.go
```diff
@@ -0,0 +1,196 @@
...
+// directory where the file will be stored. Returns nil if initialization
+// failed.
+func NewDBStore(dirpath string) *DBStore {
+	env, err := lmdb.NewEnv()
+	if err != nil {
+		log.Errorf("failed to create DB environment: %v", err)
+		return nil
+	}
+
+	if err := env.Open(path.Join(dirpath, DBStoreName), lmdb.NoSubdir, 0600); err != nil {
```
bboozzoo: @kacf IIRC you've used LMDB in cfengine, can you advise on the flags to use here and in places where `OpenRoot()` is called?

auth.go
```diff
@@ -134,12 +141,7 @@ func (m *MenderAuthManager) MakeAuthRequest() (*AuthRequest, error) {
 		return nil, errors.Wrapf(err, "failed to obtain device public key")
 	}
 
-	tentok, err := m.store.ReadAll(authTenantTokenName)
-	if err != nil {
-		return nil, errors.Wrapf(err, "failed to read tenant token")
-	}
-
-	tentok = bytes.TrimSpace(tentok)
+	tentok := bytes.TrimSpace([]byte(m.tenantToken))
```
pasinskim: There is also `strings.TrimSpace` method. This will avoid converting this to `[]byte` back and forth.

pasinskim: What is more, should we check if `m.tenantToken` is valid before/after?

dbstore.go
```diff
@@ -0,0 +1,196 @@
...
+		log.Errorf("failed to open DB environment: %v", err)
+		return nil
+	}
+
+	db := DBStore{
+		env: env,
+	}
+	return &db
```
pasinskim: Just a cosmetic thing, but can be simplified to:
```
return &DBStore{
	env: env,
}
```

dbstore.go
```diff
@@ -0,0 +1,196 @@
...
+	b, err := db.readBytes(name)
+	if err != nil {
+		return nil, err
+	}
+	return ioutil.NopCloser(b), nil
+}
+
+func (db *DBStore) readBytes(name string) (*bytes.Buffer, error) {
+	if db.env == nil {
+		panic("env is nil")
```
pasinskim: Should we really `panic` here?

dbstore.go
```diff
@@ -0,0 +1,196 @@
...
+			return nil, os.ErrNotExist
+		}
+		return nil, errors.Wrapf(err, "failed to read data for key %s", name)
+	}
+	return b, nil
+}
+
+func (db *DBStore) Remove(name string) error {
+	if db.env == nil {
+		panic("env is nil")
```
pasinskim: Same here. Panic is bit extreme IMHO.

main.go
```diff
@@ -300,19 +321,51 @@ func doBootstrapAuthorize(config *menderConfig, opts *runOptionsType) error {
 	return nil
 }
```
pasinskim: Maybe we can abstract this in separate function. This looks exact same as in `initDaemon` couple of lines below.

auth.go
bboozzoo: Yeah, totally forgot about `strings.TrimSpace()`.

As for validating the token, we could check if it's not empty (like it was before). Minor issue though, the backend does not care about the token, and frankly I don't know if it's planned to keep it like this or not in the future. IMO, once the backend starts caring, there should be a version bump in the API and we should check that in the Auth client code.

dbstore.go
bboozzoo: replaced with `ErrDBStoreNotInitialized`

dbstore.go
bboozzoo: fixed

main.go
bboozzoo: I'll try to reorganize this

bboozzoo: Cleaned up, common code moved to `commonInit`.

dbstore.go
bboozzoo: ok

dbstore.go
```diff
@@ -0,0 +1,199 @@
...
+	}
+
+	err := db.env.Update(func(txn *lmdb.Txn) error {
```
bboozzoo: Not sure I follow, you do not want to do a transaction here?

dbstore.go
```diff
@@ -0,0 +1,199 @@
...
+func (db *DBStore) WriteAll(name string, data []byte) error {
+	return db.writeBytes(name, bytes.NewBuffer(data))
+}
+
+func (db *DBStore) writeBytes(name string, data *bytes.Buffer) error {
+	if db.env == nil {
+		return ErrDBStoreNotInitialized
+	}
+
+	err := db.env.Update(func(txn *lmdb.Txn) error {
```
kacf: My point was that our `Store` API opens a new transaction for every operation. So for example, if you do `ReadAll()` followed by `WriteAll()`, this is not one transaction, but two. This can be dangerous if others are writing to the same DB (not a problem for us though), but can also be a performance bottleneck, especially for writes.

What I'm suggesting is that `Store` uses `View()` and `Update()` functions, just like LMDB does. This would allow the code that uses the DB to use transactions.

If added, I also think they should be enforced (IOW move the old `Store` read/write API into a `StoreTransaction` object, so they can only be carried out there).

So one of the functions would be something like:

```
func (self *DBStore) Update(func worker(txn *StoreTransaction)) (error)
```

What do you think?

bboozzoo: I don't think it's worth the effort. We are the only process writing to DB, the writes happen only occasionally if there's an update being processed, and even then we only write a handful of bytes. As such I see no need for an API like this ATM.

Besides the transaction cannot be enforced for `DirStore`, so `Update` cannot be a part of `Store` unless we differentiate interfaces (w/ & w/o transaction), or allow implementations to not implement methods (throw an error maybe?), or stop using `DirStore` as a `Store`. Complicates things a bit.

kacf: Alright, leave it for now then, since it's a somewhat disruptive change. An optimization/fix for later.

kacf:
Should be merged together with https://github.com/mendersoftware/meta-mender/pull/166.

PR #137, opened 2016-10-27
rootfs image installation/download progress display

bboozzoo:
Add support for installation progress display when meder is used in `-rootfs` mode. Demo:
https://asciinema.org/a/4cp2bwmdg5whlqywk7wyst3bi

@kacf @pasinskim @GregorioDiStefano

### DIFF
--- a/utils/progress.go
+++ b/utils/progress.go
@@ -1,0 +1,110 @@
+// Copyright 2016 Mender Software AS
+//
+//    Licensed under the Apache License, Version 2.0 (the "License");
+//    you may not use this file except in compliance with the License.
+//    You may obtain a copy of the License at
+//
+//        http://www.apache.org/licenses/LICENSE-2.0
+//
+//    Unless required by applicable law or agreed to in writing, software
+//    distributed under the License is distributed on an "AS IS" BASIS,
+//    WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+//    See the License for the specific language governing permissions and
+//    limitations under the License.
+package utils
+
+import (
+	"fmt"
+	"io"
+)
+
+type ProgressWriter struct {
+	Out  io.Writer // progress output
+	N    int64     // size of the input
+	c    int64     // current count
+	over bool      // set to true of writes have gone over declared N bytes
+}
+
+func (p *ProgressWriter) Write(data []byte) (int, error) {
+	n := len(data)
+
+	p.reportGeneric(n)
+	p.c += int64(n)
+	return n, nil
+}
+
+func (p *ProgressWriter) maybeWarn(then int64) {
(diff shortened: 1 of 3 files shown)


utils/progress.go
```diff
@@ -0,0 +1,98 @@
...
+
+	for ; nowDots < thenDots; nowDots++ {
+		p.Out.Write([]byte("."))
+		if nowDots != 0 && (nowDots+1)%dotsPerLine == 0 {
+			// print percentage and size after each complete line
+			var s string
+			nowSize := (nowDots + 1) * perDot
+			nowSizekB := nowSize / 1024
+			if p.N == 0 {
+				s = fmt.Sprintf(" %v kB\n", nowSizekB)
```
kacf: Minor: Perhaps use `kiB`? This would make it completely unambiguous.

utils/progress.go
```diff
@@ -0,0 +1,98 @@
...
+	thenDots := then / perDot
+
+	for ; nowDots < thenDots; nowDots++ {
+		p.Out.Write([]byte("."))
+		if nowDots != 0 && (nowDots+1)%dotsPerLine == 0 {
+			// print percentage and size after each complete line
+			var s string
+			nowSize := (nowDots + 1) * perDot
+			nowSizekB := nowSize / 1024
+			if p.N == 0 {
```
kacf: I would perhaps do:

```
if p.N == 0 || then > p.N
```

to catch errors where the expected size is miscalculated. Unlikely, but possible. We might consider a warning as well.

utils/progress.go
```diff
@@ -0,0 +1,98 @@
...
+		if thenDots == 0 {
+			p.Out.Write([]byte("."))
+			left -= 1
+		}
+		// fill remaining space with spaces
+		for i := int64(0); i < left; i++ {
+			p.Out.Write([]byte(" "))
+		}
+		// print 100%, make sure we proper size suffix just in case
+		// we arre below 1kB with the whole size
```
kacf: `s/arre/are/`

utils/progress.go
kacf: There are multiple places with `kB`, btw.

kacf:
Nice demo, btw, didn't know about that site.

utils/progress.go
bboozzoo: fixed

utils/progress.go
bboozzoo: Fixed. Warning will be printed when `Write()` goes over declared size. The warning will appear only once.

kacf:
This conflicts right now, but this is just because we had to temporarily revert #138 (see #140). Just leave it and it will un-conflict itself once we have fixed the tests.

bboozzoo:
Rebased on top of master.

kacf:
Running in Jenkins.

PR #138, opened 2016-10-27
Use artifact for command-line update installation.

pasinskim:


rootfs.go
```diff
@@ -69,9 +72,39 @@ func doRootfs(device UInstaller, args runOptionsType) error {
 		return errors.New("Error while updateing image from command line: " + err.Error())
 	}
 
-	if err = device.InstallUpdate(image, imageSize); err != nil {
-		return err
+	var installed bool
+	rp := parser.RootfsParser{
+		DataFunc: func(r io.Reader, dev string, uf parser.UpdateFile) error {
```
bboozzoo: Looks like a copy & paste from https://github.com/mendersoftware/mender/blob/master/mender.go#L446 can we wrap into a single func? Example:
```
func makeRootfsInstaller(device UInstallCommitRebooter, dt string, installed *bool) func(r io.Reader, dev string, uf parser.Updatefile) error {
    return func(r io.Reader, dev string, uf parser.Updatefile) error {
			if installed == nil {
				panic("install flag not passed")
			}
			if dev != dt {
				return errors.Errorf("unexpected device type [%v], expected to see [%v]",
					dev, dt)
			}

			if *installed {
				return errors.Errorf("rootfs image already installed")
			}

			log.Infof("installing update %v of size %v", uf.Name, uf.Size)
			err = device.InstallUpdate(ioutil.NopCloser(r), uf.Size)
			if err != nil {
				log.Errorf("update image installation failed: %v", err)
				return err
			}

			*installed = true
			return nil
    }
}

// then
	installed := false
	rp := parser.RootfsParser{
		DataFunc: makeRootfsInstaller(device, dt, &installed),
	}
```

rootfs.go
```diff
@@ -64,13 +65,10 @@ func doRootfs(device UInstaller, args runOptionsType, dt string) error {
 		log.Debugf("Feting update from file results: [%v], %d, %v", image, imageSize, err)
 	}
 
-	if image != nil {
-		defer image.Close()
-	}
-
-	if err != nil {
+	if image == nil || err != nil {
 		return errors.New("Error while updateing image from command line: " + err.Error())
```
pasinskim: Fixing!

rootfs.go
pasinskim: Yup. Makes a lot of sense in fact.

PR #141, opened 2016-11-04
Fixes for MEN-759, MEN-760

bboozzoo:
The first problem is fixed by not adding `Authorization` header to update fetch requests.

The second is addressed by removing `StateError`. It looks that `StateError` was used interchangeably with `StateFailure`, what led to subtle error when reporting update status that resulted in logs not being uploaded.

@pasinskim @kacf

### DIFF
--- a/mender.go
+++ b/mender.go
@@ -315,7 +315,7 @@ func (m *mender) doBootstrap() menderError {
 }
 
 func (m *mender) FetchUpdate(url string) (io.ReadCloser, int64, error) {
-	return m.updater.FetchUpdate(m.api.Request(m.authToken), url)
+	return m.updater.FetchUpdate(m.api, url)
 }
 
 // Check if new update is available. In case of errors, returns nil and error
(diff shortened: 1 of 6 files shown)


mender.go
```diff
@@ -315,7 +315,7 @@ func (m *mender) doBootstrap() menderError {
 }
 
 func (m *mender) FetchUpdate(url string) (io.ReadCloser, int64, error) {
-	return m.updater.FetchUpdate(m.api.Request(m.authToken), url)
+	return m.updater.FetchUpdate(m.api, url)
```
kacf: I agree with the fix, but double checking with @maciejmrowiec. Is it correct that we want no authorization when fetching the image? Some images could be slightly sensitive I suppose?

bboozzoo: I'm wondering how to distinguish whether we need to pass authorization and if so, what kind of authorization. 

For instance, if download URL has the same prefix (or at least host/port/scheme) as server/backend URL would it be ok to assume that we should use API token?

kacf:
Running in Jenkins.

mender.go
maciejmrowiec: @kacf download link does not need any tampering, if it will it's going to be documented. Links are signed by server thats why tweaking certain things invalidates it.

kacf: Alright, thanks.

PR #142, opened 2016-11-04
deployment abort handling

bboozzoo:
@kacf @pasinskim

### DIFF
--- a/state.go
+++ b/state.go
@@ -445,8 +445,10 @@ func (u *UpdateFetchState) Handle(ctx *StateContext, c Controller) (State, bool)
 		return NewUpdateErrorState(NewTransientError(err), u.update), false
 	}
 
-	// report downloading, don't care about errors
-	c.ReportUpdateStatus(u.update, client.StatusDownloading)
+	merr := c.ReportUpdateStatus(u.update, client.StatusDownloading)
+	if merr != nil && merr.IsFatal() {
+		return NewUpdateErrorState(NewTransientError(merr.Cause()), u.update), false
+	}
 
 	log.Debugf("handle update fetch state")
 	in, size, err := c.FetchUpdate(u.update.Image.URI)
@@ -496,8 +498,10 @@ func (u *UpdateInstallState) Handle(ctx *StateContext, c Controller) (State, boo
 		return NewUpdateErrorState(NewTransientError(err), u.update), false
 	}
 
-	// report installing, don't care about errors
-	c.ReportUpdateStatus(u.update, client.StatusInstalling)
+	merr := c.ReportUpdateStatus(u.update, client.StatusInstalling)
+	if merr != nil && merr.IsFatal() {
+		return NewUpdateErrorState(NewTransientError(merr.Cause()), u.update), false
+	}
 
 	log.Debugf("handle update install state")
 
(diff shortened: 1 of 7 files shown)


state.go
```diff
@@ -754,6 +760,13 @@ func (usr *UpdateStatusReportState) trySend(send SendData, c Controller) (error,
 			usr.update.ID, usr.status, usr.triesSendingReport)
 		if err := send(usr.update, usr.status, c); err != nil {
 			log.Errorf("failed to report data %v: %v", usr.status, err.Cause())
+			// fatal error means that the cause is not likely to go
+			// away with subsequent retries, just stop at once
+			// at once
```
kacf: `at once` x 2

state.go
```diff
@@ -497,7 +500,10 @@ func (u *UpdateInstallState) Handle(ctx *StateContext, c Controller) (State, boo
 	}
 
 	// report installing, don't care about errors
```
kacf: This comment is incorrect now.

state.go
```diff
@@ -446,7 +446,10 @@ func (u *UpdateFetchState) Handle(ctx *StateContext, c Controller) (State, bool)
 	}
 
 	// report downloading, don't care about errors
-	c.ReportUpdateStatus(u.update, client.StatusDownloading)
+	merr := c.ReportUpdateStatus(u.update, client.StatusDownloading)
+	if merr != nil && merr.IsFatal() {
+		return NewUpdateErrorState(NewTransientError(merr.Cause()), u.update), false
```
kacf: I'm a bit fuzzy on the terminology here. The error state is fatal because it will cause an abort of the update, right? But then why do we create a transient error here?

state_test.go
```diff
@@ -318,6 +318,25 @@ func TestStateUpdateReportStatus(t *testing.T) {
...
+
+	// pretend update was aborted at the backend, but was applied
+	// successfully on the device
+	usr = NewUpdateStatusReportState(update, client.StatusSuccess)
+	sc = &stateTestController{
+		reportError: NewFatalError(client.ErrDeploymentAborted),
+	}
+	s, c = usr.Handle(&ctx, sc)
+	assert.IsType(t, s, &ReportErrorState{})
+	assert.Equal(t, client.StatusSuccess, s.(*ReportErrorState).status)
```
kacf: This mixing of error and success also confuses me a bit. It signals that the server has reported an error. But why is the status "success"? I mean, I get that it is because we updated successfully, but why is the success code inside the error state?

state.go
bboozzoo: fixed

state.go
bboozzoo: removed the comment

state.go
kacf: Ah, I see. Yes, that makes sense. The effect of the error was that you entered the error state, and now you want to produce a "softer" error to do cleanup.

state_test.go
kacf: Ok, makes sense. But can I request that `status` is renamed to `updateStatus`? That would make it much clearer. Unless it's a huge hassle?

bboozzoo: It's doable. I'll update the patch and rebase on top of master.

bboozzoo:
ping, anyone?

kacf:
Running in Jenkins: https://ci.cfengine.com/job/yoctobuild-pr/96/

kacf:
https://ci.cfengine.com/job/yoctobuild-pr/99/

PR #152, opened 2016-11-22
Clean up reference to artifact names and device types.

kacf:
* Image ID and Yocto ID have been merged into one, and there is only one
  id now, which is the artifact name.

* device type information has been moved to a separate file in
  /var/lib/mender.

* artifact_name and device_type attributes have both been lowercased
  to match json attributes.

Fixes MEN-762, MEN-763


kacf:
This goes together with:
* https://github.com/mendersoftware/meta-mender/pull/187
* https://github.com/mendersoftware/mender-docs/pull/57
* https://github.com/mendersoftware/mender-qa/pull/55.

kacf:
Jenkins is green for all of these combined.

HACKING.md
```diff
@@ -9,21 +9,22 @@ Overview of Mender client architecture.
...
+         +------------------+     +--------------------+     +------------------------+
+         | NewDaemon        |     | NewUpdater         |     | NewMender              |
+         | Run              |     | NewHttpsClient     |     | GetState               |
+         | StopDaemon       <-----+ NewHttpClient      |     | GetCurrentArtifactName |
+         |                  |     | GetScheduledUpdate |     | LoadConfig             |
+         |                  |     | FetchUpdate        |     | GetUpdaterConfig       |
+         |                  |     | Bootstrap          |     | GetDaemonConfig        |
+         |                  |     |                    |     | GetDeviceType          |
+         |                  |     |                    |     |                        |
+         +-------------^-^--+     +--------------------+     +-----+---^--------------+
```
pasinskim: I think we should remove this file for now. It is outdated and we are not updating it.

client/client_update.go
```diff
@@ -108,16 +108,15 @@ type UpdateResponse struct {
 	Image struct {
 		URI      string
 		Checksum string
-		YoctoID  string `json:"yocto_id"`
-		ID       string
+		Name     string `json:"artifact_name"`
```
pasinskim: I know there is a separate task for this, but will this be as described here: https://tracker.mender.io/browse/MEN-734?

mender.go
```diff
@@ -64,8 +64,9 @@ const (
 )
 
 var (
-	defaultManifestFile = path.Join(getConfDirPath(), "build_mender")
-	defaultDataStore    = getStateDirPath()
+	defaultArtifactInfoFile = path.Join(getConfDirPath(), "artifact-info")
+	defaultDeviceTypeFile   = path.Join(getStateDirPath(), "device_type")
```
pasinskim: Should we unify this somehow, so that we are having `artifact_info` and `device_type` or `artifact-info` and `device-type`?

HACKING.md
kacf: Ok, will remove it!

mender.go
kacf: Yes, it should be `artifact_info`. Not sure why I didn't do it immediately, it's the only one with a dash right now. I will change it.

client/client_update.go
kacf: Alright, I have changed this, since it is what's in the spec now. But I'm not sure it should be. I've posted a comment in that ticket, so we can handle it separately.

PR #155, opened 2016-11-24
changed update check api call to match actual backend route

GregorioDiStefano:
@bboozzoo @pasinskim 

### DIFF
--- a/client/client_update.go
+++ b/client/client_update.go
@@ -161,7 +161,7 @@ func processUpdateResponse(response *http.Response) (interface{}, error) {
 }
 
 func makeUpdateCheckRequest(server string) (*http.Request, error) {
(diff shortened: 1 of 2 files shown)


client/client_update.go
```diff
@@ -162,7 +162,7 @@ func processUpdateResponse(response *http.Response) (interface{}, error) {
 }
 
 func makeUpdateCheckRequest(server string) (*http.Request, error) {
-	url := buildApiURL(server, "/deployments/device/update")
+	url := buildApiURL(server, "/deployments/next")
```
bboozzoo: should be `/deployments/device/deployments/next`

GregorioDiStefano: thanks

PR #157, opened 2016-11-24
Wait schedule

pasinskim:


bboozzoo:
Perhaps this code works, but it seems overly complex to me. I anticipated the implementation to be more like this:
```
// StateContext gains new time.Time fields - LastUpdateCheck, LastInventoryUpdate
// these keep the time when last update happened, possiblly inventory time needs to be
// refreshed/reset after an update completes

type MaybeInvUpdateState struct {
	CancellableState
	// something something
}

func (m *MaybeInvUpdateState) Handle(ctx *StateContext, _ Controller) {
	// make sure that the times are structs not pointers
	update := ctx.LastUpdateCheck
	inventory := ctx.LastInventoryUpdate

	update.Add(c.GetUpdatePollInterval())
	inventory.Add(c.GetUpdatePollInterval())

	next := struct {
		When        time.Time
		UpdateAfter *time.Time
		State       State
	}{
		time.Now(),
		nil,
	}
	if update.Before(inventory) {
		next.Time = update
		next.UpdateAfter = &ctx.LastUpdateCheck
		next.State = &updateCheckState

	} else {
		next.Time = inventory
		next.UpdateAfter = &ctx.LastInventoryUpdate
		next.State = &inventoryUpdateState
	}

	now := time.Now()
	if !next.Before(time.Now()) {
		wait := next.Sub(now)
		m.Wait(wait)
		// handle cancellation
	}

	*next.UpdateAfter = now
	return next.State
}

// then return MaybeInvUpdateState from states that transition to these inventory/update wait times
// IMO, possibly even replace UpdateCheckWaitState
```

state_scheduler.go
```diff
@@ -0,0 +1,164 @@
...
+		runNext: time.Now().Add(freq),
+		id:      key,
+	}
+	s.events = append(s.events, ev)
+
+	return nil
+}
+
+func (s *stateScheduler) RemoveState(state State, freq time.Duration) error {
+	key := getKey(freq, state)
```
kacf: Why use a key at all? The state is enough to identify the element in the list, since you only allow one of each state anyway. And having to pass both the state and the frequency into `RemoveState` is a somewhat strange API, I think the state alone should be enough.

pasinskim: I was thinking that for some reason you might want to have one state to be fired with different intervals. As example, maybe you will need inventory to be run once with zero interval after committing update, and then you will have inventory with some other interval to be run periodically. 
That's why we are having a key which is some combination of state and frequency.

pasinskim:
@kacf @bboozzoo thanks for the comments. I will leave this and https://github.com/mendersoftware/mender/pull/158 both open until Monday. Then we can decide which one to proceed with. But I guess for now we can go with the simpler solution and use this one once we will need it. I like the code as well (of course, I am the author) so we can reuse it when needed (I won't remove the branch and the code is fairly well isolated).

pasinskim:
I am closing this for now. We can reopen and continue work on this one once needed.

PR #158, opened 2016-11-25
Send inventory with different interval than update request. 

pasinskim:


pasinskim:
Merge together with https://github.com/mendersoftware/meta-mender/pull/188

kacf:
Is this basically a simpler and less generic version of #157?

pasinskim:
@kacf that's right. Let's decide which one do we want.

kacf:
Maybe we should go with this one for now. #157 has potential, but it needs some more work, and isn't strictly necessary right now.

PR #160, opened 2016-11-28
Change update response to reflect API changes.

pasinskim:
Update client to use new format of API as described here:
https://tracker.mender.io/browse/MEN-737

### DIFF
--- a/client/client_update_test.go
+++ b/client/client_update_test.go
@@ -26,35 +26,82 @@ import (
 )
 
 const correctUpdateResponse = `{
-"image": {
-"uri": "https://menderupdate.com",
-"checksum": "checksum",
-"name": "core-image-base"
-},
-"id": "13876-123132-321123"
+	"id": "deplyoment-123",
+	"artifact": {
+		"source": {
+			"uri": "https://menderupdate.com",
+			"expire": "2016-03-11T13:03:17.063+0000"
+		},
+		"device_types_compatible": ["BBB"],
+		"artifact_name": "myapp-release-z-build-123"
+	}
+}`
+
+const correctUpdateResponseMultipleDevices = `{
+	"id": "deplyoment-123",
+	"artifact": {
+		"source": {
+			"uri": "https://menderupdate.com",
+			"expire": "2016-03-11T13:03:17.063+0000"
+		},
+		"device_types_compatible": [
+			"BBB",
+			"ELC AMX",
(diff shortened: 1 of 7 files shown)


client/client_update_test.go
```diff
@@ -26,35 +26,97 @@ import (
...
+			"expire": "2016-03-11T13:03:17.063+0000"
+		},
+		"device_types_compatible": ["BBB"],
+		"artifact_name": "myapp-release-z-build-123"
+	}
+}`
+
+const missingDevicesUpdateResponse = `{
+	"id": "deplyoment-123",
+	"bad_field": "13876-123132-321123"
```
kacf: Again missing comma.

client/client_update_test.go
```diff
@@ -26,35 +26,97 @@ import (
...
 }`
 
 const malformedUpdateResponse = `{
-"image": {
-"non_existing": "https://menderupdate.com",
-"checksum": "Hello, world!",
-},
-"bad_field": "13876-123132-321123"
+	"id": "deplyoment-123",
+	"bad_field": "13876-123132-321123"
```
kacf: I think you need a comma at the end here. I think the test doesn't catch this because it's looking for failure, but it's failing for the wrong reason (parse error instead of missing field).

client/client_update_test.go
```diff
@@ -26,35 +26,97 @@ import (
...
 }`
 
 const brokenUpdateResponse = `{
-"image": {
-"uri": "https://menderupdate
-"checksum": "Hello, world!"
-},
-"id": "13876-123132-321123"
+	"id": "deplyoment-123",
+	"bad_field": "13876-123132-321123"
```
kacf: Here the comma *should* be missing, so this one is correct.

client/client_update_test.go
```diff
@@ -26,35 +26,97 @@ import (
...
 }`
 
-const missingFieldsUpdateResponse = `{
-"image": {
-"uri": "https://menderupdate.com"
-},
-"id": "13876-123132-321123"
+const missingNameUpdateResponse = `{
+	"id": "deplyoment-123",
+	"bad_field": "13876-123132-321123"
```
kacf: Again missing comma.

pasinskim:
@kacf thanks for spotting comma issues. All should be fixed now.

PR #161, opened 2016-11-28
MEN-812 Do not store state as number.

pasinskim:
So far we've been storing update state as number (`MenderState`).
There is potential risk in this method as after adding some more
states we might end up with some issues after upgrade.

Let's assume we are upgrading client from the version 1.0 to
version 1.1. In the version 1.1 some state was added which is changing
the order of states. In such a scenario, state stored in version 1.0
will be interpreted differently by version 1.1.

This is supposed to fix the ambiguity in storing and loading states.

### DIFF
--- a/mender.go
+++ b/mender.go
@@ -143,6 +143,15 @@ func (m MenderState) String() string {
 	return n
 }
 
+func StateID(name string) MenderState {
+	for k, v := range stateNames {
+		if v == name {
+			return k
+		}
+	}
+	return MenderStateError
+}
+
 type mender struct {
 	UInstallCommitRebooter
 	updater          client.Updater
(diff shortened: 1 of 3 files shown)


bboozzoo:
LGTM. Though I'd try with a custom marshaller, like this: https://play.golang.org/p/2ba8OKa463 saves calling String() repeatedly and we continue to use integers at runtime.

Come to thing of serialization, perhaps we should introduce a version format in StateData as well?

kacf:
> Come to thing of serialization, perhaps we should introduce a version format in StateData as well?

+1

At least in the database you read it from, so you can revert to the base state if you've upgraded to a new incompatible version.

pasinskim:
Agree. Adding a version makes sense. Should we merge this one as is and open new ticket to add serialization and perhaps see what we can do with custom marshaling? 
I am also OK with closing this one and extending the scope of MEN-812.

kacf:
Yes, I think so. This is already an improvement.

pasinskim:
MEN-853 created.

PR #162, opened 2016-12-06
MEN-480: Make sure U-Boot environment update is atomic.

kacf:
We need to update the environment in one operation, because otherwise
it can be in an in-between state if the device goes down unexpectedly.
The actual atomicity of the operation is ensured by U-Boot and its
redundant environment, we just need to call the right command.

One test case was removed because it's not possible to get such an
error anymore.

### DIFF
--- a/bootenv.go
+++ b/bootenv.go
@@ -16,6 +16,7 @@ package main
 import (
 	"bufio"
 	"errors"
+	"fmt"
 	"os/exec"
 	"strings"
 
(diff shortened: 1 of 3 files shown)


bootenv.go
```diff
@@ -33,22 +34,39 @@ func NewEnvironment(cmd Commander) *uBootEnv {
...
 	getEnvCmd := e.Command("fw_printenv", names...)
-	return getOrSetEnvironmentVariable(getEnvCmd)
+	return getEnvironmentVariable(getEnvCmd)
 }
 
 func (e *uBootEnv) WriteEnv(vars BootVars) error {
-	//TODO: try to make this atomic later
+	// Make environment update atomic by using fw_setenv "-s" option.
+	setEnvCmd := e.Command("fw_setenv", "-s", "-")
+	pipe, err := setEnvCmd.StdinPipe()
```
pasinskim: I recall @bboozzoo having some opinions about using pipes with command. I don't remember exactly what this was about. Maybe he can take a look as well.

bootenv.go
```diff
@@ -33,22 +34,39 @@ func NewEnvironment(cmd Commander) *uBootEnv {
...
-		setEnvCmd := e.Command("fw_setenv", k, v)
-		if _, err := getOrSetEnvironmentVariable(setEnvCmd); err != nil {
-			log.Error("Error setting U-Boot variable: ", err)
+		_, err = fmt.Fprintf(pipe, "%s %s\n", k, v)
+		if err != nil {
+			log.Error("Error while setting U-Boot variable: ", err)
 			return err
 		}
 	}
+	pipe.Close()
```
pasinskim: Maybe you can move this higher and do `defer pipe.Close()`. It is more idiomatic, and will avoid resource leaking if writing to pipe will fail.

kacf: Oh, nice catch! We can't use `defer`, since this needs to be done in order, but I need to add `Close` to the error paths.

statcommander_test.go
```diff
@@ -59,6 +60,9 @@ func TestHelperProcessSuccess(t *testing.T) {
 		return
 	}
 
+	// Drain input if there is any.
+	_, _ = ioutil.ReadAll(os.Stdin)
```
kacf: I'm unsure whether or not you need them, but I agree that stating explicitly that you don't care about the return data is a good thing here.

kacf:
Updated!

bootenv.go
pasinskim: You could also close executing a command inside a function and call defer there without worrying about calling `Close()` in each return path. But this way it will work as well.

PR #164, opened 2016-12-12
Client interval and resending improvements.

pasinskim:
Set the default poll interval to 30 minutes in case not configured.
Add retry poll interval configuration option and use it for re-sending auth requests and reports.

### DIFF
--- a/state_test.go
+++ b/state_test.go
@@ -65,6 +65,10 @@ func (s *stateTestController) GetInventoryPollInterval() time.Duration {
 	return s.pollIntvl
 }
 
+func (s *stateTestController) GetRetryPollInterval() time.Duration {
+	return s.pollIntvl
+}
+
 func (s *stateTestController) HasUpgrade() (bool, menderError) {
 	return s.hasUpgrade, s.hasUpgradeErr
 }
--- a/state_test.go
+++ b/state_test.go
@@ -34,6 +34,7 @@ type stateTestController struct {
 	bootstrapErr    menderError
 	artifactName    string
 	pollIntvl       time.Duration
+	retryIntvl      time.Duration
 	hasUpgrade      bool
 	hasUpgradeErr   menderError
 	state           State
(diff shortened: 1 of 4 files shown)


state_test.go
```diff
@@ -306,9 +303,10 @@ func TestStateUpdateReportStatus(t *testing.T) {
 	s, c = usr.Handle(&ctx, sc)
 	assert.IsType(t, s, &ReportErrorState{})
 	assert.False(t, c)
-	assert.WithinDuration(t, time.Now(), now1, 3*time.Second)
-	assert.InDelta(t, int(maxReportSendingTime/poll),
-		usr.(*UpdateStatusReportState).triesSendingReport, 100)
+	assert.WithinDuration(t, time.Now(), now1,
+		time.Duration(maxReportSendingAttempts)*time.Millisecond+time.Second)
```
kacf: Are you missing parenthesis around `time.Millisecond+time.Second` here? You will only multiply with milliseconds here, not seconds.

pasinskim: Nope. This is to wait `maxReportSendingAttempts` multiplied by the time needed for one reqyest plus some buffer (1s in this case).

kacf: Ok, I don't fully understand it, but I'll trust you! :-)

pasinskim:
Merge with https://github.com/mendersoftware/meta-mender/pull/202

kacf:
Running this together with https://github.com/mendersoftware/meta-mender/pull/202 in Jenkins.

kacf:
Hmm, this failed rather badly in Jenkins, which is a bit surprising. Running again to make sure it isn't a fluke.

kacf:
Nope, this is failing consistently, although the reason is not clear to me. Can you try rebasing both of the pull requests on latest master and see if that helps?

pasinskim:
Fixed. This was related to missing sed that replaced new config option in configuration file.

PR #168, opened 2017-01-09
Fix transition between states while errors occur

pasinskim:
This is fixing some logic errors while detecting errors. Most notably, after detecting errors, client is entering `init` state, causing some important actions not being taken.

### DIFF
--- a/state.go
+++ b/state.go
@@ -404,8 +404,11 @@ func (uc *UpdateCommitState) Handle(ctx *StateContext, c Controller) (State, boo
 	err := c.CommitUpdate()
 	if err != nil {
 		log.Errorf("update commit failed: %s", err)
-		// TODO: should we rollback?
-		return NewUpdateStatusReportState(uc.update, client.StatusFailure), false
+		// we need to perform roll-back here; one scenario is when u-boot fw utils
+		// won't work after update; at this point without rolling-back it won't be
+		// possible to perform new update
+		// as the update was not commited we can safely reboot only
+		return NewRebootState(uc.update), false
 	}
 
 	// update is commited now; report status
--- a/state.go
+++ b/state.go
@@ -337,7 +337,8 @@ func (uv *UpdateVerifyState) Handle(ctx *StateContext, c Controller) (State, boo
 
 	// start deployment logging
 	if err := DeploymentLogger.Enable(uv.update.ID); err != nil {
-		return NewUpdateErrorState(NewTransientError(err), uv.update), false
+		// just log error
+		log.Errorf("failed to enable deployment logger: %s", err)
 	}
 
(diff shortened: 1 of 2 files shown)


pasinskim:
All tests passed in Jenkins.

state.go
```diff
@@ -356,13 +357,14 @@ func (uv *UpdateVerifyState) Handle(ctx *StateContext, c Controller) (State, boo
...
-			// seems like we're running in a different image than expected from update
-			// information, best report an error
-			log.Errorf("running with image %v, expected updated image %v",
-				c.GetCurrentArtifactName(), uv.update.ArtifactName())
-			return NewUpdateStatusReportState(uv.update, client.StatusFailure), false
 		}
+		// seems like we're running in a different image than expected from update
+		// information, best report an error
+		// this can ONLY happen if the artifact name does not match information
+		// stored in `/etc/mdender/artifact_info` file
```
kacf: `s/mdender/mender/` both here and in the commit message.

state.go
```diff
@@ -405,8 +405,11 @@ func (uc *UpdateCommitState) Handle(ctx *StateContext, c Controller) (State, boo
 	err := c.CommitUpdate()
 	if err != nil {
 		log.Errorf("update commit failed: %s", err)
-		// TODO: should we rollback?
-		return NewUpdateStatusReportState(uc.update, client.StatusFailure), false
+		// we need to perform roll-back here; one scenario is when u-boot fw utils
+		// won't work after update; at this point without rolling-back it won't be
+		// possible to perform new update
+		// as the update was not commited we can safely reboot only
+		return NewRebootState(uc.update), false
```
kacf: What's the difference between RollbackState and RebootState?

pasinskim: The only difference is that you swap the active partition. But since commit was not performed in this case after reboot you will end up with the previous active anyway.

pasinskim:
I like your commit description better.

state.go
pasinskim: Fixed!

state.go
kacf: But isn't that always the case? Committing should be the last thing we do, so any reboot before that should trigger a rollback. This is perhaps not so important for this PR, I just wanna make sure we're consistent.

pasinskim: Yes, I can see now. You are *absolutely* right. Here reboot is enough. The only place, where we should use rollback state is when we can not report successful update (which is happening just after commit).

pasinskim:
Apparently, the  commit message, you've pointed out was similar in two places. I've fixed it in one only. Now both should be updated.

pasinskim:
Jenkins was green: https://ci.cfengine.com/job/yoctobuild-pr/188/consoleFull. Merging.

PR #169, opened 2017-01-11
Serialize statedata

pasinskim:


state.go
```diff
@@ -159,8 +159,10 @@ type StateRunner interface {
 	RunState(ctx *StateContext) (State, bool)
 }
 
-// state information that can be used for restring state from storage
+// StateData is state information that can be used for restring state from storage
```
kacf: Preexisting, but let's fix now: `s/restring/restoring/`

state.go
```diff
@@ -1085,8 +1095,19 @@ func LoadStateData(store Store) (StateData, error) {
...
-	return sd, err
+	if err != nil {
+		return StateData{}, err
+	}
+
+	switch sd.Version {
+	case 0, 1:
+		return sd, nil
+	default:
+		return sd, errors.New("unsupported state data version")
```
kacf: I would return `StateData{}` instead of `sd` here.

state.go
pasinskim: Right. Fixed!

state.go
pasinskim: Maybe. Will fix. The idea was that even though you have some version which is not supported, maybe you will be able to get some data from it anyway.
But I agree, that for now this is not needed.

pasinskim:
@kacf all should be fixed now.

state.go
```diff
@@ -159,12 +159,14 @@ type StateRunner interface {
 	RunState(ctx *StateContext) (State, bool)
 }
 
-// state information that can be used for restring state from storage
+// StateData is state information that can be used for restroing state from storage
```
kacf: Lol, you fixed one typo and introduced another!

pasinskim: Crap. Need more coffee in the morning before fixing typos... You know that human brain is working so that it is focusing only on first and last couple of letters in the given word completely ignoring what is in the middle: https://en.wikipedia.org/wiki/Typoglycemia

kacf: Hehe, that's interesting. I could read the wikipedia text example, could you? :-)

PR #171, opened 2017-01-16
MEN-944 fix deployment logging and add more logs

pasinskim:


state.go
```diff
@@ -693,6 +694,8 @@ type AuthorizedState struct {
 }
 
 func (a *AuthorizedState) Handle(ctx *StateContext, c Controller) (State, bool) {
+	log.Debugf("handle authorizd state")
```
kacf: `s/authorizd/authorized/`

pasinskim: Fixed!

PR #173, opened 2017-01-20
Ensure 128-bit security in client keys like the rest of the system

estenberg:


estenberg:
@bboozzoo could you have a look, please?

bboozzoo:
@GregorioDiStefano @kacf client PRs don't go through integration tests?

kacf:
Currently not, but it's on its way. I just have to fix the currently broken Yocto build.

But they will not be tested in Travis, no. We can't, because we have to build with Yocto first. So it will be a semi-manual process by me and Greg still.

estenberg:
ping? @bboozzoo you think it's safe to merge so we can test with the round of beta testing coming up?

PR #177, opened 2017-01-27
MEN-964 Wait with marking inactive partition until checking update st…

pasinskim:
…atus.

As installation step might take a while we should check is deployment
is still active after installation is done.
Only if deployment is still active we are marking inactive partition
as active one and proceeding with update.

### DIFF
--- a/installer/installer.go
+++ b/installer/installer.go
@@ -37,7 +37,7 @@ func InstallRootfs(device UInstaller) parser.DataHandlerFunc {
 			log.Errorf("update image installation failed: %v", err)
 			return err
 		}
-		return device.EnableUpdatedPartition()
+		return nil
 	}
 }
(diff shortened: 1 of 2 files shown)


kacf:
@pasinskim: Sorry, I had to revert this in #178, because it breaks the `-rootfs` argument which we use to run client acceptance test. I think it doesn't update the environment unless it can contact the server. (See [this build](https://ci.cfengine.com/job/yoctobuild-pr/262/console))

Let me know if you need additional help in resolving this, if not just resubmit your patch after you've fixed it.

GregorioDiStefano:
Also note: this change broken deployment aborts in general. 
if a deployment is `pending` and you abort, the abort is ignored, and the deployment become successful.

PR #179, opened 2017-01-31
MEN-964 Wait with marking inactive partition until checking update st…

kacf:
…atus.

As installation step might take a while we should check is deployment
is still active after installation is done.
Only if deployment is still active we are marking inactive partition
as active one and proceeding with update.

doRootfs is only called via command line and therefore takes the
shortcut straight to updating the partition.

Based on Marcin Pasinski's original commit:
f986d4a657ae4d56823474e986e788dfb0b912c5

### DIFF
--- a/installer/installer.go
+++ b/installer/installer.go
@@ -37,7 +37,7 @@ func InstallRootfs(device UInstaller) parser.DataHandlerFunc {
 			log.Errorf("update image installation failed: %v", err)
 			return err
 		}
-		return device.EnableUpdatedPartition()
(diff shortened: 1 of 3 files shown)


kacf:
Code coverage is lower, but I know that both paths are tested in our acceptance and integration tests.

GregorioDiStefano:
the original PR  broke deployment aborting in general, has that been addressed now?

kacf:
I don't know. I guess the PR will tell us.

PR #180, opened 2017-01-31
print descriptive state information string instead of integer

GregorioDiStefano:

### DIFF
--- a/mender.go
+++ b/mender.go
@@ -149,6 +149,10 @@ func (m MenderState) MarshalJSON() ([]byte, error) {
 	return json.Marshal(n)
 }
 
+func (m MenderState) String() string {
+	return stateNames[m]
+}
+
 func (m *MenderState) UnmarshalJSON(data []byte) error {
 	var s string
 	err := json.Unmarshal(data, &s)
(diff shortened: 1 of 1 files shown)


GregorioDiStefano:
@bboozzoo @pasinskim

bboozzoo:
I'm quite sure it used to print state names instead of numbers. Was this changed recently?

GregorioDiStefano:
Same here, I recall seeing the state change being a string, but we are back to showing only ID numbers..

bboozzoo:
@pasinskim your doing?

pasinskim:
I was also pretty sure we had the states printed, but I don't recall me changing this.

pasinskim:
All right, now I know what happened here. For a while we've been using states as strings in real-time to get rid of the issue related to storing and loading state between updates. 
Later, I changed this to be int again and added custom marshaling before storing and reading: https://github.com/mendersoftware/mender/pull/169

This is the root-cause of the issue.

mender.go
```diff
@@ -448,7 +448,7 @@ func (m mender) GetRetryPollInterval() time.Duration {
 }
 
 func (m *mender) SetState(s State) {
-	log.Infof("Mender state: %v -> %v", m.state.Id(), s.Id())
+	log.Infof("Mender state: %s -> %s", stateNames[m.state.Id()], stateNames[s.Id()])
```
pasinskim: I would do this bit differently.

Just add `MenderState` `String()` method that will satisfy `Stringer` interface:
```
func (ms MenderState) String() string {
	return stateNames[ms]
}
```

Here you can have then:
`log.Infof("Mender state: %s -> %s", m.state.Id(), s.Id())`

GregorioDiStefano: good suggestion, fixed

PR #181, opened 2017-02-01
Connection timeouting

pasinskim:


client/client.go
```diff
@@ -102,7 +109,38 @@ func NewApiClient(conf Config) (*ApiClient, error) {
 	return New(conf)
 }
 
-// Client initialization
+// keepaliveDialer is used to set keepalive options on the socket used for
+// connection. This is for handling connection timeouts while server
+// is not accessible after connection was alrady started.
+type keepaliveDialer struct {
+	net.Dialer
+}
```
kacf: Looks like `net.Dialer` already [has a field](https://golang.org/pkg/net/#Dialer) `KeepAlive`, could you just use that instead of the `tcpkeepalive` module?

pasinskim:
@bboozzoo thanks for the comment. This is exactly what I wanted to discuss with you. We need to figure out some reasonable timeouts. I've just set something quasi-random here to check if this is working.

client/client.go
```diff
@@ -35,6 +39,32 @@ var (
...
+	defaultClientReadingTimeout = 12 * time.Hour
+	// More granular connection timeouts
+	defaultDialTimeout           = 30 * time.Second
+	defaultTLSHandshakeTimeout   = 30 * time.Second
+	defaultResponseHeaderTimeout = 60 * time.Second
+
+	// connection keepalive options
+	connectionKeepaliveTime     = 5 * time.Second
+	connectionKeepaliveInterval = 5 * time.Second
+	connectionKeepaliveProbes   = 10
```
kacf: I don't really understand the roles off all these. I can understand the dial variant, that's for the initial TCP handshake, and the KeepaliveTime, that's for the remaining connection. But what about the others? Aren't they already covered by the TCP keepalive?

client/client.go
pasinskim: Indeed there is a `KeepAlive` field which could do the job. The only drawback is that it is quite limited. You can specify the only one interval which is used both as `keepalice_time` and `keepalive_interval`. What is more, currently there is no way to specify `keepalive_probes`, which is obviously impacting the whole connection timeout.
For the sake of simplicity, MAYBE, we can use that.

client/client.go
pasinskim: So to specify keepalive time you need three different things:
```
The procedures involving keepalive use three user-driven variables:

tcp_keepalive_time

 the interval between the last data packet sent (simple ACKs are not considered data) and the first keepalive probe; after the connection is marked to need keepalive, this counter is not used any further

tcp_keepalive_intvl

 the interval between subsequential keepalive probes, regardless of what the connection has exchanged in the meantime

tcp_keepalive_probes

 the number of unacknowledged probes to send before considering the connection dead and notifying the application layer
``` 
More you can see here for example: http://www.tldp.org/HOWTO/html_single/TCP-Keepalive-HOWTO/

kacf: Thanks, that explains the three bottom ones. But what about the other four? With the exception of Dial, they seem redundant.

client/client.go
kacf: Actually, I didn't even know that it was possible to be that granular, and the one interval you mention is what I would have expected from the API. I reckon that should be good enough, I mean, I don't think we don't need super fine grained control here, it's just a mechanism to time out in a roughly given time.

client/client.go
pasinskim: See this https://github.com/mendersoftware/mender/pull/181/commits/50936084882b2812bf189fc18fde02714ad19c44#diff-1cd41a2611da49ce63630d5a51bd82c5R41

I should probably add better description there. 
`defaultClientReadingTimeout` is the timeout for the whole connection, from start to after body is read
`defaultDialTimeout` is for establishing the connection 
`defaultTLSHandshakeTimeout` this is quite self explanatory
`defaultResponseHeaderTimeout` is for reading response headers

In some sense those might look like redundant, but those are bit different. As keepalive will eliminate stale connections it will not eliminate extremely slow connections. Those will handle the issue, when connection will be very slow, but active.

kacf: Ok, I see. The `defaultClientReadingTimeout` might be useful, but I think that `defaultTLSHandshakeTimeout` and `defaultResponseHeaderTimeout` are unnecessary. Even an extremely slow connection will finish those in reasonable time, and if they are really stalling, then they will be covered by the keepalive mechanism.

Not that I am against any of them. I'm just trying to make our lives easier by not providing timeout functionality we don't need. Because I assume these would require application logic to work.

pasinskim: Makes sense. I will remove those.

pasinskim:
The version without using external library: https://github.com/mendersoftware/mender/pull/185

PR #182, opened 2017-02-01
MEN-961 Make sure rollback happens after deployment is aborted

pasinskim:
This is changing rollback state to perform reboot itself, instead of
switching to reboot state. The reason is we are storing state
information in reboot state and we are checking if deployment is valid.
In case of aborted deployment, which was correctly detected after
device was rebooted we've been entering rollback and then reboot state.
As deployment was aborted instead of continuing with reboot we
reported error and went back to error and then init state.
No rollback happened in this scenario.

### DIFF
--- a/state.go
+++ b/state.go
@@ -1047,10 +1047,6 @@ func (e *RebootState) Handle(ctx *StateContext, c Controller) (State, bool) {
 	}
 
 	// we can not reach this point
-
-	// stop deployment logging
-	DeploymentLogger.Disable()
-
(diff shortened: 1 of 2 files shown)


kacf:
@GregorioDiStefano: I think the integration tests are not so good right now. Do you want to merge this and see if it improves things, or rather wait until you've stabilized?

GregorioDiStefano:
@kacf working on getting things working again, ill take care of this: test and then merge soon

PR #185, opened 2017-02-06
Timeouting client connection

pasinskim:


client/client.go
```diff
@@ -35,6 +37,29 @@ var (
...
+	//
+	//  It covers the entire exchange, from Dial (if a connection is not reused)
+	// to reading the body. This is to timeout long lasing connections.
+	//
+	// 4 hours shold be enough to download 2GB image file with the
+	// average download spead ~1 mbps
+	defaultClientReadingTimeout = 4 * time.Hour
+
+	// connection keepalive options
+	connectionKeepaliveTime = 10 * time.Second
```
kacf: I think 10 seconds is perhaps a tiny bit short. Links like GSM may be really crappy, and even my local wireless may sometimes exceed ten seconds if it changes channel for example. Maybe sixty seconds?

pasinskim: So 10 seconds is not the total keepalive time. 10s after the last packet was transmitted (`tcp_keepalive_time`) we will start sending keepalive packets. Those will be sent every 10s (`tcp_keepalive_intvl`) until reaching `tcp_keepalive_probes`. The last parameter is defined in `/proc/sys/net/ipv4/tcp_keepalive_probes` and by default this is 9. 
So having default settings this will be 100s which is even longer than you suggested.

pasinskim: To be clear, using default Go keepalive option is not allowing to distinguish between `tcp_keepalive_time` and `tcp_keepalive_intvl` so both are defined with the same value. See here: https://github.com/golang/go/issues/8328

kacf: I see, makes sense.

kacf:
@GregorioDiStefano will test this together with the timeout integration test.

PR #187, opened 2017-02-07
Zero update install counter.

pasinskim:
We need to reset the counter otherwise the next fetch or install
fail will be retried only once.


pasinskim:
@kacf how about now?

kacf:
Running integration checks [here](https://ci.cfengine.com/job/yoctobuild-pr/353/).

kacf:
Cherry-picked to 1.0.x as well, so it should be in 1.0.1!

PR #200, opened 2017-04-24
MEN-1126 Implement writing to UBI block devices in Mender client

mirzak:
Note that I changed the title in regard what MEN-1126 says in the Mender tracker. I believe that this tittle better describes what is done.

The implementation was written trying to be generic, that is we do not need any "extra" information in the Mender artifact to determinate if we are flashing an UBI or SD/MMC block device. 

This has been tested on my Colibri VF61 in standalone mode only. But I wanted to put it out there to get some feedback. 

I have also tested running this version of the Mender client on my Raspberry Pi which also seems to work. But more testing is certainly required before this is merged, hopefully this pull request will trigger some tests and I will see what I have broken :).

I pulled in a new dependency (go-sysfs) and this could probably be discussed. I felt that I did not need to re-invent the wheel when there was a package that did what I wanted, though we do not use "that much" but could also be useful moving forward with UBI support since all information about UBI is in sysfs. I am also not sure if I added the new dependency in `vendor`correctly but I guess you will let me know.

Eager to get your feedback and to continue with the work of adding UBI support.

(Be gently, golang was completely to me :))

### DIFF
--- a/block_device.go
+++ b/block_device.go
@@ -51,6 +51,30 @@ func (bd *BlockDevice) Write(p []byte) (int, error) {
 			return 0, err
 		}
 
+		// From <mtd/ubi-user.h>
+		//
+		// UBI volume update
+		// ~~~~~~~~~~~~~~~~~
+		//
+		// Volume update should be done via the UBI_IOCVOLUP ioctl command of the
+		// corresponding UBI volume character device. A pointer to a 64-bit update
+		// size should be passed to the ioctl. After this, UBI expects user to write
+		// this number of bytes to the volume character device. The update is finished
+		// when the claimed number of bytes is passed. So, the volume update sequence
+		// is something like:
+		//
+		// fd = open("/dev/my_volume");
+		// ioctl(fd, UBI_IOCVOLUP, &image_size);
+		// write(fd, buf, image_size);
+		// close(fd);
+		if bd.typeUBI {
+			err := setUbiUpdateVolume(out, bd.ImageSize)
+			if err != nil {
+				log.Errorf("Failed to write images size to UBI_IOCVOLUP: %v", err)
+				return 0, err
+			}
+		}
+
 		size, err := BlockDeviceGetSizeOf(out)
 		if err != nil {
 			log.Errorf("failed to read block device size: %v", err)
--- a/block_device.go
+++ b/block_device.go
@@ -21,11 +21,15 @@ import (
 )
 
 var (
-	BlockDeviceGetSizeOf BlockDeviceGetSizeFunc = getBlockDeviceSize
+	BlockDeviceGetSizeOf       BlockDeviceGetSizeFunc       = getBlockDeviceSize
+	BlockDeviceGetSectorSizeOf BlockDeviceGetSectorSizeFunc = getBlockDeviceSectorSize
 )
 
 // BlockDeviceGetSizeFunc is a helper for obtaining the size of a block device.
 type BlockDeviceGetSizeFunc func(file *os.File) (uint64, error)
+
+// BlockDeviceGetSectorSizeFunc is a helper for obtaining the sector size of a block device.
+type BlockDeviceGetSectorSizeFunc func(file *os.File) (int, error)
 
 // BlockDevice is a low-level wrapper for a block device. The wrapper implements
 // io.Writer and io.Closer interfaces.
@@ -96,4 +100,16 @@ func (bd *BlockDevice) Size() (uint64, error) {
 	defer out.Close()
 
 	return BlockDeviceGetSizeOf(out)
+}
+
+// SectorSize queries the logical sector size of the underlying block device. Automatically opens a
+// new fd in O_RDONLY mode, thus can be used in parallel to other operations.
+func (bd *BlockDevice) SectorSize() (int, error) {
+	out, err := os.OpenFile(bd.Path, os.O_RDONLY, 0)
+	if err != nil {
+		return 0, err
+	}
+	defer out.Close()
+
+	return BlockDeviceGetSectorSizeOf(out)
 }
--- a/block_device.go
+++ b/block_device.go
@@ -34,9 +34,10 @@ type BlockDeviceGetSectorSizeFunc func(file *os.File) (int, error)
 // BlockDevice is a low-level wrapper for a block device. The wrapper implements
 // io.Writer and io.Closer interfaces.
 type BlockDevice struct {
-	Path string               // device path, ex. /dev/mmcblk0p1
-	out  *os.File             // os.File for writing
-	w    *utils.LimitedWriter // wrapper for `out` limited the number of bytes written
+	Path    string               // device path, ex. /dev/mmcblk0p1
+	out     *os.File             // os.File for writing
+	w       *utils.LimitedWriter // wrapper for `out` limited the number of bytes written
+	typeUBI bool                 // Set to true if we are updating an UBI volume
 }
 
 // Write writes data `p` to underlying block device. Will automatically open
--- a/block_device.go
+++ b/block_device.go
@@ -34,10 +34,11 @@ type BlockDeviceGetSectorSizeFunc func(file *os.File) (int, error)
 // BlockDevice is a low-level wrapper for a block device. The wrapper implements
 // io.Writer and io.Closer interfaces.
 type BlockDevice struct {
-	Path    string               // device path, ex. /dev/mmcblk0p1
-	out     *os.File             // os.File for writing
-	w       *utils.LimitedWriter // wrapper for `out` limited the number of bytes written
-	typeUBI bool                 // Set to true if we are updating an UBI volume
+	Path      string               // device path, ex. /dev/mmcblk0p1
+	out       *os.File             // os.File for writing
+	w         *utils.LimitedWriter // wrapper for `out` limited the number of bytes written
+	typeUBI   bool                 // Set to true if we are updating an UBI volume
+	ImageSize int64                // image size
 }
 
 // Write writes data `p` to underlying block device. Will automatically open
--- a/device.go
+++ b/device.go
@@ -91,7 +91,7 @@ func (d *device) InstallUpdate(image io.ReadCloser, size int64) error {
 		inactivePartition = filepath.Join("/dev", inactivePartition)
 	}
 
-	b := &BlockDevice{Path: inactivePartition, typeUBI: typeUBI)
+	b := &BlockDevice{Path: inactivePartition, typeUBI: typeUBI, ImageSize: size}
 
 	if bsz, err := b.Size(); err != nil {
 		log.Errorf("failed to read size of block device %s: %v",
--- a/device.go
+++ b/device.go
@@ -90,7 +90,18 @@ func (d *device) InstallUpdate(image io.ReadCloser, size int64) error {
 		return syscall.ENOSPC
 	}
 
-	w, err := io.Copy(b, image)
+	ssz, err := b.SectorSize()
+	if err != nil {
+		log.Errorf("failed to read sector size of block device %s: %v",
+			inactivePartition, err)
+		return err
+	}
+
+	// allocate buffer based on sector size and provide it for staging
+	// in io.CopyBuffer
+	buf := make([]byte, ssz)
+
+	w, err := io.CopyBuffer(b, image, buf)
 	if err != nil {
 		log.Errorf("failed to write image data to device %v: %v",
 			inactivePartition, err)
--- a/device.go
+++ b/device.go
@@ -15,6 +15,7 @@ package main
 
 import (
 	"io"
+	"path/filepath"
 	"strconv"
 	"syscall"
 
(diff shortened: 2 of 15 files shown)


mirzak:
Aaah `make extracheck" :). On it.

kacf:
Nice!!!!

I will look at this shortly!

pasinskim:
Great stuff Mirza!
One clarification question though. I thought we are going to introduce new image type, something like `ubifs-image` instead of `rootfs-image`. In fact there are some changes in meta-mender introducing this. For the comment and the code itself it looks like we don't need new image type. Is that right?

block_device.go
```diff
@@ -45,6 +52,14 @@ func (bd *BlockDevice) Write(p []byte) (int, error) {
 			return 0, err
 		}
 
+		// This will only succeed on UBI volumes, fail silently on SD/MMC
+		//
+		// This must be here becuase UBI will verify if data written
+		// matches the size argument when the file descriptor is closed.
+		syscall.Syscall(syscall.SYS_IOCTL, out.Fd(),
+			uintptr(unsafe.Pointer(UBI_IOCVOLUP)),
+			uintptr(unsafe.Pointer(&bd.ImageSize)))
```
pasinskim: Some better explanation here would be nice why we need to do this. Also shouldn't we check the error code and fail if there is an error and we are using UBI?

mirzak:
> One clarification question though. I thought we are going to introduce new image type, something like ubifs-image instead of rootfs-image. In fact there are some changes in meta-mender introducing this. For the comment and the code itself it looks like we don't need new image type. Is that right?

The idea from the beginning was to add something like `ubifs-image` which should be an equivalent of `rootfs-image` but for UBI based devices. But as the work has progressed it has become clearer that we might not need to separate these two. Me and Kristian discussed this a bit on the mailing list and the plan is to try to avoid creating a separate image type and try to figure out in Mender client which device type we are updating and act accordingly. In the end the differences have showed to be minimal.

The references to `.ubimg` in meta-mender is only the image that is used for provisioning of new devices. The `.mender` is no different for `ubi` or `ext4`, well the data differs but the artifact contains the same attributes.

device.go
```diff
@@ -78,7 +80,18 @@ func (d *device) InstallUpdate(image io.ReadCloser, size int64) error {
...
-	b := &BlockDevice{Path: inactivePartition}
+	// This condition is to handle UBI block devices. Which is not prefixed
+	// with /dev due to the fact that the kernel root= argument does not
+	// handle UBI block devices which are prefixed with /dev
+	//
+	// Kernel root= only accepts:
+	// - ubi0_0
+	// - ubi:rootfsa
+	if strings.HasPrefix(inactivePartition, "/dev/") == false {
+		inactivePartition = path.Join("/dev", inactivePartition)
```
pasinskim: I think it might be safer to use `filepath` instead of `path`. This will remove some OS dependency if will be needed in future.

device.go
```diff
@@ -90,10 +103,35 @@ func (d *device) InstallUpdate(image io.ReadCloser, size int64) error {
...
-	w, err := io.Copy(b, image)
+	ssz, err := b.SectorSize()
 	if err != nil {
-		log.Errorf("failed to write image data to device %v: %v",
+		log.Errorf("failed to read sector size of block device %s: %v",
 			inactivePartition, err)
+		return err
+	}
+
+	buf := make([]byte, ssz)
```
pasinskim: Maybe more effective would be using `io.CopyN()` instead. Then we can simplify reading and writing to something like: 
```
	for {
		_, err := io.CopyN(b, image, ssz)

		if err == io.EOF {
			break
		} else if err != nil {
			log.Errorf("failed to read image data: %v", err)
			return err
		}
	}
```

ioctl.go
```diff
@@ -29,17 +32,48 @@ type ioctlRequestValue uintptr
...
-// If the device is not a block device NotABlockDevice error and
-// size 0 will be returned.
-func getBlockDeviceSize(file *os.File) (uint64, error) {
-	var fd uintptr = file.Fd()
-	ioctlRequest := BLKGETSIZE64
-	var blkSize uint64
+func getUbiDeviceSectorSize(file *os.File) (int, error) {
+	dev := strings.TrimPrefix(file.Name(), "/dev/")
+
+	eb_size := sysfs.Class.Object("ubi").SubObject(dev).Attribute("usable_eb_size")
```
pasinskim: We've are trying to stick to `ebSize` convention.

ioctl.go
```diff
@@ -29,17 +32,48 @@ type ioctlRequestValue uintptr
...
+
+	if eb_size.Exists() == false {
```
pasinskim: `if !eb_size.Exists() {`

ioctl.go
```diff
@@ -29,17 +32,48 @@ type ioctlRequestValue uintptr
...
+		return 0, NotABlockDevice
+	}
+
+	return int(sectorSize), nil
+}
+
+func getUbiDeviceSize(file *os.File) (uint64, error) {
+	dev := strings.TrimPrefix(file.Name(), "/dev/")
+
+	data_bytes := sysfs.Class.Object("ubi").SubObject(dev).Attribute("data_bytes")
```
pasinskim: `dataBytes` is a convention we are using.

device.go
mirzak: Got it.

ioctl.go
```diff
@@ -29,17 +32,48 @@ type ioctlRequestValue uintptr
...
+
+	if data_bytes.Exists() == false {
```
pasinskim: `if !data_bytes.Exists() {`

device.go
mirzak: I did take a look at `io.CopyN` initially but I believe it has the same limitation as `io.Copy`, and that is that it copies 32kb as maximum on each iteration, even if size argument is larger.

ioctl.go
mirzak: Got it.

ioctl.go
mirzak: Got it.

ioctl.go
mirzak: Got it

ioctl.go
mirzak: Got it

ioctl_32_bit.go
```diff
@@ -16,5 +16,11 @@
 
 package main
 
+//take from <mtd/ubi-user.h
+const UBI_IOCVOLUP ioctlRequestValue = 0x40084f00
```
pasinskim: s/take/Taken/

pasinskim: There is also `>` missing.

ioctl_64_bit.go
```diff
@@ -16,5 +16,11 @@
 
 package main
 
+//take from <mtd/ubi-user.h
```
pasinskim: s/take/Taken and missing `>`

block_device.go
mirzak: Yes, will write up an better explanation. Could probably be inspired from http://lxr.free-electrons.com/source/include/uapi/mtd/ubi-user.h#L70

The reason we are not checking error return code here is because this is an attempt to keep the code generic. This would probably be the only location where we need an:
```
if UBI_UPDATE {
    //do something
}

``` 

And I have been actively trying to avoid these and the Mender client is not currently aware of what type it is updating. The idea was also that this fails "silently" on an SD/MMC. If it fails on UBI the update will fail, not optimal probably and maybe not worth keeping it "generic" for this reason?

ioctl_32_bit.go
mirzak: This was sloppiness :)

ioctl_64_bit.go
mirzak: This was sloppiness :)

mirzak:
> Overall looks nice. There are some smaller things that you can consider changing. Also Travis is complaining about some unused code, which you should remove so that the tests can be run and we can see if this is passing in our CI. Next, maybe we can think about adding some automated UBIFS tests together with this PR as this is quite a big and risky change.

Will fix the minors. Thank you for review.

I was aware of that the Travis test where failing, but I was awaiting some feedback on the "first" commit. I tested removing the unused code locally but I also need to re-write some tests because it failed due to changes I made (removed code).

> @mirzak @kacf do you know if we can test UBI in QEMU?

QEMU does support emulating NAND [1], but not something that I have attempted to do.

[1]. https://github.com/qemu/qemu/blob/master/hw/block/nand.c

kacf:
> Next, maybe we can think about adding some automated UBIFS tests together with this PR as this is quite a big and risky change.
> @mirzak @kacf do you know if we can test UBI in QEMU?

There are separate tasks for adding QEMU flash support and acceptance tests for QEMU, [MEN-1127](https://tracker.mender.io/browse/MEN-1127) and [MEN-1128](https://tracker.mender.io/browse/MEN-1128), respectively. I don't think this in itself should block this feature. Eystein and I talked about this and it's alright to mark this as experimental/community supported until proper tests are in place.

And of course it depends on your motivation, @mirzak. I don't know if you've ever looked at the `meta-mender/tests/acceptance` directory, but this is where such tests would go. These tests are more comprehensive than unit tests, and is yet another framework to learn, so whether or not this is something you would like to work on is up to you. The tests would be a requirement for us to professionally support this, but I expect that once they're written, whether by you or by us, we could take over maintenance, since I expect the maintenance to be quite minor.

block_device.go
pasinskim: That makes lot of sense. I was thinking more about adding some run time check here. Maybe something similar to `if strings.HasPrefix(inactivePartition, "/dev/") == false`?

device.go
pasinskim: That's a good point. How about `CopyBuffer` then? There you can provide a buffer of a `ssz` length and the rest will be done "for free".

mirzak: `CopyBuffer` definitely looks interesting. Will take a closer look and test it out.

mirzak:
Updated the pull-request based on your review comments and also re-worked the "first commit" a bit as well. So that will require some extra attention from reviewers. 

The tests are still failing which I will start looking in to now as well, but wanted to push the changes while I fix the tests.

mirzak:
Tests fixed, had to mock BlockDeviceGetSectorSizeOf in the same way that you do with BlockDeviceGetSizeOf

block_device.go
```diff
@@ -45,6 +53,32 @@ func (bd *BlockDevice) Write(p []byte) (int, error) {
...
+		// close(fd);
+		if bd.typeUBI {
+			_, _, err := syscall.Syscall(syscall.SYS_IOCTL, out.Fd(),
+				uintptr(unsafe.Pointer(UBI_IOCVOLUP)),
+				uintptr(unsafe.Pointer(&bd.ImageSize)))
+			if err != 0 {
+				log.Errorf("Failed to write images size to UBI_IOCVOLUP: %v", err)
+				return 0, err
+			}
+		}
```
kacf: Could you move this syscall also to a separate function in `ioctl.go`? The reason is that this is very Linux specific, and we are considering ports to other platforms such as FreeBSD. It's easier if things like these are collected in one place.

device.go
```diff
@@ -78,7 +80,19 @@ func (d *device) InstallUpdate(image io.ReadCloser, size int64) error {
 		return err
 	}
 
-	b := &BlockDevice{Path: inactivePartition}
+	typeUBI := sysfs.Class.Object("ubi").SubObject(inactivePartition).Exists()
```
kacf: Maybe move this also to `ioctl.go`. It's not as clear a match as the syscall, but sysfs is definitely Linux specific as well.

ioctl.go
```diff
@@ -29,17 +32,48 @@ type ioctlRequestValue uintptr
...
+// Returns value in first return. Second returns error condition.
+// If the device is not a block device NotABlockDevice error and
+// value 0 will be returned.
+func ioctl(fd uintptr, request ioctlRequestValue) (uint64, error) {
+	var response uint64
 	_, _, errno := syscall.Syscall(syscall.SYS_IOCTL, fd,
-		uintptr(unsafe.Pointer(ioctlRequest)),
-		uintptr(unsafe.Pointer(&blkSize)))
+		uintptr(unsafe.Pointer(request)),
+		uintptr(unsafe.Pointer(&response)))
```
kacf: Nice refactoring!

kacf:
I'm also giving this a test spin in [our internal Jenkins](https://ci.cfengine.com/job/yoctobuild-pr/552).

mirzak:
> Overall, extremely nice work, @mirzak! Sorry for being a bit late to the party, I only have some minor comments anyway, I think you guys handled most issues already.

No need to apologize, I know that there is a lot of going on.

> Can you add a Changelog: None tag to all the commits except the last one? This is required by our new builder to make sure all changes are checked off for a changelog entry.

Will do.

>For the last commit, can you add something like Changelog: Introduce experimental support for writing to UBI volumes? This will automatically be put in the changelog for the next version.

Will do

block_device.go
mirzak: Yes, will do.

device.go
mirzak: Yes will do.

ioctl.go
mirzak: Thanks! :)

kacf:
This needs to be merged together with https://github.com/mendersoftware/meta-mender/pull/301, rerunning using [new internal job](https://ci.cfengine.com/job/yoctobuild-pr/560).

kacf:
@mirzak: The tests are failing, but I *think* it is because the branch is out of date. Could you rebase it?

mirzak:
Will rebase tonight @kacf

mirzak:
Did a rebase now, but might have done it in a haste and got a "merge" in the pull request now...

kacf:
Yeah, better avoid that. I created #213 now.

kacf:
Thanks a lot, @mirzak! Merged in #213.

PR #202, opened 2017-05-02
MEN-1195 Use verification key location from configuration file.

pasinskim:
Remove command line parameter specifying the location of the
artifact signature verification key and use the value stored
in configuration file instead.

Changelog: None

### DIFF
--- a/main.go
+++ b/main.go
@@ -42,7 +42,6 @@ type runOptionsType struct {
 	config         *string
 	dataStore      *string
 	imageFile      *string
-	verifyKey      *string
 	commit         *bool
 	bootstrap      *bool
 	daemon         *bool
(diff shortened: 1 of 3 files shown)


pasinskim:
Basically there are tests here: https://github.com/mendersoftware/mender/blob/master/installer/installer_test.go#L48. Those are covering various cases when the signature is present.

kacf:
Thanks! Testing this [in Jenkins](https://ci.cfengine.com/job/yoctobuild-pr/506/) now, and will test it with my WIP acceptance test as well.

kacf:
One failure, but I'm 99% sure it's spurious. I'm taking my chances and merging this!

PR #206, opened 2017-05-11
Add log message containing detailed info about installed artifact.

pasinskim:
Changelog: None


pasinskim:
@GregorioDiStefano should we add some more stuff here?

GregorioDiStefano:
i think this is good for now

PR #208, opened 2017-05-12
MEN-1191 Store scripts form artifact

pasinskim:


installer/script_installer.go
```diff
@@ -0,0 +1,119 @@
...
+		} else if err != nil {
+			return err
+		}
+		scrBkpDir := s.store + "-bkp"
+		err = os.Rename(s.store, scrBkpDir)
+		if err != nil {
+			return err
+		}
+
+		err = os.Rename(s.tmpStore, s.store)
```
kacf: I suspect that the `os.rename()` method might not work together with the `ioutil.TempDir()` call that you use above. I assume it will select a directory in `/tmp`, which will almost certainly be on a different filesystem than the final location.

pasinskim:
Superseded by https://github.com/mendersoftware/mender/pull/212

PR #212, opened 2017-05-19
State scripts implementation

pasinskim:
DO NOT MERGE. Testing!


pasinskim:
@kacf you can start reviewing. The code is complete, I will probably still add some more unit tests. There is one integration test failing, but I've been talking to @GregorioDiStefano about this one. It needs to be changed because of some changes in new implementation.

PR #213, opened 2017-05-30
Ubi support v2

kacf:
Rebased version of #200.


kacf:
Passed tests! (Never mind the automatic checks, they don't include the meta-mender change).

mirzak:
Woho! :)

PR #215, opened 2017-06-07
Mender state scripts implementation

pasinskim:


transition.go
```diff
@@ -0,0 +1,142 @@
...
+func (t Transition) String() string {
+	return transitionNames[t]
+}
+
+// For some states we should ignore errors as recovery is not possible
+// and we might end up with device bricked.
+func ignoreErrors(t Transition, action string) bool {
+	return t == ToIdle ||
+		t == ToArtifactRollback ||
+		t == ToArtifactRollbackReboot_Enter ||
```
kacf: `ToArtifactRollbackReboot_Leave` as well, or? At least I cannot see how you could possibly recover from this situation.

transition.go
```diff
@@ -0,0 +1,142 @@
...
+		if spl[1] != "Leave" {
+			return nil
+		}
+	}
+
+	if err := exec.ExecuteAll(name, "Error", ignoreErrors(t, "Error")); err != nil {
+		return errors.Wrapf(err, "error running error state script(s) for %v state", t)
+	}
+	return nil
+}
```
kacf: Is the only difference between these three functions the words "Enter", "Leave" and "Error"? I think you could consolidate these by having an extra argument with this word in it. Possibly with three wrappers to call it. Less code means less maintenance!

transition.go
```diff
@@ -0,0 +1,142 @@
...
+func (t Transition) Error(exec statescript.Executor) error {
+	if t == ToNone {
+		return nil
+	}
+
+	name := t.String()
+	spl := strings.Split(name, "_")
+	if len(spl) == 2 {
+		name = spl[0]
+		if spl[1] != "Leave" {
```
kacf: This should be `"Error"`, no?

state_test.go
```diff
@@ -41,6 +41,7 @@ type stateTestController struct {
 	state           State
 	updateResp      *client.UpdateResponse
 	updateRespErr   menderError
+	authorized      bool
 	authorize       menderError
```
kacf: Since you are adding `authorized`, suggest renaming `authorize` to `authorizeErr` to avoid confusion.

mender.go
```diff
@@ -327,7 +330,14 @@ func (m *mender) loadAuth() menderError {
 }
 
 func (m *mender) IsAuthorized() bool {
-	return m.authMgr.IsAuthorized()
+	if m.authMgr.IsAuthorized() {
+		log.Info("authorization data present and valid")
```
kacf: Maybe move to `Debug`? This will be hit 99% of the time for a healthy device, won't it?

state.go
```diff
@@ -300,128 +294,85 @@ func (i *IdleState) Handle(ctx *StateContext, c Controller) (State, bool) {
...
-		return NewUpdateVerifyState(sd.UpdateInfo), false
-
-		// this should not happen
+	// this should not happen
 	default:
-		log.Errorf("got invalid update state: %v", sd.Name)
-		me := NewFatalError(errors.New("got invalid update state"))
+		log.Errorf("got invalid state: %v", sd.Name)
+		me := NewFatalError(errors.Errorf("got invalid state stored: %v", sd.Name))
+
```
kacf: Won't Idle be a valid state here? If you reboot the device normally, for example?

state.go
```diff
@@ -1075,11 +1075,13 @@ func (e *RebootState) Handle(ctx *StateContext, c Controller) (State, bool) {
 
 type RollbackState struct {
 	UpdateState
+	swap bool
 }
 
-func NewRollbackState(update client.UpdateResponse) State {
+func NewRollbackState(update client.UpdateResponse, swapPartitions bool) State {
```
kacf: > Sometimes we need to go through rollback state even if rollback (swapping partitions) is not needed.

When? You're only calling this function with `swapPartitions` set to `true`.

statescript/store.go
```diff
@@ -0,0 +1,90 @@
...
+)
+
+type Store struct {
+	location string
+}
+
+func NewStore(destination string) *Store {
+	return &Store{
+		location: destination,
+	}
```
pasinskim: Not sure if I understand this one.

statescript/store.go
```diff
@@ -0,0 +1,90 @@
...
+	err := os.RemoveAll(s.location)
+	if err == nil || (err != nil && os.IsNotExist(err)) {
+		return os.MkdirAll(s.location, 0755)
+	}
+	return err
+}
+
+func (s *Store) StoreScript(r io.Reader, name string) error {
+	sLocation := filepath.Join(s.location, name)
+	f, err := os.OpenFile(sLocation, os.O_RDWR|os.O_CREATE|os.O_TRUNC, 0755)
```
pasinskim: Fixing!

installer/installer.go
```diff
@@ -85,11 +88,28 @@ func Install(art io.ReadCloser, dt string, key []byte, device UInstaller) error
 		return s.Verify(message, sig)
 	}
 
+	scr := statescript.NewStore(scrDir)
+	// we need to wipe out the scripts directory first
+	if err := scr.Clear(); err != nil {
+		log.Errorf("installer: can not remove script directory %s: %v", scrDir, err)
+		return errors.Wrap(err, "installer: can not remove script directory")
```
pasinskim: Right. Will change the message here.

transition.go
```diff
@@ -0,0 +1,142 @@
...
+	ToError
+	ToDownload
+	ToArtifactInstall
+	// should hsve Enter and Error actions
+	ToArtifactReboot_Enter
+	// should have Leave action only
+	ToArtifactReboot_Leave
+	ToArtifactCommit
+	ToArtifactRollback
+	// should hsve Enter and Error actions
```
pasinskim: Fixed!

transition.go
pasinskim: Not really. Only entering state is important and errors must be ignored. When you are leaving given state and something will fail you will simply call state.Error and then move on the the error state.

transition.go
pasinskim: Nope. Maybe I will add comment to explain the logic. Here, if we are in exit part of the given state (`ToArtifactReboot_Leave` or `ToArtifactRollbackReboot_Leave`) and something will fail we are supposed to call error actions scripts.

transition.go
pasinskim: Makes sense. Will refactor code.

state_test.go
pasinskim: Makes sense. Fixing!

state.go
pasinskim: This was added in subsequent commit which maybe should be squashed with this one. Anyway, we are using this in more places with the argument set to `false`.

state.go
pasinskim: Was superseded by changes in newer commits.

statescript/store.go
pasinskim: Fixed now! Check if we are using root or relative path was added.

pasinskim:
@kacf can you take a look at last commit? I think I got everything, but I've also added some more tests.

mender.go
kacf: What about this one? Should it be moved to `Debug` or not?

There is one below this one as well which may be a candidate for the same.

state.go
kacf: Oh, that's right. For some reason my commits were out of order... Weird.

statescript/executor.go
```diff
@@ -0,0 +1,141 @@
+// Copyright 2016 Mender Software AS
```
kacf: This should still be fixed.

pasinskim: Done!

pasinskim:
Separate feature branch containing those changes was created. After integration tests are done we can merge it to master.

statescript/store.go
```diff
@@ -0,0 +1,96 @@
...
+			"statescript: can not create script file: %v", sLocation)
+	}
+	defer f.Close()
+
+	_, err = io.Copy(f, r)
+	if err != nil {
+		return errors.Wrapf(err,
+			"statescript: can not write script file: %v", sLocation)
+	}
+	return nil
```
kacf: Btw, it just occurred to me that we need to call `f.Sync()` here, because if we don't, the U-Boot environment may be synced first, and if we lose power at that moment, we will boot into the new update *without* the scripts. So I suggest adding `f.Sync()` before `return nil`.

PR #218, opened 2017-06-19
MEN-505: Fix - Now throws an error when committing nothing.

oleorhagen:
Changelog: Title

### DIFF
--- a/device.go
+++ b/device.go
@@ -32,6 +32,7 @@ type device struct {
 	BootEnvReadWriter
 	Commander
 	*partitions
+	hasUpdateCb func(d *device) (bool, error)
 }
 
 func NewDevice(env BootEnvReadWriter, sc StatCommander, config deviceConfig) *device {
@@ -43,7 +44,12 @@ func NewDevice(env BootEnvReadWriter, sc StatCommander, config deviceConfig) *de
 		active:            "",
 		inactive:          "",
 	}
-	device := device{env, sc, &partitions}
+	device := device{
+		BootEnvReadWriter: env,
+		Commander:         sc,
+		partitions:        &partitions,
+	}
+	device.hasUpdateCb = hasUpdate
 	return &device
 }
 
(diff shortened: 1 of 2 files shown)


device.go
```diff
@@ -32,6 +32,7 @@ type device struct {
 	BootEnvReadWriter
 	Commander
 	*partitions
+	hasUpdateCb func(d *device) (bool, error)
```
bboozzoo: It looks to me that this callback is only needed for testability sake, in which case it makes no sense to have it here and instead update the tests to inject proper data as `BootEnvReadWriter` interface.

oleorhagen: @pasinskim

bboozzoo: An example of injecting `BootEnvReadWriter` with desired data is right here: https://github.com/mendersoftware/mender/pull/218/commits/f3e8bea7137b061c08ebc6c95cf30c0dba94e499#diff-eed9e366efcb961973c3dbd3d4b7d620R172

pasinskim: @bboozzoo the problem is that we need to read the env twice, and there is no way to mock env read/writer to return two different things for two different calls.

bboozzoo: Of course you can, just use a different interface:
```
// implements BootEnvReadWriter
type fakeBootEnv struct {
	readVars  BootVars
	readErr   error
	writeVars BootVars
	writeErr  error
}

func (f *fakeBootEnv) ReadEnv(...string) (BootVars, error) {
	return f.readVars, f.readErr
}

func (f *fakeBootEnv) WriteEnv(w BootVars) error {
	f.writeVars = w
	return f.writeErr
}

func Test_commitUpdate(t *testing.T) {
	device := device{}

	device.BootEnvReadWriter = &fakeBootEnv{
		readVars: BootVars{
			"upgrade_available": "1",
		},
	}

	if err := device.CommitUpdate(); err != nil {
		t.FailNow()
	}

	device.BootEnvReadWriter = &fakeBootEnv{
		readVars: BootVars{
			"upgrade_available": "1",
		},
		writeErr: errors.New("failed"),
	}
	if err := device.CommitUpdate(); err == nil {
		t.FailNow()
	}
}
```

pasinskim: @bboozzoo right, I haven't thought about that. Makes perfect sense. @oleorhagen I think what @bboozzoo is suggesting makes lot of sense.

oleorhagen: Yeh. Way prettier than our hack. I'll fix it up now.

oleorhagen: So I created a new branch, and a new pull request: #220

kacf:
Superseded by #220.

PR #219, opened 2017-06-20
MEN-1215: Changed the errormessage to more closely reflect the issue.

oleorhagen:
Changelog: Title

### DIFF
--- a/main.go
+++ b/main.go
@@ -405,7 +405,10 @@ func doMain(args []string) error {
 	switch {
 
 	case *runOptions.imageFile != "":
-		dt := GetDeviceType(defaultDeviceTypeFile)
+		dt, err := GetDeviceType(defaultDeviceTypeFile)
+		if err != nil {
+			log.Errorf("Unable to verify the existing hardware. Update will continue anyways: %v : %v", defaultDeviceTypeFile, err)
+		}
 		vKey := config.GetVerificationKey()
 		return doRootfs(device, runOptions, dt, vKey)
 
--- a/mender.go
+++ b/mender.go
@@ -213,7 +213,7 @@ func getManifestData(dataType, manifestFile string) string {
 	// This is where Yocto stores buid information
 	manifest, err := os.Open(manifestFile)
 	if err != nil {
-		log.Errorf("Can not read manifest field '%s' from file: %s", dataType, manifestFile)
+		log.Errorf("device_type file not found. Unable to verify the existing hardware. Update will continue anyways, '%s' from file: %s", dataType, manifestFile)
 		return ""
 	}
 
--- a/mender.go
+++ b/mender.go
@@ -48,7 +48,7 @@ type UInstallCommitRebooter interface {
 type Controller interface {
 	Authorize() menderError
 	Bootstrap() menderError
-	GetCurrentArtifactName() string
+	GetCurrentArtifactName() (string, error)
 	GetUpdatePollInterval() time.Duration
 	GetInventoryPollInterval() time.Duration
 	GetRetryPollInterval() time.Duration
@@ -209,12 +209,11 @@ func NewMender(config menderConfig, pieces MenderPieces) (*mender, error) {
 	return m, nil
 }
 
-func getManifestData(dataType, manifestFile string) string {
+func getManifestData(dataType, manifestFile string) (string, error) {
 	// This is where Yocto stores buid information
 	manifest, err := os.Open(manifestFile)
 	if err != nil {
-		log.Errorf("device_type file not found. Unable to verify the existing hardware. Update will continue anyways, '%s' from file: %s", dataType, manifestFile)
-		return ""
+		return "", err
 	}
 
 	scanner := bufio.NewScanner(manifest)
@@ -226,23 +225,24 @@ func getManifestData(dataType, manifestFile string) string {
 			lineID := strings.Split(line, "=")
 			if len(lineID) != 2 {
 				log.Errorf("Broken device manifest file: (%v)", lineID)
-				return ""
+				return "", fmt.Errorf("Broken device manifest file: (%v)", lineID)
 			}
 			log.Debug("Current manifest data: ", strings.TrimSpace(lineID[1]))
-			return strings.TrimSpace(lineID[1])
+			return strings.TrimSpace(lineID[1]), nil
 		}
 	}
-	if err := scanner.Err(); err != nil {
+	err = scanner.Err()
+	if err != nil {
 		log.Error(err)
 	}
-	return ""
+	return "", err
 }
 
-func (m *mender) GetCurrentArtifactName() string {
+func (m *mender) GetCurrentArtifactName() (string, error) {
 	return getManifestData("artifact_name", m.artifactInfoFile)
 }
 
-func (m *mender) GetDeviceType() string {
+func (m *mender) GetDeviceType() (string, error) {
 	return getManifestData("device_type", m.deviceTypeFile)
 }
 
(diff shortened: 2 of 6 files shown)


mender.go
```diff
@@ -213,7 +213,7 @@ func getManifestData(dataType, manifestFile string) string {
 	// This is where Yocto stores buid information
 	manifest, err := os.Open(manifestFile)
 	if err != nil {
-		log.Errorf("Can not read manifest field '%s' from file: %s", dataType, manifestFile)
+		log.Errorf("device_type file not found. Unable to verify the existing hardware. Update will continue anyways, '%s' from file: %s", dataType, manifestFile)
```
kacf: This section is shared and called by several functions with different files. I suggest you put the check and the message in the function that calls `GetDeviceType()` instead.

Also, consider including the `err.Error()` string somewhere in the log output, because you don't actually know that the file cannot be found, only that it cannot be opened, which could be because permissions are wrong. Standard practice in Mender is to do something like: `log.Errorf('%s' could not be opened: %s", file, err.Error())` (colon separating high level error from low level error).

main.go
```diff
@@ -405,7 +405,10 @@ func doMain(args []string) error {
 	switch {
 
 	case *runOptions.imageFile != "":
-		dt := GetDeviceType(defaultDeviceTypeFile)
+		dt, err := GetDeviceType(defaultDeviceTypeFile)
+		if err != nil {
```
oleorhagen: @kacf now the function throws an error.

mender.go
```diff
@@ -357,15 +357,19 @@ func (m *mender) FetchUpdate(url string) (io.ReadCloser, int64, error) {
 // that occurred. If no update is available *UpdateResponse is nil, otherwise it
 // contains update information.
 func (m *mender) CheckUpdate() (*client.UpdateResponse, menderError) {
-	currentArtifactName := m.GetCurrentArtifactName()
-	//TODO: if currentArtifactName == "" {
-	// 	return errors.New("")
-	// }
+	currentArtifactName, err := m.GetCurrentArtifactName()
+	if err != nil {
+		log.Error("could not get the current artifact name")
```
oleorhagen: A log should be good enough for notifying the user that we could not get the artifact?

main.go
oleorhagen: @kacf now the get-functions returns error instead of empty string

mender.go
oleorhagen: Appropriate to simply log it as an error? Or should an error be thrown?

mender_test.go
```diff
@@ -53,10 +53,12 @@ func Test_getArtifactName_noArtifactNameInFile_returnsEmptyName(t *testing.T) {
...
 	mender.artifactInfoFile = "artifact_info"
 
-	assert.Equal(t, "", mender.GetCurrentArtifactName())
+	artName, err := mender.GetCurrentArtifactName()
+	assert.NoError(t, err)
+	assert.Equal(t, "", artName)
 }
 
-func Test_getArtifactName_malformedArtifactNameLine_returnsEmptyName(t *testing.T) {
+func Test_getArtifactName_malformedArtifactNameLine_returnsError(t *testing.T) {
```
oleorhagen: Changed the name of the test function to reflect that it now checks for error instead of empty string. All tests are passing. Thinking this should not be an issue? 
Also, why are these test-functions of a different format. e.g. why the underscores?

state_test.go
```diff
@@ -55,8 +55,8 @@ func (s *stateTestController) Bootstrap() menderError {
 	return s.bootstrapErr
 }
 
-func (s *stateTestController) GetCurrentArtifactName() string {
-	return s.artifactName
+func (s *stateTestController) GetCurrentArtifactName() (string, error) {
+	return s.artifactName, nil // TODO return an error?
```
oleorhagen: The state test controller never throws an error. I'm thinking this is okay?

mender.go
kacf: No, I think this is ok. For delta updates we may have to change this, but we'll cross that bridge later.

mender_test.go
kacf: Not sure about the underscores, but yes, makes sense!

state.go
```diff
@@ -357,8 +357,12 @@ func (uv *UpdateVerifyState) Handle(ctx *StateContext, c Controller) (State, boo
 	}
 
 	if has {
-		if uv.update.ArtifactName() == c.GetCurrentArtifactName() {
-			log.Infof("successfully running with new image %v", c.GetCurrentArtifactName())
+		artifactName, err := c.GetCurrentArtifactName()
+		if err != nil {
+			log.Errorf("Cannot determine current artifact. Update will continue anyways: %v : %v", defaultDeviceTypeFile, err)
+		}
```
kacf: Actually, here we should throw an error. The reason is that this is the verification code before committing a new update, so this is a good place to treat it like an error, roll back, and force the creator of the artifact to fix the problem.

Above you can see how to treat errors in the state machine, and you can also ask Marcin.

state.go
```diff
@@ -359,7 +359,9 @@ func (uv *UpdateVerifyState) Handle(ctx *StateContext, c Controller) (State, boo
 	if has {
 		artifactName, err := c.GetCurrentArtifactName()
 		if err != nil {
-			log.Errorf("Cannot determine current artifact. Update will continue anyways: %v : %v", defaultDeviceTypeFile, err)
+			log.Errorf("Cannot determine current artifact. Update will note continue: %v : %v", defaultDeviceTypeFile, err)
```
oleorhagen: @kacf still logs, but also throws an error when no artifact-file can be found.

state_test.go
```diff
@@ -56,6 +56,9 @@ func (s *stateTestController) Bootstrap() menderError {
 }
 
 func (s *stateTestController) GetCurrentArtifactName() (string, error) {
+	if s.artifactName == "" {
```
oleorhagen: No longer a dummy. Throws an error when no artifact file is set.

state_test.go
```diff
@@ -577,6 +580,13 @@ func TestUpdateVerifyState(t *testing.T) {
 	assert.IsType(t, &RebootState{}, s)
 	assert.False(t, c)
 
+	// Test upgrade available and no artifact-file found
```
oleorhagen: Tests the case where an upgrade is available, but no artifact-file can be found.

state.go
```diff
@@ -357,8 +357,14 @@ func (uv *UpdateVerifyState) Handle(ctx *StateContext, c Controller) (State, boo
 	}
 
 	if has {
-		if uv.update.ArtifactName() == c.GetCurrentArtifactName() {
-			log.Infof("successfully running with new image %v", c.GetCurrentArtifactName())
+		artifactName, err := c.GetCurrentArtifactName()
+		if err != nil {
+			log.Errorf("Cannot determine current artifact. Update will note continue: %v : %v", defaultDeviceTypeFile, err)
+			me := NewFatalError(errors.Wrapf(err, "Cannot determine current artifact. Update will not continue: %v : %v", defaultDeviceTypeFile, err))
```
kacf: I suggest changing both of these to:

```
log.Errorf("Cannot determine name of new artifact. Update will not continue: %v : %v", defaultDeviceTypeFile, err)
me := NewFatalError(errors.Wrapf(err, "Cannot determine name of new artifact. Update will not continue: %v : %v", defaultDeviceTypeFile, err))
```

Just to emphasize that the issue is in the *new* artifact, since this is the only place where you actually have a shot at fixing the issue.

oleorhagen:
@kacf updated the comment

state.go
oleorhagen: @kacf sry. Fixed it up now.

kacf:
This passed in internal Jenkins.

PR #220, opened 2017-06-20
MEN-505: Fix - Now throws an error when committing nothing.

oleorhagen:
Changelog: Title

Added fix from bboozzoo.


oleorhagen:
ref: #218

pasinskim:
@GregorioDiStefano any reason why integration tests are not triggered here?

GregorioDiStefano:
@pasinskim yup, trying to move to Debian 9, instead of 8.8, this job is queued, will run soon.

PR #222, opened 2017-06-26
Men 1151

oleorhagen:
Now handles the case where it would previously crash when no sys-certs were available on the system.

### DIFF
--- a/client/client.go
+++ b/client/client.go
@@ -191,22 +191,32 @@ func loadServerTrust(conf Config) (*x509.CertPool, error) {
 		return nil, nil
 	}
 
-	certs, err := x509.SystemCertPool()
+	syscerts, err := x509.SystemCertPool()
 	if err != nil {
 		return nil, err
 	}
 
 	// Read certificate file.
-	cacert, err := ioutil.ReadFile(conf.ServerCert)
+	servcert, err := ioutil.ReadFile(conf.ServerCert)
 	if err != nil {
 		return nil, err
 	}
-	certs.AppendCertsFromPEM(cacert)
 
-	if len(certs.Subjects()) == 0 {
+	if len(servcert) == 0 {
+		return nil, errors.New("unable to find system and server certificates")
+	}
+
+	if syscerts == nil {
+		log.Warn("No system certificates found.")
+		syscerts = x509.NewCertPool()
+	}
+
+	syscerts.AppendCertsFromPEM(servcert)
+
+	if len(syscerts.Subjects()) == 0 {
 		return nil, errorAddingServerCertificateToPool
 	}
-	return certs, nil
+	return syscerts, nil
 }
 
 func loadClientCert(conf Config) (*tls.Certificate, error) {
--- a/client/client_test.go
+++ b/client/client_test.go
@@ -14,9 +14,12 @@
 package client
 
 import (
+	"io/ioutil"
 	"net"
 	"net/http"
 	"net/http/httptest"
+	"os"
+	"runtime"
 	"strings"
 	"testing"
 	"time"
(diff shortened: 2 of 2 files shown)


client/client.go
```diff
@@ -191,22 +191,34 @@ func loadServerTrust(conf Config) (*x509.CertPool, error) {
 		return nil, nil
 	}
 
-	certs, err := x509.SystemCertPool()
+	syscerts, err := x509.SystemCertPool()
+	// fmt.Println(syscerts)
```
bboozzoo: fmt?

client/client.go
```diff
@@ -191,22 +191,34 @@ func loadServerTrust(conf Config) (*x509.CertPool, error) {
...
 	if err != nil {
 		return nil, err
 	}
 
 	// Read certificate file.
-	cacert, err := ioutil.ReadFile(conf.ServerCert)
+	servcert, err := ioutil.ReadFile(conf.ServerCert)
+	// fmt.Println(servcert)
```
bboozzoo: fmt

client/client_test.go
```diff
@@ -164,3 +167,29 @@ func TestCaLoading(t *testing.T) {
 	assert.True(t, systemOK)
 	assert.True(t, oursOK)
 }
+
+func TestEmptySystemCertPool(t *testing.T) {
+	version := runtime.Version()
+	if version == "1.6" || version == "1.7" || version == "1.8" {
```
bboozzoo: does 1.8.1 also return `"1.8"` runtime?

client/client.go
```diff
@@ -191,22 +191,34 @@ func loadServerTrust(conf Config) (*x509.CertPool, error) {
...
 	if err != nil {
 		return nil, err
 	}
-	certs.AppendCertsFromPEM(cacert)
 
-	if len(certs.Subjects()) == 0 {
+	if servcert == nil && syscerts == nil {
```
bboozzoo: I'm guessing your intention is to handle a case when server certificat exists, but is empty, so `if len(servcert) == 0 { ...`  would be more appropriate.

You create a new cert pool a couple of lines down, so I don't see why checking `syscerts == nil` is needed here.

kacf: The intention is to catch when there is neither a specified certificate nor a pool of system certificates. I don't think we have handled the case where the file exists, but is empty. I'm not even sure if that would return nil or not. Might be worth checking.

GregorioDiStefano:
Integrations tests aren't running because @oleorhagen  needs to be part of the mendersoftware github group/org

client/client.go
```diff
@@ -191,22 +191,32 @@ func loadServerTrust(conf Config) (*x509.CertPool, error) {
...
 	// Read certificate file.
-	cacert, err := ioutil.ReadFile(conf.ServerCert)
+	servcert, err := ioutil.ReadFile(conf.ServerCert)
 	if err != nil {
 		return nil, err
 	}
-	certs.AppendCertsFromPEM(cacert)
 
-	if len(certs.Subjects()) == 0 {
+	if len(servcert) == 0 {
```
oleorhagen: Now doing prefix string search for version, and removed the commented out code. Also added @bboozzoo 
's suggestion of simply checking the number of bytes read from ReadFile.

client/client_test.go
```diff
@@ -164,3 +167,29 @@ func TestCaLoading(t *testing.T) {
 	assert.True(t, systemOK)
 	assert.True(t, oursOK)
 }
+
+func TestEmptySystemCertPool(t *testing.T) {
+	version := runtime.Version()
+	if strings.HasPrefix(version, "1.6") || strings.HasPrefix(version, "1.7") || strings.HasPrefix(version, "1.8") {
```
oleorhagen: Now checks the prefix of the version string to handle 1.x.x etc.

client/client.go
bboozzoo: yup, `len()` is smart enough to return 0 if `severcert == []byte{}` or `servercert == nil`

client/client_test.go
```diff
@@ -164,3 +167,29 @@ func TestCaLoading(t *testing.T) {
...
+		// Environment variable not included until version 1.9. Therefore skipping this test.
+		t.SkipNow()
+	}
+
+	tmpdir, err := ioutil.TempDir("", "nocertsfolder")
+	assert.NoError(t, err)
+
+	// Fake the environment variables, to override ssl-cert lookup
+	err = os.Setenv("SSL_CERT_DIR", tmpdir)
```
bboozzoo: This is go 1.9 goodness, right?

oleorhagen: That is right ^^

kacf:
Not sure why integration tests didn't run. Running this [internally](https://ci.cfengine.com/job/yoctobuild-pr/642).

PR #225, opened 2017-07-10
Men-1071

oleorhagen:
Added a tenant-token variable to menderconfig, and also implemented a simple test.

Also, do you want me to clean up the git history?

### DIFF
--- a/config.go
+++ b/config.go
@@ -22,6 +22,8 @@ import (
 	"github.com/pkg/errors"
 )
 
+const defaultTenantToken = "authtentoken"
+
 type menderConfig struct {
 	ClientProtocol    string
 	ArtifactVerifyKey string
@@ -38,6 +40,7 @@ type menderConfig struct {
 	ServerURL                    string
 	ServerCertificate            string
 	UpdateLogPath                string
+	TenantToken                  string
 }
 
 func LoadConfig(configFile string) (*menderConfig, error) {
@@ -87,6 +90,15 @@ func (c menderConfig) GetDeviceConfig() deviceConfig {
 
 func (c menderConfig) GetDeploymentLogLocation() string {
 	return c.UpdateLogPath
+}
+
+// GetTenantToken returns a default tenant-token if
+// no custom token is set in local.conf
+func (c menderConfig) GetTenantToken() []byte {
+	if c.TenantToken != "" {
+		return []byte(c.TenantToken)
+	}
+	return []byte(defaultTenantToken)
 }
 
 func (c menderConfig) GetVerificationKey() []byte {
--- a/main.go
+++ b/main.go
@@ -61,8 +61,6 @@ var (
 
 var defaultConfFile string = path.Join(getConfDirPath(), "mender.conf")
 
-const defaultTenantTokenFile string = "authtentoken"
-
 var DeploymentLogger *DeploymentLogManager
 
 type Commander interface {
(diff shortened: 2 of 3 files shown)


main.go
```diff
@@ -306,17 +304,20 @@ func getKeyStore(datastore string, keyName string) *Keystore {
 	return NewKeystore(dirstore, keyName)
 }
 
-func loadTenantToken(datastore string) ([]byte, error) {
+func loadTenantToken(config *menderConfig, datastore string) ([]byte, error) {
+
 	dirstore := NewDirStore(datastore)
-	raw, err := dirstore.ReadAll(defaultTenantTokenFile)
+
+	raw, err := dirstore.ReadAll(config.GetTenantToken())
```
bboozzoo: This might be slightly confusing. We're letting the user configure the name of tenant token file, but it must be located inside mender data store directory. Why give the user such ability even though the file must be inside a very specific directory?

oleorhagen: Fair point. What would you have instead? reading it directly from config?

bboozzoo: Just a thought, but this might be a good opportunity to clean up the config a bit, we have:

* `ArtifactVerifyKey` - seems to be an absolute path, 
* `DeviceKey` - relative path inside state dir, file name configurable (needed?)
* `UpdateLogPath` - absolute path to update logs?
* `TenantTokenPath` - relative inside state dir, file name configurable as well (really needed?)

Maybe we could remove device key and tenant token path from configuration and use fixed names instead. I don't think we've come up with a valid use cases for using different paths for these settings.

(Well, in theory I think there might be a case for using absolute paths for device key & tenant token, eg. when both files are accessible form some secure storage).

oleorhagen: Actually `TenantToken` is now the actual token, and not a file or a path, but I guess stopping now, and rehashing the details here is a good idea.
@kacf thoughts?

main.go
```diff
@@ -304,11 +304,16 @@ func getKeyStore(datastore string, keyName string) *Keystore {
 	return NewKeystore(dirstore, keyName)
 }
 
-func loadTenantToken(config *menderConfig, datastore string) ([]byte, error) {
+func loadTenantToken(config *menderConfig, datastore string) (raw []byte, err error) {
+
+	if raw = config.GetCustomTenantToken(); len(raw) != 0 {
+		err = nil
+		return
+	}
```
oleorhagen: @bboozzoo cleaned up configs method names to reflect what is actually going on.

main.go
```diff
@@ -304,11 +304,16 @@ func getKeyStore(datastore string, keyName string) *Keystore {
...
 
 	dirstore := NewDirStore(datastore)
 
-	raw, err := dirstore.ReadAll(config.GetTenantToken())
+	raw, err = dirstore.ReadAll(config.GetDefaultTenantTokenFileName())
```
oleorhagen: same

main.go
kacf: * `DeviceKey` - I agree, there is absolutely no need for this to be configurable. Feel free to remove this where it's used and replace with a const.
* `TenantTokenPath` - This will turn into an actual token, like @oleorhagen described, so this is a non-issue.

main.go
```diff
@@ -306,22 +304,30 @@ func getKeyStore(datastore string, keyName string) *Keystore {
...
+
+	if raw = config.GetCustomTenantToken(); len(raw) != 0 {
+		err = nil
+		return
+	}
+
 	dirstore := NewDirStore(datastore)
-	raw, err := dirstore.ReadAll(defaultTenantTokenFile)
+
+	raw, err = dirstore.ReadAll(config.GetDefaultTenantTokenFileName())
```
kacf: Since we decided to switch to a pure string in `mender.conf`, I don't think we need to load it from any file, which means that `datastore`, `GetDefaultTenantTokenFileName()` and `defaultTenantTokenFileName` can all be removed, as well as this logic.

In fact, since this now becomes much simpler, I think you can just remove the `loadTenantToken` function completely, and just call `config.GetCustomTenantToken()` directly (yey, I love code removal!).

config.go
```diff
@@ -95,12 +95,13 @@ func (c menderConfig) GetDeploymentLogLocation() string {
...
-
-func (c menderConfig) GetDefaultTenantTokenFileName() string {
-	return defaultTenantTokenFileName
+// GetTenantToken returns a default tenant-token if
+// no custom token is set in local.conf
+func (c menderConfig) GetTenantToken() []byte {
+	if c.TenantToken != "" {
+		return []byte(c.TenantToken)
+	}
+	return []byte(defaultTenantToken)
```
oleorhagen: @kacf now returns the token depending upon whether or not a token is set in local.conf. k?

kacf: Perfect!

oleorhagen:
@kacf fixed. However, I have not removed `DeviceKey`. Separate PR?

kacf:
Yes, separate PR please! I'm testing this [in internal Jenkins](https://ci.cfengine.com/job/yoctobuild-pr/670) now, since the external one is broken.

PR #227, opened 2017-07-13
Include file that was accidentally omitted in last commit.

kacf:
Probably it still passed because the file can be found in the GOPATH
instead.

Changelog: None


kacf:
Cherry-picked all copyright changes to 1.1.x as well.

PR #228, opened 2017-07-13
State scripts

kacf:


kacf:
Succeded together with https://github.com/mendersoftware/integration/pull/295 [here](https://mender-jenkins.mender.io/job/mender-builder/1693/console).

PR #235, opened 2017-07-20
State scripts latest updates and fixes

pasinskim:


pasinskim:
Passing integration tests here: https://mender-jenkins.mender.io/job/mender-builder/1532

installer/installer.go
```diff
@@ -33,7 +33,7 @@ type UInstaller interface {
 }
 
 func Install(art io.ReadCloser, dt string, key []byte, scrDir string,
```
bboozzoo: How about turning this into a struct? say
```
type InstallerConfig struct {
     DeviceType string
     Key *rsa.PublicKey
     ScriptsDir string
     AllowStateScripts bool
}
```
Then the function would become:
`func Install (from io.ReadCloser, device UInstaller, config InstallerConfig) error`, looks cleaner, doesn't it?

mender.go
```diff
@@ -529,20 +533,48 @@ func shouldTransit(from, to State) bool {
...
+		return new
+	case *UpdateStoreState:
+		if action == "Leave" {
+			new := NewUpdateStatusReportState(t.update, client.StatusFailure)
+			new.SetTransition(ToError)
+			return new
+		}
+		return NewUpdateErrorState(me, t.update)
+	case *UpdateInstallState:
+		return NewRollbackState(t.Update(), false, false)
```
bboozzoo: While you're doing a change that big, you might want to refactor NewRollbackState to take a struct:
```
type RollbackActions struct {
    SwapPartitions bool
    Reboot bool
}
```
otherwise it's hard to tell what `true, true` or `false, false` stands for. Then instead of 
`NewRollbackState(t.Update(), true, true)` you'd have:

```
NewRollbackState(t.Update(), RollbackActions{
    SwapPartitions: true,
    Reboot: true,
})
```

pasinskim:
Retested in Jenkins to make sure this is working: https://mender-jenkins.mender.io/job/mender-builder/1535. Build is green.

kacf:
Superseded by #228.

PR #236, opened 2017-07-26
Remove cert 1.1.x

GregorioDiStefano:


kacf:
Unless there is a strong reason I don't think we should put this change in a patch release. It changes the config file structure.

PR #239, opened 2017-08-15
Added retry-later functionality on top of the state-script functionality 

oleorhagen:
For now, a retry later will stop the main execution thread with a
time.Sleep for a default, or user-configured interval.

Changelog: Title

### DIFF
--- a/config.go
+++ b/config.go
@@ -32,16 +32,18 @@ type menderConfig struct {
 		Key         string
 		SkipVerify  bool
 	}
-	RootfsPartA                  string
-	RootfsPartB                  string
-	UpdatePollIntervalSeconds    int
-	InventoryPollIntervalSeconds int
-	RetryPollIntervalSeconds     int
-	StateScriptTimeoutSeconds    int
-	ServerURL                    string
-	ServerCertificate            string
-	UpdateLogPath                string
-	TenantToken                  string
+	RootfsPartA                     string
+	RootfsPartB                     string
+	UpdatePollIntervalSeconds       int
+	InventoryPollIntervalSeconds    int
+	RetryPollIntervalSeconds        int
+	StateScriptTimeoutSeconds       int
+	StateScriptRetryTimeoutSeconds  int
+	StateScriptRetryIntervalSeconds int
+	ServerURL                       string
+	ServerCertificate               string
+	UpdateLogPath                   string
+	TenantToken                     string
 }
 
 func LoadConfig(configFile string) (*menderConfig, error) {
--- a/statescript/executor.go
+++ b/statescript/executor.go
@@ -29,6 +29,15 @@ import (
 	"github.com/pkg/errors"
 )
 
+const (
+	// exitRetryLater - exit code returned if a script requests a retry
+	exitRetryLater = 21
+
+	defaultStateScriptRetryInterval time.Duration = 30 * time.Minute
+
+	defaultStateScriptRetryTimeout time.Duration = 60 * time.Second
+)
+
 type Executor interface {
 	ExecuteAll(state, action string, ignoreError bool) error
 	CheckRootfsScriptsVersion() error
@@ -39,11 +48,30 @@ type Launcher struct {
 	RootfsScriptsPath       string
 	SupportedScriptVersions []int
 	Timeout                 int
+	RetryInterval           int
+	RetryTimeout            int
+}
+
+func (l *Launcher) getRetryInterval() time.Duration {
+
+	if l.RetryInterval != 0 {
+		return time.Duration(l.RetryInterval) * time.Second
+	}
+	log.Warningf("No timeout interval set for the retry-scripts. Falling back to default: %s", defaultStateScriptRetryTimeout.String())
+	return defaultStateScriptRetryTimeout
+}
+
+func (l *Launcher) getRetryTimeout() time.Duration {
+
+	if l.RetryTimeout != 0 {
+		return time.Duration(l.RetryTimeout) * time.Second
+	}
+	log.Warningf("No total time set for the retry-scripts' timeslot. Falling back to default: %s", defaultStateScriptRetryInterval.String())
+	return defaultStateScriptRetryInterval
 }
 
 //TODO: we can optimize for reading directories once and then creating
 // a map with all the scripts that needs to be executed.
-
 func (l Launcher) CheckRootfsScriptsVersion() error {
 	// first check if we are having some scripts
 	scripts, err := ioutil.ReadDir(l.RootfsScriptsPath)
@@ -205,6 +233,55 @@ func execute(name string, timeout time.Duration) error {
 	return nil
 }
 
+func retCode(err error) int {
+	defaultFailedCode := -1
+
+	if err != nil {
+		// try to get the exit code
+		if exitError, ok := err.(*exec.ExitError); ok {
+			ws := exitError.Sys().(syscall.WaitStatus)
+			return ws.ExitStatus()
+		} else {
+			return defaultFailedCode
+		}
+	}
+	return 0
+}
+
+// Catches a script that requests a retry in a loop. Is limited by the total window given to a script demanding a
+// retry.
+func executeScript(s os.FileInfo, dir string, l Launcher, timeout time.Duration, ignoreError bool) error {
+
+	iet := time.Now()
+	for {
+		err := execute(filepath.Join(dir, s.Name()), timeout)
+		switch ret := retCode(err); ret {
+		case 0:
+			// success
+			return nil
+		case exitRetryLater:
+			if time.Since(iet) <= l.getRetryTimeout() {
+				log.Infof("statescript: %s requested a retry", s.Name())
+				time.Sleep(l.getRetryInterval())
+				continue
+			}
+			if ignoreError {
+				log.Errorf("statescript: ignoring error executing '%s': %d: %s", s.Name(), ret, err.Error())
+				return nil
+			}
+			return errors.Errorf("statescript: retry time-limit exceeded %s", err.Error())
+		default:
+			// In case of error scripts all should be executed.
+			if ignoreError {
+				log.Errorf("statescript: ignoring error executing '%s': %d: %s", s.Name(), ret, err.Error())
+				return nil
+			}
+			return errors.Errorf("statescript: error executing '%s': %d : %s",
+				s.Name(), ret, err.Error())
+		}
+	}
+}
+
 func (l Launcher) ExecuteAll(state, action string, ignoreError bool) error {
 	scr, dir, err := l.get(state, action)
 	if err != nil {
@@ -232,14 +309,9 @@ func (l Launcher) ExecuteAll(state, action string, ignoreError bool) error {
 			}
 		}
 
-		if err = execute(filepath.Join(dir, s.Name()), timeout); err != nil {
-			// In case of error scripts all should be executed.
-			if ignoreError {
-				log.Errorf("statescript: ignoring error executing '%s': %s", s.Name(), err.Error())
-			} else {
-				return errors.Errorf("statescript: error executing '%s': %s",
-					s.Name(), err.Error())
-			}
+		log.Debugf("executing script: %s", s.Name())
+		if err = executeScript(s, dir, l, timeout, ignoreError); err != nil {
+			return err
 		}
 	}
 	return nil
--- a/statescript/statescript_test.go
+++ b/statescript/statescript_test.go
@@ -16,6 +16,7 @@ package statescript
 
 import (
 	"bytes"
+	"fmt"
 	"io/ioutil"
 	"os"
 	"path/filepath"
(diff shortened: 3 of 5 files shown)


statescript/executor.go
```diff
@@ -179,8 +212,95 @@ func execute(name string, timeout time.Duration) int {
...
+	db := store.NewDBStore("/data/mender")
+	if db == nil {
+		return errors.New("failed to open database")
+	}
+	defer db.Close()
+
+	return l.executeAll(db, state, action, ignoreError)
+}
+
+func (l Launcher) executeAll(db *store.DBStore, state, action string, ignoreError bool) error {
```
oleorhagen: @pasinskim I originally split this up to do make testing easier. For now I have only unit tested my methods (which is pretty unflexible, so I might remove them, and replace it with some different tests). Still the main functionality here should be pretty straight forward, now it is just baked into the main execution path, and pauses it when a script requires a retry.

statescript/statescript_test.go
```diff
@@ -205,3 +212,107 @@ func testArtifactArrayEquals(t *testing.T, expected []string, actual []os.FileIn
 		assert.EqualValues(t, expected[i], script.Name())
 	}
 }
+
+// Unit test
+func TestRetryLaterContextUnmarshalBinary(t *testing.T) {
```
oleorhagen: As previously mentioned. Unit tests... :/

store/dbstore.go
```diff
@@ -197,3 +198,24 @@ func (dbw *DBStoreWrite) Close() error {
 func (dbw *DBStoreWrite) Commit() error {
 	return dbw.dbs.writeBytes(dbw.name, &dbw.data)
 }
+
+// Update - updates an existing dataentry, or creates a new one if none exists
```
oleorhagen: @pasinskim I added a couple of more methods to the database so that it will be easier to store say, structs etc. I know that @bboozzoo wants a text marshaller interface instead of a binary one, so that reading data out of the database manually is easier, but I think that a binary one is plenty as long as one is careful. Thoughts?

statescript/executor.go
```diff
@@ -206,14 +327,40 @@ func (l Launcher) ExecuteAll(state, action string, ignoreError bool) error {
...
+
+			ret := execute(filepath.Join(dir, s.Name()), timeout)
+
+			if ret != 0 {
+
+				// In case of error scripts all should be executed.
+				if ignoreError {
+					log.Errorf("statescript: ignoring error executing '%s': %d", s.Name(), ret)
+				} else if ret == exitRetryLater {
+					err = l.handleRetryLaterError(db, s.Name(), state, execTime)
```
pasinskim: I think `handleRetryLaterError` is misleading here. It is not an error in fact, so I suggest renaming to `handleRetryLater`.

statescript/executor.go
```diff
@@ -206,14 +327,40 @@ func (l Launcher) ExecuteAll(state, action string, ignoreError bool) error {
...
-		if ret := execute(filepath.Join(dir, s.Name()), timeout); ret != 0 {
-			// In case of error scripts all should be executed.
-			if ignoreError {
-				log.Errorf("statescript: ignoring error executing '%s': %d", s.Name(), ret)
-			} else {
-				return errors.Errorf("statescript: error executing '%s': %d",
-					s.Name(), ret)
+		retryLaterFlag := false
+
+		for { // retry-later
```
pasinskim: I think I would simplify this a bit here. You can leave the `for` loop, but how about something like:
```
for {
      switch ret := execute(filepath.Join(dir, s.Name()), timeout) {
      case 0:
        // success; just break the loop and continue
        break
      case exitRetryLater:
        // we need to re-run the script later
        time.Sleep(l.getRetryInterval())
        continue
      default:
        // some error occured
        
        // In case of error scripts all should be executed.
  			if ignoreError {
  				log.Errorf("statescript: ignoring error executing '%s': %d", s.Name(), ret)
          break
  			} else {
  				return errors.Errorf("statescript: error executing '%s': %d",
  					s.Name(), ret)
  			}
      }
    }
```
I think the above might cover the whole functionality we need ATM. I am not sure why we need to store some context inside the DB and read this all the time. All we need is probably busy waiting here.

config.go
```diff
@@ -32,16 +32,18 @@ type menderConfig struct {
...
-	UpdateLogPath                string
-	TenantToken                  string
+	RootfsPartA                     string
+	RootfsPartB                     string
+	UpdatePollIntervalSeconds       int
+	InventoryPollIntervalSeconds    int
+	RetryPollIntervalSeconds        int
+	StateScriptTimeoutSeconds       int
+	StateScriptRetryTimeoutSeconds  int
+	StateScriptRetryIntervalSeconds int
```
oleorhagen: @pasinskim as far as I can see, the StatescriptTimeoutSeconds is the default timeout for a state-script. I thought the retry-scripts were supposed to have a default timeout?

statescript/executor.go
```diff
@@ -28,6 +28,17 @@ import (
 	"github.com/pkg/errors"
 )
 
+const (
+	exitRetryLater = 21
+
+	defaultRetryWindow time.Duration = 5 * time.Second // Total time given to a retry-script
+
+	defaultRetryTimeout time.Duration = 10 * time.Second
+)
```
oleorhagen: Not sure what a sensible default will be here...

statescript/executor.go
```diff
@@ -38,6 +49,26 @@ type Launcher struct {
...
+	RetryTimeout            int
+}
+
+func (l *Launcher) getRetryInterval() time.Duration {
+
+	if l.RetryInterval != 0 {
+		return time.Duration(l.RetryInterval) * time.Second
+	}
+	log.Warningf("No timeout interval set for the retry-scripts. Falling back to default: %2s", defaultRetryTimeout.String())
+	return defaultRetryTimeout
```
oleorhagen: It is printed as a string so that I can limit the width of the print, for pretty-printing.

store/dbstore_test.go
```diff
@@ -103,4 +104,13 @@ func TestDBStore(t *testing.T) {
...
 	assert.Error(t, err)
 
+	// time implements BinaryMarshaler
+	tnow := time.Now()
+	err = d.Update("foo", tnow)
+	assert.NoError(t, err)
+
+	tret := new(time.Time)
+	err = d.Get("foo", tret)
+	assert.Equal(t, *tret, tnow)
```
oleorhagen: I'll remove this once you have reviewed :)

statescript/executor.go
```diff
@@ -206,13 +240,31 @@ func (l Launcher) ExecuteAll(state, action string, ignoreError bool) error {
...
+				return errors.New("statescript: retry time-limit exceeded")
+			default:
+				// In case of error scripts all should be executed.
+				if ignoreError {
+					log.Errorf("statescript: ignoring error executing '%s': %d", s.Name(), ret)
+				} else {
+					return errors.Errorf("statescript: error executing '%s': %d",
+						s.Name(), ret)
+				}
+				break retrylater
```
oleorhagen: when rewriting this into a switch statement, which is arguably prettier, I will have to use a break/goto, as a break will only break out of the switch. I like the esthetics of this solution. Kosher or not?

statescript/executor.go
```diff
@@ -28,6 +28,17 @@ import (
...
+const (
+	exitRetryLater = 21
+
+	defaultRetryWindow time.Duration = 5 * time.Second // Total time given to a retry-script
+
+	defaultRetryTimeout time.Duration = 10 * time.Second
+)
+
+// ErrRetryLater is the error returned by a script requesting a retry
+var ErrRetryLater = errors.New("retry script later")
```
pasinskim: I can not see this being used anywhere.

statescript/executor.go
```diff
@@ -206,13 +240,31 @@ func (l Launcher) ExecuteAll(state, action string, ignoreError bool) error {
...
+		iet := time.Now()
+
+	retrylater:
+		for {
+			switch ret := execute(filepath.Join(dir, s.Name()), timeout); ret {
+			case 0:
+				// success. Break the loop and continue
+				break retrylater
+			case exitRetryLater:
+				if time.Since(iet) <= l.getRetryTimeout() {
```
pasinskim: Not sure why do we need this check here and the timeout. Can't we simply read return code and if script is requesting retrying later we should wait the retry interval?

statescript/executor.go
```diff
@@ -206,13 +240,31 @@ func (l Launcher) ExecuteAll(state, action string, ignoreError bool) error {
...
+					log.Errorf("statescript: %s requested a retry", s.Name())
+					time.Sleep(l.getRetryInterval())
+					continue
+				}
+				return errors.New("statescript: retry time-limit exceeded")
```
pasinskim: Because I am not sure about the overall design with using two different timeouts, I can not say for sure, but probably in this case we should check `ignoreErrors` case as well and proceed with the other scripts to execute.

statescript/executor.go
```diff
@@ -179,8 +210,43 @@ func execute(name string, timeout time.Duration) int {
...
+			// In case of error scripts all should be executed.
+			if ignoreError {
+				log.Errorf("statescript: ignoring error executing '%s': %d", s.Name(), ret)
+				return nil
+			}
+			return errors.Errorf("statescript: error executing '%s': %d",
+				s.Name(), ret)
+		}
+	}
+}
```
oleorhagen: @pasinskim like we discussed the functionality has now been moved out into a function to hinder future breakage in case of misuse. Thus now a return is used to break out of the loop when needed, instead of other spurious hacks we've tried out previously.

statescript/executor.go
```diff
@@ -28,6 +28,17 @@ import (
 	"github.com/pkg/errors"
 )
 
+const (
+	exitRetryLater = 21
+
+	defaultRetryWindow time.Duration = 5 * time.Second // Total time given to a retry-script
```
pasinskim: Shouldn't retry window be longer than retry timeout? As for the defaults, I think we can do something like: retry every 60 seconds and timeout 30 minutes. What I am also thinking is (we can open separate JIRA ticket) we can have `0` timeout value which will let the script wait forever (might be risky though).

statescript/executor.go
pasinskim: Should we delete above? There are no references in the code.

statescript/executor.go
```diff
@@ -179,8 +210,43 @@ func execute(name string, timeout time.Duration) int {
...
+// retry.
+func executeScript(s os.FileInfo, dir string, l Launcher, timeout time.Duration, ignoreError bool) error {
+
+	iet := time.Now()
+	for {
+		switch ret := execute(filepath.Join(dir, s.Name()), timeout); ret {
+		case 0:
+			// success
+			return nil
+		case exitRetryLater:
```
pasinskim: I think the logic here is wrong. We should first sleep and then ignore errors if any. 
```
case exitRetryLater:
  if time.Since(iet) <= l.getRetryTimeout() {
    log.Errorf("statescript: %s requested a retry", s.Name())
    time.Sleep(l.getRetryInterval())
    continue
  }
  else if ignoreError {
    log.Errorf("statescript: ignoring error executing '%s': %d", s.Name(), ret)
    return nil
  }
  return errors.New("statescript: retry time-limit exceeded")
```

statescript/executor.go
```diff
@@ -179,8 +210,43 @@ func execute(name string, timeout time.Duration) int {
...
+			if ignoreError {
+				log.Errorf("statescript: ignoring error executing '%s': %d", s.Name(), ret)
+				return nil
+			}
+			if time.Since(iet) <= l.getRetryTimeout() {
+				log.Errorf("statescript: %s requested a retry", s.Name())
```
pasinskim: This should be the `Info` level.

statescript/executor.go
```diff
@@ -179,8 +210,43 @@ func execute(name string, timeout time.Duration) int {
...
+				return nil
+			}
+			return errors.Errorf("statescript: error executing '%s': %d",
+				s.Name(), ret)
+		}
+	}
+}
+
 func (l Launcher) ExecuteAll(state, action string, ignoreError bool) error {
+
```
pasinskim: Not a big thing, but I would rather avoid adding those empty lines every now and then in the existing code. I don't care about the new code, but if there are lot of those small not important changes it is harder to follow the changes later (when using git blame for example). Just a hint for future.

statescript/statescript_test.go
```diff
@@ -190,6 +191,36 @@ func TestExecutor(t *testing.T) {
 	sysInstallScripts, _, err = e.get("ArtifactInstall", "Leave")
 	testArtifactArrayEquals(t, scriptArr[1:], sysInstallScripts)
 	assert.NoError(t, err)
+
+	// add a script that will time-out, and die
+	filep, err := createArtifactTestScript(tmpArt, "ArtifactInstall_Leave_10_btoot", "!#/bin/bash \nsleep 2")
```
pasinskim: Are we removing this somewhere? Take a look at `defer` to do it Go way.

statescript/statescript_test.go
```diff
@@ -190,6 +191,36 @@ func TestExecutor(t *testing.T) {
...
+	assert.NoError(t, err)
+	err = l.ExecuteAll("ArtifactInstall", "Enter", false)
+	assert.Contains(t, err.Error(), "retry time-limit exceeded")
+
+	err = os.Remove(filepath.Join(tmpArt, "ArtifactInstall_Enter_66"))
+	assert.NoError(t, err)
+
+	// add a script that retries and then succeeds
+	script := fmt.Sprintf("#!/bin/bash \n sleep 1 \n if [ ! -f %s/scriptflag ]; then\n echo f > %[1]s/scriptflag\n exit 21 \nfi \n rm -f %[1]s/scriptflag\n exit 0", tmpArt)
+	_, err = createArtifactTestScript(tmpArt, "ArtifactInstall_Enter_67", script)
```
pasinskim: I think this one is not removed as well, unless we are removing the whole `tmpArt` directory when the tests are over.

oleorhagen: `tmpArt` is removed after the tests are run ;)

statescript/executor.go
oleorhagen: https://tracker.mender.io/browse/MEN-1479

pasinskim:
@oleorhagen looks good now. Just make sure to rebase as it looks like there are some conflicts here.

statescript/executor.go
```diff
@@ -205,6 +233,55 @@ func execute(name string, timeout time.Duration) error {
...
+			// success
+			return nil
+		case exitRetryLater:
+			if time.Since(iet) <= l.getRetryTimeout() {
+				log.Infof("statescript: %s requested a retry", s.Name())
+				time.Sleep(l.getRetryInterval())
+				continue
+			}
+			if ignoreError {
+				log.Errorf("statescript: ignoring error executing '%s': %d: %s", s.Name(), ret, err.Error())
```
oleorhagen: @kacf this is related to the fixup you did for statescript-error messages. This should do the same thing, don't you think?

kacf: Not sure which one you are referring to..?

oleorhagen: https://github.com/mendersoftware/mender/commit/f5ac3fcefced0c54ed348436f7e716bfa963e027

kacf: Aah, that one. Yes, I believe so!

PR #240, opened 2017-08-15
MEN-1349: Mender now logs whatever a state-script outputs to stderr

oleorhagen:
Will not log from Idle and sync states

Changelog: Title

### DIFF
--- a/statescript/executor.go
+++ b/statescript/executor.go
@@ -15,6 +15,7 @@
 package statescript
 
 import (
+	"io"
 	"io/ioutil"
 	"os"
 	"os/exec"
@@ -156,6 +157,16 @@ func execute(name string, timeout time.Duration) int {
 
 	cmd := exec.Command(name)
 
+	var stderr io.ReadCloser
+	var err error
+
+	if !strings.HasPrefix(name, "Idle") && !strings.HasPrefix(name, "Sync") {
+		stderr, err = cmd.StderrPipe()
+		if err != nil {
+			log.Errorf("statescript: %v", err)
+		}
+	}
+
 	// As child process gets the same PGID as the parent by default, in order
 	// to avoid killing Mender when killing process group we are setting
 	// new PGID for the executed script and its children.
@@ -163,6 +174,22 @@ func execute(name string, timeout time.Duration) int {
 
 	if err := cmd.Start(); err != nil {
 		return retCode(err)
+	}
+
+	var bts []byte
+	if stderr != nil {
+		bts, err = ioutil.ReadAll(stderr)
+		if err != nil {
+			log.Error(err)
+		}
+	}
+
+	if len(bts) > 0 {
+		if len(bts) > 10*1024 {
+			log.Errorf("stderr collected while running script %s [%s] (Truncated to 10KB)", name, bts[:10*1024])
+		} else {
+			log.Errorf("stderr collected while running script %s [%s]", name, string(bts))
+		}
 	}
 
 	timer := time.AfterFunc(timeout, func() {
--- a/statescript/statescript_test.go
+++ b/statescript/statescript_test.go
@@ -22,6 +22,7 @@ import (
 	"strconv"
 	"testing"
 
+	"github.com/mendersoftware/log"
 	"github.com/stretchr/testify/assert"
 )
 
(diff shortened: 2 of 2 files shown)


statescript/executor.go
```diff
@@ -152,10 +154,51 @@ func (l Launcher) getTimeout() time.Duration {
...
+	return false
+}
+
 func execute(name string, timeout time.Duration) int {
 
 	cmd := exec.Command(name)
 
+	var stderr io.ReadCloser
+	var err error
+	if !strings.Contains(name, "idle") && !strings.Contains(name, "Sync") { // do not log from these states
```
pasinskim: Shouldn't we have `Idle` here (capital letter)? What is more, I don't think we need this conditional path. It's not wrong to keep and maybe we should to be on a safe side, but the logger is not enabled if we are inside Idle or Sync state anyway, so we won't be able to log anything.

statescript/executor.go
```diff
@@ -165,6 +208,18 @@ func execute(name string, timeout time.Duration) int {
...
+	var bts []byte
+	if stderr != nil {
+		bts, err = ioutil.ReadAll(stderr)
+		if err != nil {
+			log.Error(err)
+		}
+	}
+
+	if len(bts) > 0 && scriptHasLogStorageLeft(name, len(bts)) {
+		log.Error(string(bts))
```
pasinskim: I would prefix this with some info which script is producing the logs:
`log.Errorf("stderr collected while executing script: %s [%s]", name, string(bts))`

oleorhagen:
@pasinskim more to your liking now?

statescript/executor.go
```diff
@@ -165,6 +167,19 @@ func execute(name string, timeout time.Duration) int {
...
 	}
 
+	bts, err := ioutil.ReadAll(stderr)
+	if err != nil {
+		log.Error(err)
+	}
+
+	if len(bts) > 0 {
+		if len(bts) > 10*1024*8 { // do not log more than 10KB at a time
+			log.Errorf("More than 10KB collected from stderr of %s, data will not be written to log file", name)
```
pasinskim: How about logging warning and truncating the message to 10kB, so that we are storing something always?

statescript/executor.go
```diff
@@ -165,6 +167,19 @@ func execute(name string, timeout time.Duration) int {
 		return retCode(err)
 	}
 
+	bts, err := ioutil.ReadAll(stderr)
+	if err != nil {
+		log.Error(err)
+	}
+
+	if len(bts) > 0 {
+		if len(bts) > 10*1024*8 { // do not log more than 10KB at a time
```
pasinskim: I think this should be `len(bts) > 10*1024`. `ioutil.ReadAll` is returning slice of byte (`[]byte`) which is `"byte is an alias for uint8 and is equivalent to uint8 in all ways. It is used, by convention, to distinguish byte values from 8-bit unsigned integer values."`.

statescript/executor.go
```diff
@@ -156,6 +156,8 @@ func execute(name string, timeout time.Duration) int {
 
 	cmd := exec.Command(name)
 
+	stderr, err := cmd.StderrPipe()
```
pasinskim: This could be confusing given my last comment, but I think your `if !strings.Contains(name, "Idle") && !strings.Contains(name, "Sync") {` was OK here. I would only change it to `if !strings.HasPrefix(name, "Idle") && !strings.HasPrefix(name, "Sync") {` for safety and performance reasons.

oleorhagen:
@pasinskim

statescript/executor.go
```diff
@@ -165,6 +176,28 @@ func execute(name string, timeout time.Duration) int {
...
+		bts, err = ioutil.ReadAll(stderr)
+		if err != nil {
+			log.Error(err)
+		}
+	}
+
+	if len(bts) > 0 {
+		if len(bts) > 10*1024 { // do not log more than 10KB at a time
+			buf := make([]byte, 10*1024)
+			n := copy(buf, bts)
```
pasinskim: Not sure why do we need a copy of the slice here. We can easily get rid of it and use "slice" of the original slice. See this for reference: https://blog.golang.org/go-slices-usage-and-internals.
You can simply do something like this below `log.Errorf("stderr collected while running script %s [%s] (Truncated to 10KB)", name, string(bts[:10240]))`. I would simply add new variable `maxSize := 10*1024` and then use it here. I think you could also try to limit the number in `Errorf` statement using `%.10240s`. In C you can also use something like `printf("%.*s", len, string_to_print)`, but not sure if this works in Go.

oleorhagen: I'm thinking @pasinskim that maybe logging even 10KB of what is output to stderr is a bit much? Maybe we should truncate it a little bit more? (1KB say?).
Slicing looks like a much cleaner way of doing this :)

oleorhagen:
@pasinskim as usual great suggestions :) I have now prettied if up, and we now limit the string logged in the format print, which works like a charm 👍

statescript/statescript_test.go
```diff
@@ -190,6 +191,25 @@ func TestExecutor(t *testing.T) {
...
+	res := execute(fileP.Name(), 100) // give the script plenty of time to run
+	assert.Equal(t, 0, res)
+	assert.Contains(t, buf.String(), "error data")
+
+	// write more than 10KB to stderr
+	fileP, err = createArtifactTestScript(tmpArt, "ArtifactInstall_Leave_11", "#!/bin/bash \nhead -c 81921 </dev/urandom >&2\n exit 1")
+	assert.NoError(t, err)
+	res = execute(fileP.Name(), 100)
+	assert.Equal(t, 1, res)
+	assert.Contains(t, buf.String(), "Truncated to 10KB")
```
pasinskim: How about testing if the size of the buff is indeed smaller than 10kB + couple of bytes for the error message itself?

PR #245, opened 2017-09-11
Update mender-artifact vendored dependency.

pasinskim:
Changelog: None

### DIFF
--- a/installer/installer_test.go
+++ b/installer/installer_test.go
@@ -110,8 +110,12 @@ func TestInstallWithScripts(t *testing.T) {
 
 type fDevice struct{}
 
-func (d *fDevice) InstallUpdate(io.ReadCloser, int64) error { return nil }
-func (d *fDevice) EnableUpdatedPartition() error            { return nil }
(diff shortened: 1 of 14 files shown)


kacf:
FYI, the SHA256 checksum `e3b0c44298fc1c149afbf4c8996fb92427ae41e4649b934ca495991b7852b855`, which is mentioned in the Travis log, is the sum for a null file, IOW 0 bytes, so probably the read is failing somewhere. Yes, I actually recognized it! :-D

mirzak:
> FYI, the SHA256 checksum e3b0c44298fc1c149afbf4c8996fb92427ae41e4649b934ca495991b7852b855, which is mentioned in the Travis log, is the sum for a null file, IOW 0 bytes, so probably the read is failing somewhere. Yes, I actually recognized it! :-D

You are a wizard! :)

pasinskim:
Thanks "the wizard" :1st_place_medal: . This helped to tackle the issue a lot actually. @kacf maybe you should have the new nickname: "OZ of code" it will be.

PR #246, opened 2017-09-14
Improve error messages for state scripts errors.

kacf:
Rely on the full error description instead of just the error code.

Changelog: Commit

### DIFF
--- a/statescript/statescript_test.go
+++ b/statescript/statescript_test.go
@@ -199,8 +199,8 @@ func TestExecutor(t *testing.T) {
 	log.SetOutput(&buf)
 	fileP, err := createArtifactTestScript(tmpArt, "ArtifactInstall_Leave_00", "#!/bin/bash \necho 'error data' >&2")
 	assert.NoError(t, err)
-	res := execute(fileP.Name(), 100) // give the script plenty of time to run
-	assert.Equal(t, 0, res)
+	err = execute(fileP.Name(), 100) // give the script plenty of time to run
+	assert.NoError(t, err)
 	assert.Contains(t, buf.String(), "error data")
 
 	buf.Reset()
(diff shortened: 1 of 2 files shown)


statescript/statescript_test.go
```diff
@@ -199,17 +199,17 @@ func TestExecutor(t *testing.T) {
 	log.SetOutput(&buf)
 	fileP, err := createArtifactTestScript(tmpArt, "ArtifactInstall_Leave_00", "#!/bin/bash \necho 'error data' >&2")
 	assert.NoError(t, err)
-	res := execute(fileP.Name(), 100) // give the script plenty of time to run
-	assert.Equal(t, 0, res)
+	err = execute(fileP.Name(), 100) // give the script plenty of time to run
+	assert.Equal(t, nil, err)
```
bboozzoo: There's `assert.NoError(t, err)` which is IMO clearer about the intent.

statescript/statescript_test.go
```diff
@@ -199,17 +199,17 @@ func TestExecutor(t *testing.T) {
...
 
 	buf.Reset()
 
 	// write more than 10KB to stderr
 	fileP, err = createArtifactTestScript(tmpArt, "ArtifactInstall_Leave_11", "#!/bin/bash \nhead -c 89999 </dev/urandom >&2\n exit 1")
 	assert.NoError(t, err)
-	res = execute(fileP.Name(), 100)
-	assert.Equal(t, 1, res)
+	err = execute(fileP.Name(), 100)
+	assert.NotEqual(t, nil, err)
```
bboozzoo: You can try `assert.EqualError(t, err, <message>)`, if I'm not mistaken, in this scenario the message will be `"exit status 1"`

kacf:
@bboozzoo: Thanks for review, fixed both!

> BTW. this line: cf55da1#diff-dae769a7c116a700ec3aa7606cb4efc0R164 shouldn't there be a return err right after the log?

Yes, I believe you're right! It's quite unrelated though, I will submit this separately.

kacf:
> I will submit this separately.

Err, no I won't, because the return type has changed! Update coming! :-)

kacf:
@bboozzoo: Updated, now with error return for stderr pipe failure as well!

PR #251, opened 2017-09-21
Make sure to switch the deployment logs on and off.

pasinskim:
In some states deployment logger was not switched on and off
correctly. This one makes sure we are logging only the data
while the device update is in progress.

Changelog: None

### DIFF
--- a/state.go
+++ b/state.go
@@ -321,6 +321,9 @@ type IdleState struct {
 }
 
 func (i *IdleState) Handle(ctx *StateContext, c Controller) (State, bool) {
+	// stop deployment logging
+	DeploymentLogger.Disable()
+
 	// check if client is authorized
 	if c.IsAuthorized() {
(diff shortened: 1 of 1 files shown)


state.go
```diff
@@ -828,6 +839,9 @@ func NewErrorState(err menderError) State {
 }
 
 func (e *ErrorState) Handle(ctx *StateContext, c Controller) (State, bool) {
+	// stop deployment logging
+	DeploymentLogger.Disable()
```
kacf: Should we disable the logger here? Won't that hide error messages from the deployments? Or are all those handled in `ReportErrorState`?

pasinskim: So there are two different error states: `ErrorState` and `UpdateErrorState`. For the later we are logging data.

kacf: Ah I see.

PR #252, opened 2017-09-22
set return code = 2, when there is nothing to commit

mirzak:
Also demote the print to warning.

Changelog: Title

Signed-off-by: Mirza Krak <mirza.krak@endian.se>


kacf:
Thanks!

PR #253, opened 2017-09-22
mender-device-identity: only collect MAC from ARPHRD_ETHER types 

mirzak:
The script only skipped "lo" interfaces while looking for the
lowest ifindex number. There are a lot more "virtual" interfaces [1]
beside "lo" and we could pick up an MAC address from something
that does not actually have a MAC address since ifindex number is
set based on probe order basicly.

This is of course a example script only and should probably not handle
every configuration possible out there but it should at least handle
the most common one well.

Example of picking up a strange address from a "odd" device:
./mender-device-identity
using interface /sys/class/net/sit0
mac=00:00:00:00

[1]. http://elixir.free-electrons.com/linux/latest/source/include/uapi/linux/if_arp.h

Changelog: Title

Signed-off-by: Mirza Krak <mirza.krak@endian.se>

### DIFF
--- a/support/mender-device-identity
+++ b/support/mender-device-identity
@@ -13,9 +13,10 @@
 # mac=de:ad:ca:fe:00:01
 # cpuid=1112233
 #
-# The example script collects the MAC address of a network interface
-# with the lowest ifindex, other than the loopback device 'lo'. The
-# identity data is output in the following format:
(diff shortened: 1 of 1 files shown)


kacf:
+1 for the fix itself, but this is slightly high risk, due to it possibly changing the identity data of a device. So I'll keep this in the master branch only.

PR #254, opened 2017-10-05
MEN-1511: Implement download resume feature.

kacf:
This implements the ability of the client to resume download if the
connection is broken for any reason. It does this by replacing the
socket reader with a custom reader object that keeps track of the read
offset, and if it is broken at any point, the download is
reestablished at the same point and reading continues.

This requires some custom parsing of the 'Content-Range' header, since
Go does not have this support natively on the client side. We
primarily need the offset that the server gives us back, with some
surrounding code to check for sanity of the returned header. All
according to HTTP spec.

One weakness of this method is that it is not possible to abort the
update and have the client notice this in the middle of the download,
even if the connection is broken. But this would require changing
state to check for deployment status, which would be quite hard to do
while an update is in progress mid-stream. Not impossible, but a lot
of work, and it was deemed out of scope for this initial feature. It's
also no worse than it was before, since you cannot currently abort an
update mid-download either.

Changelog: Fix 'unexpected EOF' error when downloading large updates.

Changelog: Implement ability for client to resume a download from
where it left off if the connection is broken.

### DIFF
--- a/client/client.go
+++ b/client/client.go
@@ -220,3 +220,38 @@ func buildApiURL(server, url string) string {
 	}
 	return buildURL(server) + apiPrefix + url
 }
+
+// Normally one minute, but used in tests to lower the interval to avoid
+// waiting.
+var exponentialBackoffSmallestUnit time.Duration = time.Minute
+
+// Simple algorithm: Start with one minute, and try three times, then double
+// interval (maxInterval is maximum) and try again. Repeat until we tried
+// three times with maxInterval.
+func GetExponentialBackoffTime(tried int, maxInterval time.Duration) (time.Duration, error) {
+	const perIntervalAttempts = 3
+
+	interval := 1 * exponentialBackoffSmallestUnit
+	nextInterval := interval
+
+	for c := 0; c <= tried; c += perIntervalAttempts {
+		interval = nextInterval
+		nextInterval *= 2
+		if interval >= maxInterval {
+			if tried-c >= perIntervalAttempts {
+				// At max interval and already tried three
+				// times. Give up.
+				return 0, errors.New("Tried maximum amount of times")
+			}
+
+			// Don't use less than the smallest unit, usually one
+			// minute.
+			if maxInterval < exponentialBackoffSmallestUnit {
+				return exponentialBackoffSmallestUnit, nil
+			}
+			return maxInterval, nil
+		}
+	}
+
+	return interval, nil
+}
--- a/client/update_resumer.go
+++ b/client/update_resumer.go
@@ -1,0 +1,175 @@
+// Copyright 2017 Northern.tech AS
+//
+//    Licensed under the Apache License, Version 2.0 (the "License");
+//    you may not use this file except in compliance with the License.
+//    You may obtain a copy of the License at
+//
+//        http://www.apache.org/licenses/LICENSE-2.0
+//
+//    Unless required by applicable law or agreed to in writing, software
+//    distributed under the License is distributed on an "AS IS" BASIS,
+//    WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+//    See the License for the specific language governing permissions and
+//    limitations under the License.
+
+package client
+
+import (
+	"fmt"
+	"github.com/mendersoftware/log"
+	"github.com/pkg/errors"
+	"io"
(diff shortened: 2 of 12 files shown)


kacf:
FYI, the flash failure is unrelated, some problem with pyro.

client/client.go
```diff
@@ -220,3 +220,36 @@ func buildApiURL(server, url string) string {
...
+		interval = nextInterval
+		nextInterval *= 2
+		if interval >= maxInterval {
+			if tried-c >= perIntervalAttempts {
+				// At max interval and already tried three
+				// times. Give up.
+				return 0, errors.New("Tried maximum amount of times")
+			}
+
+			// Don't use less than one minute.
```
pasinskim: Less than `exponentialBackoffSmallestUnit`. It is one minute now, but might be changed later, right.

client/client.go
```diff
@@ -220,3 +220,36 @@ func buildApiURL(server, url string) string {
...
+			if maxInterval < 1*exponentialBackoffSmallestUnit {
```
pasinskim: Do we need `1* ...` here?

client/update_resumer.go
```diff
@@ -0,0 +1,175 @@
...
+	"io/ioutil"
+	"net/http"
+	"strconv"
+	"strings"
+	"time"
+)
+
+type UpdateResumer struct {
+	stream        io.ReadCloser
+	api           ApiRequester
```
pasinskim: I would change the name to requester or similar. API is too broad IMHO.

client/update_resumer.go
```diff
@@ -0,0 +1,175 @@
...
+	"github.com/pkg/errors"
+	"io"
+	"io/ioutil"
+	"net/http"
+	"strconv"
+	"strings"
+	"time"
+)
+
+type UpdateResumer struct {
```
pasinskim: Just a note here. Given the naming convention I would assume `UpdateResumer` is an interface.

client/client.go
```diff
@@ -220,3 +220,36 @@ func buildApiURL(server, url string) string {
...
+	const perIntervalAttempts = 3
+
+	interval := 1 * exponentialBackoffSmallestUnit
+	nextInterval := interval
+
+	for c := 0; c <= tried; c += perIntervalAttempts {
+		interval = nextInterval
+		nextInterval *= 2
+		if interval >= maxInterval {
+			if tried-c >= perIntervalAttempts {
```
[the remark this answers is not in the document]
kacf: Indeed, The Go formatter is not very good IMHO, it often makes things less readable.

client/client.go
```diff
@@ -220,3 +220,36 @@ func buildApiURL(server, url string) string {
...
+				// At max interval and already tried three
+				// times. Give up.
+				return 0, errors.New("Tried maximum amount of times")
```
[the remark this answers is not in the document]
kacf: I would prefer to keep this message quite generic, because this function has potential uses outside of downloading as well. It's already used when initiating the download, not just in the middle. Since the error tends to be wrapped it's usually clear from the context what the attempts are related to.

The interval is already printed where we call this function from.

I should perhaps have mentioned that this code is preexisting, I only moved it, changed the name and added the smallest unit variable.

client/client.go
```diff
@@ -220,3 +220,36 @@ func buildApiURL(server, url string) string {
 	}
 	return buildURL(server) + apiPrefix + url
 }
+
+// Used in tests.
```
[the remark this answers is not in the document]
kacf: It's perhaps more correct to say that it's only changed in tests. It's just to avoid having to wait a whole minute. I can clarify the comment.

client/client.go
kacf: Right, I'll clarify this comment as well.

client/client.go
kacf: No. :-) I'll remove it.

client/update_resumer.go
kacf: What is the naming convention for structs? I think "UpdateResumer" is pretty clear about what it does.

client/update_resumer.go
kacf: Yep, agree. Will change!

kacf:
Thanks for the review, I'll fix the issues!

PR #255, opened 2017-11-07
MEN-1187: Support recovering from spontaneous reboots 

oleorhagen:
Here is the initial draft.
Mind you, this is gonna break a lot of acceptance-tests, and certainly the integration-tests.
However, it's main talking points are as follows:
* Stores state-data in all transitions, and recovers from a powerloss along the critical-path of and update. e.g. updatecheck -> fetch -> store -> install -> reboot.
* currently it also stores data in idle and authorize. Although I do not think that this is necessary in the final implementation, it has been handy for testing in qemu, without an update present.
* the install flip the u-boot flag has been moved into reboot now, and is the last thing that is done before the device actually reboots, in order to support recovering and re-running artifact-install_leave and reboot_enter scripts.

Never mind the old commits, they will be cleaned up prior to a merge.

### DIFF
--- a/state.go
+++ b/state.go
@@ -145,9 +145,9 @@ type StateRunner interface {
 	// Set runner's state to 's'
 	SetNextState(s State)
 	// Obtain runner's state
-	GetCurrentState(s store.Store) (State, State, TransitionStatus)
+	GetCurrentState(sctx *StateContext) (State, State, TransitionStatus)
 	// Run the currently set state with this context
-	TransitionState(from, to State, ctx *StateContext, status TransitionStatus) (State, State, bool)
+	TransitionState(from, to State, ctx *StateContext, tStatus TransitionStatus) (State, State, bool)
 }
 
 // StateData is state information that can be used for restoring state from storage
@@ -155,17 +155,15 @@ type StateData struct {
 	// version is providing information about the format of the data
 	Version int
 	// number representing the id of the last state to execute
-	Name      MenderState
-	FromState MenderState
-	ToState   MenderState
+	Name MenderState
 	// update reponse data for the update that was in progress
 	UpdateInfo client.UpdateResponse
 	// update status
 	UpdateStatus string
-	// TransitionStatus
+	// Client states
+	FromState        MenderState
+	ToState          MenderState
 	TransitionStatus TransitionStatus
-	// Error
-	Err menderError
 }
 
 const (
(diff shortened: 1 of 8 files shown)


state.go
```diff
@@ -356,10 +379,54 @@ func (i *InitState) Handle(ctx *StateContext, c Controller) (State, bool) {
 
 	log.Infof("handling loaded state: %s", sd.Name)
 
-	// chack last known state
+	i.SetTransition(sd.LeaveTransition) // Handle an unfinished leave transition
+
+	// check last known state
 	switch sd.Name {
```
pasinskim: Should we have some default here as well? I am thinking now, that maybe we will end up with the situation that after reboot (running with the new update) we will have some state that was not supported (by previous client).

state.go
```diff
@@ -611,23 +708,13 @@ func NewUpdateStoreState(in io.ReadCloser, size int64, update client.UpdateRespo
 func (u *UpdateStoreState) Handle(ctx *StateContext, c Controller) (State, bool) {
 
 	// make sure to close the stream with image data
-	defer u.imagein.Close()
+	defer func() { u.imagein.Close() }()
```
pasinskim: Why this is needed?

state.go
```diff
@@ -689,6 +779,10 @@ type FetchStoreRetryState struct {
 	err    error
 }
 
+func (fir *FetchStoreRetryState) SaveRecoveryData(lt Transition, s store.Store) {
+	// TODO - do we want to recover from this?
```
pasinskim: `TODO`?

state.go
```diff
@@ -794,6 +898,14 @@ type ErrorState struct {
 	cause menderError
 }
 
+func (e *ErrorState) SaveRecoveryData(lt Transition, s store.Store) {
+	UpdateStateData(s, StateData{
+		Name:            e.Id(),
+		LeaveTransition: lt,
+		// TODO - also needs to store menderError
```
pasinskim: `TODO`?

store/store.go
```diff
@@ -42,3 +42,13 @@ type Store interface {
 	// close the store
 	Close() error
 }
+
+type WriteAller interface {
```
pasinskim: I would rename to `AllReader` and `AllWriter`.

state.go
```diff
@@ -226,6 +235,9 @@ type State interface {
 	// Return transition
 	Transition() Transition
 	SetTransition(t Transition)
+	// SaveRecoveryData stores the data needed to recreate the state
+	// in case of powerloss
+	SaveRecoveryData(lt Transition, s store.Store)
```
pasinskim: I am wondering here. Now all the states need to implement `SaveRecoveryData`. Wouldn't it be easier to store the data for all of them by default and ignore the ones we don't need? One way could be to add an extra field to `baseState` (`storeRecoveryData bool`) or similar.

oleorhagen:
superseeded by: https://github.com/mendersoftware/mender/pull/266

PR #258, opened 2017-11-27
Report state script status to the server.

pasinskim:


mender.go
```diff
@@ -574,6 +607,28 @@ func TransitionError(s State, action string) State {
...
 }
 
+func shouldReportUpdateStatus(state MenderState) bool {
+	switch state {
+	case MenderStateUpdateInstall, MenderStateReboot,
+		MenderStateUpdateVerify, MenderStateUpdateCommit,
+		MenderStateAfterReboot, MenderStateRollback,
+		MenderStateRollbackReboot, MenderStateAfterRollbackReboot,
+		MenderStateUpdateError:
+		return true
```
kacf: Does this mirror the contents of the `stateStatus` map? Can you iterate over that one instead and return true if it matches? The reason I'm asking is that this is the kind of thing that could easily get out of sync.

statescript/executor.go
```diff
@@ -309,7 +322,22 @@ func (l Launcher) ExecuteAll(state, action string, ignoreError bool) error {
...
+			}
+		}
+		defer func() {
+			if report != nil {
+				if err = reportScriptStatus(report,
+					fmt.Sprintf("finished executing script: %s", s.Name())); err != nil {
+					log.Errorf("statescript: can not send finished status to server: %s", err.Error())
+				}
+			}
+		}()
```
kacf: You just tested for `report != nil`. I suggest putting the `defer func` block inside the previous if block, and omitting the second `if report != nil` test entirely.

mender.go
```diff
@@ -585,22 +640,46 @@ func (m *mender) TransitionState(to State, ctx *StateContext) (State, bool) {
...
+				API: m.api.Request(m.authToken),
+				URL: m.config.ServerURL,
+				Report: client.StatusReport{
+					DeploymentID: upd.ID,
+					Status:       to.Id().Status(),
+				},
+			}
+		}
+		return nil
+	}()
```
kacf: This construct is a bit confusing. If I read it correctly you are both creating and calling the function in the same statement. Can you simply avoid the function altogether and assign directly to `result` without executing `return`?

pasinskim:
So, we are reporting besides of what we've already had (basic status) `SubState` which is some extra information complementing the status one. In case of the state scripts we are sending the name of the script being executed (both when execution starts and when we are done). See https://github.com/mendersoftware/mender/pull/258/files#diff-dae769a7c116a700ec3aa7606cb4efc0R328 and https://github.com/mendersoftware/mender/pull/258/files#diff-dae769a7c116a700ec3aa7606cb4efc0R334

mender.go
pasinskim: Right. Good suggestion. I will change this one.

mender.go
pasinskim: Maybe I was trying to bee too smart here. Fixed!

statescript/executor.go
pasinskim: Thanks! Fixed.

kacf:
Aha! This was the critical piece I was missing, thanks for explaining.

PR #261, opened 2017-12-19
Add more logs about failed authorization request.

pasinskim:
This one is fixing both MEN-1660 and MEN-1661.


client/client.go
```diff
@@ -173,8 +178,6 @@ type Config struct {
 
 func loadServerTrust(conf Config) (*x509.CertPool, error) {
 	if conf.ServerCert == "" {
-		// TODO: this is for pre-production version only to simplify tests.
-		// Make sure to remove in production version.
 		log.Warn("Server certificate not provided. Trusting all servers.")
 		return nil, nil
```
kacf: What is our intention with `loadServerTrust`? My understanding is that the majority of hosted Mender users will likely want to run without a custom provided certificate, so it should revert to the system certificate pool in this case. Not revert to trusting everything, as seems to be done here. This might be a security problem.

What's your view on this?

client/client.go
```diff
@@ -145,6 +146,10 @@ func newHttpsClient(conf Config) (*http.Client, error) {
 
 	trustedcerts, err := loadServerTrust(conf)
 	if err != nil {
+		log.Error("Can not open certificate file.")
+		log.Error("Your /etc/mender/server.crt can not be opened, if you are using " +
```
kacf: You don't know if that's the case. See also the first comment in `loadServerTrust`.

I would remove this whole logging section and just keep returning the error, since you already get quite a lot of information from inside `loadServerTrust`, including more information about *why* there was an error. If you want to include the "make sure to include API gateway .... " message, then perhaps it can be moved there instead.

client/client.go
```diff
@@ -187,11 +190,24 @@ func loadServerTrust(conf Config) (*x509.CertPool, error) {
 	// Read certificate file.
 	servcert, err := ioutil.ReadFile(conf.ServerCert)
 	if err != nil {
+		log.Error("/etc/mender/server.crt is inaccessible.")
```
kacf: `log.Errorf("%s is inaccessible: %s", conf.ServerCert, err.Error())`

client/client.go
```diff
@@ -187,11 +190,24 @@ func loadServerTrust(conf Config) (*x509.CertPool, error) {
...
 		return nil, err
 	}
 
 	if len(servcert) == 0 {
-		return nil, errors.New("unable to find system and server certificates")
+		log.Error("/etc/mender/server.crt is empty.")
```
kacf: `log.Errorf("Both %s and the system certificate pool are empty.", conf.ServerCert)`

client/client.go
```diff
@@ -187,11 +190,24 @@ func loadServerTrust(conf Config) (*x509.CertPool, error) {
...
+		return nil, errors.New("server certificate is empty")
+	}
+
+	block, _ := pem.Decode([]byte(servcert))
+	if block != nil {
+		cert, err := x509.ParseCertificate(block.Bytes)
+		if err == nil {
+			log.Info("Have correct server certificate.")
```
kacf: What do you mean with "Have correct server certificate"? Do we know that for sure? Maybe drop this and print only the other two log messages.

client/client_auth.go
```diff
@@ -49,7 +52,42 @@ func (u *AuthClient) Request(api ApiRequester, server string, dataSrc AuthDataMe
...
+				default:
+					log.Errorf("Server certificate is invalid, reason: %#v", certErr.Reason)
+				}
+				return nil, errors.Wrapf(err, "certificate exists, but is invalid")
+			default:
+				log.Errorf("authorization request error: %v", certErr)
+			}
+		}
+		return nil, errors.Wrapf(err,
+			"generic error occured while executing authorization request")
```
kacf: :+1: on the whole block!

client/client.go
pasinskim: Right. Fixing!

client/client.go
pasinskim: This is definitely something we should fix, and you are right. We shouldn't trust everything in this case. I will make a separate ticket for fixing this one and will leave the `TODO` as it was.

client/client.go
pasinskim: Fixed!

client/client.go
pasinskim: Fixed!

client/client.go
pasinskim: I can drop this, but what I meant was that the certificate is there and is formatted correctly, since we can parse it. I will make the message more explicit.

pasinskim:
@kacf should be OK now. Review issues are fixed!

client/client.go
kacf: I think it can very easily be misunderstood to mean: "The certificate matches the one from the server", but this is wrong. All we know is that it parsed correctly. Maybe instead write "certificate correctly parsed"?

PR #265, opened 2018-01-23
mender-device-identity: Check if file exists before reading

nandra:
Mender on oragepi fails to run because identty script exit with error like:
/usr/share/mender/identity/mender-device-identity
cat: can't open '/sys/class/net/bonding_masters/type': Not a directory

Add check before reading type to avoid problems.

Signed-off-by: Marek Belisko <marek.belisko@open-nandra.com>


kacf:
Thanks @nandra, merged in #268.

PR #266, opened 2018-01-25
MEN-1187: Spontanaeous-reboot hardening of the client

oleorhagen:
Changelog: Title

Now the client, if losing power, will fall back to the original partition,
and run the Artifact-failure scripts, before going back to idle

### DIFF
--- a/state.go
+++ b/state.go
@@ -324,6 +324,9 @@ func (i *IdleState) Handle(ctx *StateContext, c Controller) (State, bool) {
 	// stop deployment logging
 	DeploymentLogger.Disable()
 
+	// cleanup state-data if any data is still present after an update
+	RemoveStateData(ctx.store)
+
 	// check if client is authorized
 	if c.IsAuthorized() {
 		return checkWaitState, false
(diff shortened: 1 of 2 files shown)


oleorhagen:
Welcome back! @pasinskim

mender.go
```diff
@@ -672,6 +672,12 @@ func (m *mender) TransitionState(to State, ctx *StateContext) (State, bool) {
 
 		m.SetNextState(to)
 
+		// Special case cleanup of stored state-data
```
pasinskim: Can you add some comment why do we need this.

state.go
```diff
@@ -356,24 +356,52 @@ func (i *InitState) Handle(ctx *StateContext, c Controller) (State, bool) {
...
 
 		return NewUpdateErrorState(me, sd.UpdateInfo), false
 	}
 }
 
+func committedPartition(c Controller) bool {
+
+	ua, err := c.HasUpgrade()
+	log.Infof("in committedPartition: ua: %t, err: %v", ua, err)
+	if err != nil {
```
pasinskim: Maybe we can print some error message here as this is pretty important and will impact state transition.

state.go
```diff
@@ -356,24 +356,52 @@ func (i *InitState) Handle(ctx *StateContext, c Controller) (State, bool) {
 
 	log.Infof("handling loaded state: %s", sd.Name)
 
-	// chack last known state
-	switch sd.Name {
-	// update process was finished; check what is the status of update
-	case MenderStateReboot:
+	if !committedPartition(c) {
+		// only valid entrypoint into the uncommitted partition is reboot_leave
```
pasinskim: Is `commit_enter` also a possibility here? What I am thinking is we've entered `commit_enter` state and then the reboot happened. Will we end up here as well?

state.go
```diff
@@ -356,24 +356,52 @@ func (i *InitState) Handle(ctx *StateContext, c Controller) (State, bool) {
...
+
+	// check last known state
+	switch sd.Name {
 
 	case MenderStateRollbackReboot:
 		return NewAfterRollbackRebootState(sd.UpdateInfo), false
 
-	// this should not happen
+	// Rerun commit-leave
+	case MenderStateUpdateCommit:
```
pasinskim: Do you want to force the update_commit -> init transition here? Why can't we simply skip setting the update_commit state?

state.go
```diff
@@ -922,16 +959,19 @@ func (usr *UpdateStatusReportState) Handle(ctx *StateContext, c Controller) (Sta
...
-		Name:         usr.Id(),
-		UpdateInfo:   usr.Update(),
-		UpdateStatus: usr.status,
-	}); err != nil {
-		log.Errorf("failed to store state data in update status report state: %v",
-			err)
-		return NewReportErrorState(usr.Update(), usr.status), false
+	// Do not store this if artifact-commit scripts are run when leaving the state
+	// as then the scripts will not be rerun
+	if usr.Transition() != ToArtifactCommit {
```
pasinskim: So why we are storing the state now only if we are not entering here from commit state? 
Next, the fact that we need to use `Transition()` here means we need to really think about refactoring this whole state machine concept as it is becoming quite messy.

pasinskim:
@oleorhagen maybe we can rerun integration tests as some state scripts tests are failing here. Or maybe this is expected without some modifications in the tests itself.

PR #267, opened 2018-01-28
Allow multiple digit partition numbers. 

drewmoseley:
Signed-off-by: Drew Moseley <drew.moseley@northern.tech>

### DIFF
--- a/device.go
+++ b/device.go
@@ -14,9 +14,11 @@
 package main
 
 import (
+	"fmt"
 	"io"
 	"path/filepath"
 	"strconv"
+	"strings"
 	"syscall"
 
 	"github.com/mendersoftware/log"
@@ -57,13 +59,13 @@ func (d *device) Reboot() error {
 
 func (d *device) SwapPartitions() error {
 	// first get inactive partition
-	inactivePartition, err := d.getInactivePartition()
+	inactivePartition, inactivePartitionHex, err := d.getInactivePartition()
 	if err != nil {
 		return err
 	}
 	log.Infof("setting partition for rollback: %s", inactivePartition)
 
-	err = d.WriteEnv(BootVars{"mender_boot_part": inactivePartition, "upgrade_available": "0"})
+	err = d.WriteEnv(BootVars{"mender_boot_part": inactivePartition, "mender_boot_part_hex": inactivePartitionHex, "upgrade_available": "0"})
 	if err != nil {
 		return err
 	}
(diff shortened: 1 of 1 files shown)


drewmoseley:
This corresponds to changes in https://github.com/mendersoftware/meta-mender/pull/469

device.go
```diff
@@ -137,32 +139,40 @@ func (d *device) InstallUpdate(image io.ReadCloser, size int64) error {
...
-		return "", errors.New("Error obtaining inactive partition: " + err.Error())
+		return "", "", errors.New("Error obtaining inactive partition: " + err.Error())
 	}
 
 	log.Debugf("Marking inactive partition (%s) as the new boot candidate.", inactivePartition)
 
-	partitionNumber := inactivePartition[len(inactivePartition)-1:]
-	if _, err := strconv.Atoi(partitionNumber); err != nil {
-		return "", errors.New("Invalid inactive partition: " + inactivePartition)
+	partitionSlice := strings.SplitAfter(inactivePartition, "p")
```
kacf: Repeating what is in https://github.com/mendersoftware/meta-mender/pull/469:

This won't work for `/dev/sda2` type of partitions, which is likely to be common on larger boards, such as Intel. I think you will have to count digits backwards from the end of the string.

Suggestion (untested):

```
partitionNumberDecStr := inactivePartition[len(strings.TrimRight(inactivePartition, "0123456789")):]
```

which gives you the final digits.

kacf:
Again, I don't know what's going on with the tests. Did a Fabric update suddenly break SSH connectivity? I'll be looking into this, currently *all* our tests are failing.

kacf:
I think it may have been fixed upstream. Rerunning.

Raspberry Pi is known to be broken ATM, so just ignore that one.

kacf:
I think there is definitely *something* wrong with this code, because your PR is the only one exhibiting these test problems. Maybe fix the issue I pointed out in my review first and we can take it from there.

drewmoseley:
@kacf I updated this and tested it as far as I was able.  I only have the one board that uses multiple digit partition numbers so it's possible there are other cases that may need some treatment but the Dragonboard I have on the Morty branch works.

drewmoseley:
> 	Error:		Max difference between 2018-02-06 00:10:14.063482976 +0000 UTC and 2018-02-06 00:10:14.059387126 +0000 UTC allowed is 1ms, but difference was 4.09585ms

Could it really be that this simple parsing change is taking 3+ ms?

kacf:
Sigh, the tests are such an unpredictable bunch. I have restarted it yet again.

kacf:
So this one looks like it could be legit, since it's all client failures: https://s3-eu-west-1.amazonaws.com/mender-testing-reports/integration-reports/1034/report.html

However, I will rerun once more just in case.

drewmoseley:
It seems like strange errors to have been caused by this particular change.  Any idea how that could happen?

kacf:
Oh wait, I think I know. Can you rebase your changes on top of the latest mender/master? I think this may be because you're missing Ole's reboot hardening which was merged last week.

drewmoseley:
Done.  Monitoring CI.  Fingers crossed.

kacf:
Alright, all current failures are expected due to the recent Jenkins difficulty. Merging.

drewmoseley:
@kacf what branches will this be in?  Will it get backported or only stay in master and then the next release?

kacf:
I think this can be cherry-picked. Pretty safe, and a bug fix. I opened these two:

* #273
* #274

PR #271, opened 2018-02-06
MEN-1738: log request-id in case of bad API requests

oleorhagen:
Changelog: Title

### DIFF
--- a/client/client.go
+++ b/client/client.go
@@ -16,8 +16,10 @@ package client
 import (
 	"crypto/tls"
 	"crypto/x509"
+	"encoding/json"
 	"encoding/pem"
 	"fmt"
+	"io"
 	"io/ioutil"
 	"net"
 	"net/http"
@@ -67,6 +69,45 @@ var (
 // configuration.
 type ApiRequester interface {
 	Do(req *http.Request) (*http.Response, error)
+}
+
+// APIError is an error type returned after receiving an error message from the
+// server. It wraps a regular error with the request_id - and if
+// the server returns an error message, this is also returned.
+type APIError struct {
+	error
+	reqID        string
+	serverErrMsg string
+}
+
+func NewAPIError(err error, resp *http.Response) *APIError {
+	a := APIError{
+		error: err,
+		reqID: resp.Header.Get("request_id"),
+	}
+
+	if resp.StatusCode >= 400 && resp.StatusCode < 600 {
+		a.serverErrMsg = unmarshalErrorMessage(resp.Body)
+	}
+	return &a
+}
+
+func (a *APIError) Error() string {
+
+	err := fmt.Sprintf("(request_id: %s): %s", a.reqID, a.error.Error())
+
+	if a.serverErrMsg != "" {
+		return err + fmt.Sprintf(" server error message: %s", a.serverErrMsg)
+	}
+
+	return err
+
+}
+
+// Cause returns the underlying error, as
+// an APIError is merely an error wrapper.
+func (a *APIError) Cause() error {
+	return a.error
 }
 
 type RequestProcessingFunc func(response *http.Response) (interface{}, error)
@@ -268,4 +309,16 @@ func GetExponentialBackoffTime(tried int, maxInterval time.Duration) (time.Durat
 	}
 
 	return interval, nil
+}
+
+// unmarshalErrorMessage unmarshals the error message contained in an
+// error request from the server.
+func unmarshalErrorMessage(r io.Reader) string {
+	e := new(struct {
+		Error string `json:"error"`
+	})
+	if err := json.NewDecoder(r).Decode(e); err != nil {
+		return fmt.Sprintf("failed to parse server response: %v", err)
+	}
+	return e.Error
 }
--- a/client/client_auth.go
+++ b/client/client_auth.go
@@ -1,4 +1,4 @@
-// Copyright 2017 Northern.tech AS
+// Copyright 2018 Northern.tech AS
 //
 //    Licensed under the Apache License, Version 2.0 (the "License");
 //    you may not use this file except in compliance with the License.
@@ -95,18 +95,18 @@ func (u *AuthClient) Request(api ApiRequester, server string, dataSrc AuthDataMe
 
 	switch rsp.StatusCode {
 	case http.StatusUnauthorized:
-		return nil, AuthErrorUnauthorized
+		return nil, NewAPIError(AuthErrorUnauthorized, rsp)
 	case http.StatusOK:
 		log.Debugf("receive response data")
 		data, err := ioutil.ReadAll(rsp.Body)
 		if err != nil {
-			return nil, errors.Wrapf(err, "failed to receive authorization response data")
+			return nil, NewAPIError(errors.Wrapf(err, "failed to receive authorization response data"), rsp)
 		}
 
-		log.Debugf("received response data %v", data)
+		log.Debugf("received response data:  %v", data)
 		return data, nil
 	default:
-		return nil, errors.Errorf("unexpected authorization status %v", rsp.StatusCode)
+		return nil, NewAPIError(errors.Errorf("unexpected authorization status %v", rsp.StatusCode), rsp)
 	}
 }
 
--- a/client/client_status.go
+++ b/client/client_status.go
@@ -1,4 +1,4 @@
-// Copyright 2017 Northern.tech AS
+// Copyright 2018 Northern.tech AS
 //
 //    Licensed under the Apache License, Version 2.0 (the "License");
 //    you may not use this file except in compliance with the License.
@@ -81,13 +81,13 @@ func (u *StatusClient) Report(api ApiRequester, url string, report StatusReport)
 	switch {
 	case r.StatusCode == http.StatusConflict:
 		log.Warnf("status report rejected, deployment aborted at the backend")
-		return ErrDeploymentAborted
+		return NewAPIError(ErrDeploymentAborted, r)
 	case r.StatusCode != http.StatusNoContent:
 		log.Errorf("got unexpected HTTP status when reporting status: %v", r.StatusCode)
-		return errors.Errorf("reporting status failed, bad status %v", r.StatusCode)
+		return NewAPIError(errors.Errorf("reporting status failed, bad status %v", r.StatusCode), r)
 	}
 
-	log.Debugf("status reported, response %v", r)
+	log.Debugf("status reported, response %s", r)
 
 	return nil
 }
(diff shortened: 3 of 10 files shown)


kacf:
I think the integration test failure is bogus, can be treated as passed for all practical purposes.

oleorhagen:
@pasinskim bump

client/client.go
```diff
@@ -97,7 +97,19 @@ func (ar *ApiRequest) Do(req *http.Request) (*http.Response, error) {
...
+	return logRequestIDWrapper(ar.api.Do(req))
+}
+
+// logRequestIDWrapper logs the request-id in case of an invalid response to the API
+func logRequestIDWrapper(r *http.Response, err error) (*http.Response, error) {
+	log.Info("data")
+	if err == nil {
+		rid := r.Header.Get("x-men-requestid")
+		if rid != "" && r.StatusCode != http.StatusOK {
+			log.Infof("requestid: %s ", rid)
```
pasinskim: Should we also extend this error message, to add a bit more context here?

client/client.go
```diff
@@ -97,7 +97,19 @@ func (ar *ApiRequest) Do(req *http.Request) (*http.Response, error) {
 	if req.Header.Get("Authorization") == "" {
 		req.Header.Set("Authorization", fmt.Sprintf("Bearer %s", ar.auth))
 	}
-	return ar.api.Do(req)
+	return logRequestIDWrapper(ar.api.Do(req))
+}
+
+// logRequestIDWrapper logs the request-id in case of an invalid response to the API
+func logRequestIDWrapper(r *http.Response, err error) (*http.Response, error) {
+	log.Info("data")
```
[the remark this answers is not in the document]
oleorhagen: yes, sry bout that

client/client.go
oleorhagen: yes, I was thinking the same. @maciejmrowiec, what do u want in the logs?

maciejmrowiec: we need context, do you guys use errors.Wrap in client code ?

oleorhagen: Yeh

client/client.go
```diff
@@ -97,7 +97,18 @@ func (ar *ApiRequest) Do(req *http.Request) (*http.Response, error) {
...
-	return ar.api.Do(req)
+	return logRequestIDWrapper(ar.api.Do(req))
+}
+
+// logRequestIDWrapper logs the request-id in case of an invalid response to the API
+func logRequestIDWrapper(r *http.Response, err error) (*http.Response, error) {
+	if err == nil {
+		rid := r.Header.Get("x-men-requestid")
+		if rid != "" && r.StatusCode != http.StatusOK {
+			log.Infof("requestid: %s, retCode: %d, request-url: %s", rid, r.StatusCode, r.Request.URL)
```
oleorhagen: @maciejmrowiec is this an appropriate amount of logging info?

maciejmrowiec: What about error messages from server? Are they logged somewhere?

Any error resp, from any of our API always contains this:
https://docs.mender.io/1.3/apis/device-apis/device-authentication#error

maciejmrowiec: http.StatusOK thats 200, everything in rage 200-399 are successful calls, 400-499 are client side errors, 500-599 server side errors

https://en.wikipedia.org/wiki/List_of_HTTP_status_codes

oleorhagen: it does not look like the server error messages are logged anywhere, but I figured u would use the request id to simply read that out of the server logs serverside, or do you want them to show up in the client logs as well?

and so you want only client and server side errors to log the request-id?

maciejmrowiec: In general yes, but it would good to have context what something was happening something like: `update 123: starting postinstall scripts: executing script 123.sh: reporting update progress status: request failed with message: internal server error, request_id 123 `

Then you have enough context when you are failing, what is failing, you can use request_id to get missing pieces of the puzzle from server side. 

Probably @pasinskim would be better to review this as I have very limited understanding of client logic / logging structure.

client/client.go
```diff
@@ -269,3 +271,16 @@ func GetExponentialBackoffTime(tried int, maxInterval time.Duration) (time.Durat
 
 	return interval, nil
 }
+
+// unMarshalErrorMessage unmarshals the error message contained in an
+// error request from the server.
+func unMarshalErrorMessage(r io.Reader) string {
```
pasinskim: Should be `unmarshalErrorMessage`.

client/client.go
```diff
@@ -269,3 +271,16 @@ func GetExponentialBackoffTime(tried int, maxInterval time.Duration) (time.Durat
...
+	e := new(struct {
+		Error string `json:"error"`
+		ReqID string `json:"request_id"`
+	})
+	if err := json.NewDecoder(r).Decode(e); err != nil {
+		return fmt.Sprintf("failed to parse server response: %v", err)
+	}
+	return fmt.Sprintf("request failed with error: %s : request_id: %s", e.Error, e.ReqID)
```
pasinskim: Maybe you can change the log message to something like `request (id: %s) error: %s` since this will be a part of the "failed to upload the client logs: error-code: %d : %s" message which is indicating the error anyway.

client/client_auth.go
```diff
@@ -93,6 +93,10 @@ func (u *AuthClient) Request(api ApiRequester, server string, dataSrc AuthDataMe
 
 	log.Debugf("got response: %v", rsp)
 
+	if rsp.StatusCode >= 400 && rsp.StatusCode < 600 {
```
pasinskim: The ticket says "Include request_id details in standard Mender logging", but here I can see we are logging only 4XX and 5XX errors. Are you sure this was the intention of the ticket?

oleorhagen: as far as I can understand https://docs.mender.io/1.3/apis/device-apis/device-authentication#error only these codes are errors, and thus contain an error message.

oleorhagen:
@pasinskim if you could have another quick look, that'd be great. I rewrote it around the whole wrap-error functionality instead, as wanted by majiej.

client/client.go
```diff
@@ -69,6 +71,44 @@ type ApiRequester interface {
 	Do(req *http.Request) (*http.Response, error)
 }
 
+// APIError wraps an error with the request_id - and if
```
pasinskim: `... and is also an error type returned after receiving an error message from the server`.

client/client.go
```diff
@@ -69,6 +71,44 @@ type ApiRequester interface {
...
+
+	if resp.StatusCode >= 400 && resp.StatusCode < 600 {
+		a.serverErrMsg = unmarshalErrorMessage(resp.Body)
+	}
+	return &a
+}
+
+func (a *APIError) Error() string {
+
+	err := fmt.Sprintf("(request_id: %s): %s", a.reqID, a.error.Error())
```
pasinskim: Can we move this inside `if a.serverErrMsg != ""` below? There is no need to evaluate it if there is no error as we are returning an empty string in this case anyway.

client/client.go
```diff
@@ -269,3 +309,16 @@ func GetExponentialBackoffTime(tried int, maxInterval time.Duration) (time.Durat
...
+// error request from the server.
+func unmarshalErrorMessage(r io.Reader) string {
+	e := new(struct {
+		Error string `json:"error"`
+		ReqID string `json:"request_id"`
+	})
+	if err := json.NewDecoder(r).Decode(e); err != nil {
+		return fmt.Sprintf("failed to parse server response: %v", err)
+	}
+	return fmt.Sprintf("request (id: %s) error: %s", e.ReqID, e.Error)
```
pasinskim: Does it mean we are printing sometimes `request_id` twice in one error message? One will be added here: https://github.com/mendersoftware/mender/pull/271/files#diff-1cd41a2611da49ce63630d5a51bd82c5R96 and one as the part of unmarshalling.

client/client_status.go
```diff
@@ -81,13 +81,13 @@ func (u *StatusClient) Report(api ApiRequester, url string, report StatusReport)
...
-		return ErrDeploymentAborted
+		return NewAPIError(ErrDeploymentAborted, r)
 	case r.StatusCode != http.StatusNoContent:
 		log.Errorf("got unexpected HTTP status when reporting status: %v", r.StatusCode)
-		return errors.Errorf("reporting status failed, bad status %v", r.StatusCode)
+		return NewAPIError(errors.Errorf("reporting status failed, bad status %v", r.StatusCode), r)
 	}
 
-	log.Debugf("status reported, response %v", r)
+	log.Debugf("(request_id: %s): status reported, response %s", r.Header.Get("request_id"), r)
```
pasinskim: Not sure if this will ever be the case, but do you know if we can end up with the empty `Header`? If so this might core dump if we are not checking `if Header != nil`.

client/client_update.go
```diff
@@ -78,7 +79,17 @@ func (u *UpdateClient) getUpdateInfo(api ApiRequester, process RequestProcessing
 
 	defer r.Body.Close()
 
+	respdata, err := ioutil.ReadAll(r.Body)
```
pasinskim: Why do we need whole this section now?

client/client_status.go
oleorhagen: according to @maciejmrowiec the header is a part of every request. But maybe we should remove this, just in case it changes sometime in the future?

client/client_update.go
oleorhagen: Because we can only read from the body once.

client/client.go
oleorhagen: ur right. nice find.

client/client.go
oleorhagen: agree

client/client.go
oleorhagen: changed it to return the original error: ``` 
func (a *APIError) Error() string {

	err := fmt.Sprintf("(request_id: %s): %s", a.reqID, a.error.Error())

	if a.serverErrMsg != "" {
		return err + fmt.Sprintf(" server error message: %s", a.serverErrMsg)
	}

	return err

}
```
Which it was supposed to all along.

oleorhagen: it is simply a server-message append.

client/client_update.go
oleorhagen: It is processed once in one of the process functions. And then again in NewApiError, if there is an error.

client/client_status.go
oleorhagen: I settled on removing them.

pasinskim:
Let's wait for the tests to pass.

GregorioDiStefano:
Integration tests passed (test marked as failure due to container removal post test timing out, this is due to an old version of docker being using on Google Cloud)

kacf:
Travis check still fails though. Probably it's the new license check that checks copyright year. Just make sure all the files that have changed are updated to 2018.

oleorhagen:
@maciejmrowiec should this be cherry picked anywhere?

maciejmrowiec:
no, it's a "new feature"

PR #273, opened 2018-02-22
1.3.x cherry-picks

kacf:


drewmoseley:
Nice.  I'll be able to drop some local patches for the Dragonboard with this.

PR #275, opened 2018-03-12
MEN-1806: print artifact name from the cli 

oleorhagen:
Changelog: Title

### DIFF
--- a/main.go
+++ b/main.go
@@ -14,11 +14,15 @@
 package main
 
 import (
+	"bufio"
+	"bytes"
 	"flag"
 	"fmt"
+	"io/ioutil"
 	"os"
 	"os/exec"
 	"path"
+	"path/filepath"
 	"runtime"
 	"strings"
 
@@ -48,6 +52,7 @@ type runOptionsType struct {
 	bootstrap       *bool
 	daemon          *bool
 	bootstrapForce  *bool
+	showArtifact    *bool
 	client.Config
 }
 
@@ -103,6 +108,8 @@ func argsParse(args []string) (runOptionsType, error) {
 
 	bootstrap := parsing.Bool("bootstrap", false, "Perform bootstrap and exit.")
 
+	showArtifact := parsing.Bool("show-artifact", false, "print the current artifact name to the command line and exit")
+
 	imageFile := parsing.String("rootfs", "",
 		"Root filesystem URI to use for update. Can be either a local "+
 			"file or a URL.")
@@ -135,6 +142,7 @@ func argsParse(args []string) (runOptionsType, error) {
 		bootstrap:       bootstrap,
 		daemon:          daemon,
 		bootstrapForce:  forcebootstrap,
+		showArtifact:    showArtifact,
 		Config: client.Config{
 			ServerCert: *serverCert,
 			NoVerify:   *skipVerify,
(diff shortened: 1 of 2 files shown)


estenberg:
That was fast. :)
Not sure about the option name, "current-artifact", though, it feels a bit nonstandard and cumbersome.
Maybe use some inspiration from other Unix tools or package managers?

How about -show-artifact?

@kacf @pasinskim any opinions?

main.go
```diff
@@ -271,6 +276,15 @@ func ShowVersion() {
 	os.Stdout.Write([]byte(v))
 }
 
+func PrintArtifactName() {
+	aname, err := ioutil.ReadFile("/etc/mender/artifact_info")
+	if err != nil {
+		fmt.Printf("failed to read artifact name: err: %v", err)
+		return
+	}
+	fmt.Println(aname)
```
kacf: This is the format of the file:

```
artifact_name=mender-image-1.0
```

It should only print the value, not the key. And you should assume that there may be other fields in there in the future.

main.go
```diff
@@ -103,6 +105,8 @@ func argsParse(args []string) (runOptionsType, error) {
 
 	bootstrap := parsing.Bool("bootstrap", false, "Perform bootstrap and exit.")
 
+	curArtifact := parsing.Bool("current-artifact", false, "print the current artifact name to the command line and exit")
```
kacf: Agree with Eystein, I think `show-artifact` is better than `current-artifact`.

oleorhagen:
fixed!

main.go
```diff
@@ -271,6 +278,25 @@ func ShowVersion() {
...
+	for s.Scan() {
+		if strings.HasPrefix(s.Text(), "artifact_name=") {
+			a := strings.Split(s.Text(), "=")
+			if len(a) == 2 {
+				fmt.Println(a[1])
+			} else {
+				fmt.Println("Wrong formatting of the artifact_info file")
+			}
+		}
+	}
```
kacf: There are a couple of corner cases here that I think there should produce an error:

* There is no `artifact_name=`
* There is more than one `artifact_name=`

Both can have severe consequences if they are wrong, so I think we should return error in these cases.

oleorhagen:
another fixup

kacf:
@pasinskim: Maybe we should raise the cyclomatic complexity requirement. `doMain` is not even two pages, and it still complains.

pasinskim:
@kacf I would rather not to. It forces us to keep the code simple and is set the same in all places.

kacf:
Ok @oleorhagen, you heard the man/ :-P

oleorhagen:
Do not merge yet

oleorhagen:
Now we should be good :)

main.go
```diff
@@ -391,8 +422,19 @@ func doMain(args []string) error {
...
+
+func handleCLIOptions(runOptions runOptionsType, env *uBootEnv, device *device, config *menderConfig) error {
+
 	switch {
 
+	case *runOptions.version:
+		ShowVersion()
+		return nil
+	case *runOptions.showArtifact:
+		return PrintArtifactName("/etc/mender/artifact_info")
```
kacf: Oh, can you use `getConfDirPath()` here instead of hardcoding `/etc/mender`? Sorry, I forgot about it earlier.

oleorhagen: sounds reasonable :) fixed!

GregorioDiStefano:
Your PR is based off tag: 1.3.0b1, hence why it keeps failing. 
Rebase your PR off master.

estenberg:
Is this done? Task still open: https://tracker.mender.io/browse/MEN-1806

PR #276, opened 2018-03-15
Add testify require as a dependency in the vendor-list and directory 

oleorhagen:
I feel like this is a useful addition. Any complaints?


oleorhagen:
@kacf done!

oleorhagen:
that would be surprising indeed

PR #277, opened 2018-03-26
MEN-1824: Abort upgrade if artifact name is not retrievable from arti…

oleorhagen:
…fact_info

Changelog: Title

### DIFF
--- a/mender.go
+++ b/mender.go
@@ -76,6 +76,8 @@ var (
 	defaultDataStore         = getStateDirPath()
 	defaultArtScriptsPath    = path.Join(getStateDirPath(), "scripts")
 	defaultRootfsScriptsPath = path.Join(getConfDirPath(), "scripts")
+
+	errNoArtifactName = errors.New("cannot determine current artifact name")
 )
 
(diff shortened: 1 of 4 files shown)


mender.go
```diff
@@ -449,8 +451,12 @@ func (m *mender) FetchUpdate(url string) (io.ReadCloser, int64, error) {
 // contains update information.
 func (m *mender) CheckUpdate() (*client.UpdateResponse, menderError) {
 	currentArtifactName, err := m.GetCurrentArtifactName()
-	if err != nil {
+	if err != nil || currentArtifactName == "" {
 		log.Error("could not get the current artifact name")
```
oleorhagen: The way it is written now, the client should never make it here if the artifact name is empty ( as this is handled in inventory refresh) , and thus the only error should be io. Do you wan't to panic instead of adding the 'artifact-name is empty' error message?
This also means that we maybe should treat only io-errors at this point. Thoughts?

kacf: I think we can leave it as you've done here. Always good to have an error path just in case.

PR #278, opened 2018-04-03
Regenerate keys on all key errors, not just when keys are missing.

kacf:
This is safer for autonomous nodes, and is especially important if the
client loses power while the key is being written, which results in an
unreadable key instead of a missing key.

Changelog: Title


oleorhagen:
Looks fine to me. Might still leave a log message on error though?

kacf:
Yes. I will fix that.

kacf:
Updated.

PR #283, opened 2018-04-20
FIXED: MEN-1854: HTTP error 401 is not handled by all states

alfrunes:
The responsecode for each http api-requests is explicitly checked and handles 401 by issuing for a new JWT token.

changelog: title

### DIFF
--- a/client/client.go
+++ b/client/client.go
@@ -117,11 +117,14 @@ type ApiClient struct {
 	http.Client
 }
 
-// Return a new ApiRequest sharing this ApiClient helper
-func (a *ApiClient) Request(code AuthToken) *ApiRequest {
+type ClientReauthorizeFunc func() (AuthToken, error)
+
+// Return a new ApiRequest
+func (a *ApiClient) Request(code AuthToken, req ClientReauthorizeFunc) *ApiRequest {
 	return &ApiRequest{
-		api:  a,
-		auth: code,
+		api:    a,
+		auth:   code,
+		revoke: req,
 	}
 }
 
@@ -131,14 +134,36 @@ func (a *ApiClient) Request(code AuthToken) *ApiRequest {
 type ApiRequest struct {
 	api *ApiClient
 	// authorization code to use for requests
-	auth AuthToken
+	auth   AuthToken
+	revoke ClientReauthorizeFunc
 }
 
 func (ar *ApiRequest) Do(req *http.Request) (*http.Response, error) {
 	if req.Header.Get("Authorization") == "" {
 		req.Header.Set("Authorization", fmt.Sprintf("Bearer %s", ar.auth))
 	}
-	return ar.api.Do(req)
+	r, err := ar.api.Do(req)
+	if r != nil && r.StatusCode == http.StatusUnauthorized {
+		// invalid JWT; most likely the token is expired:
+		// Try to refresh it and reattempt sending the request
+		log.Info("Device unauthorized; attempting reauthorization")
+		if jwt, e := ar.revoke(); e == nil {
+			// retry API request with new JWT token
+			ar.auth = jwt
+			// check if request had a body
+			// (GetBody is optional, and nil if body is empty)
+			if req.GetBody != nil {
+				if body, e := req.GetBody(); e == nil {
+					req.Body = body
+				}
+			}
+			req.Header.Set("Authorization", fmt.Sprintf("Bearer %s", ar.auth))
+			r, err = ar.api.Do(req)
+		} else {
+			log.Warnf("Reauthorization failed with error: %s", e.Error())
+		}
+	}
+	return r, err
 }
 
 func NewApiClient(conf Config) (*ApiClient, error) {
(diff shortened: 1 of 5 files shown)


client/client.go
```diff
@@ -118,10 +118,11 @@ type ApiClient struct {
 }
 
 // Return a new ApiRequest sharing this ApiClient helper
-func (a *ApiClient) Request(code AuthToken) *ApiRequest {
+func (a *ApiClient) Request(code AuthToken, req func() (AuthToken, error)) *ApiRequest {
```
pasinskim: Not a big issue, but we could have something like:
`type ClientReauthorizeFunc func() (AuthToken, error)
func (a *ApiClient) Request(code AuthToken, req ClientReauthorizeFunc) *ApiRequest `
This might be more readable as we are using this function signature in many places.

client/client.go
```diff
@@ -131,14 +132,34 @@ func (a *ApiClient) Request(code AuthToken) *ApiRequest {
 type ApiRequest struct {
 	api *ApiClient
 	// authorization code to use for requests
-	auth AuthToken
+	auth   AuthToken
+	revoke func() (AuthToken, error)
```
pasinskim: Considering comment above this will be `revoke ClientReauthorizeFunc`.

client/client.go
```diff
@@ -131,14 +132,34 @@ func (a *ApiClient) Request(code AuthToken) *ApiRequest {
...
 
 func (ar *ApiRequest) Do(req *http.Request) (*http.Response, error) {
 	if req.Header.Get("Authorization") == "" {
 		req.Header.Set("Authorization", fmt.Sprintf("Bearer %s", ar.auth))
 	}
-	return ar.api.Do(req)
+	r, err := ar.api.Do(req)
+	if r != nil && r.StatusCode == http.StatusUnauthorized {
+		// invalid JWT-token
+		log.Info("Device unauthorized; attempting reauthorization")
```
pasinskim: I would add some comment here explaining why we need to reauthorize the client.

client/client.go
```diff
@@ -131,14 +132,34 @@ func (a *ApiClient) Request(code AuthToken) *ApiRequest {
...
+				// if body present: retrieve body
+				if body, e := req.GetBody(); e == nil {
+					req.Body = body
+				}
+			}
+			// retry API request with new JWT token
+			req.Header.Set("Authorization", fmt.Sprintf("Bearer %s", ar.auth))
+			r, err = ar.api.Do(req)
+		} else {
+			log.Info("Reauthorization failed")
```
pasinskim: We could do `log.Infof("Reauthorization failed with error: %s", err.Error())` to get some more data why this error happened.

client/client.go
```diff
@@ -131,14 +132,34 @@ func (a *ApiClient) Request(code AuthToken) *ApiRequest {
...
 		req.Header.Set("Authorization", fmt.Sprintf("Bearer %s", ar.auth))
 	}
-	return ar.api.Do(req)
+	r, err := ar.api.Do(req)
+	if r != nil && r.StatusCode == http.StatusUnauthorized {
+		// invalid JWT-token
+		log.Info("Device unauthorized; attempting reauthorization")
+		if jwt, e := ar.revoke(); e == nil {
+			ar.auth = jwt
+			if req.GetBody != nil {
```
pasinskim: Won't this be set always since we are doing client requests only? Shouldn't we check for `req.Body` here and copy only if we have some body?

alfrunes: I think that's what tried first, but it turns out that the http.Do() call "wipes out" the body, and the only way to get it back is through the (optional) GetBody function in the Request structure.

alfrunes:
@pasinskim changes committed

client/client.go
```diff
@@ -131,14 +134,35 @@ func (a *ApiClient) Request(code AuthToken) *ApiRequest {
...
 
 func (ar *ApiRequest) Do(req *http.Request) (*http.Response, error) {
 	if req.Header.Get("Authorization") == "" {
 		req.Header.Set("Authorization", fmt.Sprintf("Bearer %s", ar.auth))
 	}
-	return ar.api.Do(req)
+	r, err := ar.api.Do(req)
+	if r != nil && r.StatusCode == http.StatusUnauthorized {
+		// invalid JWT-token; most likely expired an expired token
+		log.Info("Device unauthorized; attempting reauthorization")
```
pasinskim: `expired an expired token` does not sound good. Maybe you can add something `Most likely the token expired and we will try to refresh it and then retry sending the original request.`

alfrunes: oops..

alfrunes:
@pasinskim

pasinskim:
@alfrunes cannot see the comment fixed...

kacf:
Remember to also test it together with https://github.com/mendersoftware/integration/pull/478.

alfrunes:
@pasinskim; sorry, forgot to add the file before pushing

PR #285, opened 2018-05-01
Initial script for generating Mender client keys,

estenberg:
especially useful for preauthorizing devices.

Is this an appropriate location for this script? Note that it is not that valuable without https://tracker.mender.io/browse/MEN-1902

### DIFF
--- a/support/keygen-client
+++ b/support/keygen-client
@@ -1,8 +1,17 @@
 #!/bin/bash
 set -e
 
-FILE_NAME_PRIVATE_KEY="private.key"
-FILE_NAME_PUBLIC_KEY="public.key"
+if [[ $PREFIX_KEY = *"/"* ]]; then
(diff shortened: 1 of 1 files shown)


kacf:
Otherwise looks good, but should we think about some kind of indexing of keys? I assume this script may be used to generate lots of keys all at once. Maybe an argument or environment variable that can be used to separate/index keys?

Also, I think we should get the documentation for how to use it before merging this.

estenberg:
@kacf thanks for the review. I fixed both changelog message and added a prefix environment varialble.

Regarding getting docs first it is a bit of a chicken-and-egg problem because I need this for the docs at https://github.com/mendersoftware/mender-docs/pull/400

So could we merge this one and we follow up by merging the preauth docs after? Or did you think about some other place we should have this documented? I can not think of other use cases for this script than preauthorizing myself.

estenberg:
Docs are live: https://docs.mender.io/server-integration/preauthorizing-devices

PR #290, opened 2018-06-11
MEN-1880: log active partition before and after reboot.

oleorhagen:
Changelog: Title

### DIFF
--- a/main.go
+++ b/main.go
@@ -420,6 +420,12 @@ func doMain(args []string) error {
 
 	env := NewEnvironment(new(osCalls))
 	device := NewDevice(env, new(osCalls), config.GetDeviceConfig())
+	ap, err := device.GetActive()
+	if err != nil {
+		log.Errorf("Failed to read the current active partition: %s", err.Error())
+	} else {
+		log.Infof("Mender running on partition: %s", ap)
+	}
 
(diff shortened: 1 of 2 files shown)


main.go
```diff
@@ -420,6 +420,12 @@ func doMain(args []string) error {
 
 	env := NewEnvironment(new(osCalls))
 	device := NewDevice(env, new(osCalls), config.GetDeviceConfig())
+	ap, err := device.GetActive()
+	if err != nil {
+		log.Error("Failed to read the current active partition.")
+	} else {
+		log.Infof("Mender running image: %s", ap)
```
kacf: Maybe change to `Mender running on partition: %s`, since image sounds like artifact name, which is a different thing.

main.go
```diff
@@ -420,6 +420,12 @@ func doMain(args []string) error {
 
 	env := NewEnvironment(new(osCalls))
 	device := NewDevice(env, new(osCalls), config.GetDeviceConfig())
+	ap, err := device.GetActive()
+	if err != nil {
+		log.Error("Failed to read the current active partition.")
```
kacf: Suggestion:

```
log.Errorf("Failed to read the current active partition: %s", err.Error())
```

main.go
pasinskim: I would add somewhere the world partition `Running the Mender image from partition: %s`

PR #291, opened 2018-07-06
MEN-1905: Enable user to force an update-check locally

oleorhagen:
The user can now force an update check by either running mender with the
-check-update option, or send a signal [SIGUSR1] to the running mender process.

Changelog: Title

### DIFF
--- a/main.go
+++ b/main.go
@@ -21,10 +21,12 @@ import (
 	"io/ioutil"
 	"os"
 	"os/exec"
+	"os/signal"
 	"path"
 	"path/filepath"
 	"runtime"
 	"strings"
+	"syscall"
 
 	"github.com/mendersoftware/log"
 	"github.com/mendersoftware/mender/client"
@@ -53,6 +55,7 @@ type runOptionsType struct {
 	daemon          *bool
 	bootstrapForce  *bool
 	showArtifact    *bool
+	updateCheck     *bool
 	client.Config
 }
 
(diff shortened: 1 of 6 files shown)


main.go
```diff
@@ -471,6 +478,39 @@ func handleCLIOptions(runOptions runOptionsType, env *uBootEnv, device *device,
 	return nil
 }
 
+// updateCheck sends a sigusr1 signal to the running mender daemon.
+func updateCheck() error {
+	cmd := exec.Command("pkill", "--signal", "SIGUSR1", "mender")
```
kacf: We should use a better mechanism than this. `pkill` is dangerous to use in production software, because it can match many processes, even unrelated ones that happen to contain "mender" in their names. And it will signal the current process. And it might not even be there for restricted systems.

The classic approach is to use a pid file somewhere, but maybe an easier and more future oriented approach is to ask systemd which PID mender has. `systemctl status mender` shows it and I'm sure there is a way to get a more script-friendly variant of it (check the man page).

main.go
```diff
@@ -118,6 +121,8 @@ func argsParse(args []string) (runOptionsType, error) {
 
 	daemon := parsing.Bool("daemon", false, "Run as a daemon.")
 
+	updateCheck := parsing.Bool("check-update", false, "force update check")
```
kacf: I think this flag is important enough that it deserves an integration test. Just clone an existing test, set the polling interval to be really long, and then use this flag to speed it up.

oleorhagen:
@kacf Looks like a spurious failure to me...

PR #293, opened 2018-07-10
MEN-1972 Failover servers

alfrunes:
@kacf @pasinskim ;
This PR is not yet finished, but it would be nice you could take a look at it and give some feedback of what you think about the implemented mechanic. So far only status codes 4XX and 5XX from API requests trigger failover, as well as an unsuccessful authorization request.
I have extended mender.conf to contain an array entry Servers of arbitrary many servers, that contains only the server URL and an optional tenant token, but it is straight forward to extend to also contain multiple different paths for server certificates (as @kacf [mentioned](https://tracker.mender.io/browse/MEN-1972?focusedCommentId=89477&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-89477) this might not be necessary).

### DIFF
--- a/client/client.go
+++ b/client/client.go
@@ -71,6 +71,15 @@ type ApiRequester interface {
 	Do(req *http.Request) (*http.Response, error)
 }
 
+// MenderServer is a placeholder for a full server definition used when
+// multiple servers are given. The fields corresponds to the definitions
+// given in menderConfig.
+type MenderServer struct {
+	ServerURL string
+	// TODO: Move all possible server specific configurations in
+	//       menderConfig over to this struct. (e.g. TenantToken?)
+}
+
 // APIError is an error type returned after receiving an error message from the
 // server. It wraps a regular error with the request_id - and if
 // the server returns an error message, this is also returned.
@@ -118,18 +127,18 @@ type ApiClient struct {
 }
 
 // function type for reauthorization closure (see func reauthorize@mender.go)
-type ClientReauthorizeFunc func() (AuthToken, error)
+type ClientReauthorizeFunc func(string) (AuthToken, error)
 
 // function type for setting server (in case of multiple fallover servers)
-type ServerManagementFunc func() string
+type ServerManagementFunc func() *MenderServer
 
 // Return a new ApiRequest
-func (a *ApiClient) Request(code AuthToken, _nextServerIterator ServerManagementFunc, req ClientReauthorizeFunc) *ApiRequest {
+func (a *ApiClient) Request(code AuthToken, nextServerIterator ServerManagementFunc, reauth ClientReauthorizeFunc) *ApiRequest {
 	return &ApiRequest{
 		api:                a,
 		auth:               code,
-		nextServerIterator: _nextServerIterator,
-		revoke:             req,
+		nextServerIterator: nextServerIterator,
+		revoke:             reauth,
 	}
 }
 
@@ -148,13 +157,13 @@ type ApiRequest struct {
 
 // tryDo is a wrapper around http.Do that also tries to reauthorize
 // on a 401 response (Unauthorized).
-func (ar *ApiRequest) tryDo(req *http.Request) (*http.Response, error) {
+func (ar *ApiRequest) tryDo(req *http.Request, serverURL string) (*http.Response, error) {
 	r, err := ar.api.Do(req)
-	if r != nil && r.StatusCode == http.StatusUnauthorized {
+	if err == nil && r.StatusCode == http.StatusUnauthorized {
 		// invalid JWT; most likely the token is expired:
 		// Try to refresh it and reattempt sending the request
 		log.Info("Device unauthorized; attempting reauthorization")
-		if jwt, e := ar.revoke(); e == nil {
+		if jwt, e := ar.revoke(serverURL); e == nil {
 			// retry API request with new JWT token
 			ar.auth = jwt
 			// check if request had a body
@@ -175,29 +184,48 @@ func (ar *ApiRequest) tryDo(req *http.Request) (*http.Response, error) {
 
 // Do is a wrapper for http.Do function for ApiRequests. This function in
 // addition to calling http.Do handles client-server authorization header /
-// reauthorization, as well as attempting failover servers (if given) if
+// reauthorization, as well as attempting failover servers (if given) whenever
 // the server "refuse" to serve the request.
 func (ar *ApiRequest) Do(req *http.Request) (*http.Response, error) {
+	if ar.nextServerIterator == nil {
+		return nil, errors.New("Empty server list!")
+	}
 	if req.Header.Get("Authorization") == "" {
 		// Add JWT to header
 		req.Header.Set("Authorization", fmt.Sprintf("Bearer %s", ar.auth))
 	}
-	r, err := ar.tryDo(req)
-	// if: (failoverservers given) && (error / bad statuscode)
-	// then: cycle servers and retry request
-	if (ar.nextServerIterator != nil) &&
-		(r != nil && r.StatusCode >= 400 && r.StatusCode < 600) || (err != nil) {
-		// incrementally try listed servers
-		log.Warnf("Server %q does not serve API-Request: %q. %s",
-			req.URL.Host, req.URL.Path, "Attempting failover servers.")
-		// do {try next server} while (server refuse to serve)
-		for serverURL := ar.nextServerIterator(); serverURL != ""; serverURL = ar.nextServerIterator() {
-			// set new host (server) in request
-			req.URL.Host = serverURL
-			req.Host = serverURL
-			r, err = ar.tryDo(req)
-			if r != nil && (r.StatusCode < 400 || r.StatusCode >= 600) {
-				log.Infof("Automatically fell over to server: %q", serverURL)
+	var r *http.Response
+	var host string
+	var err error
+
+	server := ar.nextServerIterator()
+	for {
+		// Split host from URL
+		tmp := strings.Split(server.ServerURL, "://")
+		if len(tmp) == 1 {
+			host = tmp[0]
+		} else {
+			// (len >= 2) should usually be 2
+			host = tmp[1]
+		}
+
+		req.URL.Host = host
+		req.Host = host
+		r, err = ar.tryDo(req, server.ServerURL)
+		if err == nil && r.StatusCode < 400 {
+			break
+		}
+		prewHost := server.ServerURL
+		if server = ar.nextServerIterator(); server == nil {
+			break
+		}
+		log.Warnf("Server %q failed to serve request %q. Attempting %q",
+			prewHost, req.URL.Path, server.ServerURL)
+	}
+	if server != nil {
+		// reset server iterator
+		for {
+			if ar.nextServerIterator() == nil {
 				break
 			}
 		}
--- a/client/client.go
+++ b/client/client.go
@@ -117,14 +117,19 @@ type ApiClient struct {
 	http.Client
 }
 
+// function type for reauthorization closure (see func reauthorize@mender.go)
 type ClientReauthorizeFunc func() (AuthToken, error)
 
+// function type for setting server (in case of multiple fallover servers)
+type ServerManagementFunc func() int
+
 // Return a new ApiRequest
-func (a *ApiClient) Request(code AuthToken, req ClientReauthorizeFunc) *ApiRequest {
+func (a *ApiClient) Request(code AuthToken, _setNextServer ServerManagementFunc, req ClientReauthorizeFunc) *ApiRequest {
 	return &ApiRequest{
-		api:    a,
-		auth:   code,
-		revoke: req,
+		api:           a,
+		auth:          code,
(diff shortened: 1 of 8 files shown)


config.go
```diff
@@ -55,14 +84,29 @@ func LoadConfig(configFile string) (*menderConfig, error) {
 		return nil, err
 	}
 
+	if confFromFile.Servers != nil {
+		// Overwrite "active" server with first one from the list of servers.
+		confFromFile.ServerURL = confFromFile.Servers[0].ServerURL
+		confFromFile.TenantToken = confFromFile.Servers[0].TenantToken
```
kacf: So `Servers` essentially replaces `ServerURL` in the configuration file? If so, I would produce a hard error here if both are present, just to be sure there is no doubt about which server(s) gets used.

Same goes for the `TenantToken`.

config.go
```diff
@@ -55,14 +84,29 @@ func LoadConfig(configFile string) (*menderConfig, error) {
...
+			if strings.HasSuffix(confFromFile.ServerURL, "/") {
+				confFromFile.Servers[i].ServerURL =
+					strings.TrimSuffix(confFromFile.Servers[i].ServerURL, "/")
+			}
+		}
+	}
+
 	if strings.HasSuffix(confFromFile.ServerURL, "/") {
+		// Remove trailing '/', which is added back in REST-API strings.
 		confFromFile.ServerURL = strings.TrimSuffix(confFromFile.ServerURL, "/")
```
kacf: Previous comment also related to this section. This will overwrite the first entry from `Servers`, which can be confusing.

mender.go
```diff
@@ -503,14 +525,14 @@ func (m *mender) CheckUpdate() (*client.UpdateResponse, menderError) {
...
-	err := s.Report(m.api.Request(m.authToken, reauthorize(m)), m.config.ServerURL,
+	err := s.Report(m.api.Request(m.authToken, setNextServer(m), reauthorize(m)), m.config.ServerURL,
 		client.StatusReport{
 			DeploymentID: update.ID,
 			Status:       status,
 		})
 	if err != nil {
 		log.Error("error reporting update status: ", err)
-
+		// lock your machine!
```
kacf: @oleorhagen: Is this you? :-)

mender.go
```diff
@@ -539,9 +566,61 @@ func reauthorize(m *mender) func() (client.AuthToken, error) {
...
+		if m.config.ServerURL == m.config.Servers[i].ServerURL {
+			skipIdx = i
+			break
+		}
+	}
+	return func() (srvURL string) {
+		if m.config.Servers == nil {
+			return ""
+		}
+		if idx == skipIdx {
```
kacf: So the returned function keeps the state of the variables of the outer function which defined it, even after both have returned? Interesting, didn't know this...

mender.go
```diff
@@ -539,9 +566,61 @@ func reauthorize(m *mender) func() (client.AuthToken, error) {
 	}
 }
 
+// setNextServer is an iterator like function that cycles through the
+// list of available servers in mender.menderConfig.Servers and puts
+// the server as "current active", i.e. fills out the fields
+// mender.menderConfig.ServerURL and mender.menderConfig.TenantToken
+// with the corresponding fields in the server list.
+func setNextServer(m *mender) func() string {
```
kacf: Confusing name. It sounds like you're using the `m` argument to set some state somewhere, not returning anything. I would perhaps call it `nextServerIterator` or `nextServerFactory` or something like that.

config.go
kacf: Following this, I would consider only using the `Servers` list internally. IOW, instead of first trying `ServerURL` and then resorting to the list, just use the list in either case, starting with the first element. This might also simplify some of the fallback logic.

If only `ServerURL` is specified in the config file, you can just add this to the list as the only element.

mender.go
```diff
@@ -539,9 +566,61 @@ func reauthorize(m *mender) func() (client.AuthToken, error) {
...
+		return nil
+	}
+
+	var skipIdx int
+	idx := 0
+	numServers := len(m.config.Servers)
+	// find index to skip
+	for i := 0; i < numServers; i++ {
+		if m.config.ServerURL == m.config.Servers[i].ServerURL {
+			skipIdx = i
```
kacf: If we only use the `Servers` list internally, then we could probably remove the `skipIdx` logic.

config.go
alfrunes: The idea is that the whole `Servers` entry is optional, and the active one used is "loaded" into the `ServerURL`, `TenantToken` entry in the base structure. This preserves backward compatibility. Note that when the `Servers` field is specified, the `ServerURL` and `TenantToken` fields are "don't care" fields when parsing the configuration file. If `Servers` is not specified, then the client behaves as before.

client/client.go
```diff
@@ -166,6 +173,38 @@ func (ar *ApiRequest) Do(req *http.Request) (*http.Response, error) {
...
+		log.Warnf("Server does not serve API-Request: '%s'. %s",
+			req.URL.Path, "Attempting failover servers.")
+		// do {try next server} while (server refuse to serve)
+		for serverURL := ar.setNextServer(); serverURL != ""; serverURL = ar.setNextServer() {
+			// set new host (server) in request
+			req.URL.Host = serverURL
+			req.Host = serverURL
+			r, err = ar.tryDo(req)
+			if r != nil && (r.StatusCode < 400 || r.StatusCode >= 600) {
+				log.Infof("Automatically fell over to server: \"%s\"", serverURL)
```
oleorhagen: %q can be used for quoting strings ^^

mender.go
alfrunes: haha, oops...

config.go
```diff
@@ -23,29 +23,58 @@ import (
 	"github.com/pkg/errors"
 )
 
+type menderServer struct {
+	// menderServer: Placeholder for a full server definition
```
oleorhagen: Maybe move outside of the struct, so that godoc can handle it?

config.go
```diff
@@ -23,29 +23,58 @@ import (
...
+	// Rootfs device path
+	RootfsPartA string
+	RootfsPartB string
+
+	// Poll interval for checking for new updates
+	UpdatePollIntervalSeconds int
+	// Poll interval for periodically sending inventory data
+	InventoryPollIntervalSeconds int
+
+	// Global retry polling max interval for fetching update, authorize wait and update status
```
oleorhagen: Thank you!

client/client.go
```diff
@@ -117,14 +117,19 @@ type ApiClient struct {
...
 // Return a new ApiRequest
-func (a *ApiClient) Request(code AuthToken, req ClientReauthorizeFunc) *ApiRequest {
+func (a *ApiClient) Request(code AuthToken, _setNextServer ServerManagementFunc, req ClientReauthorizeFunc) *ApiRequest {
 	return &ApiRequest{
-		api:    a,
-		auth:   code,
-		revoke: req,
+		api:           a,
+		auth:          code,
+		setNextServer: _setNextServer,
```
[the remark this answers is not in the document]
alfrunes: Only to separate argument from struct field name.

mender.go
oleorhagen: haha, I believe not. @alfrunes has siblings, I think

mender.go
alfrunes: I agree, this function has been through some changes since I gave it the name.

oleorhagen: +1 `nextServerIterator`

mender.go
alfrunes: You're right, it not really necessary. Just added this so we don't repeat the same request for the current "active" server.

client/client.go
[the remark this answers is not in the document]
oleorhagen: Don't really think it's necessary here

mender.go
```diff
@@ -539,9 +566,61 @@ func reauthorize(m *mender) func() (client.AuthToken, error) {
...
+		m.config.ServerURL = m.config.Servers[i].ServerURL
+		m.config.TenantToken = m.config.Servers[i].TenantToken
+		if idx == numServers {
+			// We've tried all server, and reset to default server
+			return ""
+		} else {
+			idx++
+			// return serverURL with prefix (https?://) trimmed of
+			srv := m.config.Servers[i].ServerURL
+			_srv := strings.Split(srv, "://")
```
oleorhagen: strings.TrimPrefix?

alfrunes: Just tried make it more generic.

client/client.go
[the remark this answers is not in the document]
oleorhagen: hb https://golang.org/pkg/net/url/ ?

mender.go
oleorhagen: hb https://golang.org/pkg/net/url/ ?

config.go
alfrunes: I'm not sure if I want to replace base fields completely with the list structure, as I don't see how this simplifies the logic. Maybe I'm missing something?

mender.go
```diff
@@ -409,14 +411,34 @@ func (m *mender) Authorize() menderError {
...
-				log.Warn("can not remove rejected authentication token")
+		// If (multiple) failover servers are available, cycle through
+		// list and try to authorize.
+		nextServer := nextServerIterator(m)
+		if nextServer != nil {
+			var srv string
+			log.Warnf("Authorization resulted in error: %s", err)
+			log.Warn("Trying to authorize on failover server(s).")
+			for srv = nextServer(); srv != ""; srv = nextServer() {
+				rsp, err = m.authReq.Request(m.api, m.config.ServerURL, m.authMgr)
```
kacf: Here is an example of where I think the logic could be simplified if only using one single list: You are calling `Request` here, but it's already called once above as well. My thinking is that this could be reduced to simply entering the loop immediately, before you have called any `Request` function, and then start with the first entry, which is the primary server. Does that make sense?

alfrunes: So you're saying that we should iterate the list for each request instead of completely shifting attention from one server to another on an unsuccessful request?

mender.go
alfrunes: ??

mender.go
kacf: Well, aren't those two things kind of the same? Try one server, if it fails, try next server. So yes, I guess that's what I'm saying.

mender.go
oleorhagen: Was just thinking it could be used for parsing the url, but might be a bit over the top.

alfrunes:
@kacf; I have changed the failover mechanism as requested. It won't pass the tests in Travis due to the change in server configuration data structure, but they pass locally.

mender.go
```diff
@@ -525,23 +551,86 @@ func (m *mender) ReportUpdateStatus(update client.UpdateResponse, status string)
 	return nil
 }
 
-func reauthorize(m *mender) func() (client.AuthToken, error) {
+/* client closures */
+// see client.go: ApiRequest.Do()
+
+// reauthorize is a closure very similar to mender.Authorize(), but instead of
+// walking through all servers in the menderConfig.Servers list, it only tries
+// serverURL.
```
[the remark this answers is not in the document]
alfrunes: I discovered a potential inconsistency if you were to simply call `mender.Authorize` here:
Since `Authorize` now inevitably starts from the beginning of the `Servers` list, if a request is rejected on the first server but is valid on the second server on the list but the `jwt` token happens to be expired, but authorization request is served by the first one. 
I changed this closure in order to only attempt to authorize on the current server in the `Do` loop.

alfrunes:
@kacf does this look better?

mender.go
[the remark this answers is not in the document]
kacf: Ah, good catch!

alfrunes:
@kacf: https://github.com/mendersoftware/integration/pull/533

kacf:
Superseded by #320.

jkumarooty:
Dockers

The following is created and not started

mendersoftware/api-gateway:2.0.0
The following is created and restarting regularly

mendersoftware/deployments:1.7.0
redis:5-alpine3.8
Below is the error in the container of "mendersoftware/deployments:1.7.0"

WARNING: ca-certificates.crt does not contain exactly one certificate or CRL: skipping
time="2019-09-18T06:39:21Z" level=info msg="Deployments Service, version unknown starting up"
time="2019-09-18T06:39:21Z" level=info msg="automigrate is ON, will apply migrations" file=migrations.go func=migrations.Migrate line=48
time="2019-09-18T06:39:21Z" level=info msg="migrating deployment_service" file=migrations.go func=migrations.MigrateSingle line=70
time="2019-09-18T06:39:21Z" level=info msg="migration to version 1.2.1 skipped" db=deployment_service file=migrator_simple.go func="migrate.(*SimpleMigrator).Apply" line=125
time="2019-09-18T06:39:21Z" level=info msg="DB migrated to version 1.2.1" db=deployment_service file=migrator_simple.go func="migrate.(*SimpleMigrator).Apply" line=140
NoCredentialProviders: no valid providers in chain. Deprecated.
For verbose messaging see aws.Config.CredentialsChainVerboseErrors

 

Due to the errors we are not able to start mender and it is become the blocker.

PR #294, opened 2018-07-12
Add check for canary value in U-Boot environment.

kacf:
The way it works is that we store the 'mender_check_saveenv_canary=1'
variable in the default U-Boot environment. We do not store the
corresponding 'mender_saveenv_canary=1' value in the default
environment, but instead we write this during the boot process. When
the Mender client finds the former variable, it will look for the
second value as well, and produce an error if it is not present. If
this happens it is an indication that the U-Boot boot loader and the
user space tools do not agree on the location of the environment. It
will also catch cases where mender_setup is not run.

Also reorganize the logging and error returning to include the error
when returning higher up in the callstack, otherwise the resulting log
entry is quite misleading.

Changelog: Add automatic check for canary value in U-Boot environment
to try to detect if there is a problem in the environment setup of
U-Boot and/or the u-boot-fw-utils tools.

### DIFF
--- a/bootenv.go
+++ b/bootenv.go
@@ -1,4 +1,4 @@
-// Copyright 2017 Northern.tech AS
+// Copyright 2018 Northern.tech AS
 //
 //    Licensed under the Apache License, Version 2.0 (the "License");
 //    you may not use this file except in compliance with the License.
@@ -15,12 +15,12 @@ package main
 
 import (
 	"bufio"
-	"errors"
 	"fmt"
 	"os/exec"
 	"strings"
 
 	"github.com/mendersoftware/log"
+	"github.com/pkg/errors"
 )
 
 type uBootEnv struct {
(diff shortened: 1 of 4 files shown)


kacf:
Belong together with https://github.com/mendersoftware/meta-mender/pull/550.

bootenv.go
```diff
@@ -32,12 +32,54 @@ func NewEnvironment(cmd Commander) *uBootEnv {
...
+		return nil
+	}
+
+	value, ok := vars["mender_check_saveenv_canary"]
+	if !ok || value != "1" {
+		// Checking should not be done.
+		return nil
+	}
+
+	err_msg := "Failed mender_saveenv_canary check. There is an error in the U-Boot setup. Likely causes are: 1) Mismatch between the U-Boot boot loader environment location and the location specified in /etc/fw_env.config. 2) 'mender_setup' is not run by the U-Boot boot script"
```
oleorhagen: Nitpick: the go_centric_way_of_doing_it_is errMsg

bootenv_test.go
```diff
@@ -128,3 +132,38 @@ func Test_EnvRead_NonExisting_FailsReading(t *testing.T) {
 		t.FailNow()
 	}
 }
+
+func Test_EnvCanary(t *testing.T) {
+	runner := newTestOSCalls("var=1\nmender_check_saveenv_canary=1\nmender_saveenv_canary=0\n", 0)
+	fakeEnv := uBootEnv{&runner}
```
oleorhagen: Why not make this a table test?

partitions.go
```diff
@@ -206,8 +206,7 @@ func (p *partitions) getAndCacheActivePartition(rootChecker func(StatCommander,
 func getBootEnvActivePartition(env BootEnvReadWriter) (string, error) {
 	bootEnv, err := env.ReadEnv("mender_boot_part")
 	if err != nil {
-		log.Error(err)
```
oleorhagen: Why remove the logging?

partitions_test.go
```diff
@@ -193,7 +193,8 @@ func Test_getActivePartition_noActiveInactiveSet(t *testing.T) {
 		envCaller.output = test.fakeEnv
 		envCaller.retCode = test.fakeEnvRet
 		active, err := fakePartitions.getAndCacheActivePartition(test.rootChecker, mountedDevicesGetter)
-		if err != test.expectedError || active != test.expectedActive {
+		errorOK := (err == test.expectedError || strings.Contains(err.Error(), test.expectedError.Error()))
+		if !errorOK || active != test.expectedActive {
```
oleorhagen: assert.True ?

bootenv_test.go
kacf: I looked, but I think it's not worth it for this small test. There are only 6 test cases, and 3 of them take different code paths.

partitions.go
kacf: I tried to explain this in the commit message:

> Also reorganize the logging and error returning to include the error when returning higher up in the callstack, otherwise the resulting log entry is quite misleading.

The problem was that the error you get further up in the call stack looks completely out-of-context unless paired with this one, so since we now return the full error message, it makes sense to remove this and let the caller print all of it at once.

kacf:
Thanks! Rest of issues fixed.

oleorhagen:
Alrighty. Good to go then

kacf:
Need to fix acceptance tests first though. Some failure [there](https://github.com/mendersoftware/meta-mender/pull/550).

PR #295, opened 2018-07-18
FIX: Remove extra sigusr1 signal from updateCheck test.

oleorhagen:
Changelog: None


kacf:
Although the test cases involved look like the usual spurious ones, there is an unexpectedly large number of them failing. I have never seen more than one at a time, and here we have five! I would rerun this, just to be sure.

PR #298, opened 2018-08-10
FIX: Enabling compiling ppc64le + race condition fix

kacf:


state_test.go
```diff
@@ -325,9 +325,8 @@ func TestStateUpdateReportStatus(t *testing.T) {
 		s, c = s.Handle(&ctx, sc)
 		assert.IsType(t, &UpdateStatusReportState{}, s)
 		assert.False(t, c)
-		assert.WithinDuration(t, time.Now(), now.Add(retry), time.Millisecond)
-		now = time.Now()
 	}
+	assert.WithinDuration(t, time.Now(), now.Add(time.Duration(int64(shouldTry)*int64(retry))), time.Millisecond*10)
```
pasinskim: I would do something like 
`assert.WithinDuration(t, now, time.Now(), shouldTry * retry + time.Millisecond*10)`.

kacf: Yes, perhaps that's clearer. Will do that.

kacf:
Already checked separately on Jenkins. Merging.

PR #299, opened 2018-08-14
Fall-back to environment and conf in active part probe

mirzak:
There are situations where it nearly impossible to find the active
root-filesystem based on probing mounted devices.

One example is this:

    /dev/root on /rom type squashfs (ro,relatime)

There is no information about the actual device device used in above
example.

One additional probe step could be parsing /proc/cmdline to find the
"root=" argument but this is essentially the same as looking at
"mender_boot_part" environment variable.

Hence we add a fall-back to only using the "mender_boot_part"
environment variable and comparing that to what we have in
"mender.conf".

Changelog: Title

Signed-off-by: Mirza Krak <mirza.krak@northern.tech>

### DIFF
--- a/partitions.go
+++ b/partitions.go
@@ -191,6 +191,20 @@ func (p *partitions) getAndCacheActivePartition(rootChecker func(StatCommander, 
 
 	activePartition, err := getRootFromMountedDevices(p, rootChecker, mountedDevices, rootDevice)
 	if err != nil {
+		// If we reach this point, we have not been able to find a match
+		// based on mounted device.
+		//
+		// Fall-back to configuration and environment only!
+		if checkBootEnvAndRootPartitionMatch(bootEnvBootPart, p.rootfsPartA) {
+			p.active = p.rootfsPartA
+			log.Debug("Setting active partition from configuration and environment: ", p.active)
+			return p.active, nil
+		}
+		if checkBootEnvAndRootPartitionMatch(bootEnvBootPart, p.rootfsPartB) {
+			p.active = p.rootfsPartB
+			log.Debug("Setting active partition from configuration and environment: ", p.active)
+			return p.active, nil
+		}
 		return "", err
 	}
 	if checkBootEnvAndRootPartitionMatch(bootEnvBootPart, activePartition) {


mirzak:
I have actually find a "problem" with this solution and that is that you are not able to run:

mirzak:
Tests are failing but will address that after feedback.

partitions.go
```diff
@@ -185,21 +185,34 @@ func (p *partitions) getAndCacheActivePartition(rootChecker func(StatCommander,
...
+				p.active = activePartition
+				log.Debug("Setting active partition: ", activePartition)
+				return p.active, nil
+			}
+
+			// If we reach this point, we have not been able to find a match
+			// based on mounted device.
+			//
+			// Fall-back to configuration and environment only!
+		}
```
kacf: I'm a little bit confused about the use case and how this solves the problem. You have this in your mount output:

```
/dev/root on /rom type squashfs (ro,relatime)
```

However, neither `getMountedDevices()`[1] nor `getRootFromMountedDevices()` use this output, only `getRootCandidateFromMount()` does. `getMountedDevices()` gets the device files on the system, and then try to match this against the major and minor number of the `/` descriptor. Does squashfs not have a device file to match against?

If so, I think we should turn the matching against `p.rootfsPartA` and `p.rootfsPartB` into a special case rather than a fallback. IOW I think this code path should be taken if and only if `getRootFromMountedDevices()` returns error. Feels safer. :-) Here is my suggestion:

```
diff --git a/partitions.go b/partitions.go
index db67a30..5aba1d2 100644
--- a/partitions.go
+++ b/partitions.go
@@ -191,6 +191,20 @@ func (p *partitions) getAndCacheActivePartition(rootChecker func(StatCommander,
 
        activePartition, err := getRootFromMountedDevices(p, rootChecker, mountedDevices, rootDevice)
        if err != nil {
+               // If we reach this point, we have not been able to find a match
+               // based on mounted device.
+               //
+               // Fall-back to configuration and environment only!
+               if checkBootEnvAndRootPartitionMatch(bootEnvBootPart, p.rootfsPartA) {
+                       p.active = p.rootfsPartA
+                       log.Debug("Setting active partition from configuration and environment: ", p.active)
+                       return p.active, nil
+               }
+               if checkBootEnvAndRootPartitionMatch(bootEnvBootPart, p.rootfsPartB) {
+                       p.active = p.rootfsPartB
+                       log.Debug("Setting active partition from configuration and environment: ", p.active)
+                       return p.active, nil
+               }
                return "", err
        }
        if checkBootEnvAndRootPartitionMatch(bootEnvBootPart, activePartition) {
```

[1] `getMountedDevices()` is really incorrectly named. It should be called `getDeviceFiles()`

partitions.go
```diff
@@ -185,21 +185,34 @@ func (p *partitions) getAndCacheActivePartition(rootChecker func(StatCommander,
...
-		p.active = activePartition
-		log.Debug("Setting active partition: ", activePartition)
+
+	if checkBootEnvAndRootPartitionMatch(bootEnvBootPart, p.rootfsPartB) {
+		p.active = p.rootfsPartB
+		log.Debug("Setting active partition from configuration and environment: ", p.active)
 		return p.active, nil
 	}
 
-	log.Error("Mounted root '" + activePartition + "' does not match boot environment mender_boot_part: " + bootEnvBootPart)
```
kacf: Following the above suggestion, I think we can keep this.

partitions.go
mirzak: > getMountedDevices() gets the device files on the system, and then try to match this against the major and minor number of the / descriptor. Does squashfs not have a device file to match against?

No, squashfs nor UBI can match major/minor numbers of / descriptor.  I did a fix for UBI when that was integrated where it fall-backed to the `mount` output and comparing that to devices. In this case `mount` does not give any hints of device used and hence that it does not work.

This is a special case, where UBI volumes are used, with squashfs content on top of ubiblocks (read-only MTD blocks on top of UBI). 

I like your approach more and will try that out. I was a bit worried that my approach removes the error paths.

Thank you @kacf

kacf:
Oh, and maybe change the changelog to:

```
Changelog: Fix active partition detection when using non-native filesystems.
```

The title is correct, but quite developer centric. Makes sense?

mirzak:
Makes sense

mirzak:
@kacf, updated.

The travis failure seems to be related to an update in "check" package [1]. We need to update go version on travis?

[1]. https://github.com/opennota/check/commit/b43223f80369893b298c7cda1da077d015a28d86#diff-929eb554b63f45665f7f9bb124bebc4f

kacf:
Yes, can you add a commit to [update it](https://github.com/mendersoftware/mender/blob/master/.travis.yml#L6) to 1.10.3? Sumo is using 1.10, so it's appropriate.

mirzak:
Created #300.

Though it seems that they also fixed it in the upstream [check package](https://github.com/opennota/check/commit/39d55212d30924d4e8ace3be725503cb2ce19085#diff-9eadaea9550ec7c8d53e87981dc31bbd)

kacf:
Indeed, it passed now! :-)

I think this is a quite important bug fix, so we should cherry-pick this. Agree?

mirzak:
Agreed.

kacf:
I'll cherry-pick this, don't worry!

PR #302, opened 2018-08-24
Use ioctl constants from golang.org/x/sys/unix

tklauser:
This PR updates the vendored copy of golang.org/x/sys/unix and uses the ioctl constants from there instead of defining them locally. This allows to to support said ioctls on all GOARCHes supported by golang.org/x/sys/unix instead of just the ones listed in ioctl_{32,64bit}.go


kacf:
Rebased and merged in #306. Thanks @tklauser!

PR #303, opened 2018-08-27
MEN-2059: Add inventory scripts for rootfs type and bootloader integr…

kacf:
…ation.

Changelog: Title


kacf:
Merged as part of #304.

PR #304, opened 2018-08-27
 MEN-2060: New inventory script for "os" attribute, installed by default.

kacf:
Based on #303. Only the "os" attribute is new.

### DIFF
--- a/support/mender-inventory-os
+++ b/support/mender-inventory-os
@@ -1,0 +1,39 @@
+#!/bin/sh
+
+# Returns an "os" attribute to Mender containing the currently running OS.
+
+set -e
+
+for file in /etc/os-release /usr/lib/os-release; do
+    if [ ! -e $file ]; then
+        continue
+    fi
+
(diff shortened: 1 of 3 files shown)


support/mender-inventory-os
```diff
@@ -0,0 +1,39 @@
...
+fi
+
+if [ -e /etc/issue ]; then
+    OS="$(cat /etc/issue)"
+    if [ -n "$OS" ]; then
+        echo "os=$OS"
+        exit 0
+    fi
+fi
+
```
mirzak: Does it makes sense to set

   os="unknown"

here. That way we at least know that the script tried but did not find anything.

support/mender-inventory-os
```diff
@@ -0,0 +1,39 @@
...
+# Failed to find OS.
+exit 1
```
mirzak: Will the above (exit 1) abort the whole "inventory" stage?

PR #308, opened 2018-09-16

cosinekitty:

file will be used.

file, it will keep its defalut value ("" for a string or
0 for an integer).

After successfully loading settings, the merged values
are all available in debug output from the Mender client.

Signed-off-by: Donald D. Cross <cosinekitty@gmail.com>

### DIFF
--- a/config.go
+++ b/config.go
@@ -16,6 +16,7 @@ package main
 import (
 	"encoding/json"
 	"io/ioutil"
+	"os"
 	"strings"
 
 	"github.com/mendersoftware/log"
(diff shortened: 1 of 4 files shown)


cosinekitty:
OK, if I understand the Travis output, the reason for the failure is this one line:

```
!!! FAILED license check on ./config_test.go; make sure copyright year matches last modified year of the file
```

I will try again later today.

cosinekitty:
I see now I have made a mess by not having the correct sign-off in the commit messages. I'm going to close this and create another pull request with everything fixed (hopefully!).

PR #309, opened 2018-09-16

cosinekitty:

Changelog: Mender client now loads configuration settings from 
both /etc/mender/mender.conf and (if it exists) 
/var/lib/mender/mender.conf. The second file is located 
on the data partition, so it allows any subset of configuration
changes to survive upgrades.

file will be used.

file, it will keep its defalut value ("" for a string or
0 for an integer).

Use DEBUG severity to log final merged configuration settings.
We do not use INFO because that would expose sensitive
information (keys) by default.

Updated copyright years in config_test.go, partitions.go
to satisfy Travis CI validation.  (I don't understand
how partitions.go didn't cause a failure already.)

### DIFF
--- a/config.go
+++ b/config.go
@@ -16,6 +16,7 @@ package main
 import (
 	"encoding/json"
 	"io/ioutil"
+	"os"
 	"strings"
 
 	"github.com/mendersoftware/log"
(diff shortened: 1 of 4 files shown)


cosinekitty:
I don't believe this Travis CI failure is caused by my changes. It looks like a very stingy 1ms limit was exceeded in an unrelated test. I'm not sure what to do at this point. Any suggestions?

```
— FAIL: TestStateUpdateReportStatus (0.03s)
 Error Trace: state_test.go:316
 Error: Max difference between 2018-09-16 15:31:19.035208972 +0000 UTC and 2018-09-16 15:31:19.03271679 +0000 UTC allowed is 1ms, but difference was 2.492182ms
```

kacf:
I'm also not sure about the Travis failure, I may have to look a bit more. But let's move to the master branch first, and then see what the status is there.

cosinekitty:
OK. I am closing this pull request and creating another based on the master branch.

PR #310, opened 2018-09-18
MEN-2073

cosinekitty:

Changelog: Mender client now loads configuration settings from 
both /etc/mender/mender.conf and (if it exists) 
/var/lib/mender/mender.conf. The second file is located 
on the data partition, so it allows any subset of configuration
changes to survive upgrades.

file will be used.

file, it will keep its default value ("" for a string or
0 for an integer).

Use DEBUG severity to log final merged configuration settings.
We do not use INFO because that would expose sensitive
information (keys) by default.

Updated source code copyright years to satisfy Travis CI validation.

### DIFF
--- a/config.go
+++ b/config.go
@@ -16,6 +16,7 @@ package main
 import (
 	"encoding/json"
 	"io/ioutil"
+	"os"
 	"strings"
 
 	"github.com/mendersoftware/log"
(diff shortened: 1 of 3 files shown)


cosinekitty:
The way that I did this unfortunately makes it look like I'm trying to merge 8 commits. I will wait to see what happens with Travis CI, but I'm tempted to create yet another pull request that has only a single commit in it.

cosinekitty:
OK, the Travis CI results have test failures that I don't understand. I could use some help with this.

kacf:
@cosinekitty: I cherry-picked just that change alone to #311. It looks better now, Travis is happy (just ignore the patch coverage thing), and hopefully Jenkins will be as well, when it finishes!

PR #312, opened 2018-09-19
MEN-2083: Improve error message when running mender as non-root user

alfrunes:
This ommits changing the confusing error messages produced, but clarifies the
non-root execution attempt cases.

changelog: title

### DIFF
--- a/bootenv.go
+++ b/bootenv.go
@@ -16,6 +16,7 @@ package main
 import (
 	"bufio"
 	"fmt"
+	"os"
 	"os/exec"
 	"strings"
 
@@ -69,10 +70,17 @@ func (e *uBootEnv) checkEnvCanary() error {
 
 func (e *uBootEnv) ReadEnv(names ...string) (BootVars, error) {
 	if err := e.checkEnvCanary(); err != nil {
+		if os.Geteuid() != 0 {
+			return nil, errors.Wrap(err, "requires root privileges")
+		}
 		return nil, err
 	}
 	getEnvCmd := e.Command("fw_printenv", names...)
-	return getEnvironmentVariable(getEnvCmd)
+	vars, err := getEnvironmentVariable(getEnvCmd)
+	if err != nil && os.Getuid() != 0 {
+		return nil, errors.Wrap(err, "requires root privileges")
+	}
+	return vars, err
 }
 
 func (e *uBootEnv) WriteEnv(vars BootVars) error {
@@ -91,6 +99,9 @@ func (e *uBootEnv) WriteEnv(vars BootVars) error {
 	if err != nil {
 		log.Errorln("Could not execute fw_setenv: ", err)
 		pipe.Close()
+		if os.Geteuid() != 0 {
+			return errors.Wrap(err, "requires root privileges")
+		}
 		return err
 	}
 	for k, v := range vars {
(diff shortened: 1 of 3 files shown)


mirzak:
We should not require UID == 0, to run the Mender client. And instead we should detect "permission denied" errors.

And you actually do not need to be root to access "block devices", as they are normally setup with the following permissions:

```
$ ls -alh /dev/mmcblk0*
brw-rw---- 1 root disk 179, 0 19 sep 15.04 /dev/mmcblk0
brw-rw---- 1 root disk 179, 1 19 sep 15.04 /dev/mmcblk0p1
brw-rw---- 1 root disk 179, 2 19 sep 15.04 /dev/mmcblk0p2
brw-rw---- 1 root disk 179, 3 19 sep 15.04 /dev/mmcblk0p3
brw-rw---- 1 root disk 179, 4 19 sep 15.04 /dev/mmcblk0p4
``` 

Meaning a user that is part of the "disk" group would be allowed to write to U-boot ENV and perform the update.

alfrunes:
I have moved the permission detection to where `fw_[print|set]env` is invoked and generates an error. I'm not sure what you mean by `access()`.

bootenv.go
```diff
@@ -69,10 +75,17 @@ func (e *uBootEnv) checkEnvCanary() error {
 
 func (e *uBootEnv) ReadEnv(names ...string) (BootVars, error) {
 	if err := e.checkEnvCanary(); err != nil {
+		if os.Geteuid() != 0 {
+			return nil, permissionDeniedError("fw_printenv")
```
kacf: Sorry to be pedantic, but we don't really know that permissions are the issue here. It could also be that the binary is not found, for example, and here we're hiding the original error. Here's what I suggest:

```
if os.Geteuid() != 0 {
    return nil, errors.Wrap(err, "requires root privileges")
}
```

That way you preserve both errors, and the user is more likely to figure out what went wrong. Note that I'm avoiding the phrase "Permission denied" and just stating that we require root privileges, because we don't know if that's the issue.

Similar for the rest of the diff.

mirzak: Are we not able to catch the error code from "fw_setenv" invocation? I am pretty sure it returned the correct error code (EACCES) when invoked from the CLI.

mirzak: Tested, it prints the correct error string but returns 1.

```
root@colibri-imx7-emmc:~# /sbin/fw_setenv mirza test                                                                                        
Write error on /dev/mmcblk0boot0: Operation not permitted                                                                                   
Error: can't write fw_env to flash                                                                                                          
root@colibri-imx7-emmc:~# echo $?                                                                                                           
1
```

alfrunes: Made the changes @kacf requested.

kacf:
And regarding that, I have no clue what's wrong with Travis. Can you just try "Restart Build" on the Travis page?

alfrunes:
Already tried that; looks like some version inconsistency in a go package
```
# golang.org/x/tools/go/packages
../../../golang.org/x/tools/go/packages/packages.go:695: undefined: types.SizesFor
```

alfrunes:
Looks like the binary_size test needs an increment..

alfrunes:
@kacf; Looks like the binary size test needs an increment upstream. Can you fix it?

kacf:
Superseded by #328.

PR #313, opened 2018-09-20
MEN-2152: Log whether artifact's digital signature is valid

cosinekitty:
Changelog: Print a message to the mender log when the
mender client has confirmed the authenticity of an
artifact's digital signature.

When the Mender client authenticates a downloaded
artifact's digital signature using a public key configured
by the property "ArtifactVerifyKey" in mender.conf,
it now logs the following message:
"installer: authenticated digital signature of artifact"
This allows the developer to confirm that verification
of mender.sig took place, and that this Mender client
will reject any future downloads bearing a forged signature.

When the Mender client is not configured with a public
key for authenticating downloaded artifacts, log
the following message:
"no public key was provided for authenticating the artifact"
This informs the developer that no signature verification
took place. If signature verification was intended, it
reveals something went wrong and needs to be investigated.

### DIFF
--- a/installer/installer.go
+++ b/installer/installer.go
@@ -53,6 +53,7 @@ func Install(art io.ReadCloser, dt string, key []byte, scrDir string,
 		ar = areader.NewReaderSigned(art)
 	} else {
 		ar = areader.NewReader(art)
+		log.Info("no public key was provided for authenticating the artifact")
 	}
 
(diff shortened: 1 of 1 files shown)


cosinekitty:
Will submit another PR with the copyright year in installer.go fixed.

PR #318, opened 2018-09-26
MEN-2159: Fix update check not working under BusyBox.

kacf:
Changelog: Title


oleorhagen:
Had never occured to me that this could be an issue...

PR #321, opened 2018-09-28
Add Community Code of Conduct

michaelatmender:
Changelog: Title

Signed-off-by: Michael Clelland <michael.clelland@mender.io>

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -174,6 +174,10 @@ if desired.
   Changelog: None
   ```
 
+## Contributor Code of Conduct
+
+We have a [Code of Conduct](https://github.com/mendersoftware/mender/blob/master/code-of-conduct.md) that applies to all contributors and participants to the Mender project.
+
 
 ## Let us work together with you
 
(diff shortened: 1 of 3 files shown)


mirzak:
Why so many deviations from the original?

michaelatmender:
> Why so many deviations from the original?

I just followed a suggestion in the JIRA ticket to reuse like this. Here's the [link and discussion](https://tracker.mender.io/browse/MEN-2117?focusedCommentId=91365&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-91365)

mirzak:
Missed that discussion, nevermind that remark. 

We should add links from README.md to CoC. And probably it should be mentioned in the contribution.md as well. What do you think?

estenberg:
I would just drop the section from README.md, as it links to contributing.md which links to code of contduct. This should be enough IMHO.

The main issue is that there is not just one README.md, but each repo has one, and we should try to keep them consistent to avoid user confusion.

@mirzak WDYT?

mirzak:
Sounds reasonable to me

PR #322, opened 2018-10-01
Added unit-tests for mender-artifactV3 installation.

oleorhagen:
Changelog: None


oleorhagen:
Unit-tests added for installing a mender-artifactV3. This commit should be seen along with:
https://github.com/mendersoftware/mender-artifact/pull/138

installer/installer_test.go
```diff
@@ -70,25 +76,55 @@ func TestInstallSigned(t *testing.T) {
 	err = Install(art, "vexpress-qemu", []byte(PublicRSAKey), "", new(fDevice), true)
 	assert.NoError(t, err)
 
-	// have a key but artifact is unsigned
-	art, err = MakeRootfsImageArtifact(2, false, false)
+	// have a key but artifact is v1
+	art, err = MakeRootfsImageArtifact(1, false, false)
```
kacf: Testing v1 is all good, but shouldn't we keep the original case with v2?

In general I think we should only *add* tests for v3, not remove tests for older types, at least not until [MEN-2156](https://tracker.mender.io/browse/MEN-2156) is on the table.

oleorhagen: I'm sorry, should've commented on this. The test was simply removed because the exact same lines were present in TestInsallNoSignature. Do you want that fix in a seperate PR? @kacf

kacf: That would be great, yes!

oleorhagen: #326

kacf: Thanks!

kacf:
Should we use this also as a feature branch?

PR #323, opened 2018-10-02
Detect if inactive part is mounted and unmount

alfrunes:
This implementation parses /proc/self/mounts if the given inactive partition
is mounted, and if so unmounts the partition before initiating update.

backlog task ID: MEN-2084
changelog: title


mirzak:
I am not completely convinced that we should resolve this for the user and I am leaning more towards just failing the update straight away if we detect that it is mounted.

This is an "abnormal" state and we should give user opportunity to fix it.

But if we want to move forward with the current approach we should at least raise some warnings, before  performing the "umount".

kacf:
Hmm tricky. I think I would argue that unmounting is ok in this case, for a couple of reasons:

* The filesystem is more or less a duplicate of the root filesystem we're currently on, so it's very unlikely that the system has any use for it.
* The filesystem has been explicitly listed in `mender.conf` as the second root filesystem, so it's expected that it's messed with.

@mirzak: Or do you still think it's too magical? We could also advise people to use a state script instead, but it feels like one of those cases where where 99.9% would need that, and we just annoy everyone by requiring it.

device.go
```diff
@@ -86,6 +108,16 @@ func (d *device) InstallUpdate(image io.ReadCloser, size int64) error {
 		return err
 	}
 
+	// Make sure the file system is not mounted (MEN-2084)
+	if checkMounted(inactivePartition) != "" {
+		err = syscall.Unmount(inactivePartition, syscall.MNT_FORCE)
```
kacf: We should not use `MNT_FORCE`, but rather a gentle unmount and fail if that doesn't work. Force-unmounting could mean that some process actually has open files there, in which case we should definitely fail instead.

alfrunes:
Added a warning before unmounting removed the `MNT_FORCE` flag.

mirzak:
> @mirzak: Or do you still think it's too magical? We could also advise people to use a state script instead, but it feels like one of those cases where where 99.9% would need that, and we just annoy everyone by requiring it.

I think it is OK.

But definitely should be more verbose about it (warnings), something like:

- We have detected that the inactive part is mounted! This usually means that there is some "auto-mount" service (e.g udisks2) that mounts all available block devices. It is recommend to blacklist the partitions used by Mender to avoid any issues.

- Will attempt to "umount" and proceeding with update.

kacf:
Agreed. Can you please add that as well, @alfrunes?

alfrunes:
@kacf: Done.

PR #325, opened 2018-10-08
FIX: Do not enter infinite loop when inventory has not been sent.

oleorhagen:
Changelog: None


oleorhagen:
@mirzak yes

oleorhagen:
I will reply to the list in just a second.

oleorhagen:
Do not merge this one yet, as it does not seem to work!

oleorhagen:
This will not be the right approach for fixing this I'm afraid this will be a little bit harder, especially in the case where the client is not authorized beforehand. The problem is that the client in that case never will enter the state in a `enter` transition, and thus will simply go straight to error state, and no waiting will be done.

PR #327, opened 2018-10-15
Allow rootfsPartA and rootfsPartB to be symlinks

mdavis777:
Changelog: Title

Signed-off-by: Michael Davis <michael.davis@essvote.com>

### DIFF
--- a/partitions.go
+++ b/partitions.go
@@ -18,6 +18,7 @@ import (
 	"path"
 	"strings"
 	"syscall"
+	"path/filepath"
 
 	"github.com/mendersoftware/log"
 	"github.com/pkg/errors"
@@ -69,9 +70,9 @@ func (p *partitions) getAndCacheInactivePartition() (string, error) {
 		return "", err
 	}
 
-	if active == p.rootfsPartA {
+	if resolveLink(active) == p.rootfsPartA {
 		p.inactive = p.rootfsPartB
-	} else if active == p.rootfsPartB {
+	} else if resolveLink(active) == p.rootfsPartB {
 		p.inactive = p.rootfsPartA
 	} else {
 		return "", ErrorPartitionNoMatchActive
@@ -228,4 +229,15 @@ func getBootEnvActivePartition(env BootEnvReadWriter) (string, error) {
 
 func checkBootEnvAndRootPartitionMatch(bootPartNum string, rootPart string) bool {
 	return strings.HasSuffix(rootPart, bootPartNum)
+}
+
+func resolveLink(path string) string {
+	// If the supplied path is not a link the original path is returned
+	resolvedPath, err := filepath.EvalSymlinks(path)
+	// This would only happen if supplied a link that goes nowhere or creates a loop
+	if err != nil {
+		log.Warnf("Could not resolve path link: %s Attempting to continue", path)
+		return path
+	}
+	return (resolvedPath)
 }
--- a/device.go
+++ b/device.go
@@ -66,8 +66,8 @@ func NewDevice(env BootEnvReadWriter, sc StatCommander, config deviceConfig) *de
 	partitions := partitions{
 		StatCommander:     sc,
 		BootEnvReadWriter: env,
-		rootfsPartA:       config.rootfsPartA,
-		rootfsPartB:       config.rootfsPartB,
+		rootfsPartA:       resolveLink(config.rootfsPartA),
+		rootfsPartB:       resolveLink(config.rootfsPartB),
 		active:            "",
 		inactive:          "",
 	}


mdavis777:
Just a little information on this.  
Linux implements Labels / UUID / PartUUID as links to /dev/sdX#.
So all that is really needed to support the use of those paths is to resolve it back into /dev/sdX# and continue on as normal.

kacf:
But why even do that? Can't we just open the link directly?

mdavis777:
Sadly no. It fails on some of the error checking and can't find root. This seemed an easier more global solution than going to rewrite the checking with a new case.

kacf:
Ok, it may be that we need it then, but let's resolve https://github.com/mendersoftware/meta-mender/pull/635 first, and we can recheck afterwards.

kacf:
Although I'm not against this change, I really would like to understand better why it's needed, after our discussion in https://github.com/mendersoftware/meta-mender/pull/635. The reason is that I'm a bit worried about partition detection not working correctly if the link is resolved instead of keeping the original name. Any logs or examples of failures would be helpful!

mdavis777:
I will retest with the final  [mendersoftware/meta-mender#635](https://github.com/mendersoftware/meta-mender/pull/635) and pull the logs

mdavis777:
That is the first error it hits.  It may end up hitting more down the road.

mdavis777:
I updated the request to handle either of them being links.
I tried to put both resolves in getAndCacheInactivePartition, unfortunately it fails on the check for setting the inactive partition later on.

partitions.go
```diff
@@ -69,9 +70,9 @@ func (p *partitions) getAndCacheInactivePartition() (string, error) {
 		return "", err
 	}
 
-	if active == p.rootfsPartA {
+	if resolveLink(active) == p.rootfsPartA {
```
mdavis777: resolveLink could just as easily be called in getRootCandidateFromMount.

kacf: True, but I think it's a bit clearer if they stay here.

Besides, to address your comment about different systems returning different results for the mount device, could we not simply check both here? Aka:

```
if active == p.rootfsPartA || resolveLink(active) == p.rootfsPartA {
```

kacf:
> It is failing here because "active" comes back as /dev/sda2 and it is comparing it to p.rootfsPartA (/dev/disk/by-partuuid/.,..)
> 
> This is why resolving the link fixes it because it resolves /dev/disk/by-partuuid/.,. back into /dev/sda2

Thanks for checking, that's exactly what I wanted to know!

partitions.go
mdavis777: I originally had this but it doesn't add any value.  All four possible combinations of links and paths are currently handled.  Basically if either active or  p.rootfsPartX are links they are reduced down to /dev/sdX. All comparisons are done at this base level.

mdavis777:
Anything else I can improve upon?

partitions.go
kacf: Ah right, I see it now.

kacf:
No, sorry, I was away last week. I think this looks good, I will start testing now.

mdavis777:
Does it look like this will be able to make it into 1.7?

kacf:
I think so. The beta won't have it, since I've been having some trouble with testing this (still not sure if the code is to blame or our tests are just acting up). But I consider the handling of symlinks more of a bugfix, so this can also be added to 1.7 after the beta. WIP at https://github.com/mendersoftware/meta-mender/pull/641

kacf:
Yep, known issue, we've already fixed this in latest master.

mirzak:
This change actually causes the same problem it fixes for partuuid when used together with LVM volumes. LVM volumes are links to /dev/dm-X devices

```
$ ls /dev/mender/rootfs1 -alh
lrwxrwxrwx. 1 root root 7 Dec 18 23:00 /dev/mender/rootfs1 -> ../dm-0
```

And the stable part in above is `/dev/mender/rootfs1`, and this is the information I want to have in my mender.conf file. But as it is now it resolves the link, which results in that it sets boot candidate based on the dm-X device which obviously goes wrong here. Also there is no guarantees that you will get the same dm-X number, as it depends on what other LVM volumes you currently have in your system.

This leans towards having an option, "-resolv-links" ?

mdavis777:
I guess I am a little confused.  As long as /dev/mender/rootfs1 the link in the mount list and the link in the conf resolves to the same place it should work.  

I am not sure about the logic that sents the boot candidate as I never touched that part.  I guess I would start by looking if that can be expanded

kacf:
I think neither the `-resolv-links` option nor the path specific exceptions are good solutions. The former puts the burden on the user, and the latter seems hacky and error prone.

Does it help to use `resolvLink()` only when doing comparisons, and leaving the original definition in the struct unchanged?

mirzak:
Some additional context,

```
INFO[0000] Start updating from local image file: [mender-20000000.mender]  module=rootfs
DEBU[0000] Fetching update from file results: [&{0xc000059620}], 543932928, <nil>  module=rootfs
Installing update from the artifact of size 543932928
INFO[0000] no public key was provided for authenticating the artifact  module=installer
DEBU[0000] checking if device [device] is on compatibile device list: [device]
  module=installer
DEBU[0000] installing update mender-20000000.fs.img of size 4294967296  module=installer
DEBU[0000] Trying to install update of size: 4294967296  module=device
DEBU[0000] Active partition: /dev/mapper/mender-rootfs1  module=partitions
DEBU[0000] Detected inactive partition /dev/dm-1, based on active partition /dev/mapper/mender-rootfs1  module=partitions
INFO[0000] opening device /dev/dm-1 for writing          module=block_device
INFO[0000] partition /dev/dm-1 size: 4294967296          module=block_device
```

Something that is also odd is that there seems to be a mix of resolved/not-resolved,
e.g:

```
DEBU[0000] Active partition: /dev/mapper/mender-rootfs1  module=partitions
DEBU[0000] Detected inactive partition /dev/dm-1, based on active partition /dev/mapper/mender-rootfs1
```

Below the problem is visible, it sets boot candidate to `1` based on `/dev/dm-1`.

```
................................  99% 530432 KiB
.......................INFO[0295] wrote 4294967296/4294967296 bytes of update to device /dev/dm-1  module=device
          100% 531184 KiB
DEBU[0295] installer: successfully read artifact [name: 20000000; version: 2; compatible devices: [device]]  module=installer
DEBU[0295] Inactive partition: /dev/dm-1                 module=partitions
DEBU[0295] Marking inactive partition (/dev/dm-1) as the new boot candidate.  module=device
INFO[0295] Enabling partition with new image installed to be a boot candidate: 1  module=device
DEBU[0295] Have U-Boot variable: mender_check_saveenv_canary=  module=bootenv
DEBU[0295] List of U-Boot variables:map[mender_check_saveenv_canary:]  module=bootenv
DEBU[0295] Marking inactive partition as a boot candidate successful.  module=device
```

And the links.

```
$ ls -alh /dev/mapper/mender-rootfs*
lrwxrwxrwx. 1 root root 7 Dec 19 08:58 /dev/mapper/mender-rootfs1 -> ../dm-0
lrwxrwxrwx. 1 root root 7 Dec 19 09:05 /dev/mapper/mender-rootfs2 -> ../dm-1
```

mdavis777:
>Does it help to use resolvLink() only when doing comparisons, and leaving the original definition in the struct unchanged?
No that would just shift the issue back to partuuid. It would use the last part of the uuid as the new part number. 

I don't know that there is really a way except for handling lvms as a special case. LVM and maybe raid the only ones I can think of that might no want to be resolved.  In either case detecting them by /dev/dm* and /dev/md* are the easiest most reliable way I can think of. I am sure you could find some application to system call and parse, but I would bet they are doing the same thing.

mdavis777:
That or using the information from the grub(hopefully ubi has it too) and mender conf files and not parsing it based on the /dev path at all.

kacf:
@mdavis777: Yes, I see what you mean.

I still think we should avoid the extra `-resolv-links` option, but maybe we can make a variation of your second suggestion: choosing to dereference based on which path it is. However, I would suggest that we make it specific to `/dev/disk/by-partuuid` instead of `/dev/mapper`, because `/dev/mapper` is so generic, and we don't know what other issues we can cause by flipping this back and forth. So far I believe PARTUUID is the only one that needs to be dereferenced. Does it sound sensible?

mirzak:
Sensible to me.

mdavis777:
/dev/disk/by-partuuid is the only one that needs to be resolved now, but I could see future uses for by-uuid and by-label.  I suppose those can be dealt with if / when they are supported.  That or resolve anything that comes from /dev/disk/*.

mdavis777:
@kacf: I know they aren't stable now that's why I said future use. Reinstating the label on update would be a good future addition that might make it usable. I agree fixing the resolve if/when that happens is probably the best plan.

kacf:
I created [MEN-2302](https://tracker.mender.io/browse/MEN-2302).

PR #330, opened 2018-10-22
Added first draft documentation for update modules.

kacf:
Changelog: None


Documentation/update-modules-v3-file-api.md
```diff
@@ -0,0 +1,227 @@
...
+Update modules are executables that are placed in `/usr/lib/mender/modules/v3`
+directory, where `v3` is a reference to the version of the protocol. Mender will
+look in the directory with the same version as the version of the artifact being
+processed.
+
+
+States and execution flow
+-------------------------
+
+Update modules are modelled after the same execution flow as state scripts, and
```
estenberg: Link state scripts to https://docs.mender.io/artifacts/state-scripts ?

Documentation/update-modules-v3-file-api.md
```diff
@@ -0,0 +1,227 @@
...
+* `ArtifactRollback` has executed
+
+`ArtifactFailure` executes whenever:
+
+* Either of `ArtifactInstall`, `ArtifactReboot` or `ArtifactCommit` has
+  failed
+* Executes after `ArtifactRollback` and `ArtifactRollbackReboot`, if they
+  execute at all
+
+`Cleanup` executes unconditionally at the end of all the other states,
```
estenberg: Is it possible to detect which state we are cleaning up? Otherwise it would seem hard to know what to clean up?

Documentation/update-modules-v3-file-api.md
```diff
@@ -0,0 +1,227 @@
...
+
+The `header-augment` directory functions exactly as the `header` directory, but
+is taken from the `header-augment.tar.gz` file from the artifact.
+
+**Warning:** Be very careful with using contents from `header-augment` because
+it contains unsigned data. Unless you specifically need unsigned data in your
+update module (for example for a binary diff that depends on the device it is
+targeted against), it is advised not to use the `header-augment` directory.
+
+### `tmp`
```
estenberg: :+1:

Documentation/update-modules-v3-file-api.md
```diff
@@ -0,0 +1,227 @@
+Update modules v3 protocol
+==========================
+
+Update modules are executables that are placed in `/usr/lib/mender/modules/v3`
+directory, where `v3` is a reference to the version of the protocol. Mender will
```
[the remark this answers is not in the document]
kacf: Yep, i'll clarify that!

Documentation/update-modules-v3-file-api.md
```diff
@@ -0,0 +1,227 @@
...
+look in the directory with the same version as the version of the artifact being
```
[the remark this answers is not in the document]
kacf: Oh interesting, there was a whole range of things here I didn't know. Will fix!

Documentation/update-modules-v3-file-api.md
kacf: Yep!

Documentation/update-modules-v3-file-api.md
kacf: No, in general modules are not informed of which state they came from, which is something that's generally discouraged in state design. But the idea with `Cleanup` is that the modules would only clean up temporary files they've used (for example `rm -rf /var/tmp/modem_firmware_blobs`). "Cleanup" as in "cleaning up a failed update" is not what's intended here, that should rather happen in ArtifactRollback. Does that make sense?

I can try to make it a bit clearer that it's temporary files that should be cleaned up here, not failed updates.

kacf:
Updated.

Documentation/update-modules-v3-file-api.md
```diff
@@ -0,0 +1,234 @@
+Update modules v3 protocol
+==========================
+
+Update modules are executables that are placed in `/usr/lib/mender/modules/v3`
```
cosinekitty: This is a great start. It would be nice to include more explanation about the purpose of the scripts and why somebody would want to create them. For example, when/why does the `Download` script run? Does it provide an alternative way to perform the download, or does it run before or after the existing Mender download? The reader first needs to understand the purpose to decide how these scripts might be relevant to their needs. Then they can proceed to the technical details.

estenberg: Agreed! I think the purpose of this doc is more architecture/technical design, though. So when it comes to actually writing modules we would need separate tutorials, potentially with references to this document.

Documentation/update-modules-v3-file-api.md
estenberg: Ah, yes. Makes sense!

kacf:
I added a new commit now where I completely separated augmented artifacts from non-augmented ones, to make sure that the security implications are well understood by those who choose to use them.

Documentation/update-modules-v3-file-api.md
```diff
@@ -0,0 +1,422 @@
...
+augmented content if your update module is prepared to handle untrusted input
+and still guarantee a trusted result (more about best practices below).
+
+### Marking augmented artifacts as supported
+
+If you wish to accept `augment` files you need to implement this calling
+interface:
+
+```bash
+./update-module SupportsAugmentedArtifacts
```
estenberg: I wonder if this is potentially dangerous or could lead to unexpected behavior for update modules that don't support augmented..?
If update-module only supports "install", for example and assumes it would only be called with that, could it install / make some system changes that were not intended?

Maybe a bit far-fetched example as it should be able to support other stages, but just a bit concerned what could happen if the module is not implemented "safely" and then it gets called (as root) with an argument it doesn't support..

I don't really have a good proposal to detect this in run-time, it feels more like it should be part of some "manifest" or "registry" of update modules. Right now it could just state the update module and whether it supports augments or not, in the future it can be extended to check for rollback capability, reboot behavior, etc.

Documentation/update-modules-v3-file-api.md
```diff
@@ -0,0 +1,422 @@
...
+context it needs to install the augmented artifact.
+
+To permit specific header fields to be overridden, Mender calls the update
+module like this:
+
+```bash
+./update-module PermittedAugmentedHeaders
+```
+
+The module is expected to return a JSON structure like this:
```
estenberg: JSON might be a bit tricky in shell/C, as we know.. maybe relates to my "manifest" note above.

Documentation/update-modules-v3-file-api.md
kacf: Handling unknown arguments is [a required part of the spec](https://github.com/mendersoftware/mender/pull/330/files#diff-1122e6d8a5474b1546480d6943246a7aR192) and is important for being able to extend update modules in the future. Besides, all update modules will be called with a minimum of four different arguments: Download, ArtifactInstall, ArtifactCommit and Cleanup. So they need to be prepared for that, even if they only implement one of them.

Documentation/update-modules-v3-file-api.md
oleorhagen: Wasn't the file api kind of an effort to be free from JSON?

kacf: I think JSON is sort of unavoidable. The problem is that all the headers are JSON too, and some are free form, so the structure isn't necessarily known in advance. Therefore it's hard to represent in a flat model.

I think it's ok to use JSON here, given that this is advanced usage. Most update modules are not expected to deal with `PermittedAugmentedHeaders`, or even most of the headers that are stored in the file API. A package installer or app deployment module doesn't need any of this, for example. Besides, this is just printing, but if you return anything here, you will have to parse JSON as well, because that's what you will get in the artifact. Mender can't help you there, because it only knows its own headers.

Documentation/update-modules-v3-file-api.md
kacf: > I don't really have a good proposal to detect this in run-time, it feels more like it should be part of some "manifest" or "registry" of update modules. Right now it could just state the update module and whether it supports augments or not, in the future it can be extended to check for rollback capability, reboot behavior, etc.

One reason for not using a manifest: When doing package installs, you often don't know whether you need a reboot until after the package is installed. For example, kernel updates will set a flag, which you can see in your UI as "your machine needs to reboot". So I think this should be dynamic, with Mender asking after the install stage: "Do you need to reboot now?"

kacf:
Added reboot and rollback API, two update modules, and marked clearly which sections will actually be implemented in iteration 1.

Documentation/update-modules-v3-file-api.md
estenberg: OK! If you think it's safe then let's keep it that way. The simpler, the better. :)

support/modules/pkg-install
```diff
@@ -0,0 +1,26 @@
...
+        RPMS=
+        DEBS=
+        for file in "$FILES"/files/*; do
+            case "$file" in
+                *.rpm)
+                    RPMS="$RPMS $file"
+                    ;;
+                *.deb)
+                    DEBS="$DEBS $file"
+                    ;;
```
estenberg: Nice and short - and could be shorter! :)
From my knowledge I think it is pretty rare to want to deploy both .deb and .rpm. You typically pick one format and stick with that, and I think deb is more common.

So I would simply rename to deb-install and remove the rpm parts (we can create another for that later).

mirzak: Just a FYI. 

From a Yocto perspective "rpm" and "ipk" are the supported package types that are part of the Yocto test suite. So in this context I would pick one of these over "deb".

kacf: Since both appear to be important candidates, I split the module into one for each type, `deb-install` and `rpm-install`.

kacf:
Updated with some new modifications after discovering things during implementation:

```
commit 1c5b04f5d3b15b265f8ba892c3df1ebfb6a35d81
Date:   Wed Jan 2 14:17:51 2019

    module docs: Fix design flaw in streaming API regarding `streams-list`
    
    We don't know the order of the streams upfront, so we cannot provide
    the `streams-list` pipe originally envisioned in the spec. Instead,
    provide a `stream-next` file which returns exactly one entry, the next
    one, which is then meant to be repeatedly read until it returns empty.
    
    Also, provide a few more convenience entries in the file API, and
    clarify that the file API is subject to more entries in the future, as
    part of API extensions.
    
    Changelog: None
    
commit 1a4cf80b0a002125b734296dc9635ad4429e768f
Date:   Wed Jan 2 14:24:43 2019

    Remove `files` from update modules spec file tree.
    
    It has been removed from the v3 spec already.
    
    Changelog: None

commit 44913db387c3976d21e4e29a2ff6d28694677335
Date:   Wed Jan 2 14:48:47 2019

    Module docs: Improve behavior of ArtifactReboot state.
    
    Changelog: None
    
```

Documentation/update-modules-v3-file-api.md
```diff
@@ -0,0 +1,665 @@
...
+In general, augmented payloads typically mean that the payload must be
+transformed into something (binary delta into full binary image, encrypted image
+into decrypted image, etc), and it is typically the result of that
+transformation that needs to be verified with a checksum.
+
+
+Future possibilities
+====================
+
+Her are described some things that have not been planned in detail, but that may
```
lluiscampos: Typo: Here

kacf: Fixed locally.

Documentation/update-modules-v3-file-api.md
```diff
@@ -0,0 +1,665 @@
...
+    makes a backup of the current system. Depending on the size of the
+    filesystem, this could be much more time consuming than the update itself
+
+`ArtifactRollback` then executes whenever:
+
+* the `SupportsRollback` call has returned a non-`No` response
+  * For `AutomaticDualRootfs`, only Mender's internal variants are called
+* `ArtifactInstall` has executed successfully
+* `ArtifactReboot`, `ArtifactVerifyReboot` or `ArtifactCommit` fails
+
```
lluiscampos: I think this documentation requires and update. ArtifactRollback is being called also when ArtifactInstall fails.

kacf: Thanks, also fixed locally.

support/modules/rpm-install
```diff
@@ -0,0 +1,13 @@
+#!/bin/sh
+
+set -e
+
+STATE="$1"
+FILES="$2"
+
+case "$STATE" in
+    ArtifactInstall)
+        rpm -U "$FILES"/files/*.deb
```
jeremytrimble: above should probably be `*.rpm` not `*.deb`

kacf: Good eye! Fixed locally.

kacf:
Superseded by #358.

PR #336, opened 2018-12-03
MEN-2131: Provide command line interface to force inventory update. 

oleorhagen:
Changelog: Title

### DIFF
--- a/daemon.go
+++ b/daemon.go
@@ -14,8 +14,6 @@
 package main
 
 import (
-	"os"
-
 	"github.com/mendersoftware/log"
 	"github.com/mendersoftware/mender/datastore"
 	"github.com/mendersoftware/mender/store"
@@ -26,11 +24,11 @@ import (
 // Config section
 
 type menderDaemon struct {
-	mender      Controller
-	stop        bool
-	sctx        StateContext
-	store       store.Store
-	updateCheck chan bool // state-machine interrupt.
+	mender       Controller
+	stop         bool
+	sctx         StateContext
+	store        store.Store
+	forceToState chan State
 }
 
 func NewDaemon(mender Controller, store store.Store) *menderDaemon {
@@ -38,11 +36,12 @@ func NewDaemon(mender Controller, store store.Store) *menderDaemon {
 	daemon := menderDaemon{
 		mender: mender,
 		sctx: StateContext{
-			store:    store,
-			rebooter: system.NewSystemRebootCmd(system.OsCalls{}),
+			store:      store,
+			rebooter:   system.NewSystemRebootCmd(system.OsCalls{}),
+			wakeupChan: make(chan bool, 1),
 		},
-		store:       store,
-		updateCheck: make(chan bool, 1),
+		store:        store,
+		forceToState: make(chan State, 1),
 	}
 	return &daemon
 }
@@ -69,21 +68,15 @@ func (d *menderDaemon) Run() error {
 	var toState State = d.mender.GetCurrentState()
 	cancelled := false
 	for {
-		// If signal SIGUSR1 received, check for update straight away.
+		// If signal SIGUSR1 or SIGUSR2 is received, force the state-machine to the correct state.
 		select {
-		case sig := <-d.updateCheck:
-			log.Debugf("received signal: %t", sig)
-			_, err := LoadStateData(d.sctx.store)
-			// No previous state stored, means no update was in progress,
-			// and we can safely force an update check.
-			if err != nil && os.IsNotExist(err) {
-				if d.mender.IsAuthorized() {
-					toState = updateCheckState
-				} else {
-					toState = initState
-				}
-				d.mender.SetNextState(toState)
+		case nState := <-d.forceToState:
+			switch toState.(type) {
+			case *IdleState, *CheckWaitState:
+				log.Infof("Forcing state machine to: %s", nState)
+				toState = nState
 			}
+
 		default:
 			// Identity op - do nothing.
 		}
--- a/daemon_test.go
+++ b/daemon_test.go
@@ -217,7 +217,7 @@ func TestDaemonRun(t *testing.T) {
 
 		}
 
-		pollInterval := time.Duration(10) * time.Millisecond
+		pollInterval := time.Duration(30) * time.Second
 
 		dtc := &daemonTestController{
 			stateTestController{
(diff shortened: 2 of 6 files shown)


oleorhagen:
Integration: https://github.com/mendersoftware/integration/pull/553

estenberg:
Nice! Just on the naming side, I think "inventory update" may be a bit confusing:
* we are not changing anything on the device (the "update" is really on the server)
* we do not do an OTA update (as opposed to check-update)

There is some description here, that could help with CLI option naming: https://docs.mender.io/1.6/client-configuration/configuration-file/polling-intervals

What about "send-inventory"?

main.go
```diff
@@ -533,6 +545,14 @@ func runDaemon(d *menderDaemon) error {
...
 					ws.Wake()
 				}
+			} else if s == syscall.SIGUSR2 {
+				log.Debug("SIGUSR2 signal received.")
+				d.inventoryUpdate <- true
+				// If the state machine is in a wait state - force a wake-up.
+				ws, ok := d.mender.GetCurrentState().(WaitState)
+				if ok {
+					ws.Wake()
+				}
```
kacf: I have to admit I don't follow all the state transition machinery, but is there a potential race condition here? This is a separate go routine, so can it enter the wait state between `if ok` and the previous statement? If it happened it would mean that the wait state would be detected as "not waiting", but it would in fact be waiting.

oleorhagen: Yes, there would be. But the window would be on the order of milliseconds, so do you really think it is necessesary to handle? The worst that can happen here as far as I can see is that you have to rerun the command with a probability of 1/10e5 or less. My reasoning:
1. State-machine enters wait-state in between the check and the if statement -> no harm done. Simply rerun.
2. The state machine leaves the wait-state just in between. A wake call will be sent, but the wait-state will never be consumed by the state-machine, as it creates a new wait-state upon the next run.

kacf: > But the window would be on the order of milliseconds, so do you really think it is necessesary to handle?

I'm afraid I do. :-) We've already had quite a few encounters with "things that should not happen", because we have used emulated devices that slow things down. On a single core embedded device which is under load, milliseconds can easily turn into seconds, and we've seen this in practice.

Luckily, I think the fix here is quite easy: Just call `Wake()` unconditionally. There is a chance that it will be stuck in the `Wake()` call for a little bit if the client is not in a wait state, but I reckon this is fine, because:

1. This is a go routine, so it doesn't hold up anything else.
2. The client is guaranteed to return to a wait state eventually.
3. Calling `Wake()` one time too many will simply spin the loop one extra time, but won't actually do anything.

Do you agree?

oleorhagen:
@kacf, I have now added syncronization to the state-machine logic. See if you agree.

In essence it is locking the state-machine during transitions, and then only allowing manipulation of the state-machine when there is no change in the state-machine. 
This solves (as far as I can see all the troublesome corner-cases. 
1) The state-machine can no longer enter a sleep state, just as the signal is consumed. 
2) If the state-machine leaves a sleep-state before a wake-call is sent, the garbage-collector will simply take care of the remaining clean-up.

Does this make sense?

main.go
oleorhagen: > > But the window would be on the order of milliseconds, so do you really think it is necessesary to handle?
> 
> I'm afraid I do. :-) We've already had quite a few encounters with "things that should not happen", because we have used emulated devices that slow things down. On a single core embedded device which is under load, milliseconds can easily turn into seconds, and we've seen this in practice.
> 
> Luckily, I think the fix here is quite easy: Just call `Wake()` unconditionally. There is a chance that it will be stuck in the `Wake()` call for a little bit if the client is not in a wait state, but I reckon this is fine, because:
> 
> 1. This is a go routine, so it doesn't hold up anything else.
> 2. The client is guaranteed to return to a wait state eventually.
> 3. Calling `Wake()` one time too many will simply spin the loop one extra time, but won't actually do anything.
> 
> Do you agree?

I do like this approach, and as added weight to your argument, I dislike mine.

However, and please tell me I am wrong here:

Won't the problem be that not all states have a wait method, and as such:

* The user tries to send a wake-up call, but the mender-client have not yet entered a wait-state, which means that the cast will fail, and no wake call is ever sent, hence we cannot send it one times too many. Just as this is happening, the client enters a wait-state, and the window of opportunity for weirdness to happen is still here?

Please correct me if I am wrong here. I would love not to merge this.

kacf: Darn, you're absolutely right. Well spotted. This is surprisingly complicated.

New suggestion, which is perhaps much simpler: Make a dedicated signal channel which is global, and listen to that one as well as the wakeup channel. Then post the wakeup call on this one instead. It will always exist, so there is no race condition, and the next WaitState is guaranteed to act on it. Global variables are not great, but I think it can be justified in this case since signals are also global, and it also means that you can get rid of the mutex, which would be really nice to avoid.

What do you think?

daemon.go
```diff
@@ -64,28 +63,24 @@ func (d *menderDaemon) shouldStop() bool {
...
-				log.Debugf("received signal: %t. Forcing device from initState to updateCheck\n", toState)
-				if d.mender.IsAuthorized() {
-					toState = updateCheckState
-				} else {
-					toState = initState
-				}
-				d.mender.SetNextState(toState)
+		case nState := <-d.forceToState:
+			switch toState.(type) {
+			case *IdleState, *CheckWaitState:
```
oleorhagen: As far as I understand, there will only be two states where we want to force the next state in the client.
1) From idle - The base case.
2) If we are in either a update-check- or inventory-update-wait state, then upon the state-machine getting woke, it will return checkWait state, which also should be forced, as in this case we do not care about the polling intervals. Alas, the next state should be forced.

All other states are left untouched.

daemon.go
```diff
@@ -64,28 +63,24 @@ func (d *menderDaemon) shouldStop() bool {
...
+				toState = nState
 			}
+
 		default:
 			// Identity op - do nothing.
 		}
-		toState, cancelled = d.mender.TransitionState(toState, &d.sctx)
+
+		currentState = d.mender.TransitionState(toState, &d.sctx)
+		toState, cancelled = currentState.Handle(&d.sctx, d.mender)
```
oleorhagen: Moved the handling of the state out of the transition function. In my head this makes more sense. Tell me if you disagree.

state.go
```diff
@@ -140,15 +140,16 @@ type StateContext struct {
 	lastInventoryUpdateAttempt time.Time
 	lastAuthorizeAttempt       time.Time
 	fetchInstallAttempts       int
+	wakeupChan                 chan bool
```
oleorhagen: The wakeup channel is now global, and passed into the wait functions by reference, which should resolve the synchronization issues we've been facing.

state.go
```diff
@@ -282,15 +283,16 @@ func NewWaitState(id MenderState, t Transition) WaitState {
 	return &waitState{
 		baseState: baseState{id: id, t: t},
 		cancel:    make(chan bool),
-		wakeup:    make(chan bool),
+		// wakeup: is a global channel shared between all states, defined in TODO
```
oleorhagen: Will resolve if you agree that stateContext is the best place to keep the global wakeup channel.

oleorhagen:
@lluiscampos I would appreciate your prying eyes on this one :)

lluiscampos:
I'll look at it tomorrow :+1: 

@kacf, I can take over this PR to unload you a bit ;)

kacf:
> I can take over this PR to unload you a bit ;)

:+1:

daemon.go
kacf: I kind of agree, but the update_modules branch, and in particular the big state test, relies pretty heavily on the old behavior. Maybe we can save this for another day?

daemon.go
kacf: Agree, I like this method much better!

daemon_test.go
```diff
@@ -179,25 +179,38 @@ func TestDaemonRun(t *testing.T) {
...
+				inventPollIntvl: pollInterval,
+				state:           idleState,
+			},
+			0,
+		}
+		daemon := NewDaemon(dtc, store.NewMemStore())
+		dtc.authorized = true
+		daemon.StopDaemon()                                         // Stop after a single pass.
+		go func() { daemon.forceToState <- inventoryUpdateState }() // Force inventoryUpdate state.
+		time.Sleep(time.Millisecond * 1)                            // Make sure the signal has been sent.
```
kacf: Propose using *much* longer sleeps here. This is a typical thing that will fail spuriously in Travis. Maybe make the pollInterval 30 seconds, and this one 1 second?

state.go
kacf: :+1:

state.go
kacf: Yep, that's even better than my original suggestion to make it truly global.

olepor:
@kacf Spurious, should be all good to go (finally). If ur happy ofc

oleorhagen:
@kacf Looks spurious to me, however these tests are new to me.

lluiscampos:
Please review the commit message, I don't think we need all that in the Changelog. And also remove the extra signature

main.go
```diff
@@ -103,6 +104,8 @@ func argsParse(args []string) (runOptionsType, error) {
 
 	updateCheck := parsing.Bool("check-update", false, "force update check")
 
+	updateInventory := parsing.Bool("send-inventory", false, "force inventory update state")
```
lluiscampos: Maybe "state" can be removed from the help string, so that it is consistent with "force update check"

main.go
```diff
@@ -163,6 +167,10 @@ func argsParse(args []string) (runOptionsType, error) {
 		return runOptions, nil
 	}
 
+	if *updateInventory {
+		return runOptions, nil
```
lluiscampos: Please update comment in line 161

main.go
```diff
@@ -412,10 +420,13 @@ func doMain(args []string) error {
...
 		return err
 	}
-	// Do not run anything else if update-check is triggered.
+	// Do not run anything else if update-check or inventory-update is triggered.
 	if *runOptions.updateCheck {
 		return updateCheck(exec.Command("kill", "-USR1"), exec.Command("systemctl", "show", "-p", "MainPID", "mender"))
 	}
+	if *runOptions.updateInventory {
+		return updateCheck(exec.Command("kill", "-USR2"), exec.Command("systemctl", "show", "-p", "MainPID", "mender"))
+	}
```
lluiscampos: Is it a valid behavior to specify both `check-update` and `send-inventory`?

oleorhagen: Good eyes :eyes:

oleorhagen: And that is a no.

oleorhagen: Fixed in https://github.com/mendersoftware/mender/pull/336/commits/fabeaad473da28d76c172493a10b3ef198196b62

oleorhagen:
> Approved.
> 
> I still think it will be nice to split your Changelog paragraph, though.

Changelog is only the Title :)

lluiscampos:
> Changelog is only the Title :)

Great! Sorry I didn't realized.

kacf:
> Looks spurious to me, however these tests are new to me.

It's because it's outdated. I rebased and submitted [here](https://github.com/mendersoftware/mender/pull/392) to save you some time.

oleorhagen:
superseeded by: https://github.com/mendersoftware/mender/pull/392

PR #338, opened 2018-12-13
MEN-2195: Bugfix: State-script error code in Sync-Enter causes infini…

oleorhagen:
…te loop

Changelog: Title

### DIFF
--- a/state.go
+++ b/state.go
@@ -135,10 +135,10 @@ import (
 // StateContext carrying over data that may be used by all state handlers
 type StateContext struct {
 	// data store access
-	store                store.Store
-	lastUpdateCheck      time.Time
-	lastInventoryUpdate  time.Time
-	fetchInstallAttempts int
+	store                      store.Store
+	lastUpdateCheckAttempt     time.Time
+	lastInventoryUpdateAttempt time.Time
+	fetchInstallAttempts       int
 }
 
 type StateRunner interface {
@@ -580,7 +580,6 @@ type UpdateCheckState struct {
 
 func (u *UpdateCheckState) Handle(ctx *StateContext, c Controller) (State, bool) {
 	log.Debugf("handle update check state")
-	ctx.lastUpdateCheck = time.Now()
 
 	update, err := c.CheckUpdate()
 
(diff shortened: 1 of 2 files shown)


state.go
```diff
@@ -835,6 +834,18 @@ func (cw *CheckWaitState) Handle(ctx *StateContext, c Controller) (State, bool)
 		return cw.Wait(next.state, cw, wait)
 	}
 
+	// (MEN-2195): Set the last udpate/inventory check time to now, as an error in an enter script will
+	// hinder these states from ever running, and thus causing an infinite loop if the script
+	// keeps returning the same error.
+	switch (next.state).(type) {
+	case *InventoryUpdateState:
+		log.Info("Inventory update state next!")
```
oleorhagen: Moved the update of the `lastInventoryUpdate` and `lastUpdateCheck' to here, and renamed them, as they are now more semantically like attemps (although they functioned like this previously as well, only the time was set upon reaching the state. Still it was oblivious to whether or not the action was successful or not).

oleorhagen:
This does only fix the bug after the client has been authorized. In fact I know of at least one more just like this, prior to authorization. I don't know if there are more. Maybe this should be tested almost like a fuzzy test of some sort? I guess my question is: Should the scope be expanded for this ticket, or should I contain the fix for this commit to this special instance?
@kacf  @dellgreen

kacf:
Given that the case of "prior to authorization" is likely to be quite common (every new device, basically), I think we should fix that too, so yes, it should be included in the scope, I think.

Still, I think you can merge this PR now, and start working on the other bug separately. This PR is already reviewed, tested and passed. No reason to keep it waiting.

oleorhagen:
You don't want an integration test for this?

kacf:
Oh right. Yes, that would be great! :-)

oleorhagen:
haha, you've ruined me!

state.go
```diff
@@ -835,6 +834,18 @@ func (cw *CheckWaitState) Handle(ctx *StateContext, c Controller) (State, bool)
 		return cw.Wait(next.state, cw, wait)
```
lluiscampos: We need to update `lastInventoryUpdateAttempt` and `lastUpdateCheckAttempt` also inside this if clause.

Otherwise such variables will never be updated in the normal workflow (i.e. when we want to actually wait the interval).

lluiscampos:
Suppressed by #356

PR #341, opened 2018-12-18
CONTRIBUTING: add guidelines on PR structuring

mchalski:
saw some issues in a recent contribution, didn't have a place to point the person to - so adding it now.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -33,8 +33,8 @@ maintainers and relevant project members upfront. Start a discussion via our
 [Google group mailing
 list](https://groups.google.com/a/lists.mender.io/forum/#!forum/mender).
 
-Using commit signoffs and changelog tags is mandatory for all commits, see the
-next sections for details.
+Using commit signoffs and changelog tags is mandatory for all commits; we also
+encourage that each commit is small and cohesive. See the next sections for details.
 
 
 ### Sign your work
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.md
```diff
@@ -174,6 +174,93 @@ if desired.
...
+
+#### Example
+
+* Bad:
+
+```
+Refactor X.
+
+Changelog: None
+```
```
mirzak: Would prefer if the examples where not as abstract.

And the examples are not following the guidelines, as they are missing sign-off lines :)

kacf: I would at least vote for doing one or the other: Either include both changelog and signoff, or omit both, since in this context I think the changelog is also noise. I'm leaning slightly towards including both (people might only glance at one or two examples), but omitting them is fine by me if you want less noise.

I didn't really react to the abstractness. I think it's a pretty understandable workflow.

kacf:
Otherwise big :+1: from me!

CONTRIBUTING.md
mirzak: > as for specific examples, I can provide a backend related one - and this is the client's repo - don't want to confuse folks. could you procure a client-related example?

I would not go that far, to actually point out existing commits and provide them as examples and having separate for client and backend. I believe that we can share them.

PR #342, opened 2018-12-19
MEN-2285: Periodically Sync() when writing image to disk, don't ignore Sync() failures.

jeremytrimble:
This PR supersedes #339.

See: https://tracker.mender.io/browse/MEN-2285

Testing showed that writing to the block device in increments of single sectors (512 bytes on my eMMC device) contributed the prevalence of an OOM condition in the kernel when allocating DMA scatter-gather descriptors.
    
Rather than making a zillion small writes and expecting the kernel to coalesce them for us when we're easily able put this data into already (virtually) contiguous memory in userspace, we write to the block device in "medium-size chunks" (around 1 MiB) that are a power-of-two multiple of the sector size (obviating any concerns about sector misalignment causing reduced performance or undue wear on the underlying storage).

I have verified empirically (watching the `write` syscalls through `strace`) that with these changes the writes are actually going into the kernel with the intended size (1 MiB in my case).

Without the "medium-sized writes" change (even calling `write` in increments of 32768 bytes), I still saw the OOM issue on my kernel.

### DIFF
--- a/chunked_copy_test.go
+++ b/chunked_copy_test.go
@@ -1,4 +1,4 @@
-// Copyright 2018 Northern.tech AS
+// Copyright 2019 Northern.tech AS
 //
 //    Licensed under the Apache License, Version 2.0 (the "License");
 //    you may not use this file except in compliance with the License.
--- a/chunked_copy_test.go
+++ b/chunked_copy_test.go
@@ -1,0 +1,77 @@
+// Copyright 2018 Northern.tech AS
+//
+//    Licensed under the Apache License, Version 2.0 (the "License");
+//    you may not use this file except in compliance with the License.
+//    You may obtain a copy of the License at
+//
+//        http://www.apache.org/licenses/LICENSE-2.0
+//
+//    Unless required by applicable law or agreed to in writing, software
+//    distributed under the License is distributed on an "AS IS" BASIS,
+//    WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+//    See the License for the specific language governing permissions and
+//    limitations under the License.
+package main
+
+import (
+	"bytes"
+	"fmt"
+	"strings"
+	"testing"
+)
+
+type MyWriter struct {
+	writtenChunkSizes []int
+	writtenData       bytes.Buffer
+}
+
+func (mw *MyWriter) Write(data []byte) (n int, err error) {
+	mw.writtenChunkSizes = append(mw.writtenChunkSizes, len(data))
+	n, err = mw.writtenData.Write(data)
+	return
+}
+
+func (mw *MyWriter) GetTotalOutput() []byte {
+	return mw.writtenData.Bytes()
+
+}
+
+func TestChunkedCopy(t *testing.T) {
+
+	CHUNK_SIZE := int64(337)
+
+	bldr := strings.Builder{}
+
+	for i := 0; i < 1000; i++ {
+		bldr.WriteString(fmt.Sprintf("x%07d", i)) // can ignore return value
+	}
+
+	input_bytes := []byte(bldr.String())
+
+	mw := &MyWriter{}
+
+	bytes_written, err := chunkedCopy(mw, bytes.NewBuffer(input_bytes), CHUNK_SIZE) // use a strange chunk size just to prove it works
+
+	// a Copy() routine should return err == nil on EOF.
+	if err != nil {
+		t.Errorf("chunkedCopy: expected err to be nil but got: %v", err)
+	}
+
+	if int(bytes_written) != len(input_bytes) {
+		t.Errorf("chunkedCopy: did not copy full input: expected %v, got %v", len(input_bytes), bytes_written)
+	}
+
+	output_bytes := mw.GetTotalOutput()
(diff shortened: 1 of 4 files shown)


device.go
```diff
@@ -96,6 +97,50 @@ func (d *device) SwapPartitions() error {
...
+			if writeErr != nil {
+				return total_written, writeErr
+			}
+
+			if bytesWritten != bytesRead {
+				panic(fmt.Sprintf(
+					"Unexpected short write: attempted %v bytes but only wrote %v",
+					bytesRead,
+					bytesWritten,
+				))
```
kacf: Should return error here, not panic.

chunked_copy_test.go
```diff
@@ -0,0 +1,77 @@
...
+	}
+
+	output_bytes := mw.GetTotalOutput()
+
+	if !bytes.Equal(output_bytes, input_bytes) {
+		t.Errorf("chunkedCopy: output did not match input")
+	}
+
+	// This is what we really want to check. All of the calls to Write (except the last) must be of size CHUNK_SIZE
+	for i := 0; i < len(mw.writtenChunkSizes)-2; i++ {
```
kacf: If you're testing all but the last entry, shouldn't this be `i < len(mw.writtenChunkSizes)-1`, not `i < len(mw.writtenChunkSizes)-2`? You're testing all but the last two entries here.

jeremytrimble: Good catch -- thanks for reviewing.

device.go
jeremytrimble: Updated `.travis.yml` in subsequent commit.

kacf:
I have submitted a [rebased PR](https://github.com/mendersoftware/mender/pull/344), but I'm having some test issues. They are not related to your PR, so I will take a look at them, but just FYI I won't have time until the end of next week.

jeremytrimble:
Any idea when the rebased PR (#344) will be merged?

kacf:
Sorry, we've had some internal issues that have kept me (and will still keep me) very busy. Don't worry, I *will* get to this, but it might be a couple of weeks out.

kacf:
@jeremytrimble: Now the first and third patches have been merged, since those are obvious improvements. However, we have also [just now](https://github.com/mendersoftware/mender/pull/351) merged a pull request which updates the net/http2 support, which is known to have memory leaks, according to [MEN-2335](https://tracker.mender.io/browse/MEN-2335). Given that the "periodic sync" patch is quite intrusive, performance wise, and should not be necessary unless there is a bug in the storage layer, I would like to see if this improves the situation first. It's after all possible that the OOM condition was due to Mender eating all the memory, not the MMC driver, and the MMC driver was just a victim.

Do you have the chance to try out the latest mender/master and see if the OOM condition has improved? Make sure to pull the latest meta-mender/master as well to get the right checksum (or you can [manually update the checksum](https://github.com/mendersoftware/meta-mender/commit/0bcf6fa26bfe6f45abab57dbc32ad98f2b25cf28)).

oleorhagen:
@jeremytrimble, did you get another look at this? :)

jeremytrimble:
@kacf, @oleorhagen:  No, sorry I haven't had a chance to re-test with the latest master.

Anecdotally, I remember monitoring the mender process' memory consumption before my change and never saw significant memory usage or the memory usage growth trend indicative of a memory leak.  I don't remember the system's available free memory going low enough to be concerning.  I believe the OOM condition was due to a shortage in the "kmalloc" pool within the kernel, which allocates physically-contiguous pages, so it may have been driven more by fragmentation than by overall memory consumption.

I also question whether the "periodic sync" is actually all that intrusive to performance -- do you have data that supports this claim?  In my testing (and I'd guess in most use cases) I believe the bottleneck is always the network throughput and not the speed at which data is written to disk.

Additionally, this "periodic sync" behavior had to be changed in my (forthcoming) updates to support checkpointing (https://tracker.mender.io/browse/MEN-2293), and it is absolutely required there.

It's up to you whether you want to:
1. Accept the change in this PR, knowing that it will eventually be replaced when the checkpoint PR is merged
2. Or to close this PR and just wait for the checkpointing changes.

I believe the most dire issue in MEN-2285 has already been merged in 38d2a79288, so it's up to you.

kacf:
> I remember monitoring the mender process' memory consumption before my change and never saw significant memory usage or the memory usage growth trend indicative of a memory leak.

Ok, I believe this is already enough evidence that the issue was not because of Mender's memory leak (it would have been huge to cause a storage layer OOM condition).

> I also question whether the "periodic sync" is actually all that intrusive to performance -- do you have data that supports this claim? In my testing (and I'd guess in most use cases) I believe the bottleneck is always the network throughput and not the speed at which data is written to disk.

No hard data no, but it would be reasonable to assume that the kernel is better at sync'ing intelligently than a user space program. But since the writes are guaranteed not to be fragmented, there may be little or no difference.

The point is anyway moot, since it sounds like merging the patch is a good idea regardless (even if it is slow, it's better than OOM!). I'll pick this up as soon as I'm done with the update module merge.

kacf:
Superseded by #362.

PR #344, opened 2019-01-04
MEN-2285: Rebased PR for #342.

kacf:


kacf:
One possible spurious error (will probably rerun): https://s3-eu-west-1.amazonaws.com/mender-testing-reports/integration-reports-qemux86-64-uefi-grub/2518/report.html

kacf:
Note that there is still one patch missing from #342. WIP.

PR #345, opened 2019-01-07
MEN-2301: Version files are now allowed to contain a newline character. 

oleorhagen:
Changelog: Version files are now allowed to contain a newline character. Also
some minor changes, as readVersion now accepts an io.Reader, and files are
opened outside of the function. This means that the error message is now
consistent for all the uses of readVersion.

### DIFF
--- a/statescript/executor.go
+++ b/statescript/executor.go
@@ -99,12 +99,25 @@ func (l Launcher) CheckRootfsScriptsVersion() error {
 		return nil
 	}
 
-	ver, err := readVersion(filepath.Join(l.RootfsScriptsPath, "version"))
+	versionFilePath := filepath.Join(l.RootfsScriptsPath, "version")
+	f, err := os.Open(versionFilePath)
 	if err != nil && os.IsNotExist(err) {
-		// no version
-		return errors.New("statescript: missing rootfs scripts version file")
+		errmsg := "statescript: The statescript version file is missing. This file needs to be " +
+			"present and contain a single number representing which version of the statescript " +
+			"support the client is using in order to successfully run statescripts on the client"
+		return errors.New(errmsg)
 	} else if err != nil {
-		return errors.Wrap(err, "statescript: can not read rootfs scripts version")
+		return errors.Wrap(err, "statescript")
+	}
+	ver, err := readVersion(f)
+	if _, ok := err.(readVersionParseError); ok {
+		errmsg := "statescript: Failed to parse the version file in the statescript directory (%s). " +
+			"The file needs to contain a single integer signifying which version of the statescript " +
+			"support which this client is using"
+		return fmt.Errorf(errmsg, err)
+	}
+	if err != nil {
+		return errors.Wrap(err, "statescript")
 	}
 
 	for _, v := range l.SupportedScriptVersions {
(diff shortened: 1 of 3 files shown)


statescript/executor.go
```diff
@@ -88,12 +88,23 @@ func (l Launcher) CheckRootfsScriptsVersion() error {
...
 
-	ver, err := readVersion(filepath.Join(l.RootfsScriptsPath, "version"))
+	versionFilePath := filepath.Join(l.RootfsScriptsPath, "version")
+	f, err := os.Open(versionFilePath)
 	if err != nil && os.IsNotExist(err) {
-		// no version
-		return errors.New("statescript: missing rootfs scripts version file")
+		errmsg := `statescript: The statescript version file is missing. This file needs to be
+present and contain a single number representing which version of the statescript support the client is using
+in order to successfully run statescripts on the client`
```
kacf: Will these turn into real newlines? I think they will end up looking quite ugly in the log (literal `\n`). Maybe break with normal spaces, and in the code, with `+` concatenated strings instead?

statescript/executor.go
```diff
@@ -88,12 +88,23 @@ func (l Launcher) CheckRootfsScriptsVersion() error {
...
+		return errors.New(errmsg)
 	} else if err != nil {
-		return errors.Wrap(err, "statescript: can not read rootfs scripts version")
+		return errors.Wrap(err, "statescript")
+	}
+	ver, err := readVersion(f)
+	if _, ok := err.(readVersionParseError); ok {
+		errmsg := `statescript: Failed to parse the version file in the statescript directory (%s). The file needs to contain a single integer signifying which version of the statescript support which this client is using`
```
kacf: Suggest breaking this into concatenated strings also.

statescript/executor.go
```diff
@@ -116,8 +127,8 @@ func matchVersion(actual int, supported []int, hasScripts bool) error {
 		}
 	}
 
-	return errors.Errorf("statescript: supported versions does not match "+
-		"(supported: %v; actual: %v)", supported, actual)
+	return errors.Errorf("statescript: The version read from the version file in the statescript directory does not match the versions supported by the client, please make sure the file is present and formatted correctly "+
```
kacf: And here (this one is partially broken already).

statescript/executor.go
```diff
@@ -88,12 +88,23 @@ func (l Launcher) CheckRootfsScriptsVersion() error {
...
+		errmsg := `statescript: The statescript version file is missing. This file needs to be
+present and contain a single number representing which version of the statescript support the client is using
+in order to successfully run statescripts on the client`
+		return errors.New(errmsg)
 	} else if err != nil {
-		return errors.Wrap(err, "statescript: can not read rootfs scripts version")
+		return errors.Wrap(err, "statescript")
+	}
+	ver, err := readVersion(f)
+	if _, ok := err.(readVersionParseError); ok {
```
kacf: Nice special case message!

statescript/executor.go
```diff
@@ -296,10 +315,14 @@ func reportScriptStatus(rep *client.StatusReportWrapper, statusReport string) er
 
 func (l Launcher) ExecuteAll(state, action string, ignoreError bool,
 	report *client.StatusReportWrapper) error {
+	// Only execute the state-scripts if the version file is present.
+	if err := l.CheckRootfsScriptsVersion(); err != nil {
```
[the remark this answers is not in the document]
kacf: No, this is not right. The rootfs version file only applies to the rootfs scripts (the ones without `Artifact` in the name), so it can be missing even if the artifact is using artifact scripts. So the check needs to happen later as it already does, when we know which type of script we need to run.

oleorhagen:
@kacf. Spurious failure. Should be good now :)

PR #346, opened 2019-01-15
MEN-2302: Only resolve symlinks in /dev/disk/by-partuuid.

oleorhagen:
Changelog: Title

### DIFF
--- a/partitions.go
+++ b/partitions.go
@@ -1,4 +1,4 @@
-// Copyright 2018 Northern.tech AS
+// Copyright 2019 Northern.tech AS
 //
 //    Licensed under the Apache License, Version 2.0 (the "License");
 //    you may not use this file except in compliance with the License.
@@ -70,9 +70,9 @@ func (p *partitions) getAndCacheInactivePartition() (string, error) {
 		return "", err
 	}
 
-	if resolveLink(active) == p.rootfsPartA {
+	if maybeResolveLink(active) == p.rootfsPartA {
 		p.inactive = p.rootfsPartB
-	} else if resolveLink(active) == p.rootfsPartB {
+	} else if maybeResolveLink(active) == p.rootfsPartB {
 		p.inactive = p.rootfsPartA
 	} else {
 		return "", ErrorPartitionNoMatchActive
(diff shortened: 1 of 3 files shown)


oleorhagen:
If I did not misunderstand the discussion completely, this should be it.

partitions.go
```diff
@@ -231,13 +231,17 @@ func checkBootEnvAndRootPartitionMatch(bootPartNum string, rootPart string) bool
...
+	resolvedPath, err := filepath.EvalSymlinks(unresolvedPath)
 	// This would only happen if supplied a link that goes nowhere or creates a loop
 	if err != nil {
-		log.Warnf("Could not resolve path link: %s Attempting to continue", path)
-		return path
+		log.Warnf("Could not resolve path link: %s Attempting to continue", unresolvedPath)
+		return unresolvedPath
 	}
-	return (resolvedPath)
+	// Only resolve /dev/disk/by-partuuid/
```
mirzak: I would probably extend this comment with a why, for future you and @kacf  :). But up to you, the trace ability back to the original discussion is there

partitions.go
```diff
@@ -231,13 +231,17 @@ func checkBootEnvAndRootPartitionMatch(bootPartNum string, rootPart string) bool
 	return strings.HasSuffix(rootPart, bootPartNum)
 }
 
-func resolveLink(path string) string {
+func resolveLink(unresolvedPath string) string {
```
kacf: I like the variable rename!

I would also rename `resolveLink` to `maybeResolveLink` or something similar, just so that future readers know to look up the function and not blindly assume that it always resolves links.

partitions.go
kacf: +1

partitions_test.go
```diff
@@ -223,3 +225,14 @@ func Test_getAllMountedDevices(t *testing.T) {
...
 }
+
+func TestMaybeResolveLink(t *testing.T) {
+	// If path is not a symlink, the original path should be returned.
+	tmp, err := ioutil.TempDir("", "maybeResolveLink")
+	require.Nil(t, err)
+	resolvedPath := maybeResolveLink(tmp)
+	assert.Equal(t, tmp, resolvedPath)
+
+	resolvedPath = maybeResolveLink("/dev/disk/by-partuuid")
```
oleorhagen: Sham test. Question is: Is it necessary to test this? And if yes, I'm not entirely sure as to how this should be done robustly.

kacf: I suppose you can make a link and a non-link just to be sure. Right now it appears to be testing just two non-links.

But you're right that this won't test the heart of the problem, which is the complex interaction between PARTUUID, Mender's partition indexes, and the bootloader's partition indexes. This needs to be tested at the acceptance test level. I looked into this when the PARTUUID feature was originally merged, but we need to build a specific image to test this because it must be based on using PARTUUID. It would be good to have such a test, but I'm also ok with merging without it, since it is quite a lot of work, and PARTUUID was a contributed feature.

kacf:
@oleorhagen: Could you prioritize getting this PR ready when you come in? This one is quite important to get fixed before the final I think.

partitions_test.go
```diff
@@ -223,3 +226,20 @@ func Test_getAllMountedDevices(t *testing.T) {
...
+	assert.Equal(t, "/dev/disk/by-partuuid", resolvedPath)
+
+	tmpsym := filepath.Join(tmp, "foobar")
+	require.NoError(t, os.Symlink(tmp, tmpsym))
+	resolvedPath = maybeResolveLink(tmpsym)
```
oleorhagen: @kacf as we discussed. This tests if a symlink is not resolved, when it is not /dev/disk/by-partuuid. And if I did not misunderstand you, the rest would need an integration test. Good to go?

kacf: Yes, it's just really hard to test this in a robust way without a lot of work. So let's settle for this.

PR #349, opened 2019-01-29
[WIP] MEN-2311 client DB backwards migration

pasinskim:


state.go
```diff
@@ -164,7 +164,8 @@ type StateData struct {
 
 const (
 	// name of key that state data is stored under across reboots
-	stateDataKey = "state"
+	stateDataKey       = "state"
+	stateDataKeyCustom = "state-custom"
```
kacf: Suggest renaming to "uncommitted" instead of "custom" to make it a bit more clear.

state.go
```diff
@@ -1307,11 +1313,19 @@ func StoreStateData(store store.Store, sd StateData) error {
...
 
+	return store.WriteAll(stateDataKeyCustom, data)
+}
+
+func CommitStateData(store store.Store) error {
+	data, err := store.ReadAll(stateDataKeyCustom)
+	if err != nil {
+		return err
+	}
 	return store.WriteAll(stateDataKey, data)
```
kacf: The read and write should be done in one DB transaction to avoid race conditions. It's true there shouldn't be other writers to the database at this time, but who knows what will happen in the future, and I'd rather get our DB handling correct from the start instead of dealing with the long term consequences like in CFEngine.

Unfortunately our DB layer is not good at multi-operation transactions, even though LMDB supports it, so we may have to extend the interface a little bit.

state.go
```diff
@@ -1332,9 +1346,42 @@ func LoadStateData(store store.Store) (StateData, error) {
...
+
+	customData, err := loadData(store, stateDataKeyCustom)
+	if err != nil {
+		return StateData{}, err
+	}
+
+	if key == customData.UpdateInfo.ID {
+		return customData, nil
+	}
+	return data, nil
```
kacf: Once we have proper transactions I think these reads should be done in one transaction as well.

store/dbstore.go
```diff
@@ -166,6 +166,13 @@ func (db *DBStore) Remove(name string) error {
 		}
 
 		if err := txn.Del(dbi, []byte(name), nil); err != nil {
+			// don't return error if the entry we are trying to remove
+			// does not exits
+			if lmdbErr, ok := err.(*lmdb.OpError); ok {
+				if lmdbErr.Errno == lmdb.NotFound {
+					return nil
+				}
+			}
```
kacf: :+1:

state.go
kacf: Actually, I think I can implement the transaction interface, since I need it for something else right now. Then you can use it afterwards.

state.go
pasinskim: Fixing!

state.go
kacf: I had to rework this quite a bit to fit into my branch, and while doing so I discovered an additional issue that we should handle: This method only works once. If we upgrade the database schema again, then we can't reuse this method. The reason is that the key lies in the old client not using the uncommitted key, and the new client using it. But this doesn't work if the client always uses it. We would have to invent an additional "superuncommitted" key!

I solved this in the following way: I changed the uncommitted key to only be used when the original key has a version that is different from the version used by the current client. This always keeps the old client using the normal key, and the new client, once it discovers the normal key has an older version, uses the uncommitted key. You can check out the work [here](https://github.com/kacf/mender/commit/f630e73981397dd8808c12761b99a01473b7d906).

kacf: The rest of the logic regarding the `UpdateInfo.ID` is the same.

kacf:
Since this is now covered in my update_modules branch, I will close this. However, the migration test still remains to get running, so MEN-2311 is not done.

PR #350, opened 2019-02-04
MEN-2214: Test that both intervals work independently

lluiscampos:
Extended the existing unit test to use different poll intervals for
update and inventory. This way we can verify that they are independent
and work as intended.

The interface stateTestController required modifications accordingly

Changelog: None


lluiscampos:
Done.

PR #361, opened 2019-03-08
MEN-2195: State scripts loop bugfix port to 1.7.x

lluiscampos:


lluiscampos:
The failing test is a known spurious server bug. Merging.

PR #362, opened 2019-03-12
MEN-2285: Use FlushingWriter to automatically Sync() data to the bloc…

kacf:
…k device as we write.

Changelog: None
Signed-off-by: Jeremy Trimble <jeremy.trimble@twosixlabs.com>

### DIFF
--- a/block_device.go
+++ b/block_device.go
@@ -1,4 +1,4 @@
-// Copyright 2017 Northern.tech AS
+// Copyright 2018 Northern.tech AS
 //
 //    Licensed under the Apache License, Version 2.0 (the "License");
 //    you may not use this file except in compliance with the License.
@@ -14,6 +14,7 @@
 package main
 
 import (
+	"io"
 	"os"
 
 	"github.com/mendersoftware/log"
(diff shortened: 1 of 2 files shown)


kacf:
This is the final piece of #342.

block_device.go
```diff
@@ -34,11 +35,59 @@ type BlockDeviceGetSectorSizeFunc func(file *os.File) (int, error)
...
+	} else if fw.unflushedBytesWritten >= fw.FlushIntervalBytes {
+		err = fw.Sync()
+	}
+
+	return rv, err
+}
+
+func (fw *FlushingWriter) Sync() error {
+	err := fw.WF.Sync()
+	fw.unflushedBytesWritten = 0
```
lluiscampos: I think we should reset it to 0 only if `err != nil`

kacf: Disagree. If you actually get `err != nil` (which means you're likely in a world of trouble, but that's a different discussion), then not setting it to zero here will essentially cause a `Sync()` loop, where you constantly call it over and over as long as you write. There is likely nothing you can do if you get an error anyway, so might as well reset it.

kacf:
Only a spurious error observed. Fixed comment in a followup commit and merged manually.

block_device.go
lluiscampos: Makes sense. Thanks

PR #363, opened 2019-03-13
Enable coverage reports for sub packages as well.

lluiscampos:
This is important for us since some packages are being mainly tested
by being included in higher level package code.

Changelog: None

(cherry picked from commit 37afcbb9a143e9884b6619328b66ab1c1256fa10)

### DIFF
--- a/block_device.go
+++ b/block_device.go
@@ -1,4 +1,4 @@
-// Copyright 2018 Northern.tech AS
+// Copyright 2019 Northern.tech AS
 //
 //    Licensed under the Apache License, Version 2.0 (the "License");
 //    you may not use this file except in compliance with the License.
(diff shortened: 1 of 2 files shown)


block_device.go
```diff
@@ -1,4 +1,4 @@
-// Copyright 2018 Northern.tech AS
+// Copyright 2019 Northern.tech AS
```
kacf: Hehe, oops...

lluiscampos:
The code report fails to see the diff, I think because master cannot build due to the License year.

I manually had a look on the last coverage report on master and it is 76ish%, so we got 2% up including the submodules.

Let's see the comparison with update_modules once we merge!

kacf:
Yep!

PR #369, opened 2019-03-26
Don't push network interfaces without a mac address to inventory

RickP:
Some network interfaces may not have a mac address (e.g. CAN adapters). If these are pushed to the hosted mender backend it returns a 400-error and nothing is written to the inventory.

### DIFF
--- a/support/mender-inventory-network
+++ b/support/mender-inventory-network
@@ -37,7 +37,9 @@ for devpath in $SCN/*; do
     if [ $dev = "lo" ]; then
         continue
     fi
-    echo "mac_$dev=$(cat $devpath/address)"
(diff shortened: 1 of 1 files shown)


mirzak:
Thanks for taking care of this. One small procedural requirement, you need to add a `Signed-off-by:` tag to your commit message and you need an change log entry, in your case it can be:

    Changelog: Title 

Without these the Travis tests will fail.

More info at https://github.com/mendersoftware/mender/blob/master/CONTRIBUTING.md

RickP:
OK, I amended the commit, I hope this is not too messy.

RickP:
Doesn't seem to work, I'll close this and make a new PR

mirzak:
Force push is also acceptable.

PR #370, opened 2019-03-26
Don't push network interfaces without a mac address to inventory

RickP:
Some network interfaces may not have a mac address (e.g. CAN adapters). If these are pushed to the hosted mender backend it returns a 400-error and nothing is written to the inventory.

Changelog: Title
Signed-off-by: Rick Pannen <rick.pannen@thinventions.de>


RickP:
It failed, let's try once more :)

PR #373, opened 2019-03-29
Fix File Install UM to not wipe out dest_dir on single file installs

lluiscampos:
For the single file install use case, the user will probably want to add
the file in a directory rather than to replace the contents of it with
the file.

Removed also the "install" dependency, using mkdir instead

Changelog: Title

### DIFF
--- a/support/modules/file-install
+++ b/support/modules/file-install
@@ -6,8 +6,10 @@ STATE="$1"
 FILES="$2"
 
 prev_files_tar="$FILES"/tmp/prev_files.tar
+tmp_dest_dir="$FILES"/tmp/tmp_dest_dir
 update_files_tar="$FILES"/files/update.tar
 dest_dir_file="$FILES"/files/dest_dir
+single_file_file="$FILES"/files/single_file
 
 case "$STATE" in
 
@@ -21,19 +23,33 @@ case "$STATE" in
 
     ArtifactInstall)
         dest_dir=$(cat $dest_dir_file)
-        [[ -d $dest_dir ]] || install -d $dest_dir
-        tar -cf ${prev_files_tar} -C ${dest_dir} .
-        rm -rf ${dest_dir}
-        mkdir -p ${dest_dir}
+        single_file=$(cat $single_file_file)
+        test -z "$dest_dir" -o -z "$single_file" && exit 1
+        mkdir -p $dest_dir
+        if [ $single_file == "true" ]; then
+            mkdir -p $tmp_dest_dir
+            filename=$(tar -tf ${update_files_tar})
+            [[ -f ${dest_dir}/${filename} ]] && tar -cf ${prev_files_tar} -C ${dest_dir} ${filename}
+        else
+            tar -cf ${prev_files_tar} -C ${dest_dir} .
+            rm -rf ${dest_dir}
+            mkdir -p ${dest_dir}
+        fi
         tar -xf ${update_files_tar} -C ${dest_dir}
         ;;
 
     ArtifactRollback)
         dest_dir=$(cat $dest_dir_file)
+        single_file=$(cat $single_file_file)
+        test -z "$dest_dir" -o -z "$single_file" && exit 1
+        if [ $single_file == "true" ]; then
+            filename=$(tar -tf ${update_files_tar})
+            rm ${dest_dir}/${filename}
+        else
+            rm -rf ${dest_dir}
+            mkdir -p ${dest_dir}
+        fi
         [[ -f $prev_files_tar ]] || exit 0
-        [[ -n "$dest_dir" ]] || exit 1
-        rm -rf ${dest_dir}
-        mkdir -p ${dest_dir}
         tar -xf ${prev_files_tar} -C ${dest_dir}
         ;;
 esac
--- a/support/modules/file-install
+++ b/support/modules/file-install
@@ -24,13 +24,16 @@ case "$STATE" in
     ArtifactInstall)
         dest_dir=$(cat $dest_dir_file)
         single_file=$(cat $single_file_file)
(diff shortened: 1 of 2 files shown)


lluiscampos:
@mirzak @kacf please have a look.

Here we can discuss the details of the implementation

lluiscampos:
See the failure on Travis...

We have a 2.0.0b1 tagged release that does not pass our sanity checks for `gofmt`

I am fixing it now, but I wonder how could this happen.

kacf:
> I am fixing it now, but I wonder how could this happen.

I think this got overlooked while I was fixing integration tests. I fixed those, but that re-broke the unit tests, and I never went back here to check it, since this is on a feature branch. This will all get better once we have merged this.

In the meantime, I encourage you to just rebase on `update_modules` since it has been fixed there, and I have also pulled in the latest changes from master.

mirzak:
This looks ok, but I wanted to clarify what I was trying to do. Which was not a single file install, instead I wanted to deploy this structure,

```
$ tree
.
├── etc
│   └── app.conf
└── usr
    ├── bin
    │   └── app
    └── share
        └── app
            └── data.txt
```

Which erased my rootfs, and this will still do it. Maybe this is out of scope of this module to handle, in which case we need to clarify the description a bit which currently states

> The File Install Update Module installs a user defined file tree structure (files and subfolders) into a given destination directory in the target

kacf:
Yes, the problem with that approach is that it cannot delete files. But of course there are valid use cases for that too. I just think we need to be clear about it.

lluiscampos:
> This looks ok, but I wanted to clarify what I was trying to do. Which was not a single file install, instead I wanted to deploy this structure,
> 

I see now your use case and makes sense as well. But I think that should be a different Update Module, as it cannot support Rollback the same way this one does. It could implement a different strategy for Rollback, though, like removing everything installed, but to restore files I think it is complicated.

No matter what this PR has a fix that I think it is important. And for your use case I suggest just to document it properly in the Mender Hub article

lluiscampos:
Rebased :+1:

I'll merge after the tests pass

lluiscampos:
Discussed in person with @kacf 

We will still make some changes:
* Never accept "/" as a dest_dir. This use case can't be supported with this Update Module
* Modify the backup/restore for single files to only delete/restore *the file*

lluiscampos:
Submitted new changes for the Update Module. I have separated now the logic for single file and file tree install. Please @mirzak have a look (Kristian is on PTO)

lluiscampos:
See [MEN-2442](https://tracker.mender.io/browse/MEN-2442)

support/modules/file-install
```diff
@@ -46,6 +47,7 @@ case "$STATE" in
             filename=$(tar -tf ${update_files_tar})
             rm ${dest_dir}/${filename}
         else
+            test "$dest_dir" == "/" && exit 0
```
mirzak: Is the exit code correct here? Success?

support/modules/file-install
```diff
@@ -21,19 +23,35 @@ case "$STATE" in
 
     ArtifactInstall)
         dest_dir=$(cat $dest_dir_file)
-        [[ -d $dest_dir ]] || install -d $dest_dir
-        tar -cf ${prev_files_tar} -C ${dest_dir} .
-        rm -rf ${dest_dir}
-        mkdir -p ${dest_dir}
+        single_file=$(cat $single_file_file)
+        test -z "$dest_dir" -o -z "$single_file" && exit 1
```
mirzak: The `exit 1` are very silent and would prefer some text here. Otherwise the log from a failed update would be useless.

support/modules/file-install
```diff
@@ -21,19 +23,35 @@ case "$STATE" in
...
+        mkdir -p $dest_dir
+        if [ $single_file == "true" ]; then
+            mkdir -p $tmp_dest_dir
+            filename=$(tar -tf ${update_files_tar})
+            [[ -f ${dest_dir}/${filename} ]] && tar -cf ${prev_files_tar} -C ${dest_dir} ${filename}
+        else
+            test "$dest_dir" == "/" && exit 1
```
mirzak: This test should probably done as early as possible, no point in doing anything if above condition is true. Do not know if it is valid to have at the top outside of the states?

support/modules/file-install
lluiscampos: For the Roolback step yes, as there is nothing we can rollback to.

So what it will happen for an update with dest_dir `/` is that the ArtifactInstall will exit 1, then mender client will perform an rollback (a *successful* rollback), and then back to idle. The update will be marked as failed, though, but the client will stay in a consistent state.

support/modules/file-install
lluiscampos: I still want to support deploying single files to `/`, so the check needs to be done here.

support/modules/file-install
lluiscampos: This should never happen if you create the artifact with the provided generator tool.

But you are right about the silent part and I would prefer some text also :) I'll fix.

support/modules/file-install
lluiscampos: Also, when I originally wrote the Update Modules I had these statements out of the states.

What happens is that for some states the files are not present. So if we want to execute the script with -e (exit on errors), we cannot have the checks in the top level context.

mirzak: Thanks for explaining

support/modules/file-install
mirzak: Thanks for explaining

support/modules/file-install
lluiscampos: Done

mirzak:
Nice, no more comments from me

lluiscampos:
Thanks for your review! Merging

PR #375, opened 2019-04-12
Update modules: Implement `NeedsArtifactReboot` -> `Automatic`.

kacf:
The motivation behind is change is to promote the use of `Automatic`
in all update modules. We want this because in order to support
multiple payloads, an update module should not reboot the host in its
`ArtifactReboot` script. Doing so means that `ArtifactReboot` scripts
in later payloads do not get to execute. Instead it should leave this
to Mender using the `Automatic` mechanism. Multiple payloads have not
been implemented yet, but since `Automatic` needs to be supported by
the update module itself, we implement this part now so that we don't
accumulate too many modules without it before we get multiple payload
support.

Note that the most important test of this feature is by changing the
`rootfs-image-v2` script, which is then used by the integration test
which already exists.

Changelog: Title

MEN-2011

### DIFF
--- a/state_test.go
+++ b/state_test.go
@@ -35,6 +35,8 @@ import (
 	"github.com/mendersoftware/mender/installer"
 	"github.com/mendersoftware/mender/statescript"
 	"github.com/mendersoftware/mender/store"
+	"github.com/mendersoftware/mender/system"
+	stest "github.com/mendersoftware/mender/system/testing"
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
 )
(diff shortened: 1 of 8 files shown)


state_test.go
```diff
@@ -4781,3 +4783,43 @@ func TestDBSchemaUpdate(t *testing.T) {
...
+	DeploymentLogger = NewDeploymentLogManager(tempDir)
+	defer func() {
+		DeploymentLogger.Disable()
+		DeploymentLogger = nil
+	}()
+
+	log.Log = log.New()
+	log.AddHook(NewDeploymentLogHook(DeploymentLogger))
+	defer func() {
+		log.Log = log.New()
```
lluiscampos: Why do we do this?

kacf: Actually needing to do this is an indication that there is a different test which is not cleaning up after itself (in particular the first `log.New()`). The cleanup is needed because `log` doesn't have a way to unregister the hook afterwards, so we just have to create a new `log.Log` instead.

I can add a comment about it.

lluiscampos: I see. Thanks for the clarification. A comment would be nice, so that may encourage someone to do the proper clean-up in the future.

PR #376, opened 2019-04-12
MEN-2442: Split file-install into two Update Modules

lluiscampos:
Changelog: Deprecate old file-install Update Module and create instead
two new ones: single-file-install and file-tree-install. These have a
simpler logic and clearer scope. See MEN-2442 for details.


lluiscampos:
@mirzak Hi, as you reported the problem in the first place, can you please also have a look at the new modules?

mirzak:
Looks good

PR #379, opened 2019-04-23
Simplify names of update modules.

estenberg:
Already updated at https://hub.mender.io/c/update-modules


estenberg:
We probably need to adjust some build scripts as well, at least for the Mender client debian package?

Also, I noted that there were some dead links at https://hub.mender.io/c/update-modules after we removed the update_modules branch. I switched everything on Mender Hub to master now as they should rarely change, at least not substantially.

estenberg:
Also we should backport to 2.0.x once approved.

lluiscampos:
The package recipe uses `make install`, so no changes are required.

I have verified that the new package correctly includes the new Update Modules.

PR #386, opened 2019-04-30
standalone: Fix artifact committing not working after upgrading from 1.x.

kacf:
Fixes MEN-2465.

Changelog: Title


kacf:
Testing happens in https://github.com/mendersoftware/meta-mender/pull/728.

lluiscampos:
The drop in coverage is the new function not being unit tested. It is tested at acceptance level, so I am okey with that (we will eventually get the acceptance test coverage in our QA metrics).

kacf:
Merging this. The standalone code is not used in the integration tests at all, so it's clear this is a spurious failure.

PR #387, opened 2019-04-30
MEN-2378: Print warning on an invalid server certificate.

oleorhagen:
Changelog: Title Instead of bailing out on an invalid server certificate,
the client now prints a verbose error message, and then starts up the daemon,
as opposed to dying.

### DIFF
--- a/main.go
+++ b/main.go
@@ -66,6 +66,10 @@ var (
 		"- must give exactly one from: %s", actionArguments)
 	errMsgIncompatibleLogOptions = errors.New("One or more " +
 		"incompatible log options specified.")
+
+	errMissingServerCertstr = "IGNORING ERROR: The client server-certificate can not be loaded error: (%s). The client will " +
+		"continue running, but will not be able to communicate with the server. If this is not your intention " +
+		"please add a valid server certificate"
 )
 
 var DeploymentLogger *DeploymentLogManager
@@ -390,9 +394,14 @@ func initDaemon(config *menderConfig, dev installer.DualRootfsDevice, env instal
 	mp.dualRootfsDevice = dev
 
 	controller, err := NewMender(config, *mp)
-	if controller == nil {
-		mp.store.Close()
-		return nil, errors.Wrap(err, "error initializing mender controller")
+	if err != nil {
+		// Ignore server certificate error  (See: MEN-2378)
+		if _, ok := errors.Cause(err).(*client.ClientServerCertificateError); ok {
+			log.Warningf(errMissingServerCertstr, err)
+		} else {
+			mp.store.Close()
+			return nil, errors.Wrap(err, "error initializing mender controller")
+		}
 	}
 
 	if *opts.bootstrapForce {
--- a/main_test.go
+++ b/main_test.go
@@ -465,3 +465,27 @@ func TestInitDaemon(t *testing.T) {
 	require.Nil(t, err)
 	assert.Error(t, handleCLIOptions(runOpts, &installer.UBootEnv{}, dualRootfs, &menderConfig{}))
 }
+
+// Tests that the client will boot with an error message in the case of an invalid server certificate.
+func TestInvalidServerCertificateBoot(t *testing.T) {
+	tdir, err := ioutil.TempDir("", "invalidcert-test")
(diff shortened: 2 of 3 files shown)


main_test.go
```diff
@@ -102,6 +102,30 @@ func TestRunDaemon(t *testing.T) {
...
 }
 
+// Tests that the client will boot with an error message in the case of an invalid server certificate.
+func TestInvalidServerCertificateBoot(t *testing.T) {
+	tdir, err := ioutil.TempDir("", "invalidcert-test")
+	require.Nil(t, err)
+
+	fmt.Printf("%v", log.Log)
+	logBuf := bytes.NewBuffer(nil)
+	log.SetLevel(log.DebugLevel)
```
lluiscampos: I don't think we need to set the level to Debug for this test

lluiscampos: By the way, here for instance you are not defering a call to restore the original level, meaning "you are not cleaning up after yourself". This might trigger errors like the one you showed me, where the order of the test cases alter the results.

main.go
```diff
@@ -390,9 +390,16 @@ func initDaemon(config *menderConfig, dev installer.DualRootfsDevice, env instal
...
-		return nil, errors.Wrap(err, "error initializing mender controller")
+	if err != nil {
+		// Ignore server certificate error  (See: MEN-2378)
+		if _, ok := err.(*client.ClientServerCertificateError); ok {
+			mp.store.Close()
+			return nil, errors.Wrap(err, "error initializing mender controller")
+		}
+		errstr := "IGNORING ERROR: The client server-certificate can not be loaded error: (%v). The client will " +
+			"continue running, but will not be able to communicate with the server. If this is not your intention " +
+			"please add a valid server certificate"
```
lluiscampos: The message is very long, and still I think there is something missing.

In my opinion, we should add that the client is going to rollback if this situation persists.

oleorhagen: Yes, ideally. However this has been discussed here already ;)
https://tracker.mender.io/browse/MEN-2378

main_test.go
oleorhagen: You are absolutely right  of course. Sloppy. I'm sry.

main_test.go
```diff
@@ -465,3 +465,28 @@ func TestInitDaemon(t *testing.T) {
...
 }
+
+// Tests that the client will boot with an error message in the case of an invalid server certificate.
+func TestInvalidServerCertificateBoot(t *testing.T) {
+	tdir, err := ioutil.TempDir("", "invalidcert-test")
+	require.Nil(t, err)
+
+	logBuf := bytes.NewBuffer(nil)
+	oldLog := log.Log
+	log.Log = log.New()
```
oleorhagen: Replaced the whole thing with a brand new logger, and it fixed it, believe it or not...

main.go
lluiscampos: Right.

main.go
```diff
@@ -390,9 +390,17 @@ func initDaemon(config *menderConfig, dev installer.DualRootfsDevice, env instal
...
-	if controller == nil {
-		mp.store.Close()
-		return nil, errors.Wrap(err, "error initializing mender controller")
+	if err != nil {
+		// Ignore server certificate error  (See: MEN-2378)
+		if _, ok := errors.Cause(err).(*client.ClientServerCertificateError); ok {
+			errstr := "IGNORING ERROR: The client server-certificate can not be loaded error: (%v). The client will " +
+				"continue running, but will not be able to communicate with the server. If this is not your intention " +
+				"please add a valid server certificate"
+			log.Warningf(errstr, err)
```
kacf: Please use `%s` and `err.Error()` instead. I've fixed so many logging bugs that were not caught by the static code analyzers because `%v` accepts any argument.

main_test.go
lluiscampos: :+1:

main.go
oleorhagen: Haha, did not know this was an issue. Fixing.

oleorhagen:
@kacf Good to go now.

oleorhagen:
@kacf good to go now.

PR #388, opened 2019-04-30
Added a PR-template.

oleorhagen:
Changelog: None

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -1,0 +1,6 @@
+<Commit Title>
+
+(A functionality change would need a message. Spelling etc would not).
+Changelog: None | Title | <description>
+
+Signed-off-by: User Name <name@example.com>


.github/pull_request_template.md
```diff
@@ -0,0 +1,6 @@
+<Commit Title>
+
+(A functionality change would need a message. Spelling etc would not).
+Changelog: None | Title | <description>
```
kacf: I suggest describing it like this instead:

```
Changelog: <description>

# <description> can be:
# * a literal description
# * "None" for no changelog
# * "Title" to use the commit title
```

The reason is that the changelog checker will accept anything with three words or more, so what you wrote is actually a valid changelog. And I'm *sure* someone will leave it like that. :-)

oleorhagen: I just found that we were missing a PR-template from here:
https://github.com/mendersoftware/mender/community
Also, this should help us avoid having to tell every new contributor to add a
signed off line, for one?

oleorhagen: I thought it was for committing, but maybe ur right actually. I used to think of them as the same thing, but they're actually not I guess.

kacf: Yes, it would be nice if we could force a commit hook into the user's repository when he clones it, but Git doesn't allow that (for quite obvious security reasons). :-)

So even though this is a good incentive, I think it ultimately may not be that valuable, since it will appear at the wrong stage in the process. Should we just close it?

Hmm, I wonder if it's possible to add a message that will appear after a clone. Something like:

```
Cloned repository 'mender'
Message from repository: Please run 'mender/extra/install-commit-hook'
```

or something like that.

PR #392, opened 2019-05-06
MEN-2131: Provide command line interface to force inventory update. 

kacf:
Provide command line interface to force inventory update. The client
now is able to be forced into an inventory update check state by sending a
SIGUSR2 signal to the running daemon. One quirk of the implementation is that
the forcedWakeUp channel is now global in the stateContext, therefore, an
unconsumed signal will force an extra wake-up at a later stage, if not consumed
straight away. However, this was chosen as a compromise in order to avoid
relying on synchronization primitives. Furthermore, an updatecheck or inventory-
update can only be forced from idle-state or the update- and inventory-update
states themselves. This done so that the state-machine cannot be forced from
random states into doing an update-check. The original state-machine logic has
to be respected.

Changelog: Title

### DIFF
--- a/daemon.go
+++ b/daemon.go
@@ -14,8 +14,6 @@
 package main
 
 import (
-	"os"
-
 	"github.com/mendersoftware/log"
 	"github.com/mendersoftware/mender/datastore"
 	"github.com/mendersoftware/mender/store"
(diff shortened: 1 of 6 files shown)


oleorhagen:
@kacf, good to go along with https://github.com/mendersoftware/integration/pull/553

PR #393, opened 2019-05-08
Make the device type file location configurable

spencercw:
This adds a new 'DeviceTypeFile' parameter to /etc/mender.conf to enable the device_type file to be positioned somewhere other that its default location of /var/lib/mender/device_type (aka /data/mender/device_type).

In our setup the data partition will be completely erased upon factory reset, or in case of filesystem corruption. We have a dedicated read-only partition for permanent data such as hardware revisions, so this patch allows us to point Mender there instead.

### DIFF
--- a/config.go
+++ b/config.go
@@ -39,6 +39,8 @@ type menderConfigFromFile struct {
 	// Rootfs device path
 	RootfsPartA string
 	RootfsPartB string
+	// Path to the device type file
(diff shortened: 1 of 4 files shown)


lluiscampos:
Hi @spencercw and thanks for your contribution.

Give us a couple of days to attend it. For now, I have started the integration tests for the PR to check the changes.

PR #396, opened 2019-05-13
Update modules fixes

kacf:
```
commit e2ff07c375bde3b7a0d6d4bd89f107b0bd28ab90
Date:   Mon May 13 11:00:34 2019

    Artifact gen: Support argument passthrough to `mender-artifact`.
    
    Use `--` to signal that remaining arguments should be passed directly
    to `mender-artifact`.
    
    Changelog: Commit
    
commit 0e62b78c176fdd91e0b94b6796fa2e6dce358369
Date:   Thu May 9 14:45:39 2019

    single-file module: Make sure permissions are preserved.
    
    Also make sure that backup preserves permissions.
    
    Changelog: Commit
    
commit 888299194bc37ac571d2c1c5b57157f689479316
Date:   Thu May 9 14:07:01 2019

    module-artifact-gen: Document that one can specify more than one device_type
    
    Changelog: None
    
```

### DIFF
--- a/support/modules-artifact-gen/directory-artifact-gen
+++ b/support/modules-artifact-gen/directory-artifact-gen
@@ -7,7 +7,7 @@ show_help() {
 
 Simple tool to generate Mender Artifact suitable for directory Update Module
 
-Usage: $0 [options] directory
+Usage: $0 [options] directory [-- [options-for-mender-artifact] ]
 
     Options: [ -n|artifact-name -t|--device-type -d|--dest-dir -o|--output_path -h|--help ]
 
@@ -17,6 +17,8 @@ Usage: $0 [options] directory
         --output-path       - Path to output file. Default: file-install-artifact.mender
         --help              - Show help and exit
         directory           - File tree to bundle in the update
+
+Anything after a '--' gets passed directly to the mender-artifact tool.
 
 EOF
 }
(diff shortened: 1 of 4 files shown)


support/modules/single-file
```diff
@@ -28,7 +29,7 @@ case "$STATE" in
         mkdir -p $tmp_dest_dir
         if test -f ${dest_dir}/${filename}
         then
-            if ! cp ${dest_dir}/${filename} ${tmp_dest_dir}
+            if ! cp -a ${dest_dir}/${filename} ${tmp_dest_dir}
```
lluiscampos: Why do we need these?

support/modules-artifact-gen/directory-artifact-gen
```diff
@@ -7,17 +7,19 @@ show_help() {
 
 Simple tool to generate Mender Artifact suitable for directory Update Module
 
-Usage: $0 [options] directory
+Usage: $0 [options] directory [-- [options-for-mender-artifact] ]
```
lluiscampos: This is fine.

However, makes me wonder if we should do this even simpler. The generator could only parse the parameters required (only `dest-dir`) and just pass the rest to the tool, so that it will fail there is something mandatory is missing (`artifact-name` or `device-type` for example).

support/modules/single-file
kacf: Normal `cp` is not guaranteed to preserve all file attributes (umask can interfere). This makes sure that the backup has exactly the same file attributes as the original.

support/modules-artifact-gen/directory-artifact-gen
kacf: Yes, that is an option, but then the user has to pass those arguments after the `--` which is perhaps a bit more user unfriendly? Especially given that they are required options. I don't think we should pass all unknown arguments to the tool, because it makes it risky to add arguments to the gen script later (namespace overlap). Does that make sense?

lluiscampos: Makes sense. Specially the last bit which I didn't think about.

Good.

kacf:
Great, thanks!

PR #400, opened 2019-05-22
Remove deploy logic from Travis CI

lluiscampos:
We do our deployments only from the main Mender QA pipeline (Jenkins)
and the current AWS key in Travis is erroring, so better to clean it up.

Changelog: None


.travis.yml
```diff
@@ -10,47 +10,6 @@ go:
...
-        - AWS_S3_ACL_DEFAULT=private
-
-        # Perform deployment steps if the build is performed for build: user/repo_name
-        # $TRAVIS_REPO_SLUG could be used if want to perform deploy on any repo.
-        # Currently perform deployment steps only of building main repository.
-        # Can be overridden by travis private variable set in travis settings: DEPLOY_REPO_OVERRIDE
-        - DEPLOY_REPO_DEFAULT=mendersoftware/mender
-
-        # Binary name to be uploaded to github release.
-        - GITHUB_RELEASE_BINARY=mender
```
lluiscampos: The GitHub releases seems to be broken also. I think we should fix it as there might be users looking for binaries directly on GitHub.

What do you think @kacf ?

kacf: Yes, absolutely!

lluiscampos: Created [MEN-2545](https://tracker.mender.io/browse/MEN-2545)

PR #401, opened 2019-05-22
Govendor update logrus

lluiscampos:
Two vendor updates: `mendersoftware/log` and `sirupsen/logrus`


mzedel:
as a side note and just since sirupsen already fell on my feet while building:
could we also remove the Sirupsen folder and just keep the sirupsen one? (unless there are dependencies that still rely on the uppercase variant?)

lluiscampos:
> could we also remove the Sirupsen folder and just keep the sirupsen one? (unless there are dependencies that still rely on the uppercase variant?)

I have removed it!

PR #404, opened 2019-06-04
Add macos support for single-file-artifact-gen

mzedel:
Changelog: None
Signed-off-by: Manuel Zedel <manuel.zedel@northern.tech>

### DIFF
--- a/support/modules-artifact-gen/single-file-artifact-gen
+++ b/support/modules-artifact-gen/single-file-artifact-gen
@@ -156,7 +156,11 @@ echo "$dest_dir" > $dest_dir_file
 echo "$filename" > $filename_file
 
 # Create permissions file in plain text
-stat -c %a "${file}" > $permissions_file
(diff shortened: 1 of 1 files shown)


kacf:
Pull request #404... For just a second, I thought I came to the wrong page. :-)

support/modules-artifact-gen/single-file-artifact-gen
```diff
@@ -156,7 +156,11 @@ echo "$dest_dir" > $dest_dir_file
 echo "$filename" > $filename_file
 
 # Create permissions file in plain text
-stat -c %a "${file}" > $permissions_file
+if [[ "$OSTYPE" == "darwin"* ]]; then
```
lluiscampos: I didn't know about this env var. I would have suggested using `uname -c`

mzedel: from what I have seen, uname might produce differently capitalized results while the env var should be at least somewhat reliable - not entirely sure though

PR #405, opened 2019-06-11
Restore error code 2 behavior when there is nothing to commit.

kacf:
The exact error is picked up in main.go.

Changelog: Title


oleorhagen:
@kacf, should be good to go!

PR #408, opened 2019-06-20
Qa 75 production gitlab pipeline

mzedel:


.gitlab-ci.yml
```diff
@@ -0,0 +1,65 @@
+# This file is a template, and might need editing before it works on your project.
```
lluiscampos: Please remove this comment and the one at line 5

.gitlab-ci.yml
```diff
@@ -0,0 +1,65 @@
...
+  allow_failure: true
+  script:
+    - go fmt $(go list ./... | grep -v /vendor/)
+    - go vet $(go list ./... | grep -v /vendor/)
+    - go test -race $(go list ./... | grep -v /vendor/)
+
+test:checks:
+  stage: test
+  script:
+    - make extracheck
```
lluiscampos: make extracheck will do `go vet` and `go fmt`. I think it is better to remove the test:format stage and, if any, add the `go test -race` to the Makefile's test target

.gitlab-ci.yml
```diff
@@ -0,0 +1,65 @@
...
+    - make coverage
+    - mkdir -p tests/unit-coverage && find . -name 'coverage.txt' -exec cp --parents {} ./tests/unit-coverage \;
+    - tar -cvf $CI_PROJECT_DIR/unit-coverage.tar tests/unit-coverage
+  tags:
+    - mender-qa-slave
+  artifacts:
+    expire_in: 30min
```
lluiscampos: Why such a short expire time? I think it is better with 24h, as we are not going to publish it at least to have time to consult it

.gitlab-ci.yml
```diff
@@ -0,0 +1,65 @@
...
+    paths:
+      - mender
+
+publish:tests:
+  stage: publish
+  dependencies:
+    - test:checks
+  script:
+    - tar -xvf unit-coverage.tar
+    # - bash <(curl -s https://codecov.io/bash) -s ./tests/unit-coverage;
```
lluiscampos: At some point we will do this from GitLab. Do you think we should create a task for it (for all repos) as a reminder?

.gitlab-ci.yml
kacf: Indeed, I would consider 2w, as with integration tests. Sometimes you just don't have time to analyze right away. A weekend and a planning session, that's already four days gone.

.gitlab-ci.yml
kacf: Do you mean that coverage analysis is part of Gitlab too?

lluiscampos: The approach is that what today is done by Travis (unit testing, code coverage, etc) will be openly done by Gitlab in each repo individual Pipeline, and what it is done by Jenkins (mender QA) will be closely done by Gitlab in mender QA Pipeline.

This can still be further discussed, but it seems natural to me to not maintain Travis build recipes after the migration.

mzedel: Apart from what @lluiscampos said, you can only get the overall percentage: https://docs.gitlab.com/ee/user/project/pipelines/settings.html#test-coverage-parsing in a badge https://docs.gitlab.com/ee/user/project/pipelines/settings.html#test-coverage-report-to show in the repo, but the analysis for the PR (as done by codecov.io) does not exist from Gitlab

@lluiscampos ticket created (QA-81) - but this should wait until we are convinced

kacf: Yes, my point was whether we would still use codecov.io. And I guess we will, if Gitlab only offers one simple number.

.gitlab-ci.yml
```diff
@@ -0,0 +1,55 @@
...
+
+test:
+  stage: test
+  script:
+    - make extracheck
+    - make coverage
+    - mkdir -p tests/unit-coverage && find . -name 'coverage.txt' -exec cp --parents {} ./tests/unit-coverage \;
+    - tar -cvf $CI_PROJECT_DIR/unit-coverage.tar tests/unit-coverage
+  tags:
+    - mender-qa-slave
```
lluiscampos: We need to remove this tag or share the runner with the project. I vote for just removing the tag, as I would like all the open Pipelines to run in public GitLab runners

lluiscampos: See job here: https://gitlab.com/Northern.tech/Mender/mender/-/jobs/240224541

mzedel: Unfortunately the checks seem to be a bit too much for the public runners, since they failed most of the time. So for now I enabled the shared runner or should we maybe add an additional like-public runner of our own?

PR #409, opened 2019-06-25
Fix `/bin/lsb_release` not being picked up by inventory script.

kacf:
Changelog: Title

### DIFF
--- a/support/mender-inventory-os
+++ b/support/mender-inventory-os
@@ -19,13 +19,15 @@ for file in /etc/os-release /usr/lib/os-release; do
     fi
 done
 
-if [ -x /usr/bin/lsb_release ]; then
-    OS="$(/usr/bin/lsb_release -sd)"
-    if [ -n "$OS" ]; then
-        echo "os=$OS"
-        exit 0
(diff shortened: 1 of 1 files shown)


kacf:
This will be part of https://github.com/mendersoftware/meta-mender/pull/768.

support/mender-inventory-os
```diff
@@ -19,13 +19,15 @@ for file in /etc/os-release /usr/lib/os-release; do
     fi
 done
 
-if [ -x /usr/bin/lsb_release ]; then
-    OS="$(/usr/bin/lsb_release -sd)"
-    if [ -n "$OS" ]; then
-        echo "os=$OS"
-        exit 0
+for lsb_release in /bin/lsb_release /usr/bin/lsb_release; do
```
lluiscampos: Okey, but can't we just do something like: `if [ -x ``which lsb_release`` ]; then...` ?

kacf: In principle I agree, but it's a bit tricky with the testing, which relies on which order these are attempted. I very much doubt this will be a problem on real devices, so I think I'll leave it like this for now to save time.

lluiscampos: Ok

PR #414, opened 2019-07-17
rootfs-image-v2: Make sure to follow the spec regarding `stream-next`.

kacf:
We should to read the final empty entry to make sure the client does
not block.

Changelog: commit


kacf:
@drewmoseley: Ref our earlier conversation about `rootfs-image-v2` being slightly wrong.

drewmoseley:
Looks good.  Thanks for this. I'll get around to updating the docs for clarity at some point.

PR #415, opened 2019-07-19
Make VersionString() follow line of sight

avegner:
Just a simple fix to make VersionString() simpler :)


kacf:
I think this is just personal preference, there is really no difference between the two. If it ain't broke, don't fix it! ;-)

PR #424, opened 2019-07-30
Update vendored dependencies for client.

kacf:
The `unix.UBI_IOCVOLUP` constant has vanished from the repository, and
isn't even in their history anymore. Weird...

Changelog: Title


kacf:
> I cannot really find these defines in mtd/ubi-user.h, can you point me to it?

Full disclosure: I didn't actually get them from there, I reintroduced [this commit](b7ad7b726dd00c126a866eb0b17bd76b30d218af), so I assume they *originally* came from there. At least it means that it used to work before.

lluiscampos:
I see. Thank you.

PR #425, opened 2019-08-06
Modify rootfs to 'Image file'

ajithpv:
Since, the Mender 2.0 can take both file and rootfs as update files, it is better to update the log with a generic name (Image file). Otherwise, 'file update' error also will show as "rootfs called without needed parameters"

Changelog: None
Signed-off-by: Ajith P V ajithpv@outlook.com

### DIFF
--- a/standalone.go
+++ b/standalone.go
@@ -46,7 +46,7 @@ func doStandaloneInstall(device *deviceManager, args runOptionsType,
 	var upclient client.Updater
 
 	if args == (runOptionsType{}) {
-		return errors.New("rootfs called without needed parameters")
+		return errors.New("Image file called without needed parameters")
 	}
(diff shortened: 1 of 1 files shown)


standalone.go
```diff
@@ -46,7 +46,7 @@ func doStandaloneInstall(device *deviceManager, args runOptionsType,
 	var upclient client.Updater
 
 	if args == (runOptionsType{}) {
-		return errors.New("rootfs called without needed parameters")
+		return errors.New("Image file called without needed parameters")
```
lluiscampos: The original intention here was to point to the name of the command line option, which was `rootfs`. I think we should change it to `install`, so that it still corresponds to the CLI.

PR #426, opened 2019-08-06
Update rootfs to install

ajithpv:
Since, the Mender 2.0 can take both file and rootfs as artifact, it is better to update the log with the corresponding CLI option (i.e. install). Otherwise, ' file update ' error also will show as "rootfs called without needed parameters"!

Changelog: None
Signed-off-by: Ajith P Venugopal <ajithpv@outlook.com>

### DIFF
--- a/standalone.go
+++ b/standalone.go
@@ -46,7 +46,7 @@ func doStandaloneInstall(device *deviceManager, args runOptionsType,
 	var upclient client.Updater
 
 	if args == (runOptionsType{}) {
-		return errors.New("rootfs called without needed parameters")
+		return errors.New("install called without needed parameters")
 	}
 
 	log.Debug("Starting device update.")


ajithpv:
@lluiscampos  - I have closed the previous PR and created a new one with the feedback.
PS: The old PR got some issues and broken from my local machine.

lluiscampos:
@ajithpv Thanks for re-submitting. Your commit message is still wrong, as the `Signed-off-by:` tag is still not the same as the Author, it is missing the `<>` symbols around the e-mail.

Are you manually editing this? It should be automatically written by git when using `git commit -s`.

lluiscampos:
Also for future reference, you don't need to close a PR and create a new one. You can always modify your commits and force-push to your branch.

ajithpv:
> @ajithpv Thanks for re-submitting. Your commit message is still wrong, as the `Signed-off-by:` tag is still not the same as the Author, it is missing the `<>` symbols around the e-mail.
> 
> Are you manually editing this? It should be automatically written by git when using `git commit -s`.

My bad. Since, this is a single line of change that too on the log print, I used Github itself for the PR (instead of local branch). I understood that, it is very much annoying to see multiple PR (closing and opening a new one) for a single line of change :expressionless:

ajithpv:
@lluiscampos  The check for _vexpress-qemu_ shows failed with below error:
_Exception: Could not determine MACHINE or MENDER_MACHINE value._

I'm wondering how the log change could leads to this error! :thinking:

lluiscampos:
@ajithpv I cannot explain it neither. But I agree it is not related to your changes, so I'll merge this PR and keep investigating in master.

Thanks for your contribution

PR #427, opened 2019-08-10
module/single-file: make sure filename is well defined

texierp:
Without this commit, ArtifactRollback() state is not
executed until the end.

Fixes:

root@nitrogen8m:~# mender -rollback -debug
...
INFO[0000] Update module output: + test -f /var/lib/mender/modules/v3/payloads/0000/tree/tmp/dest_dir/  module=modules
INFO[0000] Update module output: + exit 0                module=modules

Changelog: Commit

Signed-off-by: Pierre-Jean Texier <pjtexier@koncepto.io>

### DIFF
--- a/support/modules/single-file
+++ b/support/modules/single-file
@@ -47,9 +47,9 @@ case "$STATE" in
         ;;
 
     ArtifactRollback)
+        filename=$(cat $filename_file)
         test -f $tmp_dest_dir/$filename || exit 0
         dest_dir=$(cat $dest_dir_file)
-        filename=$(cat $filename_file)
         test -z "$dest_dir" -o -z "$filename" && \
             echo "Fatal error: dest_dir or filename are undefined." && exit 1
         rm ${dest_dir}/${filename}


lluiscampos:
@texierp good catch, thank you!

The commit message is good, but a too long IMO to include everything as a Changelog entry. Can you please summarize the change (from user point of view) in the Changelog tag?

lluiscampos:
Running integration tests in [Jenkins 4244](https://mender-jenkins.mender.io/job/mender-builder/4244/)

texierp:
Thanks for the review @lluiscampos 

Sure I will do!
Changelog: Title is good for you ?

lluiscampos:
> Thanks for the review @lluiscampos
> 
> Sure I will do!
> Changelog: Title is good for you ?

My suggestion: `module/single-file: fix rollback state by correctly defining filename`, or something like that. My point is that it should refer to the rollback state fix, which is the effect that the user might have seen.

You can edit the title and use `Changelog: Title`, or leave the title as it is and add such line in Changelog tag; as you wish.

texierp:
Ok thanks for the suggestion ;)

texierp:
Updated :)

texierp:
Thanks to you ;)

PR #428, opened 2019-08-14
Bad option passed to stat in single-file-artifact-gen fixed

merlin-northern:
ChangeLog:title

### DIFF
--- a/support/modules-artifact-gen/single-file-artifact-gen
+++ b/support/modules-artifact-gen/single-file-artifact-gen
@@ -155,8 +155,11 @@ echo "$dest_dir" > $dest_dir_file
 # Create single_file file in plain text
 echo "$filename" > $filename_file
 
+STAT_HAS_f=1;
+stat -f %A "${file}" >/dev/null 2>&1 || STAT_HAS_f=0;
+
 # Create permissions file in plain text
-if [[ "$OSTYPE" == "darwin"* ]]; then
+if [ $STAT_HAS_f -eq 1 ]; then
   stat -f %A "${file}" > $permissions_file
 else
   stat -c %a "${file}" > $permissions_file


support/modules-artifact-gen/single-file-artifact-gen
```diff
@@ -157,7 +157,7 @@ echo "$filename" > $filename_file
 
 # Create permissions file in plain text
 if [[ "$OSTYPE" == "darwin"* ]]; then
-  stat -f %A "${file}" > $permissions_file
+  stat -c %A "${file}" > $permissions_file
```
lluiscampos: This was long debugged by @mzedel and we decided on `-c %A` for a reason.

If I recall correctly, the issue was that "-c" did not behave the same way in Linux and in Mac OS. The [commit message](https://github.com/mendersoftware/mender/commit/60c7360d8bec6a9d863335898bcf915e0eadfdda) does not help much...

merlin-northern: > This was long debugged by @mzedel and we decided on `-c %A` for a reason.
> 
> If I recall correctly, the issue was that "-c" did not behave the same way in Linux and in Mac OS. The [commit message](https://github.com/mendersoftware/mender/commit/60c7360d8bec6a9d863335898bcf915e0eadfdda) does not help much...

I meant:
```bash
touch f; stat -c %A f;
-rw-r--r--
```

-f option does something else (--file-system display file system status instead of file status)
```bash
#touch f; stat -f %A f
stat: cannot read file system information for '%A': No such file or directory
  File: "f"
    ID: 100000400000019 Namelen: ?       Type: apfs
Block size: 4096       Fundamental block size: 4096
Blocks: Total: 29545461   Free: 2141099    Available: 1712250
Inodes: Total: 9223372036854775807 Free: 9223372036852788739
```

merlin-northern: it is just a typo nothing more

lluiscampos: It is not a typo, that I am sure about. You might be using a different version of stat than @mzedel. I wonder why this is not standard in different Mac OS.

The permissions file is expected to save the permissions in octal format, not in string. For example, in Linux:
```
touch f; stat -c %a f
644
```

lluiscampos:
Stopping builds in Jenkins and GitLab due to the PR being reworked.

support/modules-artifact-gen/single-file-artifact-gen
```diff
@@ -155,8 +155,14 @@ echo "$dest_dir" > $dest_dir_file
...
+set +e;
+STAT_HAS_f=0;
+stat -f %A "${file}" >/dev/null 2>&1;
+[ $? -eq 0 ] && STAT_HAS_f=1;
+set -e;
+
 # Create permissions file in plain text
-if [[ "$OSTYPE" == "darwin"* ]]; then
+if [ "$STAT_HAS_f" == "1" ]; then
   stat -f %A "${file}" > $permissions_file
```
mzedel: does the existence of the `-f` flag also imply the `%A`?

merlin-northern: > does the existence of the `-f` flag also imply the `%A`?

I do not know:) but look, it check if the command was successful. if `$?==0` then %A was supported. do you think we should add checking if output matched `^[0-9]+$`?

support/modules-artifact-gen/single-file-artifact-gen
```diff
@@ -155,8 +155,14 @@ echo "$dest_dir" > $dest_dir_file
 # Create single_file file in plain text
 echo "$filename" > $filename_file
 
+set +e;
+STAT_HAS_f=0;
+stat -f %A "${file}" >/dev/null 2>&1;
+[ $? -eq 0 ] && STAT_HAS_f=1;
+set -e;
```
lluiscampos: I suggest doing all this into a two liner like:
```
STAT_HAS_f=1
stat -f %A "${file}" >/dev/null 2>&1 || STAT_HAS_f=0
```

support/modules-artifact-gen/single-file-artifact-gen
```diff
@@ -155,8 +155,14 @@ echo "$dest_dir" > $dest_dir_file
...
+
 # Create permissions file in plain text
-if [[ "$OSTYPE" == "darwin"* ]]; then
+if [ "$STAT_HAS_f" == "1" ]; then
```
lluiscampos: Mmm. I guess this works, but why to check it as an string? I suggest doing a check for integer:

```
[ $STAT_HAS_f -eq 1 ]
```

lluiscampos: And thanks for changing `[[` with the basic `test`.

support/modules-artifact-gen/single-file-artifact-gen
mzedel: na, I just saw that on some example file:
```
stat -f %a LICENSE 
1565790353
stat -f %A LICENSE
644
```
the generation will fail... so the additional check would be helpful, although in my opinion it should be on the person modifying their OS to live with the consequences 😜

support/modules-artifact-gen/single-file-artifact-gen
merlin-northern: > I suggest doing all this into a two liner like:
> 
> ```
> STAT_HAS_f=1
> stat -f %A "${file}" >/dev/null 2>&1 || STAT_HAS_f=0
> ```

thanks. [modified](https://github.com/mendersoftware/mender/pull/428/commits/2ca77cc03487655e5f25525df2e1fb1518fc3e41) as requested

support/modules-artifact-gen/single-file-artifact-gen
merlin-northern: > na, I just saw that on some example file:
> 
> ```
> stat -f %a LICENSE 
> 1565790353
> stat -f %A LICENSE
> 644
> ```
> 
> the generation will fail... so the additional check would be helpful, although in my opinion it should be on the person modifying their OS to live with the consequences 😜

lots of sense here as well. check if you like [it](https://github.com/mendersoftware/mender/pull/428/commits/2ca77cc03487655e5f25525df2e1fb1518fc3e41)

support/modules-artifact-gen/single-file-artifact-gen
merlin-northern: thanks. all good suggestions.

support/modules-artifact-gen/single-file-artifact-gen
```diff
@@ -155,8 +155,16 @@ echo "$dest_dir" > $dest_dir_file
 # Create single_file file in plain text
 echo "$filename" > $filename_file
 
+set +e;
```
lluiscampos: Disabling exit on error is now unnecessary, as you are capturing the command error with a pipe ( `|| ...`)

support/modules-artifact-gen/single-file-artifact-gen
```diff
@@ -155,8 +155,16 @@ echo "$dest_dir" > $dest_dir_file
...
+STAT_HAS_f=1;
+stat -f %A "${file}" >/dev/null 2>&1 || STAT_HAS_f=0;
+if [ $STAT_HAS_f -eq 1 ]; then
+ stat -f %A "${file}" | grep -qE '^[0-7]{3,4}$' || STAT_HAS_f=0;
```
lluiscampos: This second check is redundant and not caching anything extra IMO: There might be numeric results and still not be the permissions in octel format (for instance, using %a)

Furthermore, @mzedel also suggested not to add it in his comment.

merlin-northern:
> * You need to sqash your commits into a single nice commit. We try to keep as clean history as possible in all main repos

I am the reviewer I find it easier to read and better when there is a separate commit with correction after review. I think I [addressed](https://github.com/mendersoftware/mender/pull/428/commits/154d89d8a9cbe24ef0aa56983110a632e6364af5) all your comments.

PR #429, opened 2019-08-18
Add demo module for ipk package updates

drewmoseley:
Changelog: Title

Signed-off-by: Drew Moseley <drew.moseley@northern.tech>

### DIFF
--- a/support/modules/ipk
+++ b/support/modules/ipk
@@ -1,0 +1,13 @@
+#!/bin/sh
+
+set -e
+
+STATE="$1"
+FILES="$2"
+
+case "$STATE" in
+    ArtifactInstall)
+        opkg install "$FILES"/files/*.ipk
+        ;;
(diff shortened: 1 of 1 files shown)


mirzak:
I believe that we should add this to,

      https://github.com/mendersoftware/mender-update-modules

lluiscampos:
> I believe that we should add this to,
> 
> ```
>   https://github.com/mendersoftware/mender-update-modules
> ```

I agree. We should add it in the other repo so that we get more traction there and other users decide to share their modules.

drewmoseley:
Is there a Yocto recipe for that repo?

mirzak:
Not a generic recipe that will install all the modules, but there is a place to put recipes.

https://github.com/mendersoftware/meta-mender-community/tree/sumo/meta-mender-update-modules

PR #430, opened 2019-08-27
MEN-2721: FIX: `mender -sender-inventory` next state fix.

oleorhagen:
Previously the `mender -send-inventory` command did not force the state
machine to the correct state, since `checkWaitState` returns the next state
to be handled, and not itself. Therefore the woken state machine would simply
handle the next expected event, and not `send-inventory` as this was
completely random whether would happen or not. Unlucky users therefore
reported this as having to run `-send-inventory` twice in the cases that
`send-inventory` was not the next event scheduled to happen.

Changelog: None

### DIFF
--- a/daemon.go
+++ b/daemon.go
@@ -72,9 +72,11 @@ func (d *menderDaemon) Run() error {
 		select {
 		case nState := <-d.forceToState:
 			switch toState.(type) {
-			case *IdleState, *CheckWaitState:
+			case *IdleState, *CheckWaitState, *UpdateCheckState, *InventoryUpdateState:
 				log.Infof("Forcing state machine to: %s", nState)
 				toState = nState
+			default:
+				log.Debugf("State machine does not force state transitions from: %s state", toState)
 			}
 
 		default:


oleorhagen:
Hmm... Fix, fix.. Do u want it `fixed` ? :p

daemon.go
```diff
@@ -72,9 +72,11 @@ func (d *menderDaemon) Run() error {
 		select {
 		case nState := <-d.forceToState:
 			switch toState.(type) {
-			case *IdleState, *CheckWaitState:
+			case *IdleState, *CheckWaitState, *UpdateCheckState, *InventoryUpdateState:
```
kacf: I still think there may be a problem here: If the state machine is on its way to the `UpdateCheckState` state, and you have a `forceToState` containing `InventoryUpdateState`, you end up skipping the `UpdateCheckState` entirely.

How about saving whatever you get from `forceToState` in a variable, and then handle it once the original condition, `IdleState || CheckWaitState` is true? This means you may have to wait through several states before it actually gets handled, but this is how I would expect it to work, naively. And you are guaranteed it will only be handled once you are in an idle or waiting state.

oleorhagen: Hmm, I see. Makes sense I guess. Originally I thought of this as a general 'forcing the state machine into another time, sort of'. But I guess the call-queue approach you are thinking of (at least what I am inferring from it) makes more sense. But...
Would it not be the other way around? 
First handling the forced state, storing, and then handling the original state, which has now been 'pushed back' in the call queue? If that is how you are picturing it?

oleorhagen: Also, the state that is returned, and not run, is not really supposed to run in the first place, as in all likelihood the wait was cancelled, and I think (will have to double check), that the way it is done now, an `updatecheck` is handled at the original time it was scheduled. So in a way, implementing this queing will mess with the original timeline. Am I wrong here?

kacf: Never mind, I misunderstood how the states are returned. The current approach is good, although we could think about forcing an inventory update at the end of an artifact install.

daemon.go
```diff
@@ -72,9 +72,11 @@ func (d *menderDaemon) Run() error {
...
 				log.Infof("Forcing state machine to: %s", nState)
 				toState = nState
+			default:
+				log.Debugf("State machine does not force state transitions from: %s state", toState)
```
kacf: Approved, but I would consider turning this into `Errorf` and making it a bit more user friendly, like: `Cannot check update or update inventory while in %s state`

PR #432, opened 2019-09-05
DEBUG: verifying stope stale builds functionality

lluiscampos:
Do not merge


lluiscampos:
The fix seems to work fine.

PR #435, opened 2019-09-12
MEN-2676: Set hard limit(10) for client update status report retries

oleorhagen:
Changelog: Title

### DIFF
--- a/state.go
+++ b/state.go
@@ -1319,8 +1319,19 @@ func (usr *UpdateStatusReportRetryState) Cancel() bool {
 	return usr.WaitState.Cancel()
 }
 
+// retry no more than 10 times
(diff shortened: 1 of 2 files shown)


kacf:
Alright, looks good. There are a bit more failures in the integration tests than I would expect though, I can't really say if it's spurious, when it's this many.

Also remember that you cannot both use `Changelog: Title` and a message in the same changelog tag. Just move the tag to a `Changelog: Commit` on its own line and it will include the whole commit message, including the title.

oleorhagen:
Yeh, I'm not feeling too fuzzy about this one either. I'll try n have a closer look.

oleorhagen:
@kacf, passed now. With no changes. Are our tests acting up again?

PR #436, opened 2019-09-30
Develop "mender config" tool for interactive configuration

alfrunes:
(MEN-2418)
Also freshed up the CLI a bit (MEN-2806).


main.go
```diff
@@ -425,30 +296,624 @@ func initDaemon(config *menderConfig, dev installer.DualRootfsDevice, env instal
...
+		// Overlapping globalflag
+		// -- command flag takes precedence over global flag.
+		opts.configPath = c.GlobalString("config")
+	}
+	stdin := bufio.NewReader(os.Stdin)
+
+	state := "deviceType"
+
+	// Regex filters
+	validDevice_re, err := regexp.Compile("^[A-Za-z0-9-_]+$")
```
lluiscampos: Style: avoid underscores in GO. This should be `validDeviceRE` or `validDeviceRegex`

main.go
```diff
@@ -425,30 +296,624 @@ func initDaemon(config *menderConfig, dev installer.DualRootfsDevice, env instal
...
+	}
+	validURL_re, err := regexp.Compile(`(http|https):\/\/(\w+:{0,1}\w*@)?` +
+		`(\S+)(:[0-9]+)?((\/\S+?\/)*)(\/|\/([\w#!:.?+=&%@!\-\/]))?`)
+	if err != nil {
+		return nil, errors.Wrap(err, "Unable to compile regex;")
+	}
+	validIP_re, err := regexp.Compile(
+		`([0-9]{1,3}\.){3}[0-9]{1,3}(:[0-9]{1,5})?`)
+
+	promptUser := func(prompt string) (string, error) {
```
lluiscampos: I think this should be moved out as an independent function

alfrunes:
> +1 on the cli flag upgrade!
> 
> I think the whole config wizard section should have its own source file though.

I was planning on refactoring all command related stuff out of `main.go`.

cli.go
```diff
@@ -0,0 +1,126 @@
...
+	"os"
+	"os/user"
+	"strings"
+
+	"github.com/mendersoftware/log"
+	"github.com/mendersoftware/mender/installer"
+	"github.com/mendersoftware/mender/system"
+
+	"github.com/pkg/errors"
+	"gopkg.in/urfave/cli.v2"
```
lluiscampos: You need to add this dependency with `govendor`

lluiscampos: And most importantly: using v2 will break compatibility with Debian, which will lead to a) our package no being updated in upstream or b) us assuming even more maintenance work (adding such dep in Debian).

Is there a good reason for the upgrade?

lluiscampos: The dependency is still missing. Nothing in the pipeline will run if we are not able to compile the target.

lluiscampos: Done

setup.go
```diff
@@ -0,0 +1,815 @@
...
+	rspNoDevice      = "Device type cannot be blank: "
+	rspInvalidDevice = `The device type \"%s\" contains spaces or special
+Characters. Please try again: `
+	rspSelectYN     = "Please select Y or N: "
+	rspInvalidEmail = "\n\"%s\" does not appear to be a " + // NOTE: format
+		"valid email address.\nPlease enter a valid email address: "
+	rspHMLogin = "We couldn’t find a Hosted Mender account with those " +
+		"credentials.\nPlease try again: "
+	rspConnectionError = `There was a problem connecting to hosted.mender.io.
+Please check your device’s connection and try again.\n`
```
lluiscampos: This message (probably others!) does not format correctly, as the \n is not escaped.

```
There was a problem connecting to hosted.mender.io.
Please check your device’s connection and try again.\nEmail:
```

LIC_FILES_CHKSUM.sha256
```diff
@@ -27,6 +27,7 @@ dad2b0b2cc2dbdbf95ad5d800ef7588956e74dc2479014829d42be295125c25d  vendor/github.
 fde7d610b9b95fc5a6304055c4dae951025b630aaa42a24e95ebf76675ae832c  vendor/github.com/stretchr/objx/LICENSE.md
 ffa15bdce332058a03a1d923910864fb6e58bf6df66a0e3914284725b327183e  vendor/github.com/ungerik/go-sysfs/LICENSE
 95b8ef9c4137a8f75ddd3101ffdc4cfd594fa875b261697b68baddc16b0e537c  vendor/github.com/konsorten/go-windows-terminal-sequences/LICENSE
+da277af11b85227490377fbcac6afccc68be560c4fff36ac05ca62de55345fd7  vendor/github.com/urfave/cli
```
lluiscampos: We are missing `LICENSE` at the end here, that is why the test is failing

setup.go
```diff
@@ -0,0 +1,817 @@
...
+			return errors.Wrap(err, "Error creating "+
+				"authorization request.")
+		}
+		authReq.SetBasicAuth(username, password)
+		response, err = client.Do(authReq)
+		if response != nil {
+			defer response.Body.Close()
+		}
+
+		if err != nil {
```
lluiscampos: Here, and possibly in many other places, we should log the actual error at debug level.

While testing, I am having errors and I cannot tell if it is my username or what

setup.go
```diff
@@ -0,0 +1,822 @@
...
+		case stateHostedMenderCredentials:
+			state, err = opts.askHostedMenderCredentials(ctx, stdin)
+
+		case statePolling:
+			state, err = opts.askPollingIntervals(ctx, stdin)
+		}
+		if err != nil {
+			return err
+		}
+	} // END for {state}
```
lluiscampos: We need some simple message here that says to the user that the configuration is done. As it is now, it exists silently unless you have passed the `--run-daemon` flag

setup.go
```diff
@@ -0,0 +1,828 @@
...
+		config.RetryPollIntervalSeconds = opts.retryPollInterval
+	}
+	config.HttpsClient.Certificate = opts.serverCert
+	config.TenantToken = opts.tenantToken
+
+	// Make sure serverURL and devicetypefile is set
+	config.Servers = []client.MenderServer{
+		{
+			ServerURL: opts.serverURL},
+	}
```
lluiscampos: Is this formatted correctly?

setup.go
```diff
@@ -0,0 +1,828 @@
...
+	// Regex: $1: schema, $2: URL, $3: path
+	re, err := regexp.Compile(`(https?://)?(.*)(/.*)?`)
+	if err != nil {
+		log.Warn("Unable to compile regular expression for parsing " +
+			"server URL.")
+		return
+	}
+	// strip schema and path
+	host := re.ReplaceAllString(opts.serverURL, "$2")
+	fmt.Println(host)
```
lluiscampos: Debug?

setup.go
```diff
@@ -0,0 +1,828 @@
...
+	if _, err := f.Read(lastChar); err == nil &&
+		lastChar[0] == byte('\n') {
+		route = route[1:]
+	}
+
+	_, err = f.WriteString(route)
+	if err != nil {
+		log.Warnf("Unable to add route %q to \"/etc/hosts\": %s",
+			route, err.Error())
+	}
```
lluiscampos: I would write an info at the end of this function. Being something that you run as root, it is nice to get notified of system wide changes. If I am evaluating Mender for demo is okey for the setup to modify `/etc/hosts`, but I would for sure want to clean it up later...

setup.go
```diff
@@ -0,0 +1,828 @@
...
+		"server, in seconds: [1800] "
+	promptRetryPoll = "\nThe retry poll interval defines the number of " +
+		"seconds between each attempt the client tries to communicate " +
+		"with the server. Note that the client may attempt more often " +
+		"initially to enable rapid upgrades, but will gradually fall " +
+		"back to this value if the server is busy.\n" +
+		"Set the retry poll interval [300]: "
+	promptInventoryPoll = "Set the inventory poll interval - the frequency " +
+		"with which the client will send inventory data to the server, " +
+		"in seconds: [28800] "
```
lluiscampos: General comment. Sorry I forgot in the previous round.

Most of the times that we ask for something that has a default, it will look like this:
- tool: Set some parameter [default value]
- me: type something wrong
- tool: Incorrect, please try again:
- me: mmm... what happens now if I press enter?

My point is that in the error/retry messages we should still have the default values printed... Do you agree? And what do you think @michaelatmender?

cli.go
```diff
@@ -0,0 +1,580 @@
...
+			},
+			Flags: []cli.Flag{
+				cli.StringFlag{
+					Name:        "config, c",
+					Destination: &runOptions.setupOptions.configPath,
+					Value:       defaultConfFile,
+					Usage:       "`PATH` to configuration file."},
+				cli.StringFlag{
+					Name:  "data, d",
+					Usage: "Mender state data `DIR`ECTORY path.",
```
lluiscampos: Typo: "`DIR`ECTORY"

cli.go
```diff
@@ -0,0 +1,580 @@
...
+					Destination: &runOptions.setupOptions.hostedMender,
+					Usage:       "Setup device towards Hosted Mender."},
+				cli.BoolFlag{
+					Name:        "demo, D",
+					Destination: &runOptions.setupOptions.demo,
+					Usage:       "Use demo configuration."},
+				cli.BoolFlag{
+					Name:  "run-daemon, r",
+					Usage: "Run daemon after setup."},
+			},
```
lluiscampos: Many different options leading to utilize lower and upper case of the same letters. I think that for most of them are not really helping in clarifying the usage (`-4` for ip, `-I` for inventory?).

Except for the most important ones (IMO demo, daemon and device-type) I think we should keep the rest only with the long version with double `--`. A command produced with so many small flags will be totally unreadable.

cli.go
```diff
@@ -0,0 +1,580 @@
...
+		cli.StringFlag{
+			Name:        "config, c",
+			Usage:       "Configuration `FILE` path.",
+			Value:       defaultConfFile,
+			Destination: &runOptions.config},
+		cli.StringFlag{
+			Name:        "fallback-config, C",
+			Usage:       "Fallback configuration `FILE` path.",
+			Value:       defaultFallbackConfFile,
+			Destination: &runOptions.fallbackConfig},
```
lluiscampos: This is very tricky! `-c` and `-C` for almost the same thing...

cli.go
```diff
@@ -0,0 +1,580 @@
...
+			Destination: &runOptions.bootstrapForce},
+		cli.BoolFlag{
+			Name:        "no-syslog, S",
+			Usage:       "Disable logging to syslog.",
+			Destination: &runOptions.logOptions.noSyslog},
+		cli.BoolFlag{
+			Name:        "skipverify, k",
+			Usage:       "Skip certificate verification.",
+			Destination: &runOptions.Config.NoVerify},
+	}
```
lluiscampos: Again, I think I would skip most of the short forms here also.

It is a matter to be where possible consistent across our products. Like -k in mender-artifact has relation with keys, here is about skipping something...

My criteria would be: for anything looks like is going to be used a lot manually (for instance log level), enable the short forms. For the rest use only the long ones...

setup.go
```diff
@@ -0,0 +1,826 @@
...
+
+	promptDeviceType   = "Device type (for example raspberrypi3-raspbian): "
+	promptConfirmDev   = "Confirm devicetype as %s? [Y/n] " // NOTE: format
+	promptHostedMender = "\nAre you connecting this device to Hosted Mender? [Y/n] "
+	promptCredentials  = "\nEnter your Hosted Mender credentials"
+	promptDemoMode     = "\nDemo mode uses short poll intervals and assumes the " +
+		"default demo server setup. (Recommended for testing.)\n" +
+		"Do you want to run the client in demo mode? [Y/n] "
+
+	promptServerIP   = "\nSet the IP of the Mender Server: [127.0.0.1] "
```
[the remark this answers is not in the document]
alfrunes: On a somewhat unrelated note, if I may add my own opinion -- I would fuse the `stateServerIP` and `stateServerURL` to prompt for a valid email or IP address. Adding a local route to `/etc/hosts` enforces root privileges for running setup. By simply configuring the server IP in the `mender.conf:Servers:ServerURL` prefixed with the appropriate schema `http/https` we can achieve the same result without enforcing privileged access.

Anyway, +1 on adding `const` to all default values.

setup.go
```diff
@@ -0,0 +1,826 @@
...
+	rspInvalidDevice = `The device type \"%s\" contains spaces or special
+Characters. Please try again: `
+	rspSelectYN     = "Please select Y or N: "
+	rspInvalidEmail = "\n\"%s\" does not appear to be a " + // NOTE: format
+		"valid email address.\nPlease enter a valid email address: "
+	rspHMLogin = "We couldn’t find a Hosted Mender account with those " +
+		"credentials.\nPlease try again: "
+	rspConnectionError = `
+There was a problem connecting to hosted.mender.io.
+Please check your device’s connection and try again.`
```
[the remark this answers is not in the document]
alfrunes: Correct me if I'm wrong, but I believe single quote is the correct apostrophe character.

setup.go
```diff
@@ -0,0 +1,828 @@
...
+		"if any (path to file, for example /etc/mender/server.crt): "
+	promptUpdatePoll = "\nSet the update poll interval - the frequency with " +
+		"which the client will send an update check request to the " +
+		"server, in seconds: [1800] "
+	promptRetryPoll = "\nThe retry poll interval defines the number of " +
+		"seconds between each attempt the client tries to communicate " +
+		"with the server. Note that the client may attempt more often " +
+		"initially to enable rapid upgrades, but will gradually fall " +
+		"back to this value if the server is busy.\n" +
+		"Set the retry poll interval [300]: "
```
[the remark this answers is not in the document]
alfrunes: Yeah, I pulled this straight out of the docs - forgot to summarize the prompt a little more compactly.

setup.go
```diff
@@ -0,0 +1,828 @@
...
+	config.HttpsClient.Certificate = opts.serverCert
+	config.TenantToken = opts.tenantToken
+
+	// Make sure serverURL and devicetypefile is set
+	config.Servers = []client.MenderServer{
+		{
+			ServerURL: opts.serverURL},
+	}
+	if config.DeviceTypeFile == "" {
+		config.DeviceTypeFile = "/etc/mender/device_type"
```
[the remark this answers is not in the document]
alfrunes: Not if you're configuring an entirely new file. But yeah, I changed it to the `defaultDeviceTypeFile`.

setup.go
alfrunes: As I mentioned above, I would rather merge the `stateServerURL` and `stateServerIP` and prompt for an IP or a general URL (with with the schema set) and add this to the `mender.conf` file instead of having to modify `/etc/hosts'.

cli.go
```diff
@@ -0,0 +1,580 @@
...
+		Usage:       "manage and start the Mender client.",
+		Version:     ShowVersion(),
+	}
+	app.Commands = []cli.Command{
+		{
+			Name:  "bootstrap",
+			Usage: "Perform bootstrap and exit.",
+			Flags: []cli.Flag{
+				cli.BoolFlag{
+					Name:        "forcebootstrap, F",
```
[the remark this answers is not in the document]
alfrunes: I used uppercase for this option as `-f` is conventionally used for filepath string argument, whereas forcebootstrap is a bool flag.

cli.go
alfrunes: It's actually not a typo, the ampersands uses these characters in the usage string as:
`--data DIR, -d DIR          Mender state data DIRECTORY path.`

cli.go
alfrunes: I would certainly recommend using the long option in any application which may possibly read by others (e.g. scripts), on the other hand I like having a short "lazy" way to provide flags. Some of the flag choices were chosen on based of other unix applications such as `-k` and `-4` from `curl`. We don't have that large repertoire of flags that I think it is necessary to exclude the short option, however if you strongly disagree I would like you to be more specific about which flags should be included/excluded/changed.

alfrunes: So here's what I've done:
For the setup flags I've kept the short option for `config `data` and `server-cert` which are all shared with the global options.
For the global options I've removed short option for `no-syslog` and `skipverify`.

setup.go
```diff
@@ -0,0 +1,826 @@
...
+	stateHostedMenderCredentials
+	statePolling
+	stateDone
+)
+
+const ( // Prompt constants
+	promptWizard = `Mender Client Setup
+===================
+
+Setting up the Mender client: the client runs on your device as a daemon, and will regularly poll the server to check for updates and report its inventory data.
```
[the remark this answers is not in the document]
lluiscampos: :+1:

setup.go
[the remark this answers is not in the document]
lluiscampos: Yeah, that is what I meant: single quote. It seems like it is the same character according to Wikipedia. I always thought that in code we type them as `'`.

setup.go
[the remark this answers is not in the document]
lluiscampos: On my system, with no previous config file, the file was created in `/var/lib/...`, which meant to me that `config.DeviceTypeFile` was not empty on start-up.

Ok

cli.go
lluiscampos: Nice! I had no idea. Thanks

setup.go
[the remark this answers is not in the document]
alfrunes: When configuring a demo/non-hosted setup we ask for server IP, making a static route from the default url to the given IP in `/etc/hosts` and then setting the `Servers::ServerURL` in `mender.conf` to this default url (`https://docker.mender.io`). Instead, I would suggest that we treat this case the same as the non-demo case asking for the ServerURL but we allow the host field of the URL to take either a named server or an IP address (for instance https://127.0.0.1) and configuring this as the serverURL in `mender.conf`. In practice the two approaches produces the same result, but the latter doesn't necessarily require root privileges.

cli.go
```diff
@@ -0,0 +1,580 @@
...
+
+func setupCLI(args []string) error {
+	runOptions := &runOptionsType{}
+	// Filter commandline arguments for backwards compatability.
+	// FIXME: Remove argument filtering in Mender v3.0
+	args = filterArguments(args)
+
+	app := &cli.App{
+		Before:      runOptions.handleLogFlags,
+		Copyright:   "Copyright 2019 Northern.tech AS",
```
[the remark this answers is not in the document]
alfrunes: The whole field is probably redundant and can be removed - this field in the help option has no actual implications whatsoever.

setup.go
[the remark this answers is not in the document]
lluiscampos: Thanks for clarifying. Conceptually I like very much your idea, but I suspect that there might be some other reason for having this ugly hack in /etc/hosts that we are not considering here.

Can you check this with Kristian?

cli.go
lluiscampos: I had no strong opinion on this. But thanks for your effort in simplifying it. I will review once again but now I have to learn about `go generator` tool... :)

setup.go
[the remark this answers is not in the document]
alfrunes: @kacf; is there something I'm missing here? (see comments above)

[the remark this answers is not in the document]
kacf: Yes, IP doesn't work, because TLS handshake requires a host name, and the host name has to be the same as the one you generated the certificate for. The demo certificate we ship is signed for `docker.mender.io`.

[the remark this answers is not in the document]
alfrunes: Ohh, right. Then I guess we'll leave it as it is.

setup.go
```diff
@@ -0,0 +1,831 @@
...
+		"which the client will send an update check request to the " +
+		"server, in seconds: [1800]" // (defaultUpdatePoll)
+	promptRetryPoll = "\nSet the retry poll interval - the frequency with " +
+		"which the client tries to communicate with the server (note: " +
+		"the client may attempt more often initially based on the " +
+		"previous intervals, but will fall back to this value if the" +
+		"server is busy) [300]" // (defaultRetryPoll)
+	promptInventoryPoll = "Set the inventory poll interval - the " +
+		"frequency with which the client will send inventory data to " +
+		"the server, in seconds: [28800]" // (defaultInventoryPoll)
```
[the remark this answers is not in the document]
alfrunes: I tried utilizing the default values above, but it turns out that if you cast the integer constants to strings it becomes a corrupted unicode character probably because the compiler interprets the value as an int without resolving the string context.
Alternatively I could make the texts variables and use strconv.Atoi, but I'd rather keep them as constants to avoid unnecessary imports.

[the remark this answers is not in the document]
lluiscampos: Yep. That is more or less what I thought. Hence I am ok with the hardcoded :)

cli_test.go
```diff
@@ -0,0 +1,471 @@
...
+
+func TestLoggingOptions(t *testing.T) {
+	err := setupCLI([]string{"mender", "-log-level", "crap", "-commit"})
+	assert.Error(t, err, "'crap' log level should have given error")
+	// Should have a reference to log level.
+	assert.Contains(t, err.Error(), "Level")
+
+	//err = setupCLI([]string{"mender", "-info", "-log-level", "debug"})
+	//assert.Error(t, err, "Incompatible log levels should have given error")
+	//assert.Contains(t, err.Error(), errMsgIncompatibleLogOptions.Error())
```
lluiscampos: This test file has a bit of dead code (this and above)

alfrunes:
I started a new test pipeline: https://gitlab.com/Northern.tech/Mender/mender-qa/pipelines/90581030

PR #439, opened 2019-10-07
Mender refactoring

alfrunes:
Did some cleanup work in this repository, splitting the humongous main package into smaller semantically meaningful subpackages. Moreover, I made an additional main_minimal.go that leaves out the cli package and only starts the daemon (including only a subset of the flags of the full cli package).

PS: This PR is rooted on top of #436


oleorhagen:
Nice, I like it! Haven't looked at anything else than the new folder setup though, but makes sense. Inventory could perhaps been a part of device mby?

license_test.go
```diff
@@ -15,8 +15,10 @@
 package main
 
 import "testing"
+import "os"
 import mt "github.com/mendersoftware/mendertesting"
 
 func TestLicenses(t *testing.T) {
+	os.Setenv("CHKSUM_FILE", "vendor/LIC_FILES_CHKSUM.sha256")
```
kacf: This should be removed.

alfrunes:
The whole minimal thing was mainly a curiosity thing to see how much overhead the cli package generates. It uses the same functions as the full main function isolating only the daemon command, so separating/moving these functions to a new/other package would make the overhead of maintaining it close to nothing.
However, the size on x86-64 for the minimal build is 10446144 bytes compared to 13952512 bytes for the full package, which is not that significant IMO so I agree that we don't really need the minimal option.

alfrunes:
Rebased to the updated master branch.
@kacf @oleorhagen; there are some changes I'd like your comment on.
I did a couple of changes in the `state.go` where I've made all state structure definitions private (to maintain consistency) and gathered all exposed states into a global `StateCollection` structure.
Moreover, the global DeploymentLogger is contained in app - I think we should make this private as is quite nasty to initialize the deploymentlogger variable from a separate package (`conf`), but I didn't include that in this PR as the work has already been quite involved.
Other than that I am open to changes in the overall directory structure.

Makefile
```diff
@@ -36,6 +36,10 @@ ifeq ($(LOCAL),1)
 TAGS += local
 endif
 
+ifeq ($(MAKECMDGOALS),minimal)
```
oleorhagen: I thought we decided on not adding the minimal client option?

oleorhagen:
Also, you have changed the folder setup a bit from last time right? Could you just quickly describe what has changed? :) @alfrunes

Makefile
alfrunes: Nice catch! Forgot to revert those lines.

kacf:
This is quite difficult to review due to its size, and a lot have changed since the last version. I will need a bit more time.

lluiscampos:
I am dropping from this one. Two colleges already on the PR, plus the changes are not small...

cli/cli_test.go
```diff
@@ -86,24 +84,43 @@ func TestRunDaemon(t *testing.T) {
...
+	pieces.AuthMgr = app.NewAuthManager(app.AuthManagerConfig{
+		AuthDataStore: pieces.Store,
+		KeyStore:      store.NewKeystore(pieces.Store, conf.DefaultKeyFile),
+		IdentitySource: &dev.IdentityDataRunner{
+			Cmdr: stest.NewTestOSCalls("mac=foobar", 0),
+		},
+	})
 
 	for name, test := range tests {
-		mender := newDefaultTestMender()
```
kacf: Since you have added some repeated code above, could the `newDefaultTestMender()` function be moved to `github.com/mendersoftware/mender/tests`? That would make it usable from both places, and you wouldn't need this extra bit.

alfrunes: I tried, but imposes an import cycle that's not possible to break (due to the dependency on the app package).

PR #441, opened 2019-10-16
QA-81: GitLab Pipeline completition

lluiscampos:
The important commits are:

```
commit f4e622190b87bec352661e46257180bfd0fd6511
Author: Lluis Campos <lluis.campos@northern.tech>
Date:   Wed Oct 16 13:21:29 2019 +0200

    Add mechanism to trigger mender-qa with PUBLISH_ARTIFACTS set
    
    For master and release branches, mender Pipeline will trigger a
    customized build of mender-qa which will build/test/publish the client
    image(s).
    
    As a side effect, it will also publish everything else: all server
    services images plus mender-artiact and mender-cli binaries.

commit 5f639b4ea95b085e15f455cd2ef403dc555c37f3
Author: Lluis Campos <lluis.campos@northern.tech>
Date:   Wed Oct 16 13:44:54 2019 +0200

    Unify test/build in a single stage
    
    The compile job is not really important, as the binary is not going to
    be published anywhere. Just add a make call in the test stage to make
    sure the Makefile is sane.
    
    Changelog: None
    
```


lluiscampos:
@mzedel  @kacf This is now ready for a review and discussion. Few things:

- I tried to use the built-in pipeline functionality for triggering other pipelines. See doc [here](https://about.gitlab.com/blog/2019/07/24/cross-project-pipeline/) and [here](https://docs.gitlab.com/ee/ci/yaml/README.html#trigger-premium). It just doesn't work so I used a custom curl call with token.

- See my text in the commit wrt publishing server + tools along with mender client.

- (related to the previous) And one other thing that needs discussion: this will not work for release branches. For a release branch in theory I should invoke release_tool to get the versions of all the other components and pass them along... But if I could workaround the previous comment (i.e. publish only the mender client) then I could go with only passing `MENDER_REV=CI_COMMIT_REF_NAME`. Do you guys see my point? Any good suggestion?

Thanks!

mzedel:
Great work (as with all your Gitlab migration efforts!)! Given that there should be the nice multi project pipelines otherwise, I'm sad to hear that the official approach didn't work. Just for the sake of understanding why you went with the curl calls I would have loved to see the steps you tried to have a starting point for later investigations/ a bug report.
As for the the trigger here and publish everything approach... it at least shouldn't hurt (at least for backend and gui), since the images all have the commit/ image hash and would most likely end up linking to other pushed images. 
But what I could imagine as a solution would be sending a `PUBLISH_ARTIFACTS_REPO` variable or deriving that from the trigger variable in Gitlab (to e.g. do something like [this](https://docs.gitlab.com/ee/api/jobs.html#get-job-artifacts))... and then run a different "single repo publishing job" for this specific repo only...
In short: publishing everything should be fine, there might be something better possible if there has to be something done...

.gitlab-ci.yml
```diff
@@ -21,6 +20,7 @@ test:
   script:
     - make extracheck
     - make coverage
+    - make build V=1
```
kacf: I think we should test just `make` instead. This is what is typically used to build.

lluiscampos: Agreed :+1: 

`make build V=1` was what we were using in Travis

kacf:
> * See my text in the commit wrt publishing server + tools along with mender client.
> 
> * (related to the previous) And one other thing that needs discussion: this will not work for release branches. For a release branch in theory I should invoke release_tool to get the versions of all the other components and pass them along... But if I could workaround the previous comment (i.e. publish only the mender client) then I could go with only passing MENDER_REV=CI_COMMIT_REF_NAME. Do you guys see my point? Any good suggestion?

Right, I did not think of release branches. While being able to publish only the client would be nice, it does have the unfortunate side effect of producing yet another configuration we need to maintain. There are no perfect solutions here that I can see, but I think my vote would go to using the release tool. Then we will at least reuse what we use for releases.

lluiscampos:
It took few attempts, but I got it now working. @kacf can you look at commit 65a8d8c?

.gitlab-ci.yml
```diff
@@ -12,14 +12,15 @@ before_script:
...
-  - build
   - publish
+  - trigger
 
 test:
   stage: test
   script:
     - make extracheck
     - make coverage
+    - make build
```
kacf: `make build` is still here. Did you remove `V=1` by mistake?

.gitlab-ci.yml
```diff
@@ -29,27 +30,67 @@ test:
...
-    - echo "publish mender binary here"
\ No newline at end of file
+    # The embbedded script below will
+    # for each integration version including mender version $CI_COMMIT_REF_NAME:
+    # * construct a curl formatted string concatenating each repo and revision like:
+    #   '-F variables[repoA_REV]=X.X.x -F variables[repoB_REV]=X.X.x ...'
+    # * trigger a Mender QA pipeline to build/test/publish client
+    - for integ_version in $(./extra/release_tool.py
+                             --integration-versions-including mender
+                             --version 2.1.x); do
```
kacf: Hardcoded `2.1.x`?

.gitlab-ci.yml
lluiscampos: I miss-read your original comment. Good eye

.gitlab-ci.yml
lluiscampos: Ouch! That was while debugging...

lluiscampos: It should be `$CI_COMMIT_REF_NAME`. Fixed now

lluiscampos:
> This is severely pushing the limits of the `gitlab-ci.yml` file. I would consider putting this in a dedicated script. This should make it much easier to test locally as well.

Moved to an script on mender-qa. See https://github.com/mendersoftware/mender-qa/pull/321

Thanks for the good suggestion!

lluiscampos:
It works. This is the triggered Pipeline during test: https://gitlab.com/Northern.tech/Mender/mender-qa/pipelines/90246836

PR #442, opened 2019-10-24
WIP: Add mender dist packages trigger

lluiscampos:
Do not review yet

PR #443, opened 2019-10-25
MEN-2315: Interactive mender configuration

lluiscampos:
Supersedes #442 

It will be merged together with [mender-dist-packages' feature branch MEN-2315-interactive-mender-config](https://github.com/mendersoftware/mender-dist-packages/tree/MEN-2315-interactive-mender-config).

Please see individual commits.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -4,6 +4,7 @@ bindir=/usr/bin
 datadir ?= /usr/share
 sysconfdir ?= /etc
 systemd_unitdir ?= /lib/systemd
+docexamplesdir ?= /usr/share/doc/mender-client/examples
 
 GO ?= go
 GOFMT ?= gofmt
@@ -68,7 +69,7 @@ build: mender
 mender: $(PKGFILES)
 	$(GO) build $(GO_LDFLAGS) $(BUILDV) $(BUILDTAGS)
 
-install: install-bin install-conf install-identity-scripts install-inventory-scripts install-modules install-systemd
+install: install-bin install-conf install-identity-scripts install-inventory-scripts install-modules install-systemd install-examples
 
 install-bin: mender
 	install -m 755 -d $(prefix)$(bindir)
@@ -101,8 +102,12 @@ install-systemd:
 	install -m 755 -d $(prefix)$(systemd_unitdir)/system
 	install -m 0644 support/mender.service $(prefix)$(systemd_unitdir)/system/
 
+install-examples:
+	install -m 755 -d $(prefix)$(docexamplesdir)
+	install -m 0644 support/demo.crt $(prefix)$(docexamplesdir)/
+
 uninstall: uninstall-bin uninstall-conf uninstall-identity-scripts uninstall-inventory-scripts \
-	uninstall-modules uninstall-modules-gen uninstall-systemd
+	uninstall-modules uninstall-modules-gen uninstall-systemd uninstall-examples
 
 uninstall-bin:
 	rm -f $(prefix)$(bindir)/mender
@@ -139,6 +144,10 @@ uninstall-modules-gen:
 uninstall-systemd:
 	rm -f $(prefix)$(systemd_unitdir)/system/mender.service
 	-rmdir -p $(prefix)$(systemd_unitdir)/system
+
+uninstall-examples:
+	rm -f $(prefix)$(docexamplesdir)/demo.crt
+	-rmdir -p $(prefix)$(docexamplesdir)
 
 clean:
 	$(GO) clean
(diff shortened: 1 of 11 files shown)


Makefile
```diff
@@ -4,6 +4,7 @@ bindir=/usr/bin
 datadir ?= /usr/share
 sysconfdir ?= /etc
 systemd_unitdir ?= /lib/systemd
+docexamplesdir ?= /usr/share/doc/mender-client/examples/
```
oleorhagen: In general I think we should stick to one or the other, meaning `/` at the end or not, as it is easy to burn yourself on this, which I think you have done below, when doing `$(dockexamplesdir)/foo`

Makefile
```diff
@@ -199,7 +199,7 @@ coverage:
 
 .PHONY: build clean get-tools test check \
 	cover htmlcover coverage \
-	install install-bin install-conf install-datadir install-demo install-identity-scripts \
-	install-inventory-scripts install-modules install-modules-gen install-systemd \
+	install install-bin install-conf install-datadir install-identity-scripts \
```
oleorhagen: This is growing quite big an unruly I think. What if we added a `.PHONY` to each target instead? It is quite verbose, and I know I'm basically the only proponent north of equator it seems. What do you think?

README.md
```diff
@@ -100,8 +100,8 @@ possible to use [Update
 Modules](https://docs.mender.io/artifacts/update-modules) and update
 other parts of the system.
 
-In order to connect to a Mender server, you either need to get a [Hosted
-Mender](https://hosted.mender.io/) account, or [set up a server
+In order to connect to a Mender server, you either need to get a [Mender
+Professional](https://hosted.mender.io/) account, or [set up a server
```
oleorhagen: +1

Makefile
lluiscampos: Good catch, thanks.

Makefile
lluiscampos: I don't understand what it means to be a "proponent north of equator". Other than that I agree and I am putting all of them in individual `.PHONY` targets

Makefile
lluiscampos: You have a point for the sake of consistency, but keep in mind that `//` is harmless and it is very common in `Makefile` (we have the same situation when `prefix` is empty).

oleorhagen: It is not harmless, and I don't think that Make automatically trims the double slashes for you, thus say `rm foo/bar` != `rm foo//bar`

lluiscampos: `rm foo/bar` == `rm foo//bar` in my Terminal

examples/mender.conf.demo
```diff
@@ -1,7 +1,7 @@
 {
   "InventoryPollIntervalSeconds": 5,
   "RetryPollIntervalSeconds": 30,
-  "ServerURL": "https://hosted.mender.io/",
-  "TenantToken": "Paste your Hosted Mender token here",
+  "ServerURL": "https://docker.mender.io",
+  "ServerCertificate": "/usr/share/doc/mender-client/examples/demo.crt",
```
kacf: Maybe we can talk about it when you get to the office, but I'm a little bit confused by this. If we have the setup command, why do we need this file at all? And if we do need it, should it not still point to our first class citizen, `hosted.mender.io`?

lluiscampos: (I am WFH today)

I don't really know what to do with this file. It is very much open for discussion.

It is still used in
- `Dockerfile.daemon`, where we want to have the demo certificate for a demo server.
- `mender-convert` (v1) where we want to have demo timing settings but `hosted.mender.io`.
- Also in the upstream Debian packages both conf files are installed as examples (when we unify recipes, we will do this also in our own deb package).

So, overall I thought it would be a good idea to keep the demo as both demo settings and demo server. The only broken piece with this decision would be mender-convert (v1).

kacf: `Dockerfile.daemon` I would discount completely. Tests should not influence what examples we install.

IMHO I think Debian examples should also point to `hosted.mender.io`. It is, ultimately, where we want people to go, even if the slightly fake `docker.mender.io` is also an option. This is especially true because `docker.mender.io` won't work without additional configuration. So I think both files should point to Hosted.

If we really want to provide an example, I think we should make a separate one, `mender.conf.custom-server`, or something like that.

kacf: I think that's ok.

lluiscampos:
Merging into a feature branch [MEN-2315-interactive-mender-config](https://github.com/mendersoftware/mender/tree/MEN-2315-interactive-mender-config)

PR #446, opened 2019-10-30
[WIP] Added codecov.yml

oleorhagen:
Testing out the codecov.yml configuration capabilities, no need to review atm

### DIFF
--- a/codecov.yml
+++ b/codecov.yml
@@ -1,0 +1,8 @@
+comment:
+  layout: "reach, diff, files"
+  behavior: default
+  require_changes: false  # if true: only post the comment if coverage changes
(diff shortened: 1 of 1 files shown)


kacf:
It's not really clear which problem is being solved. Why is it needed?

(I know about general problems with codecov, but how does this PR address it?)

oleorhagen:
It solves nothing really, except adding the pretty graph above. Also, it would make it easier to configure codecov, with say tresholds at a later point. But if deemed unnecessary, I will close.

kacf:
That's the explanation I needed! I have no strong feelings either way, since I have never actually used the comment reports and always go to the link.

PR #447, opened 2019-10-30
Fix `mender --version` option

lluiscampos:
It was returning unknown. The definition of Version was moved to a conf
package while refactoring in 8a75a8c and the Makefile was not updated
accordingly.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -25,7 +25,7 @@ TOOLS = \
 VERSION = $(shell git describe --tags --dirty --exact-match 2>/dev/null || git rev-parse --short HEAD)
 
 GO_LDFLAGS = \
-	-ldflags "-X main.Version=$(VERSION)"
+	-ldflags "-X github.com/mendersoftware/mender/conf.Version=$(VERSION)"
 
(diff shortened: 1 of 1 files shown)


lluiscampos:
Missing from PR #439 

@alfrunes @kacf Could you please look at this ASAP?

lluiscampos:
Shouldn't this be caught by some meta-mender acceptance test? I saw it in mender-dist-packages acceptance tests.

lluiscampos:
Stopping the Pipeline after the build stage. (The error was not caught in the first PR, so there is nothing to check there).

lluiscampos:
> It is a part of the release checklist to check the string. It has broken in several quite hard-to-catch ways over the years, so maybe that it the best place still.

Okey. At least now we can catch this through the `mender-dist-packages` master (and release) builds.

kacf:
> Okey. At least now we can catch this through the mender-dist-packages master (and release) builds.

We can catch it, but at that point it will be too late to fix it, because the repo will already be tagged.

PR #449, opened 2019-11-04
Handle implicit flag dependencies in `mender setup` command

alfrunes:
changelog: none

### DIFF
--- a/cli/cli.go
+++ b/cli/cli.go
@@ -55,8 +55,10 @@ const (
 )
 
 const (
-	errMsgAmbiguousArgumentsGiven = "Ambiguous arguments given - " +
+	errMsgAmbiguousArgumentsGivenF = "Ambiguous arguments given - " +
 		"unrecognized argument: %s"
+	errMsgConflictingArgumentsF = "Conflicting arguments given, only one " +
+		"of the following flags may be given: {%q, %q}"
 )
 
 func ShowVersion() string {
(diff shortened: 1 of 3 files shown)


cli/cli.go
```diff
@@ -448,12 +450,29 @@ func (runOptions *runOptionsType) handleCLIOptions(ctx *cli.Context) error {
...
+		} else if ctx.IsSet("server-ip") {
+			ctx.Set("demo", "true")
+		}
+		ctx.Set("mender-professional", "false")
+	}
+
+	if ctx.IsSet("update-poll") || ctx.IsSet("inventory-poll") ||
+		ctx.IsSet("retry-poll") {
+		ctx.Set("demo", "false")
+	}
```
lluiscampos: :+1:

cli/cli.go
```diff
@@ -448,12 +450,29 @@ func (runOptions *runOptionsType) handleCLIOptions(ctx *cli.Context) error {
...
+	// Check implicit flag dependencies
+	if ctx.IsSet("server-url") || ctx.IsSet("server-ip") {
+		if ctx.IsSet("server-url") && ctx.IsSet("server-ip") {
+			return errors.Errorf(errMsgConflictingArgumentsF,
+				"server-url", "server-ip")
+		} else if ctx.IsSet("server-ip") {
+			ctx.Set("demo", "true")
+		}
+		ctx.Set("mender-professional", "false")
+	}
```
lluiscampos: I still have this strange case. Setting `server-ip` and any (or all) of the interval flags (i.e. --update-poll 123) will then prompt me for serverURL.

alfrunes: Yeah, I can see the changes due to setting --update-poll takes precedence over `--server-ip` what is the expected behavior in this case? Should I raise a conflicting flags error? should we treat `--server-ip` as `--server-url`?

alfrunes:
@lluiscampos; I've updated the commit to handle demo and polling-interval flags.

PR #453, opened 2019-11-26
Update to Mender-Artifact version 3, and fully support it in the client

oleorhagen:


oleorhagen:
@lluiscampos do you want this in a feature branch also? I figured this was fine on master.

lluiscampos:
Which is the JIRA ticket associated to this PR? Please add it also to the main commit (so that it is referenced when we generate the changelog)

lluiscampos:
> @lluiscampos do you want this in a feature branch also? I figured this was fine on master.
> 
> This is added as a part of the binary support for depends and provides.

I agree. It makes sense to get it in master.

oleorhagen:
Have not been able to reproduce the failures locally. Therefore considering the tests passed.

kacf:
> I agree. It makes sense to get it in master.

The feature lives in `provides_and_depends`. This merge has had the unfortunate effect that the client in master will accept provides and depends, but it will do the wrong thing (ignore them).

This in turn caused https://github.com/mendersoftware/mender-binary-delta/pull/38 to pass, but it should have failed.

I'm not sure what's the best solution here: A couple of suggestion that come to mind:

* Merge `provides_and_depends` into master. This might (and probably will) cause https://github.com/mendersoftware/mender-binary-delta to start failing. I'd be surprised if this all works on the first attempt. Then we need to make additional fixes afterwards
* The slightly more correct thing, IMHO: Revert this, and merge it into `provides_and_depends` instead. Then do the same in https://github.com/mendersoftware/mender-binary-delta and resubmit the PR with https://github.com/mendersoftware/mender-binary-delta/blob/master/tests/integration/Dockerfile#L1 set to the `provides_and_depends` branch for testing. Then deal with any failures on the branch instead.

Regardless of which, something needs to happen, since right now the client is broken in a very dangerous way: It will work, but will do the wrong thing.

oleorhagen:
binary-delta-pipeline, after `provides_and_depends` been merged to master: https://gitlab.com/Northern.tech/Mender/mender-binary-delta/pipelines/100523698

kacf:
Exciting!

lluiscampos:
> > I agree. It makes sense to get it in master.
> 
> The feature lives in `provides_and_depends`. This merge has had the unfortunate effect that the client in master will accept provides and depends, but it will do the wrong thing (ignore them).

Good catch, @kacf. Sorry we missed this last sprint. Good to have you back :)

oleorhagen:
So the tests passed :raised_hands: 
However, it seems the new `qemu-rofs` image was not updated, so we have to count this test as a false positive (the worst kind!) :-1:

lluiscampos:
> So the tests passed raised_hands
> However, it seems the new `qemu-rofs` image was not updated, so we have to count this test as a false positive (the worst kind!) -1

I will be looking into it later today and ping you once is published.

PR #457, opened 2019-12-04
Setup fixup

alfrunes:
Rename `--mender-professional` flag to `--hosted-mender`, and adding a `quiet` option to suppress informative prompts.
I also removed the `run-daemon` option from setup flags as it seems redundant.

### DIFF
--- a/cli/setup.go
+++ b/cli/setup.go
@@ -45,7 +45,7 @@ type setupOptionsType struct {
 	invPollInterval    int
 	retryPollInterval  int
 	updatePollInterval int
-	menderProfessional bool
+	hostedMender       bool
 	demo               bool
 }
 
@@ -53,12 +53,12 @@ type setupOptionsType struct {
 const ( // state enum
 	stateDeviceType = iota
 	stateDeviceTypeConfirm
-	stateMenderProfessional
+	stateHostedMender
 	stateDemoMode
 	stateServerURL
 	stateServerIP
 	stateServerCert
-	statemenderProfessionalCredentials
+	stateCredentials
 	statePolling
 	stateDone
 	stateInvalid = -1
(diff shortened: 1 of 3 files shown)


cli/setup.go
```diff
@@ -92,25 +92,21 @@ const (
...
-		"minutes.\nLog in to your Mender UI at %s to accept it.\n"
-	promptDeviceType         = "Device type (for example raspberrypi3-raspbian): "
-	promptConfirmDev         = "Confirm devicetype as %s? [Y/n] " // NOTE: format
-	promptmenderProfessional = "\nAre you connecting this device to Mender " +
-		"Professional? [Y/n] "
-	promptCredentials = "\nEnter your Mender Professional credentials"
+		"device type and settings for communicating with the server.\n"
+	promptDone         = "Mender setup successfully."
+	promptDeviceType   = "Device type (for example raspberrypi3-raspbian): "
+	promptConfirmDev   = "Confirm devicetype as %s? [Y/n] " // NOTE: format
```
lluiscampos: As @mirzak and me suggested in the chat, I think we can remove this confirmation step

cli/setup.go
```diff
@@ -727,7 +723,9 @@ func doSetup(ctx *cli.Context, config *conf.MenderConfigFromFile,
 	}
 
 	// Prompt 'wizard' message
-	fmt.Println(promptWizard)
+	if !ctx.Bool("quiet") {
+		fmt.Println(promptWizard)
+	}
```
lluiscampos: Easy-peasy :+1:

cli/setup.go
```diff
@@ -92,25 +92,21 @@ const (
 	demoRetryPoll         = 30
 	demoUpdatePoll        = 5
 	demoServerCertificate = "/usr/share/doc/mender-client/examples/demo.crt"
-	menderProfessionalURL = "https://hosted.mender.io"
+	hostedMenderURL       = "https://hosted.mender.io"
 
 	// Prompt constants
 	promptWizard = "Mender Client Setup\n" +
 		"===================\n\n" +
 		"Setting up the Mender client daemon: the daemon will " +
```
lluiscampos: As we have removed the "daemon" term below, I suggest we remove it here also. I think it is enough with "The Mender client will regularly poll..."

alfrunes: I'll stick with ..."The client will regularly poll..." No need to repeat Mender everywhere :)

PR #458, opened 2019-12-04
Add verbosity to curl call for triggering deb package Pipeline

lluiscampos:
Add verbosity to curl call for triggering deb package Pipeline: GitLab json response contains the URL of the new Pipeline.


lluiscampos:
Stopping mender-qa Pipeline for this PR.

PR #460, opened 2019-12-07
MEN-2969: Fix: Store provide data across reboots in standalone mode

oleorhagen:
Previously the standalone mode would lose the provide data from the Artifact, as
the data was not stored, for when the device was rebooted.

Changelog: Now the client stores Artifact provides parameters across reboots in
standalone mode. Previously this data was ignored, and hence upgrading with an
Artifact with provides parameters these were lost.

### DIFF
--- a/app/standalone.go
+++ b/app/standalone.go
@@ -219,7 +219,7 @@ func doStandaloneInstallStates(art io.ReadCloser, key []byte,
 		return err
 	}
 
-	err = standaloneStoreArtifactState(device.Store, standaloneData.artifactName, installers)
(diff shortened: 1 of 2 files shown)


oleorhagen:
@kacf, yeh, maybe it should :P However, more importantly, I am still not sure if it should store artifact-group?

kacf:
You're right, it should.

oleorhagen:
> You're right, it should.

Haha, see! No unit test needed!

haha, good. I was really unsure if i understood this stuff or not for a while there.

Fixing, and testing now.

PR #461, opened 2019-12-10
New command: `snapshot dump` to dump current rootfs

alfrunes:
The new command creates a rootfs snapshot of the currently running
rootfs. The command is supported for ext[3|4], xfs, ReiserFS and JFS.
Moreover, the command will only run provided that state data is stored
on a different partition.

changelog: title


cli/snapshot.go
```diff
@@ -0,0 +1,169 @@
...
+	IOCTL_FITHAW_MAGIC   = 0xC0045878 // _IOWR('X', 120, int)
+)
+
+type SnapshotOptions struct {
+	ArtifactName      string
+	ArtifactGroup     string
+	CompatibleDevices []string
+	Depends           []string
+	Provides          []string
+}
```
kacf: Unused struct.

cli/snapshot.go
```diff
@@ -0,0 +1,169 @@
...
+
+type mountInfo struct {
+	mountID    int
+	mountPoint string
+	mountSrc   string
+	fsType     string
+}
```
kacf: Unused struct.

cli/snapshot.go
```diff
@@ -0,0 +1,169 @@
...
+
+	fsDevMajor := unix.Major(stat.Dev)
+	fsDevMinor := unix.Minor(stat.Dev)
+
+	linkPath, err := os.Readlink(
+		fmt.Sprintf("/dev/block/%d:%d", fsDevMajor, fsDevMinor))
+	if err != nil {
+		return "", errors.Wrap(err, "Error resolving device file path")
+	}
+	devPath := fmt.Sprintf("/dev/block/%s", linkPath)
```
kacf: Suggest doing this instead of `Readlink` and `Sprintf`, since it won't work if `linkPath` is absolute:

```
devPath, err := filepath.EvalSymlinks(fmt.Sprintf("/dev/block/%d:%d", fsDevMajor, fsDevMinor))
```

cli/snapshot.go
```diff
@@ -0,0 +1,169 @@
...
+		go func() {
+			select {
+			case <-sigChan:
+			case <-thawChan:
+			}
+			// FIXME: handle error
+			thawFs("/")
+			// Invoke os signal handler
+			signal.Stop(sigChan)
+			unix.Kill(os.Getpid(), unix.SIGINT)
```
kacf: Suggest keeping the value from `sigChan` and printing which one you received, which is often quite useful information (`SIGPIPE` may indicate that SSH died, for example).

cli/snapshot.go
```diff
@@ -0,0 +1,169 @@
...
+		}()
+		freezeFs("/")
+		unix.Sync()
```
kacf: Will this work? I would expect this to potentially block indefinitely after `freezeFs` has been called. This sequence of events is fine:

```
freezeFs("/")
unix.Sync()
--CONTEXT SWITCH--
<Some other process calls write()
```

However, if you're unlucky, I think this next one will block forever:

```
freezeFs("/")
--CONTEXT SWITCH--
<Some other process calls write()
--CONTEXT SWITCH--
unix.Sync()
```

I think `Sync()` should not be necessary though. Digging into the kernel source, it appears that [`Sync()` is called for you, as part of the freeze](https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/tree/fs/super.c#n1729), and I assume this is atomic, which is better than you can hope for by calling it manually.

cli/snapshot.go
```diff
@@ -0,0 +1,169 @@
...
+
+		f, err := os.Open(rootDev)
+		if err != nil {
+			thawChan <- 1
+			errChan <- err
+		}
+		defer f.Close()
+		io.Copy(os.Stdout, f)
```
kacf: Should handle the error from `io.Copy`.

cli/snapshot.go
```diff
@@ -0,0 +1,169 @@
...
+		return err
+	}
+	if rootDev == dataDev {
+		return errors.Errorf(
+			"State data store (%s) is located on rootfs partition",
+			runOpts.dataStore)
+	}
+
+	errChan := make(chan error)
+	go func() {
```
kacf: Why the outer `go func()`? It seems to just return an error at the end, so it's pretty synchronous.

cli/snapshot.go
```diff
@@ -0,0 +1,169 @@
...
+	err = unix.IoctlSetInt(fd, IOCTL_FIFREEZE_MAGIC, 0)
+	if err != nil {
+		return errors.Wrap(err, "Error freezing fs from writing")
+	}
+
+	done := make(chan bool)
+	go func() {
+		// Run this in separate go routine in case a new write arrived
+		// just after calling FIFREEZE
+		unix.Syncfs(fd)
```
kacf: Same as previously, should not be necessary.

cli/snapshot.go
```diff
@@ -0,0 +1,169 @@
...
+		fmt.Sprintf("/dev/block/%d:%d", fsDevMajor, fsDevMinor))
+	if err != nil {
+		return "", errors.Wrap(err, "Error resolving device file path")
+	}
+	devPath := fmt.Sprintf("/dev/block/%s", linkPath)
+
+	return devPath, nil
+}
+
+func freezeFs(fsRootPath string) error {
```
kacf: For consistency, and unless it causes import headaches, you could consider putting `freezeFs` and `thawFs` in `system/ioctl.go`.

cli/snapshot.go
```diff
@@ -0,0 +1,169 @@
...
+}
+
+type mountInfo struct {
+	mountID    int
+	mountPoint string
+	mountSrc   string
+	fsType     string
+}
+
+func (runOpts *runOptionsType) DumpSnapshot(ctx *cli.Context) error {
```
kacf: As mentioned in [MEN-2931](https://tracker.mender.io/browse/MEN-2931), we might want to ensure that logging calls don't go to the stream anywhere in here. This is also for future modifications, a logging call can easily be added without knowing that this will go to the stream.

So I suggest calling `log.SetOutput(os.Stderr)` to switch it to standard error. Does this make sense?

cli/snapshot.go
```diff
@@ -0,0 +1,169 @@
...
+	rootDev, err := getFsDevFile("/")
+	if err != nil {
+		return err
+	}
+	dataDev, err := getFsDevFile(runOpts.dataStore)
+	if err != nil {
+		return err
+	}
+	if rootDev == dataDev {
```
kacf: :+1:

cli/snapshot.go
alfrunes: Some leftovers from a previous snapshot draft.

cli/snapshot.go
alfrunes: Ohh, I was confused after I read somewhere that all writes *after* calling `FIFREEZE` would be blocked, so I added a Sync to make sure that all cached writes would be flushed (for good measure). But I see I should probably consolidated with the sources first :)

cli/snapshot.go
alfrunes: Yup, I definitely should

alfrunes:
@kacf I've reworked this PR a bit. I did some grooming of `cli/cli.go` and `system/ioctl.go` and added the `ssh` command as well (I can move the latter to a separate PR if you like), as well as addressing the above comments.

system/ioctl.go
```diff
@@ -86,80 +96,103 @@ func getUbiDeviceSize(file *os.File) (uint64, error) {
...
-// value 0 will be returned.
-func ioctlRead(fd uintptr, request ioctlRequestValue) (uint64, error) {
-	var response uint64
-	_, _, errno := syscall.Syscall(syscall.SYS_IOCTL, fd,
-		uintptr(unsafe.Pointer(request)),
-		uintptr(unsafe.Pointer(&response)))
-
-	if errno == syscall.ENOTTY {
-		// This means the descriptor is not a block device.
-		// ENOTTY... weird, I know.
```
kacf: The comment about `ENOTTY` has gotten lost and I think this is quite unobvious information. Could you make sure this is preserved somewhere?

utils/progress_bar.go
```diff
@@ -0,0 +1,122 @@
+package utils
```
kacf: Needs license header.

utils/progress_bar.go
```diff
@@ -0,0 +1,122 @@
...
+	pChars := strings.Repeat(pb.pchar, numPChars)
+	spaces := strings.Repeat(" ", progWidth-numPChars)
+	_, err := pb.out.WriteString(fmt.Sprintf(
+		"%s[%s%s]%s", pb.prefix, pChars, spaces, suffix))
+	return err
+}
+
+// TODO: Create a separate utility source file with funcs like this
+func ShortSize(bytes uint64) string {
+	var suffixes = [...]string{"B", "KB", "MB", "GB", "TB", "UWOT?!"}
```
kacf: Hehe, I would stop at `TB`. Also, for complete inambiguity, I would use `"KiB", "MiB", "GiB", "TiB"`, since these are in fact [Kibibytes](https://en.wikipedia.org/wiki/Kibibyte), etc. Yes, the name is silly, but it is the only unambiguous option. :-)

cli/cli.go
```diff
@@ -308,6 +313,26 @@ func SetupCLI(args []string) error {
...
+					Usage:       "Write rootfs to ssh host",
+					UsageText: "mender snapshot ssh " +
+						"[command options] -- " +
+						"<ssh options>",
+					Action: func(ctx *cli.Context) error {
+						return runOptions.DumpSSH(ctx)
+					},
+					Flags: []cli.Flag{
+						cli.StringFlag{
+							Name: "file, f",
```
kacf: I propose using `--outfile` instead of `--file`, since we already use the `--file` argument in mender-artifact to mean input file. Less confusing.

`-f` is actually a valid ssh option, but I think that's ok, since it's unlikely to be used in this context, and it's still possible to specify after the `--`. Switching to `-o` is tempting, but we should probably not, because `-o` is a quite relevant argument to use (for example to specify `-o StrictHostKeyChecking=false`).

So in other words, my proposal is `"outfile, f"`. Makes sense?

cli/snapshot.go
```diff
@@ -15,6 +17,42 @@ import (
...
+		if args[i] == "-i" {
+			identitySet = true
+		}
+	}
+	if !identitySet {
+		// If called with `sudo` set identity to caller default
+		uid := os.Getenv("SUDO_UID")
+		usr, err := user.LookupId(uid)
+		if err == nil {
+			args = append(args, "-i", usr.HomeDir+"/.ssh/id_rsa")
```
kacf: This file is not guaranteed to exist, and in fact, it's quite unlikely on an embedded device. In this case I assume that the command will fail. Whether or not it will work out of the box depends on policy, so I think I would consider adding this as a suggestion to the command help text instead of doing anything with it here.

cli/cli.go
alfrunes: Hmm... Or `"out-file, o"`?
It's a bit awkward, but you have to force ssh arguments to be positional (i.e. after `--`) otherwise the cli package caches it and throws an unrecognized option error.

cli/cli.go
```diff
@@ -282,6 +289,27 @@ func SetupCLI(args []string) error {
...
+			Description: snapshotDescription,
+			Subcommands: []cli.Command{
+				{
+					Name:  "dump",
+					Usage: "Dumps rootfs to stdout.",
+					Action: func(ctx *cli.Context) error {
+						// Expected to return ENOTTY
+						_, err := unix.IoctlGetTermios(
+							int(os.Stdout.Fd()),
+							unix.TCGETS)
```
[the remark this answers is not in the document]
alfrunes: But this was intended only for the dump command - there will not be an error here for the ssh command since we don't redirect stdout in that case.

alfrunes:
@kacf I trimmed off the ssh commit and split the modifications to `system/ioctl.go` to a separate commit. I also addressed the above comments and squashed it in to the main commit.

cli/cli.go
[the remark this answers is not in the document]
kacf: Oh right, it's being reused. Ok, I guess this is pretty small, so we can keep it here.

cli/cli.go
kacf: Aha! Yes, agree, if they have to be positional anyway, then we should definitely use `-o`.

alfrunes:
Increasing the memory in qemu seems to resolve the issues in the qemux86-64 acceptance tests - perhaps we're already close to the 256M limit.
Regarding the integration tests I have no idea - it seems that the device fails to reboot. I tried downgrading the `unix` package to the previous version, but this didn't change anything. Other than that I don't see how any other contributions of this PR should affect anything in the client - so I'm out of ideas of what to do with this.

kacf:
Oh, now I see it. Something has gone wrong in the ioctl refactoring. Look at this message:

> ```
> native sector size of block device /dev/mmcblk0p3 is 234881024, we will write in chunks of 234881024
> ```

It's trying to write in chunks of 234M each, which is the whole filesystem, basically. And then it tries to allocate that...

alfrunes:
Ahhh! Wrong ioctl magic for getting the sector size. Thanks for pointing that out :+1:

kacf:
@alfrunes: There is also an acceptance test at https://github.com/mendersoftware/meta-mender/pull/888 now, so we can test it together with that PR before merging this. I tested it locally, and it worked with your changes from earlier today.

alfrunes:
Ok. I found another bug, so I'll just wait and see if the `vexpress_qemu` passes before starting a new one.

kacf:
The failure is almost certainly because of the parallel test issue which has been discussed the Slack channel. I have started [a new integration test run](https://gitlab.com/Northern.tech/Mender/mender-qa/pipelines/106822649) with a lower parallel factor, so hopefully that one will pass.

kacf:
@alfrunes: Alright the last one passed. The code coverage can be ignored for this particular PR, since this is ioctl level functionality, and we test in the acceptance tests, which do not count code coverage.

Shall I merge this and https://github.com/mendersoftware/meta-mender/pull/888 ?

alfrunes:
@kacf Go ahead :+1:

PR #462, opened 2019-12-10
Remove device type confirmation from setup unit tests

lluiscampos:
Remove device type confirmation from setup unit tests

Fixes the current hung in CI when running the unit tests.

Missing from commit aeb3db2 and overlooked at PR #457. The reason for
the unit tests not to run in the original PR is unknown.


lluiscampos:
Stopping the integration Pipeline

PR #463, opened 2019-12-16
Mender setup fixes

kacf:


cli/setup.go
```diff
@@ -851,30 +842,43 @@ func (opts *setupOptionsType) maybeAddHostLookup() {
...
 			return
 		}
 	}
-	route := fmt.Sprintf("\n%-15s %s\n", opts.serverIP, host)
 
 	// Seek to last character
 	_, err = f.Seek(-1, os.SEEK_END)
 	if err != nil {
-		log.Warnf("Unable to add route %q to \"/etc/hosts\": %s",
+		log.Warnf("Unable to add route \"%s\" to \"/etc/hosts\": %s",
```
oleorhagen: Confused:
Why changing `%q` to explicitly quote a string? 
They're the same thing

cli/setup.go
```diff
@@ -851,30 +842,43 @@ func (opts *setupOptionsType) maybeAddHostLookup() {
...
 			route, err.Error())
 	}
-	// Remove newline from route string if there allready is one
+	routeLine := "\n" + route + "\n"
+	// Remove newline from routeLine string if there allready is one
```
oleorhagen: s/allready/already

cli/setup.go
```diff
@@ -851,30 +842,43 @@ func (opts *setupOptionsType) maybeAddHostLookup() {
...
 		lastChar[0] == byte('\n') {
-		route = route[1:]
+		routeLine = routeLine[1:]
 	}
 
-	_, err = f.WriteString(route)
+	_, err = f.WriteString(routeLine)
 	if err != nil {
-		log.Warnf("Unable to add route %q to \"/etc/hosts\": %s",
+		log.Warnf("Unable to add route \"%s\" to \"/etc/hosts\": %s",
```
oleorhagen: Same with `%q`

cli/setup.go
lluiscampos: From the commit message:
```
* Use %s instead of %q. This is actually not critical after the
  newlines were removed, but is still semantically better: Quoted
  strings are definitely not appropriate to copy-paste into
  `/etc/hosts`.
```

oleorhagen: haha, thx @lluiscampos. However, I still don't get it. The output is the exact same

kacf: Actually the `%q` has nothing to do with the quotes, these were missing in both bases. The `%q` has to do with newlines. With newlines, in the `%s` case, you get:

```
Unable to add route 
127.0.0.1 docker.mender.io
 to \"/etc/hosts\"
```

With `%q`, you get:

```
Unable to add route \n127.0.0.1 docker.mender.io\n to \"/etc/hosts\"
```

In the end I decided to remove the newlines from the output, which is why this is not critical to fix. However, from a general perspective it's clear that `%s` gives more correct output for this particular case, since `\n` will be misinterpreted inside `/etc/hosts`. If it were a C-string, then `%q` would have been more appropriate.

kacf:
Fixed the typo, but it has no effect on tests, so I cancelled the pipeline. Original result is [here](https://gitlab.com/Northern.tech/Mender/mender-qa/pipelines/103406438).

PR #465, opened 2019-12-18
Refactor the block-device interface

oleorhagen:
When reviewing this, ignore the commit history. It is easiest to just look at the code as a whole. 

What this code does is two-fold, and this will be rewritten into two commits reflecting this before final review. However, here is what has been done.

* Refactored the block-device type to use a file-like interface with Open, Close and Write methods.
* The underlying writer is a chain of writers, each with their own responsibilies, however, only one may do any manipulation of the underlying file-descriptor (the innermost one).
* Add an optimized writer, which only writes 'clean frames' to the block-device. Meaning that only byte-frames which have to change on the device are written.

Hope this makes sense. 

Please only consider this PR a template for discussion, and ignore the leftover TODO's and spurious log messages. These will be removed before submitting the PR (along with a cleaner git-history).


oleorhagen:
Related ticket: https://tracker.mender.io/browse/MEN-2939

installer/block_device.go
```diff
@@ -195,3 +252,155 @@ func (bd *BlockDevice) SectorSize() (int, error) {
...
+	payloadBuf := make([]byte, frameSize)
+
+	//
+	// Read len(b) bytes from the block-device
+	//
+	n, err = io.ReadFull(bd.blockDevice, payloadBuf)
+	if err != nil {
+		log.Errorf("Failed to read a full frame of size: %d from the block-device", err)
+		return 0, err
+	}
```
kacf: You should also check what `n` returned. It's not entirely clear what should happen if `n != frameSize` though. It's an unlikely event, but also has very bad consequences if it does. My proposal would be to treat this also as an error.

installer/block_device.go
```diff
@@ -195,3 +252,155 @@ func (bd *BlockDevice) SectorSize() (int, error) {
...
+	log.Infof("Skipping writing of clean frame. Reporting: %d bytes as written", n)
+	// No need to write a clean frame
+	bd.totalFrames += 1
+	return n, err
+}
+
+func (obw *OptimizedBlockDeviceWriter) Close() error {
+	s := "The optimized block-device writer wrote a total of %d frames, " +
+		"where %d frames did need to be rewritten"
+	log.Infof(s, obw.totalFrames, obw.dirtyFrames)
```
kacf: Nice statistics! :+1:

installer/block_device.go
```diff
@@ -195,3 +252,155 @@ func (bd *BlockDevice) SectorSize() (int, error) {
...
+	}
+
+	return rv, err
+}
+
+func (fw *FlushingWriter) Sync() error {
+	err := fw.File.Sync()
+	fw.unflushedBytesWritten = 0
+	return err
+}
```
kacf: Confused. Isn't `FlushingWriter` the `BlockDevicer` at the bottom of the layers? Doesn't it need to support `Seek()` then?

installer/block_device.go
oleorhagen: This is already handled in `readFull`. Which will return an error if that is the case.

installer/block_device.go
oleorhagen: Yeh, nice idea. I'm struggling a bit with the wording though.

installer/block_device.go
oleorhagen: It does support seek by composition

oleorhagen:
@kacf Since we both seemed to agree on the chained-writer approach, I have now implemented this  solution, and the PR is finally ready for your proper review!

installer/block_device.go
```diff
@@ -36,140 +40,193 @@ type BlockDeviceGetSectorSizeFunc func(file *os.File) (int, error)
...
-// For instance, an os.File is a WriteSyncer.
-type WriteSyncer interface {
+// BlockDevicer is a file-like interface for the block-device.
+type BlockDevicer interface {
+	io.Reader
 	io.Writer
-	Sync() error // Commits previously-written data to stable storage.
+	io.Closer
+	io.Seeker
+	Sync() error
```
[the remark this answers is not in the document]
oleorhagen: What else would sync do?

installer/block_device.go
```diff
@@ -36,140 +40,193 @@ type BlockDeviceGetSectorSizeFunc func(file *os.File) (int, error)
...
-	FlushIntervalBytes uint64               // Force a flush to disk each time this many bytes are written
+	Path string // Device path, ex. /dev/mmcblk0p1
+	w    io.WriteCloser
 }
 
-// A WriteSyncer is an io.Writer that also implements a Sync() function which commits written data to stable storage.
-// For instance, an os.File is a WriteSyncer.
-type WriteSyncer interface {
+// BlockDevicer is a file-like interface for the block-device.
+type BlockDevicer interface {
```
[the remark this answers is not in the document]
oleorhagen: Hmm, disagree on this one actually. This is supposed to be the block-device file. So the defining features are write, read, seek, etc. 
Name is still up for grabs though. But I don't think it's gonna be block-device-syncer.

installer/block_device.go
kacf: :+1:

installer/block_device.go
kacf: Ah right, I missed that.

installer/block_device.go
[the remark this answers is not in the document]
kacf: It's frequently misunderstood, people tend to think that if you haven't synced, then other processes can't read it yet. I'd keep the comment, just to be friendly to future inexperienced readers.

installer/block_device.go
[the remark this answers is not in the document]
kacf: Yes, looking a second time, I think I also agree. It supports syncing as a natural consequence of being a block device.

oleorhagen:
Ignoring spurious integration test failure.

PR #469, opened 2020-01-07
New command: `snapshot ssh`

alfrunes:
New sub-command that redirects the snapshot to a remote host over ssh.

I also included a quiet flag to the `snapshot [dump|ssh]` command


alfrunes:
After discussion with the rest of the team we decided not to include the ssh sub-command after all.
(PR superseded by #480)

PR #471, opened 2020-01-09
README.md: Corrected a typo (intance -> instance)

olehermanse:
Changelog: None


olehermanse:
> You need to sign your commit using -s when committing.

@lluiscampos There is no such option when committing from github.com. Of course I could fix it by cloning and committing from command line, or manually entering the text in the GitHub UI. But I would reconsider this policy, it should be as easy as possible to contribute tiny changes like this.

PR #473, opened 2020-01-13
Add `go get` command to README for compilation.

estenberg:


README.md
```diff
@@ -88,6 +88,7 @@ To install Mender on a device from source, please run the following commands
 inside the cloned repository:
 
 ```
+go get
```
lluiscampos: This will download the dependencies instead of using the vendored ones. It will most likely "work", but it is not what we want.

Instead, the user should clone the repo in an specific path inside its own GOPATH. I'll launch a new PR with the recommended steps to clarify my point.

lluiscampos: See #474

lluiscampos:
Closing, see comment above. We should instead merge #474

estenberg:
Ah, yes it did fail for me in mysterious ways relating to dependencies. :) 
Thanks!

PR #474, opened 2020-01-14
Instruct the user to clone the repo inside GOPATH before running make

lluiscampos:
Supersedes #473

### DIFF
--- a/README.md
+++ b/README.md
@@ -84,8 +84,14 @@ make TAGS=nolzma
 
 ### Steps
 
-To install Mender on a device from source, please run the following commands
-inside the cloned repository:
+To install Mender on a device from source, first clone the repository in the correct folder
+structure inside your `$GOPATH` (typically `$HOME/go`):
+
+```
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -84,8 +84,14 @@ make TAGS=nolzma
 
 ### Steps
 
-To install Mender on a device from source, please run the following commands
-inside the cloned repository:
+To install Mender on a device from source, first clone the repository in the correct folder
+structure inside your GOPATH (typically `$HOME/go`):
+
+```
+git clone https://github.com/mendersoftware/mender.git $HOME/go/src/github.com/mendersoftware/mender
```
estenberg: Maybe use $GOPATH instead of $HOME/go then?

lluiscampos: :+1:

lluiscampos:
> Not sure if most people have GOPATH set, but I would assume so. If so, I'd recommend just cloning into $GOPATH to make it clear.

Changed go $GOPATH as you suggested. If the reader does not have it set, it seems quite obvious for he/she to just set it to HOME/go as we have in parenthesis.

Thanks for the input. Merging!

PR #480, opened 2020-01-16
Refactor and improvement of snapshot command

alfrunes:
* Added handling of ALL terminating signals to the unfreeze rootfs
before invoking the system (default) signal handler, this is important
because if the application terminates without executing the signal
handler the entire system may freeze.
* Refactored the signal handler, removing redundant go channel.
Improved progress bar - NewProgressBar no longer returns nil if the
output is not a tty device but defaults to a width of 32 bytes. Removed
colon prefix for the progress bar on dump.
* Improved log/error messages and cli help/usage descriptions. Remove
duplicate command definition of "snapshot".
* Added --quiet flag to dump sub-command suppressing ALL output


cli/cli.go
```diff
@@ -308,6 +308,27 @@ func SetupCLI(args []string) error {
...
+							unix.TCGETS)
+						if err == nil {
+							return errDumpTerminal
+						}
+						return runOptions.
+							CopySnapshot(ctx, os.Stdout)
+					},
+				},
+			},
+		},
```
kacf: I think this commit is not supposed to be here. This causes the flag to be introduced twice, because this commit already exists in the history.

utils/progress_bar.go
```diff
@@ -31,33 +31,34 @@ const (
 )
 
 const (
-	defaultTerminalWidth = 32
+	defaultTerminalWidth = 80
```
kacf: This doesn't agree with what you wrote in the commit message.

main.go
```diff
@@ -28,7 +28,7 @@ func doMain() int {
 			log.Warnln(err.Error())
 			return 2
 		} else {
-			log.Errorln(err.Error())
+			log.Fatalln(err.Error())
```
kacf: This should be in a separate commit.

.gitlab-ci.yml
```diff
@@ -2,6 +2,7 @@ image: golang:1.13
 
 variables:
   REPO_NAME: github.com/mendersoftware/mender
+  GIT_DEPTH: 0
```
kacf: This should be in a separate commit.

alfrunes:
@kacf; thanks for the review. I've addressed your comments and added some more features based on observations from using it with an actual device. Most significantly I added the "watchdog go routine" that keeps the system from freezing on invalid usage (e.g. piping to a process that writes to the filesystem). I also added two new flags `--compression` and `--fs-path` for convenience.

kacf:
It's not a good idea to force push anymore. The changes I tried to merge in https://github.com/alfrunes/mender/pull/2 are now easily lost. Also, https://github.com/mendersoftware/mender/pull/486 looks like it might pass, and I think we should merge that so that we actually get this in in time for the sprint review. Can you add changes on top of that instead?

kacf:
Dammit, I was wrong. It didn't pass.. :-( But still, please make your changes on top of the old changes. It's really difficult to review when all commits are updated.

alfrunes:
@kacf Sorry about all the back and forth with the refactoring... I rebased this one on top of the feature branch so it should be ready for merge as long as the tests passes.

kacf:
> Sorry about all the back and forth with the refactoring... I rebased this one on top of the feature branch so it should be ready for merge as long as the tests passes.

Ok. For now they are not passing, but I haven't identified the cause yet. One failure appears to be specific to the test, which I'm working on. The other failure appears to be some kind of hang when encountering the expire timer. But the message about unfreezing the filesystem is printed, so I don't quite understand why it doesn't recover.

For now I prefer to keep a stable base, not update my branch and just focus on getting it passing, unless you haven specific fixes which you think would address the issue I described. Once the test actually passes we can start putting your new code on top and see how it behaves.

alfrunes:
@kacf found the issue with the frozen main thread, should be fixed now.

kacf:
Superseded by #493.

PR #481, opened 2020-01-20
MEN-3047: Fix crash when specified certificate can't be opened.

kacf:
```
commit 282ccabca28c4c21006c420b1c0dcfe0e8d41bfa
Date:   Mon Jan 20 12:17:00 2020

    MEN-3047: Fix crash when specified certificate can't be opened.
    
    It behaves essentially the same as if the certificate was not
    specified, and uses only the system CA certificates (though with a big
    warning in the log). However, if it cannot find *any* certificates, it
    will fail with an error message.
    
    Changelog: Commit
    
commit 1c8202451e89ee8768e2b2483d0a71fe5c20a5ec
Date:   Thu Jan 16 14:21:59 2020

    setup: Treat read-only filesystem the same as normal permission error.
    
    Changelog: None
    
commit b3fe9dd71905e2c4983fb3a33c339d40bea5e736
Date:   Thu Jan 16 13:14:37 2020

    Fix segfault after temporary file creation fails.
    
    It is an error to access the file object when the TempFile call has
    failed. It won't exist in this case anyway, so the call is not needed.
    
    Changelog: Fix segfault when running `mender setup` on a read-only
    filesystem.
    
```

### DIFF
--- a/cli/cli.go
+++ b/cli/cli.go
@@ -23,6 +23,7 @@ import (
 	"path"
 	"runtime"
 	"strings"
+	"syscall"
 
 	"github.com/mendersoftware/log"
 	"github.com/mendersoftware/mender/app"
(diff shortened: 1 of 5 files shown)


kacf:
Never mind, this is wrong, will fix and resubmit.

PR #482, opened 2020-01-20
MEN-3047: Fix crash when specified certificate can't be opened.

kacf:
```
commit 8b74abe167b13ca3dc02cd179d2b37a8bef63959
Date:   Mon Jan 20 14:04:58 2020

    MEN-3047: Fix crash when specified certificate can't be opened.
    
    Changelog: Fix crash when specified certificate can't be opened.
    Both the `ServerCertificate` setting and the system certificates are
    now optional, in the sense that the client will run without them.
    However, the client will not be able to connect without the right
    certificates, so the main usecase of this change is to have a workable
    client that will roll back if connections can't be made, instead of
    exiting.
    
    Also redid the `TestEmptySystemCertPool` test. This test actually
    never worked as intended, because Go caches the system certificates
    behind the scenes, so it only worked if you ran the test by itself.
    The pass condition was not specific enough to catch this failure.
    Reworked the test to work with an interface instead.
    
commit b3fe9dd71905e2c4983fb3a33c339d40bea5e736
Date:   Thu Jan 16 13:14:37 2020

    Fix segfault after temporary file creation fails.
    
    It is an error to access the file object when the TempFile call has
    failed. It won't exist in this case anyway, so the call is not needed.
    
    Changelog: Fix segfault when running `mender setup` on a read-only
    filesystem.
    
```

### DIFF
--- a/cli/commands.go
+++ b/cli/commands.go
@@ -52,12 +52,6 @@ type runOptionsType struct {
 	setupOptions setupOptionsType // Options for setup subcommand
 }
 
-const (
-	errMissingServerCertstr = "IGNORING ERROR: The client server-certificate can not be loaded error: (%s). The client will " +
-		"continue running, but will not be able to communicate with the server. If this is not your intention " +
-		"please add a valid server certificate"
-)
-
 func commonInit(config *conf.MenderConfig, opts *runOptionsType) (*app.MenderPieces, error) {
 
 	tentok := config.GetTenantToken()
@@ -195,13 +189,8 @@ func initDaemon(config *conf.MenderConfig,
 
 	controller, err := app.NewMender(config, *mp)
 	if err != nil {
-		// Ignore server certificate error  (See: MEN-2378)
-		if _, ok := errors.Cause(err).(*client.ClientServerCertificateError); ok {
-			log.Warningf(errMissingServerCertstr, err)
-		} else {
-			mp.Store.Close()
-			return nil, errors.Wrap(err, "error initializing mender controller")
-		}
+		mp.Store.Close()
+		return nil, errors.Wrap(err, "error initializing mender controller")
 	}
 
 	if opts.bootstrapForce {
--- a/client/client.go
+++ b/client/client.go
@@ -33,10 +33,14 @@ import (
 
 const (
 	apiPrefix = "/api/devices/v1/"
-)
 
-var (
-	errorAddingServerCertificateToPool = errors.New("Error adding trusted server certificate to pool.")
+	errMissingServerCertF = "IGNORING ERROR: The client server-certificate can not be " +
+		"loaded: (%s). The client will continue running, but may not be able to " +
+		"communicate with the server. If this is not your intention please add a valid " +
+		"server certificate"
+	errMissingCerts = "No trusted certificates. The client will continue running, but will " +
+		"not be able to communicate with the server. Either specify ServerCertificate in " +
+		"mender.conf, or make sure that CA certificates are installed on the system"
 )
 
 var (
@@ -276,21 +280,10 @@ func newHttpClient() *http.Client {
 	return &http.Client{}
 }
 
-type ClientServerCertificateError struct {
-	err error
-}
-
-func (s *ClientServerCertificateError) Error() string {
-	return errors.Wrapf(s.err, "cannot initialize server trust").Error()
-}
-
 func newHttpsClient(conf Config) (*http.Client, error) {
 	client := newHttpClient()
 
-	trustedcerts, err := loadServerTrust(conf)
-	if err != nil {
-		return nil, &ClientServerCertificateError{err}
-	}
+	trustedcerts := loadServerTrust(&conf)
 
 	if conf.NoVerify {
 		log.Warnf("certificate verification skipped..")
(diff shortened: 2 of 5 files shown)


kacf:
@estenberg, @oleorhagen, @lluiscampos: Ideally, I would like several eyes on this one, since it is security sensitive. At least check that what I wrote in the commit message makes sense.

cli/commands.go
```diff
@@ -195,13 +189,8 @@ func initDaemon(config *conf.MenderConfig,
 
 	controller, err := app.NewMender(config, *mp)
 	if err != nil {
-		// Ignore server certificate error  (See: MEN-2378)
-		if _, ok := errors.Cause(err).(*client.ClientServerCertificateError); ok {
```
oleorhagen: Why not return an explicit error and catch it in a type-switch like before?

kacf: `loadServerTrust` cannot return error anymore, it always returns a valid list of certificates, even if it is an empty one. We need the list, so the alternative would be to return both a list *and* an error, which I think goes against common Go patterns.

client/client_test.go
```diff
@@ -215,30 +213,30 @@ func TestCaLoading(t *testing.T) {
...
-	err = os.Setenv("SSL_CERT_FILE", tmpdir+"idonotexist.crt") // fakes a non existing cert-file
-	assert.NoError(t, err)
-
 	conf := Config{
 		ServerCert: "server.crt",
 	}
 
-	certs, err := loadServerTrust(conf)
-	assert.NoError(t, err)
-	assert.NotZero(t, certs)
```
lluiscampos: If I understood your code and commit message, with the previous implementation this test should have been expecting `Error`?

client/client.go
```diff
@@ -316,52 +309,56 @@ type Config struct {
...
-		if err == nil {
-			log.Infof("API Gateway certificate (in PEM format): \n%s", string(servcert))
-			log.Infof("Issuer: %s, Valid from: %s, Valid to: %s",
-				cert.Issuer.Organization, cert.NotBefore, cert.NotAfter)
-		}
+		// Ignore server certificate error  (See: MEN-2378)
+		log.Warnf(errMissingServerCertF, err.Error())
 	}
 
 	if syscerts == nil {
```
oleorhagen: I guess this is not necessary anymore then?

client/client.go
```diff
@@ -316,52 +309,56 @@ type Config struct {
...
 		log.Warn("No system certificates found.")
 		syscerts = x509.NewCertPool()
 	}
 
-	syscerts.AppendCertsFromPEM(servcert)
+	if len(servcert) > 0 {
+		block, _ := pem.Decode([]byte(servcert))
```
oleorhagen: Question: `servcert` is already `[]byte`, no?

client/client.go
```diff
@@ -316,52 +309,56 @@ type Config struct {
...
-		// tls module to use system certs.
-		return nil, nil
-	}
+type systemCertPoolGetter interface {
+	GetSystemCertPool() (*x509.CertPool, error)
+}
+
+type systemCertPool struct{}
+
+func (systemCertPool) GetSystemCertPool() (*x509.CertPool, error) {
```
oleorhagen: Suggestion: This is an initializer right?
How about `systemCertPool.New()` instead?

client/client.go
```diff
@@ -316,52 +309,56 @@ type Config struct {
...
+	if len(servcert) > 0 {
+		block, _ := pem.Decode([]byte(servcert))
+		if block != nil {
+			cert, err := x509.ParseCertificate(block.Bytes)
+			if err == nil {
+				log.Infof("API Gateway certificate (in PEM format): \n%s", string(servcert))
+				log.Infof("Issuer: %s, Valid from: %s, Valid to: %s",
+					cert.Issuer.Organization, cert.NotBefore, cert.NotAfter)
+			}
+		}
```
oleorhagen: Question: 
Should we warn on non-PEM encoded certificates?

cli/commands.go
oleorhagen: Suggestion (feel free to ignore): 
Maybe a prettier API would be.
`systemCertPool.New()`
Then
`systemCertPool.Load()`
And the load just doesn't return an error?

estenberg:
@kacf sorry but I am completely useless when it comes to PRs of the client.

Is there any known security issue you'd like input on? If so, if you can phrase it with words / a model I will help and provide input. :)

client/client.go
kacf: No, this is the actual loading code. This calls `x509.SystemCertPool()`, the initializer is `x509.NewCertPool()` (which we don't expose, since we don't need it). For this reason we should avoid the name `New()`.

client/client.go
kacf: It is necessary, because this initializes the list to an empty one if `GetSystemCertPool()` returned `nil`. I should add a test case for that, actually...

client/client.go
kacf: This was preexisting, but I can try to remove it.

client/client.go
kacf: You're right! This is a warning we should have, I'll add it.

client/client_test.go
kacf: Well, kind of. It's clear that no one ever tested the failure case. `SSL_CERT_DIR` actually makes no difference in this test, because `ServerCert: "server.crt` is a valid certificate, and hence you will get `NoError` no matter what you put there. I guess it was attempting to verify that this would always be successful.

But it did not test what would happen if both certificate sources are empty, which is the new case I added below. This would have been `Error` with the old code, but is now `len == 0` instead.

And yet a complicating factor is that using `NotZero` was useless here, as even a failed call to `SystemCertPool` returns a non-nil (but empty) map, and hence `NotZero` would return true in all cases.

kacf:
@estenberg: I guess the main feedback I want is on this block in the topmost commit:

> Fix crash when specified certificate can't be opened.
> Both the `ServerCertificate` setting and the system certificates are now optional, in the sense that the client will run without them. However, the client will not be able to connect without the right certificates, so the main usecase of this change is to have a workable client that will roll back if connections can't be made, instead of exiting.

From a security perspective I cannot see anything wrong with this logic. Can you?

client/client.go
oleorhagen: Alright, I misunderstood. Thank you for the explanation.

estenberg:
> @estenberg: I guess the main feedback I want is on this block in the topmost commit:
> 
> > Fix crash when specified certificate can't be opened.
> > Both the `ServerCertificate` setting and the system certificates are now optional, in the sense that the client will run without them. However, the client will not be able to connect without the right certificates, so the main usecase of this change is to have a workable client that will roll back if connections can't be made, instead of exiting.
> 
> From a security perspective I cannot see anything wrong with this logic. Can you?

This looks fine to me. As long as nobody can "inject" a certificate (e.g. it would use some insecure default certificate if none are found, or accept all certificates if none are provided) this should be fine. 

I would expect the TLS implementation to simply error out if the received certificate can not be verified (also if no root certificate store exists). Something else would seem like a pretty serious security bug in the TLS implementation/library.

kacf:
Yes, there is a separate setting to accept all certificates, merely using an empty list does not trigger this behavior, so it should be safe.

PR #484, opened 2020-01-22
2.1.x cherry-picks

kacf:


kacf:
The yellow bullet is from an earlier build. Merging.

PR #488, opened 2020-01-27
FIX: Main thread freezes if watchdog expires.

kacf:
changelog: none

Based on a patch by: Alf-Rune Siqveland <alf.rune@northern.tech>


kacf:
@alfrunes: Check this out. Your [refactored branch](https://github.com/mendersoftware/mender/pull/480) failed in a variety of ways which the old branch didn't. So I backported the freeze fix instead of using it as is. I think it's correct, since the other part you changed was already correct in the old branch, so the change became smaller. This passed with a manual run on a Gitlab runner when I tried it, but I will start one test run from scratch also to be sure.

PR #490, opened 2020-01-29
Port to new and smaller cli package

alfrunes:
Request for comment: what do you think?
The binary size is reduced significantly as this cli package only depends on built-in golang packages (in addition to the unix/windows packages in the golang `x` repository).


alfrunes:
This PR is branched on top of #489, only the last three commits provide changes.

oleorhagen:
I think this should be discussed in the ticket: https://tracker.mender.io/browse/MEN-3065 if this is really what we want to do.

And if we go with this? Maybe it should be a library of it's own, and standardized across all our CLI's?

oleorhagen:
Quickly had a look over the cli lib though, and I think it looks nice! Kudos!

kacf:
I will look more at this once we have the base snapshot feature merged.

alfrunes:
@kacf ?

kacf:
Oops, sorry, I didn't mean to close this. It was in my list of tabs "related to" the snapshot feature and I closed it by mistake!

kacf:
Wait, that's not what happened. It was auto-closed because I deleted the feature branch! You need to reopen it against the master branch now that we merged the main feature. :-)

PR #492, opened 2020-01-30
MEN-2587: Added Client provides to deployments/next API call

oleorhagen:
This will add the device provides from the datastore, encode them in the request
body, and send them along with the deployments/next API call, in which the
deployments endpoint will decide upon what to do. If it is to treat it like an
old API call, or handle the provides is up to the backend.

Changelog: None

### DIFF
--- a/app/mender.go
+++ b/app/mender.go
@@ -330,11 +330,11 @@ func verifyArtifactDependencies(depends, provides map[string]interface{}) error 
 func (m *Mender) CheckUpdate() (*datastore.UpdateInfo, menderError) {
 	currentArtifactName, err := m.GetCurrentArtifactName()
 	if err != nil || currentArtifactName == "" {
-		log.Error("could not get the current artifact name")
+		log.Error("could not get the current Artifact name")
 		if err == nil {
 			err = errors.New("artifact name is empty")
 		}
-		return nil, NewTransientError(fmt.Errorf("could not read the artifact name. This is a necessary condition in order for a mender update to finish safely. Please give the current artifact a name (This can be done by adding a name to the file /etc/mender/artifact_info) err: %v", err))
+		return nil, NewTransientError(fmt.Errorf("could not read the Artifact name. This is a necessary condition in order for a Mender update to finish safely. Please give the current Artifact a name (This can be done by adding a name to the file /etc/mender/artifact_info) err: %v", err))
 	}
 
 	deviceType, err := m.GetDeviceType()
@@ -640,9 +640,9 @@ func (m *Mender) InventoryRefresh() error {
 	artifactName, err := m.GetCurrentArtifactName()
 	if err != nil || artifactName == "" {
 		if err == nil {
-			err = errors.New("artifact name is empty")
+			err = errors.New("Artifact name is empty")
 		}
-		errstr := fmt.Sprintf("could not read the artifact name. This is a necessary condition in order for a mender update to finish safely. Please give the current artifact a name (This can be done by adding a name to the file /etc/mender/artifact_info) err: %v", err)
+		errstr := fmt.Sprintf("could not read the artifact name. This is a necessary condition in order for a Mender update to finish safely. Please give the current Artifact a name (This can be done by adding a name to the file /etc/mender/artifact_info) err: %v", err)
 		return errors.Wrap(errNoArtifactName, errstr)
 	}
 
--- a/app/mender.go
+++ b/app/mender.go
@@ -324,9 +324,9 @@ func verifyArtifactDependencies(depends, provides map[string]interface{}) error 
 	return nil
 }
 
-// Check if new update is available. In case of errors, returns nil and error
-// that occurred. If no update is available *UpdateInfo is nil, otherwise it
-// contains update information.
+// CheckUpdate Check if new update is available. In case of errors, returns nil
+// and error that occurred. If no update is available *UpdateInfo is nil,
+// otherwise it contains update information.
 func (m *Mender) CheckUpdate() (*datastore.UpdateInfo, menderError) {
 	currentArtifactName, err := m.GetCurrentArtifactName()
 	if err != nil || currentArtifactName == "" {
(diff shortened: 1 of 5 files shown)


oleorhagen:
I think this warrants an integration test also actually.

app/mender.go
```diff
@@ -324,27 +324,33 @@ func verifyArtifactDependencies(depends, provides map[string]interface{}) error
...
-// that occurred. If no update is available *UpdateInfo is nil, otherwise it
-// contains update information.
+// CheckUpdate Check if new update is available. In case of errors, returns nil
+// and error that occurred. If no update is available *UpdateInfo is nil,
+// otherwise it contains update information.
 func (m *Mender) CheckUpdate() (*datastore.UpdateInfo, menderError) {
 	currentArtifactName, err := m.GetCurrentArtifactName()
 	if err != nil || currentArtifactName == "" {
-		log.Error("could not get the current artifact name")
+		log.Error("could not get the current Artifact name")
```
kacf: These changes should be in a separate commit.

client/client_update.go
```diff
@@ -189,12 +190,19 @@ func makeUpdateCheckRequest(server string, current CurrentUpdate) (*http.Request
...
+
+	r := bytes.NewBuffer(providesBody)
+
 	ep := "/deployments/device/deployments/next"
 	if len(vals) != 0 {
 		ep = ep + "?" + vals.Encode()
 	}
 	url := buildApiURL(server, ep)
-	req, err := http.NewRequest(http.MethodGet, url, nil)
+	req, err := http.NewRequest(http.MethodGet, url, r)
```
kacf: The new API call is a POST request. You need to try the POST first, and fall back to GET if you get a 404. The latter should not have the artifact_provides (Enterprise only).

lluiscampos:
Please do not merge this one until we have completely merged snapshot feature (pull #487)

oleorhagen:
@kacf Should be better now. I promise.

And another task to your review backlog. I'm sure you're excited.

client/client_update.go
```diff
@@ -64,22 +66,44 @@ func (u *UpdateClient) GetScheduledUpdate(api ApiRequester, server string,
...
+		log.Debugf("Failed sending device provides to the backend: Error: %v", err)
+		return nil, errors.Wrapf(err, "enterprise update check request failed")
 	}
 
 	defer r.Body.Close()
 
+	if r.StatusCode != http.StatusOK && r.StatusCode != http.StatusNoContent {
+		if r.StatusCode == http.StatusNotFound {
+			// 404 - Fall back to the old GET request
+			log.Debug("device provides not accepted by the server")
```
kacf: I would rather say "POST request not accepted by the server"

client/client_update.go
```diff
@@ -64,22 +66,44 @@ func (u *UpdateClient) GetScheduledUpdate(api ApiRequester, server string,
...
-		return nil, errors.Wrapf(err, "update check request failed")
+		log.Debugf("Failed sending device provides to the backend: Error: %v", err)
+		return nil, errors.Wrapf(err, "enterprise update check request failed")
 	}
 
 	defer r.Body.Close()
 
+	if r.StatusCode != http.StatusOK && r.StatusCode != http.StatusNoContent {
+		if r.StatusCode == http.StatusNotFound {
+			// 404 - Fall back to the old GET request
```
kacf: @kjaskiewiczz: What do you think about this logic? The idea is that after the `deployments/next` POST request, it will fall back to a GET request if, and only if, the status code is 404.

Which was my suggestion. But now I'm a bit uneasy. Is it resilient enough? Can proxies mess this up, for example, by returning a different error code? What do you think?

client/test/server.go
```diff
@@ -317,17 +318,25 @@ func (cts *ClientTestServer) updateReq(w http.ResponseWriter, r *http.Request) {
 	log.Infof("got update request %v", r)
 	cts.Update.Called = true
 
+	// Enterprise client device provides post is not supported yet
+	if r.Method == "POST" {
+		w.WriteHeader(404)
+		return
+	}
```
kacf: It would be nice to test that this fallback works.

oleorhagen: I would say this is actually tested in: https://github.com/mendersoftware/mender/blob/7e716aee59d0f7ff77008e3bb2f4f52906644e8c/app/mender_test.go#L303

kacf: Doh! Of course it is, I was thinking of it backwards. :-P

client/client_update.go
kjaskiewiczz: the logic itself looks ok, but I don't understand the idea of falling back to the GET request

oleorhagen: @kjaskiewiczz how would you want it done from the client side?

kacf: > I don't understand the idea of falling back to the GET request

@kjaskiewiczz: The client doesn't know whether it's talking to an Enterprise or Open Source server. It has to fall back to the Open Source GET method to keep working with Open Source servers.

kjaskiewiczz: right! thanks for clarification
proxies shouldn't modify 404 response

kacf: Ok, thanks!

oleorhagen:
This was run off of integration test master, but a branch `artifact-depends-i4` has now been spun off of integration master for further use with this feature branch.

kacf:
There is something weird with the history here. Are you sure the `artifact-depends-i4` branch is up to date with master?

oleorhagen:
> There is something weird with the history here. Are you sure the `artifact-depends-i4` branch is up to date with master?

I am sure it is not. 

So I guess you'd rather want the feature branch merged with master, and the PR clean?

I did it this way, since there is nothing in the feature branch already, so I don't really see it matters that much. Also created a matching branch in integration now, for peace and quiet.

kacf:
> So I guess you'd rather want the feature branch merged with master, and the PR clean?

It doesn't have to be. But you can't both have the feature branch behind master, and base your own work on top of master. Either your work needs to be based on top of the feature branch, or the feature branch must merge master into it.

client/client_update.go
```diff
@@ -86,7 +86,7 @@ func (u *UpdateClient) getUpdateInfo(api ApiRequester, process RequestProcessing
 	defer r.Body.Close()
 
 	if r.StatusCode != http.StatusOK && r.StatusCode != http.StatusNoContent {
-		if r.StatusCode == http.StatusNotFound {
+		if r.StatusCode == http.StatusNotFound || r.StatusCode == http.StatusMethodNotAllowed {
```
kacf: This commit makes me uneasy. Don't get me wrong, this is correct, but my concern earlier about only falling back on certain error codes is already being slowly confirmed. Maybe we should just fall back on at least all of the 4xx error codes. The risk of the Open Source version breaking is really high here... What do you think?

oleorhagen: Most definitely! What I thought about initially. But since you wanted 404 explicitly I figured this was safe. I'll let the tests run this time though, so I don't have to wait an entire new run.

oleorhagen: How about just falling back on all codes not 200, just to be sure?

oleorhagen: I mean, `200` and `204` is handled on the above line already right? So let's just always fall back in any other case.

oleorhagen:
@kacf settled for the `fail on all error codes approach`. Can you have a last lookover?

PR #493, opened 2020-02-03
Snapshot continued

kacf:
Mainly resubmit of #489, with some fixes on top.


kacf:
@alfrunes, @oleorhagen: Whoever gets time to review first, only the [last commit](https://github.com/mendersoftware/mender/pull/493/commits/cc399319b6fb899a72b413e8d2be5263473fdb79) is new, the rest has been reviewed already. At the time of writing, this [passed everything](https://gitlab.com/Northern.tech/Mender/mender-qa/pipelines/114924300) except integration tests, which I believe to be spurious and am rerunning.

kacf:
> My only concern is if this generates any significant overhead - considering the freeze handler is blocked for at every iteration of the copy loop.

Copying from a mounted/frozen filesystem is slower than a non-mounted one, but I can't say if this is due to our code, or that the kernel has more sanity checking to do. In either case, even though the difference was detectable, it was small. This was when I was piping to `/dev/null`, in other words the fastest sink you could have; when piping to the network, the difference would likely be even smaller.

kacf:
Alright, finally passed. The one failure is because of [this commit](https://github.com/mendersoftware/integration/commit/028359f3a450b5083dacea4817787d02a9e657e9), we are testing a slightly out of date client with a new integration test. This will be fixed as soon as we merge, so I won't rebase and run the whole pipeline again just for this.

PR #496, opened 2020-02-06
Rename setup flag --server-cert to --trusted-certs

alfrunes:
The two flags are the same with different names within the two flag
scopes.

changelog: none


kacf:
I think we should keep the original, since we are referring to it as "Server Certificate" everywhere else, including [in Yocto](https://docs.mender.io/2.2/artifacts/yocto-project/building-for-production#including-the-client-certificates) and [the client configuration](https://docs.mender.io/2.2/client-configuration/configuration-file/configuration-options#servercertificate).

If we do want to change it it *must* be done before the final, but my vote would to not do it.

PR #497, opened 2020-02-10
Make interactive setup device type default to configured device type

alfrunes:


cli/setup.go
```diff
@@ -154,12 +154,22 @@ const (
...
+			return "unknown"
+		}
+		devType = string(hostName)
+		devType = strings.Trim(devType, "\n")
+	} else {
+		devType = string(deviceType)
+		idx := strings.Index(devType, "=")
+		if idx >= 0 && len(deviceType) > idx {
+			devType = devType[idx+1:]
+		}
```
lluiscampos: I think you can reuse [GetDeviceType](https://github.com/mendersoftware/mender/blob/master/device/device.go#L143) from `device.go`.

cli/setup.go
```diff
@@ -154,12 +154,22 @@ const (
 // ---------------------------- END Setup constants ----------------------------
 
 func getDefaultDeviceType() string {
-	hostName, err := ioutil.ReadFile("/etc/hostname")
+	var devType string
+	deviceType, err := ioutil.ReadFile(conf.DefaultDeviceTypeFile)
 	if err != nil {
```
lluiscampos: Also, generally speaking, I would expect the logic to represent the preference with something like:
- try reading `conf.DefaultDeviceTypeFile`
- if ok, parse and get the `devType`
- else try `/etc/hostname`
- else unknown

lluiscampos:
Also, I think this small change is claiming for a unit test :wink:

cli/setup.go
```diff
@@ -153,13 +155,17 @@ const (
 
 // ---------------------------- END Setup constants ----------------------------
 
-func getDefaultDeviceType() string {
-	hostName, err := ioutil.ReadFile("/etc/hostname")
+func getDefaultDeviceType(ctx *cli.Context) (devType string) {
+	devType, err := device.GetDeviceType(path.
+		Join(ctx.GlobalString("data"), "device_type"))
```
lluiscampos: Is this correct? TBH I don't know exactly what `ctx.GlobalString("data")` is, but I thought we can use `conf.DefaultDeviceTypeFile` (used already below at `saveConfigOptions`).

alfrunes: It's the value from the `--data` flag which gives the path to the directory containing the `device_type` file. It defaults to the same value, I just figured it might be a "more correct" value in case the flag is set.

lluiscampos: Thanks for the clarification :+1:

PR #498, opened 2020-02-11
Make device type example match an actual real life example.

kacf:
Changelog: None


kacf:
Cherry-picked to 2.2.x as well.

PR #499, opened 2020-02-18
MEN-3065: Shrink cli library

lluiscampos:
Switch to forked urfave/cli and make use of new "minimal" version

The fork contains a patch with which we remove the text/template
dependency with compile time flags. Modified Makefile accordingly.

ATM this fork is on lluiscampos personal GitHub space, it will be moved
to mendersoftware once we have the OK of the team.

See https://github.com/lluiscampos/cli/commit/b850ab5be871e67785792371878f996576462e63

Changelog: Remove text/template dependency from the cli library reducing
mender client binary size by approximately 20%


oleorhagen:
I don't have time to look this over now. But from just scrolling through it, it looks real nice! And if switching to v2,  and reducing the binary size by 20%, this is not much to pay!

I like it a lot at first glance

lluiscampos:
> I like it a lot at first glance

Thanks!

But don't fool yourself: the actual removal of the library is not pretty. So please have a closer look to the fork work and let's discuss possible improvements there.

vendor/github.com/urfave/cli/help_minimal.go
```diff
@@ -0,0 +1,399 @@
...
+func checkSubcommandHelp(c *Context) bool {
+	if c.Bool("h") || c.Bool("help") {
+		_ = ShowSubcommandHelp(c)
+		return true
+	}
+
+	return false
+}
+
+func checkShellCompleteFlag(a *App, arguments []string) (bool, []string) {
```
alfrunes: I don't believe we have completion enabled, and it's kind of awkward how you need to generate them in the first place, so I wouldn't bother...

vendor/github.com/urfave/cli/help_minimal.go
```diff
@@ -0,0 +1,399 @@
...
+   [global options] command [command options] [arguments...]
+
+VERSION:
+   %s
+
+DESCRIPTION:
+   %s
+
+COMMANDS:
+%s
```
alfrunes: Shouldn't there be an indentation here and below?

lluiscampos: Thank for for reviewing this!

The indentation is actually [here](https://github.com/mendersoftware/mender/pull/499/files#diff-0999ccdefedfd9b0aa22020c07ffdc51R298) and [here](https://github.com/mendersoftware/mender/pull/499/files#diff-0999ccdefedfd9b0aa22020c07ffdc51R302).

vendor/github.com/urfave/cli/help_minimal.go
lluiscampos: For this `help_minimal.go`, I replaced printHelpCustom with printHelpCustomNoTemplate but left the rest of the file intact. It does not have a real impact and I thought this would help maintenance, as a the git diffs will be easier to apply to both files.

lluiscampos:
Actually, I just realized we have a problem with the dependencies management, see Pipeline failure.

The issue is that urfave/cli v2 requries go modules, as it is using versioned imported paths. I tried but it doesn't seem to be easy to workaround this with `govendor`.

Unless someone shows me how, I will need to slightly rework this PR to base the fork on v1.x and drop the v2 update changes.

As for an update to mender code base to use gomodules, this is something that we should do soon but for master, not to be cherry-picked for the final Mender 2.3.0 release

vendor/github.com/urfave/cli/help_minimal.go
```diff
@@ -0,0 +1,399 @@
...
+   %s [command options] [options]
+
+OPTIONS:
+%s
+`
+
+// printHelpCustomNoTemplate is the default implementation of HelpPrinterCustom.
+//
+// Substitutes printHelpCustom from help.go removing templates support
+func printHelpCustomNoTemplate(out io.Writer, _ string, data interface{}, _ map[string]interface{}) {
```
kacf: If you added this function, I propose putting it in a new file to truly minimize the diff.

lluiscampos: Good idea. I'll rework

Makefile
```diff
@@ -181,7 +181,7 @@ gofmt:
 
 govet:
 	echo "-- checking with govet"
-	$(GO) vet -unsafeptr=false
+	$(GO) vet $(BUILDTAGS) -unsafeptr=false
```
oleorhagen: I guess all the build invocations will need the build-tags now?

oleorhagen: lol, fixed while I was commenting :p

lluiscampos:
> Right, but this is a more general problem. I suppose more and more libraries will transition to this, so we will need to make the move sooner or later. However, I agree that for our purposes, and especially 2.3.0 final, it is safer and less disruptive to stay on v1.

Latest news in this topic is that @oleorhagen has found a workaround. As we are not really using md2man dependency, as long as we pass the same build tags for `go vet` or `go test`, the checks should succeed and not vendoring md2man should go unnoticed.

Then, as a follow-up we can create a ticket for actually starting using gomodules (both in client and in all backend services, IMO).

With this in place, we could then merge (after rework) this PR, i.e. getting urfave/cli v2. Strictly speaking we shouldn't do this from beta to final, but the upgrade was done in the context of reducing binary size (and an easier to maintain fork).

So if the tests pass now, I vote for merging it with the upgrade.

> Apart from that, I suggest adding some kind of README or something to document how the fork was made and how to update it.

:+1:

cli/cli.go
```diff
@@ -148,18 +148,19 @@ OPTIONS:
...
-					Name:        "forcebootstrap, F",
+				&cli.BoolFlag{
+					Name:        "forcebootstrap",
+					Aliases:     []string{"F"},
 					Usage:       "Force bootstrap.",
 					Destination: &runOptions.bootstrapForce},
 			},
 			Action: func(ctx *cli.Context) error {
-				if len(ctx.Args()) > 0 {
+				if ctx.Args().Len() > 0 {
```
alfrunes: All these checks can be moved down to a single check in `commonCLIHandler` function to reduce the cyclometric complexity of this function.

vendor/github.com/urfave/cli/docs.go
```diff
@@ -1,3 +1,5 @@
+//+build !cli_minimal
```
alfrunes: I would abandon the `cli_minimal` tag and just use `//+build ignore`. This way `make` and `go build` are consistent.

vendor/github.com/urfave/cli/help_minimal.go
```diff
@@ -0,0 +1,399 @@
...
+DESCRIPTION:
+   %s
+
+COMMANDS:
+%s
+GLOBAL OPTIONS:
+%s
+`
+
+var cmdHelpTextFmt = `NAME:
```
alfrunes: This text format should conditionally include a sub-commands section.

main_test.go
```diff
@@ -51,15 +51,15 @@ func TestBinarySize(t *testing.T) {
...
+	const maxSize int64 = 9103464
 
 	cmd := exec.Command("go", "version")
 	version, err := cmd.CombinedOutput()
 	require.NoError(t, err)
 	t.Logf("Go version: %s", string(version))
 
 	programName := "/tmp/mender"
-	cmd = exec.Command("go", "build", "-o", programName)
+	cmd = exec.Command("go", "build", "-tags", "cli_minimal", "-o", programName)
```
alfrunes: I'm not a fan of having build tags in the default go build command. I would rather invert the tag logic or simply use the ignore tag instead.

vendor/github.com/urfave/cli/docs.go
kacf: Agree. Especially `go test -run Something` is really annoying if the tags don't work.

lluiscampos: This is a great suggestion :+1:

cli/cli.go
lluiscampos: Done. I'll push it in the upcoming rework.

main_test.go
lluiscampos: :+1:

vendor/github.com/urfave/cli/help_minimal.go
lluiscampos: I don't understand, do you mean that in a command help I should list the subcommands somehow? I don't see that in the [original template](https://github.com/urfave/cli/blob/master/template.go#L38).

Also, I just realized that our code [here](https://github.com/mendersoftware/mender/blob/master/cli/cli.go#L118) was redefining the subcommand template, but it will not have any effect anymore so I am removing it.

alfrunes: It's a bit confusing, but the SubCommandHelpTemplate is used if the command has any sub-commands: https://github.com/urfave/cli/blob/master/help.go#L190
I'm not sure why they do it like this when they can easily merge the templates, but I would suggest utilizing printf to just conditionally add a command section if any sub-command is present in the struct.

lluiscampos: Thanks!

I think that upstream uses SubCommandHelpTemplate for actually "commands with subcommands". That terminology seems to be all around help.go

Anyway, this was great because I actually found that with the "minimal" version these "commands with subcommands" were wrongly printed, not only missing the subcommands (as I think you were pointing out) but actually using the "app" format. See [this line](https://github.com/lluiscampos/cli/commit/0d6847ae9ffc664d55a3c8bf0d3395e9fb127c15#diff-c403ca73c89e0b507f69b2d17779ab6eR78-R79) to better understand what I mean.

lluiscampos:
I just pushed the last rework which includes all your suggestions. Please have a second look at it.

Testing it together with https://github.com/mendersoftware/meta-mender/pull/923 in [Pipeline 119287767](https://gitlab.com/Northern.tech/Mender/mender-qa/pipelines/119287767).

After your review (and tests passing), I will move the fork to mendersoftware, and update this PR accordingly.

lluiscampos:
Testing again in [Pipeline 119599807](https://gitlab.com/Northern.tech/Mender/mender-qa/pipelines/119599807)

cli/cli.go
lluiscampos: Please have a close look at [this commit](https://github.com/mendersoftware/mender/pull/499/commits/b5f95c79e3188d3c93ef5278d72adc87e0c4a210).

main_test.go
kacf: This can be removed now, can't it?

lluiscampos: Thanks, good catch. I missed cleaning this one in the rework.

oleorhagen:
btw, are when are we moving this to mendersoftware?

lluiscampos:
> Just a tip for other reviewers, viewing the commit "Switch to forked urfave/cli and make use of new "minimal" version" was much easier with this command (you'll have to adapt it to your remotes):
> 
> ```shell
> git show lluiscampos/MEN-3065-shrink-cli-library~3 -C -C -M
> ```

Update: `~4` after the last push

lluiscampos:
> btw, are when are we moving this to mendersoftware?

I just did :)

lluiscampos:
> Nitpick: s/cyclometic/cyclomatic in [b5f95c7](https://github.com/mendersoftware/mender/commit/b5f95c79e3188d3c93ef5278d72adc87e0c4a210)

Thanks! Fixed :+1:

lluiscampos:
Last force-push contains the fork pointing to `mendersoftware`, a new unit test, and few typos fixes. 

IMO this is now ready to be merged (I will let all tests to run one more time).

lluiscampos:
Testing in [Pipeline 120471474](https://gitlab.com/Northern.tech/Mender/mender-qa/pipelines/120471474).

lluiscampos:
Last push is only rebasing on top of latest master, which introduced only  #497. Not rerunning the tests.

vendor/vendor.json
```diff
@@ -146,10 +138,20 @@
...
-			"revision": "e6cf83ec39f6e1158ced1927d4ed14578fda8edb",
-			"revisionTime": "2019-08-04T02:37:52Z"
+			"revision": "2eb686fe967b14b3e204f0eff66427e1fdafcf6d",
+			"revisionTime": "2020-02-18T09:03:34Z",
+			"tree": true,
+			"version": "v2.1.1-minimal",
+			"versionExact": "v2.1.1-minimal"
+		},
+		{
+			"path": "github.com/urfave/cli/v2",
```
alfrunes: Can we remove this entry? The only file that depends on this path is ignored.

vendor/vendor.json
```diff
@@ -146,10 +138,20 @@
 			"revisionTime": "2019-06-13T14:39:42Z"
 		},
 		{
-			"checksumSHA1": "z6OakgyT4iR/xAKnnko9jXdJPlI=",
+			"checksumSHA1": "3HWurU2V4/n3BjF7UpJu132zGrk=",
+			"origin": "github.com/mendersoftware/cli",
 			"path": "github.com/urfave/cli",
```
alfrunes: I would replace the path to `github.com/mendersoftware/cli` as the packages differ.

lluiscampos: Do you mean modifying then all the `import` statements as well? I don't know what the recommended practice is here. Or package is just a fork, so I thought the best would be to respect original path.

vendor/vendor.json
lluiscampos: Sure, `govendor` added this automatically.

PR #500, opened 2020-02-24
Fixes various logging nitpicks

oleorhagen:
Some minor things which annoyed me when using Mender myself.

* Some log messages start with a lowercase, whilst others do not
* Wording in some places, uses words like we
* Minor wording choices

Changelog: Title

### DIFF
--- a/statescript/executor.go
+++ b/statescript/executor.go
@@ -79,7 +79,7 @@ func (l Launcher) getTimeout() time.Duration {
 	if l.Timeout != 0 {
 		return time.Duration(l.Timeout) * time.Second
 	}
-	log.Debugf("statescript: timeout for executing scripts is not defined; using default of %s seconds", defaultStateScriptTimeout)
+	log.Debugf("statescript: The timeout for executing scripts is not defined; using default of %s seconds", defaultStateScriptTimeout)
 	return defaultStateScriptTimeout
 }
 
@@ -188,7 +188,7 @@ func (l Launcher) get(state, action string) ([]os.FileInfo, string, error) {
 			if len(file.Name()) == len(re.FindString(file.Name())) {
 				scripts = append(scripts, file)
 			} else {
-				log.Warningf("script format mismatch: '%s' will not be run ", file.Name())
+				log.Warningf("Script format mismatch: '%s' will not be run ", file.Name())
 			}
 		}
 	}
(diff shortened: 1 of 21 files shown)


oleorhagen:
Really not much to it. Commit message says it all.

statescript/executor.go
```diff
@@ -243,9 +243,9 @@ func execute(name string, timeout time.Duration) error {
 
 	if len(bts) > 0 {
 		if len(bts) > 10*1024 {
-			log.Errorf("stderr collected while running script %s [%s] (Truncated to 10KB)", name, bts[:10*1024])
+			log.Errorf("The standard-error collected while running script %s [%s] (Truncated to 10KB)", name, bts[:10*1024])
 		} else {
-			log.Errorf("stderr collected while running script %s [%s]", name, string(bts))
+			log.Errorf("The standard-error output collected while running script %s [%s]", name, string(bts))
```
lluiscampos: To be consistent, remove "output" from the second case

statescript/executor.go
```diff
@@ -79,7 +79,7 @@ func (l Launcher) getTimeout() time.Duration {
 	if l.Timeout != 0 {
 		return time.Duration(l.Timeout) * time.Second
 	}
-	log.Debugf("statescript: timeout for executing scripts is not defined; using default of %s seconds", defaultStateScriptTimeout)
+	log.Debugf("Statescript: The timeout for executing scripts is not defined; using default of %s seconds", defaultStateScriptTimeout)
```
lluiscampos: I think these are weird. I would suggest using `statescript` (as it was), `StateScript` or `State script`.

statescript/executor.go
```diff
@@ -243,9 +243,9 @@ func execute(name string, timeout time.Duration) error {
 
 	if len(bts) > 0 {
 		if len(bts) > 10*1024 {
-			log.Errorf("stderr collected while running script %s [%s] (Truncated to 10KB)", name, bts[:10*1024])
+			log.Errorf("The standard-error collected while running script %s [%s] (Truncated to 10KB)", name, bts[:10*1024])
```
kacf: Usage of "The" here is weird, I think the original was better. But maybe a compromise? `Stderr messages collected...`

oleorhagen: How about: `Collected stderr while running...` ?

kacf: :+1:

oleorhagen:
Latest pipeline along with integration/832: https://gitlab.com/Northern.tech/Mender/mender-qa/pipelines/126121288

PR #502, opened 2020-02-28
Fix "State transition loop detected" when retrying status update.

kacf:
This solves the issue where retrying the status report many times
increases the state loop counter too far and we detect it as a state
loop, even though Mender would actually have finished a few states
later.

This will cause real state loops to take much longer to detect, but
they will *eventually* be detected. Real state loops are quite rare,
since they can only happen either if the system is constantly
rebooting spontaneously, or if an ArtifactVerifyRollbackReboot keeps
returning failure. The spontaneous reboot flow diagram in the Update
Module documentation is a good guide for understanding how it can
happen.

Changelog: Title

(cherry picked from commit eb15ff94bf4ed57c866d8f04d384e024e1696047)


kacf:
Cancelling acceptance/integration tests, since this change is unlikely to affect them, and we will rerun them with our upcoming build tag anyway.

PR #505, opened 2020-03-09
Removed scopestack and mendersoftware/log

oleorhagen:


oleorhagen:
Related PR: https://github.com/mendersoftware/meta-mender/pull/942

oleorhagen:
Pipeline: https://gitlab.com/Northern.tech/Mender/mender-qa/pipelines/124524325

cli/cli.go
```diff
@@ -346,8 +346,7 @@ func SetupCLI(args []string) error {
 		&cli.StringFlag{
 			Name:    "log-modules",
 			Aliases: []string{"m"},
-			Usage: "`LIST` of logging modules (levels) to " +
-				"include in the output.",
+			Usage: "Warning: log-modules is deprecated.",
```
kacf: Maybe "warning" is a bit strong since this is only the help screen. I would also add for clarity that it has no effect.

cli/cli.go
```diff
@@ -577,10 +579,27 @@ func (runOptions *runOptionsType) handleLogFlags(ctx *cli.Context) error {
...
+		// mendersoftware/log addSysLogHook. The code there also had a weird comment
+		// about setting up the syslog hook which I did not understand:
+		//
+		// comment:
+		// Add the syslog hook to the logger. This is better than adding it directly,
+		// for the reasons described in the loggingHookType comments.
+		//
+		// ==============================================================================
+		hook, err := logrus_syslog.NewSyslogHook("", "",
+			syslog.LOG_DEBUG, "mender")
```
kacf: Why can't you reuse the global log level? I think logging at debug level is not a good idea, because this will flood device logs.

cli/cli.go
```diff
@@ -577,10 +579,27 @@ func (runOptions *runOptionsType) handleLogFlags(ctx *cli.Context) error {
...
+		// NOTE: Always logging at the debug level to the syslog
+		//
+		// ==============================================================================
+		// REVIEWER-NOTE: This was changed slightly from the existing functionality in:
+		// mendersoftware/log addSysLogHook. The code there also had a weird comment
+		// about setting up the syslog hook which I did not understand:
+		//
+		// comment:
+		// Add the syslog hook to the logger. This is better than adding it directly,
+		// for the reasons described in the loggingHookType comments.
```
kacf: I probably wrote it, and I still don't understand it! :laughing:

cli/cli_test.go
```diff
@@ -183,7 +184,7 @@ func TestLoggingOptions(t *testing.T) {
 	err = SetupCLI([]string{"mender", "-no-syslog"})
 	// Just check that the flag can be specified.
 	assert.True(t, err == nil)
-	assert.False(t, strings.Contains(buf.String(), "syslog"))
+	// assert.False(t, strings.Contains(buf.String(), "syslog"))
```
kacf: Forgotten TODO?

cli/cli_test.go
```diff
@@ -196,10 +197,10 @@ func TestVersion(t *testing.T) {
 	// pretend we're stdout now
 	os.Stdout = tfile
 
-	// running with stdout pointing to temp file
+	// running with stderr pointing to temp file
```
kacf: Uhm, no..?

cli/cli_test.go
```diff
@@ -196,10 +197,10 @@ func TestVersion(t *testing.T) {
...
 	err = SetupCLI([]string{"mender", "-version"})
 
-	// restore previous stdout
+	// restore previous stderr
```
kacf: And here.

cli/cli_test.go
```diff
@@ -430,37 +431,13 @@ func TestGetMenderDaemonPID(t *testing.T) {
...
-	// Test with failing init daemon
-	set := flag.NewFlagSet("", 0)
-	command := &cli.Command{Name: "daemon"}
-	ctx := cli.NewContext(nil, set, nil)
-	ctx.Command = command
-	runOpts := runOptionsType{
-		logOptions: logOptionsType{logLevel: "info"},
-	}
-	assert.Error(t, runOpts.handleCLIOptions(ctx))
-}
```
kacf: Why was this removed? Is it not appropriate anymore?

cli/cli_test.go
```diff
@@ -430,37 +431,13 @@ func TestGetMenderDaemonPID(t *testing.T) {
...
 // Tests that the client will boot with an error message in the case of an invalid server certificate.
 func TestInvalidServerCertificateBoot(t *testing.T) {
 	tdir, err := ioutil.TempDir("", "invalidcert-test")
 	require.Nil(t, err)
 
 	logBuf := bytes.NewBuffer(nil)
-	defer func(oldLog *log.Logger) { log.Log = oldLog }(log.Log) // Restore standard logger
-	log.Log = log.New()
+	var hook = logtest.NewGlobal()
 	log.SetLevel(log.WarnLevel)
```
kacf: Do you need `defer hook.Reset()` here?

cli/cli.go
oleorhagen: Yes, that's what I would think also. But this is how it is now. I thought it weird also, so figured I'd see if anyone knew the reason. 

Anywho, since u also think it weird, changing it. 

This was also reported on MH I think. I think I even created a ticket for it.

Here it is: https://tracker.mender.io/browse/MEN-3135

Great opportunity for two birds with one stone here :raised_back_of_hand:

cli/cli_test.go
oleorhagen: wtf. Wonder how this got in here

cli/cli_test.go
oleorhagen: This seemed to me to be tested in `daemon_test.go` also. So I saw no need for it  to be duplicated in `cli_test`.

cli/cli_test.go
oleorhagen: Yes, you're right. Should probably be here. Although not stricly necessary, as the hook is only used in this function. I'm still undecided whether or not we should make some sort of utility test logger package, or maybe a test-global hook, or keep it like this. 

Thoughts?

cli/cli.go
oleorhagen: So, something along the lines of:
`Note: log-modules have no effect`?

cli/cli_test.go
oleorhagen: brain-fart?

cli/cli.go
oleorhagen: Okay, so I've looked into this a bit closer, and I still can't figure out how this is supposedly wrong.

So, my findings:

The `syslog.LOG_DEBUG`, is the priority given to the syslog message, with the std, `LOG_KERNEL` as the facility into which it logs.

But, this does not mean that this controls what is written to the syslog, only the severity given to it. In fact, the syslog hook in `logrus` will log all messages, no matter what is set as `log.Level`, or what the `syslog.Level` is here in the line above.

So this brings the question. Do we want to change this?

oleorhagen: This is the `logrus.syslog` hook:
```
func (hook *SyslogHook) Levels() []logrus.Level {
	return logrus.AllLevels
}
```
And this is the go syslog doc:
https://godoc.org/log/syslog#Priority

log/syslog/syslog.go
```diff
@@ -0,0 +1,46 @@
...
+	"log/syslog"
+
+	"github.com/sirupsen/logrus"
+	logrus_syslog "github.com/sirupsen/logrus/hooks/syslog"
+)
+
+type SyslogHook struct {
+	loglevel logrus.Level
+	*logrus_syslog.SyslogHook
+}
```
oleorhagen: @kacf This is all it takes (: 
Go isn't that bad :1st_place_medal:

cli/cli.go
kacf: Good investigation! Ideally, yes, we should change it, but I think you should keep that work out of this pull request, since it's not really related; this has always been like this, right?

cli/cli_test.go
kacf: :+1:

cli/cli_test.go
kacf: Yes, but it has global effects, doesn't it? So resetting it is important.

I don't think we should make a global test hook, because the test functions have different requirements. I think this particular call is better to do per test.

log/syslog/syslog.go
kacf: You're right, it works "by accident", because we are lucky (or because the guy knows this limitation), and [`Fire()` doesn't call `Levels()` anywhere](https://github.com/sirupsen/logrus/blob/master/hooks/syslog/syslog.go#L28-L51). If it had, it would not have worked. :-)

But this means you have already fixed it then, and my previous comment no longer applies?

Just to be clear, we still log everything at one level inside the syslog, is that right? This only controls which ones of the Go-side levels get logged at all?

cli/cli.go
kacf: Maybe `-log-modules is accepted for compatibility with old versions but has no effect`

log/syslog/syslog.go
oleorhagen: > You're right, it works "by accident", because we are lucky (or because the guy knows this limitation), and [`Fire()` doesn't call `Levels()` anywhere](https://github.com/sirupsen/logrus/blob/master/hooks/syslog/syslog.go#L28-L51). If it had, it would not have worked. :-)
> 

As far as I can tell, the `Levels` call are only used when the hook is loaded. So the levels is sort of: `This hook will subscribe to these log-levels. Then when that log-level is called, the logger goes through a map with the current log-level called. If this map has a hook pointer, then that hook is fired. Therefore, with this change, the syslog should not be called for levels below what is set by the user on the CLI now.`

> But this means you have already fixed it then, and my previous comment no longer applies?
> 
This should fix it yes :)

> Just to be clear, we still log everything at one level inside the syslog, is that right? This only controls which ones of the Go-side levels get logged at all?

It seems to me from the `Fire()` function in the syslog-hook that it will actually log at the appropriate levels. So the initializer level is basically a no-op. Except for the `LOG_USER` which we now added.

kacf: :+1:

cli/cli_test.go
oleorhagen: Come to think of it, the effects should be local actually, as the hook can not be reach after it has passed out of scope. So the `defer hook.Reset()`, is actually unnecessary.

kacf: It has [global effects](https://godoc.org/github.com/sirupsen/logrus/hooks/test#NewGlobal), namely changing the global logger. Otherwise it would not be very useful, how would the production code use the hook, otherwise?

oleorhagen: Yes, I thought this way first also. But it only adds the `hook` to the global logger. And the specific hook is only reachable (in this case) within the scope we attach it to the global logger. Which makes it local, since no one else is using this hook.

Does that make sense?

oleorhagen:
Latest pipeline: https://gitlab.com/Northern.tech/Mender/mender-qa/pipelines/126265791

oleorhagen:
Latest pipeline: https://gitlab.com/Northern.tech/Mender/mender-qa/pipelines/126287102

cli/cli_test.go
oleorhagen: Latest commit has the reset added in a defer statement

kacf: Right, I see what you mean. After looking at the actual code, I don't think there is a way to remove the hook actually (I have observed this earlier as well), so after a long test run you may actually have a gazillion hooks attached to the logger that do nothing other than use memory. Not a problem with our code, it's just a design flaw of logrus.

Since you've added it, let's keep the defer for now anyway, but I see your point.

PR #509, opened 2020-03-15
Fix Stat_t.Dev/Rdev type assumption

andhe:
The unix.Major and unix.Minor functions takes an argument of type uint64 and the code currently passes in either stat.Rdev or stat.Dev into these. While stat.Dev and stat.Rdev are often defined as uint64 this is not so for every architecture. See for example mips64el where the fields instead are uint32 [1]. This causes a build failure (atleast with golang 1.14) on architectures where the assumtion doesn't hold true. This change simply casts the value to uint64 before passing it into unix.Major/Minor to fix the build problem.

[1]: https://sources.debian.org/src/golang-1.14/1.14-2/src/syscall/ztypes_linux_mips64le.go/#L107

Changelog: Title

Signed-off-by: Andreas Henriksson <andreas@fatal.se>

### DIFF
--- a/system/ioctl.go
+++ b/system/ioctl.go
@@ -162,7 +162,7 @@ func GetDeviceIDFromPath(path string) ([2]uint32, error) {
 		unix.S_IFIFO,
 		unix.S_IFSOCK:
 		return [2]uint32{
-			unix.Major(stat.Rdev), unix.Minor(stat.Rdev)}, nil
+			unix.Major(uint64(stat.Rdev)), unix.Minor(uint64(stat.Rdev))}, nil
 
 	// If path refers to a regular file, then st_dev gives the device number
 	// of the underlying block device.
@@ -170,7 +170,7 @@ func GetDeviceIDFromPath(path string) ([2]uint32, error) {
 		unix.S_IFREG,
 		unix.S_IFLNK:
 		return [2]uint32{
-			unix.Major(stat.Dev), unix.Minor(stat.Dev)}, nil
+			unix.Major(uint64(stat.Dev)), unix.Minor(uint64(stat.Dev))}, nil
 	}
 	return [2]uint32{^uint32(0), ^uint32(0)},
 		fmt.Errorf("invalid stat(2) st_mode %04X", devType)


andhe:
A minor semi-related nitpick I noticed. Please note a few lines above where `stat` variable is defined:

`var stat stat` [1](https://sources.debian.org/src/mender-client/2.2.0-2/system/ioctl.go/#L147)

I don't understand where the `stat` *type* comes from. Shouldn't it instead be:

`var stat Stat_t`

???

Note how the line below [takes a *Stat_t argument according to documentation](https://golang.org/pkg/syscall/#Stat):

`if err := sys.Stat(path, &stat); err != nil {` [2](https://sources.debian.org/src/mender-client/2.2.0-2/system/ioctl.go/#L149)

On the other hand the compiler doesn't complain about it so I guess it's just me not understanding it, but anyway I thought I'd mention it as I think defining it to Stat_t would be easier to read/understand.

andhe:
A full build log where this problem triggers can be found at:

https://buildd.debian.org/status/fetch.php?pkg=mender-client&arch=mips64el&ver=2.2.0-2&stamp=1584242225&raw=0

(You can navigate to revelant build logs on debian infastructure via:
 * https://tracker.debian.org/mender-client (replace mender-client with any relevant package name)
 * Click build: logs in Links box in right side of the page.
 * Click on the red/green button in the table for the architecture you're interested in.
   (Alternatively click on `old` in the rightmost field in the table to see older build logs.)

.)

lluiscampos:
Running tests on [Pipeline 126590056](https://gitlab.com/Northern.tech/Mender/mender-qa/pipelines/126590056).

lluiscampos:
> A minor semi-related nitpick I noticed. Please note a few lines above where `stat` variable is defined:
> 
> `var stat stat` [1](https://sources.debian.org/src/mender-client/2.2.0-2/system/ioctl.go/#L147)
> 
> I don't understand where the `stat` _type_ comes from. Shouldn't it instead be:
> 
> `var stat Stat_t`
> 
> ???
> 
> Note how the line below [takes a *Stat_t argument according to documentation](https://golang.org/pkg/syscall/#Stat):
> 
> `if err := sys.Stat(path, &stat); err != nil {` [2](https://sources.debian.org/src/mender-client/2.2.0-2/system/ioctl.go/#L149)
> 
> On the other hand the compiler doesn't complain about it so I guess it's just me not understanding it, but anyway I thought I'd mention it as I think defining it to Stat_t would be easier to read/understand.

I was not part of this, but I found out that it is defined here [[1](https://github.com/mendersoftware/mender/blob/master/system/ioctl_linux.go#L28)] and that it was introduced in this commit [[2](https://github.com/mendersoftware/mender/pull/493/commits/108544abfd087d7da73401459a2b3431ff1d225b)]. I also think that the naming of the type is not clear, but I would leave that out of this PR.

PR #510, opened 2020-03-16
Add artifact_name and device_type to POST /next body.

alfrunes:
I also took the liberty to rebase the branch on top of current master.

### DIFF
--- a/LIC_FILES_CHKSUM.sha256
+++ b/LIC_FILES_CHKSUM.sha256
@@ -3,7 +3,7 @@
 beb140be4cd64599bedc691a55b2729c9cc611a4b9d6ec44e01270105daf18a2  vendor/github.com/mendersoftware/mendertesting/LICENSE
 ceb1b36ff073bd13d9806d4615b931707768ca9023805620acc32dd1cfc2f680  vendor/github.com/mendersoftware/log/LICENSE
 ceb1b36ff073bd13d9806d4615b931707768ca9023805620acc32dd1cfc2f680  vendor/github.com/mendersoftware/scopestack/LICENSE
-beb140be4cd64599bedc691a55b2729c9cc611a4b9d6ec44e01270105daf18a2  vendor/github.com/mendersoftware/mender-artifact/LICENSE
+32714818ad6f98ee0185a52e23a475d89122e3efd2b2c26c733781c28e798c99  vendor/github.com/mendersoftware/mender-artifact/LICENSE
 #
 # BSD 2 Clause license.
 8d427fd87bc9579ea368fde3d49f9ca22eac857f91a9dec7e3004bdfab7dee86  vendor/github.com/pkg/errors/LICENSE
--- a/Makefile
+++ b/Makefile
@@ -12,7 +12,7 @@ V ?=
 PKGS = $(shell go list ./... | grep -v vendor)
 PKGFILES = $(shell find . \( -path ./vendor -o -path ./Godeps \) -prune \
 		-o -type f -name '*.go' -print)
-PKGFILES_notest = $(shell echo $(PKGFILES) | tr ' ' '\n' | grep -v _test.go)
+PKGFILES_notest = $(shell echo $(PKGFILES) | tr ' ' '\n' | grep -v '\(client/test\|_test.go\)' )
 GOCYCLO ?= 15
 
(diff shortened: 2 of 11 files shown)


alfrunes:
NOTE: There are a chain of fixes coming up, I'll ping you again once the PR is ready for review.

utils/compare.go
```diff
@@ -0,0 +1,54 @@
...
+//    Unless required by applicable law or agreed to in writing, software
+//    distributed under the License is distributed on an "AS IS" BASIS,
+//    WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+//    See the License for the specific language governing permissions and
+//    limitations under the License.
+
+package utils
+
+import (
+	"reflect"
```
kacf: I wonder if this package is big. Have you tested if it increases binary size significantly?

We might not need the comparison function to be so generic if all we support is strings and lists of strings.

kacf:
With the rebasing, I take it that it's the last three commits that are yours?

app/mender.go
```diff
@@ -277,45 +282,41 @@ func (m *Mender) FetchUpdate(url string) (io.ReadCloser, int64, error) {
...
-			if depend == "" {
-				continue
-			}
-		default:
-			return errors.Errorf(
-				"Invalid type for dependency with name %s", key)
-		}
 		if p, ok := provides[key]; ok {
 			switch depend.(type) {
+			case []interface{}:
```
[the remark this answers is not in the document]
alfrunes: JSON lists of strings will unmarshal to `[]interface{}` type and need to be explicitly cast to string (the type assertion is done in utils.ElemInSlice function).

utils/compare.go
alfrunes: Comparing the size with master, the unstripped difference is ~2K, where as the stripped diff is approximately 1K.

alfrunes:
^ I was just about to ping you and @oleorhagen and tell you it's ready for review and only the top 3 commits introduce new changes.

utils/compare.go
kacf: Hehe, ok, I believe we can live with that! :stuck_out_tongue_winking_eye:

utils/compare.go
```diff
@@ -0,0 +1,54 @@
...
+	if rSlice.Kind() != reflect.Slice {
+		return false, ErrNotSlice
+	}
+	n := rSlice.Len()
+	for i := 0; i < n; i++ {
+		sliceElem := rSlice.Index(i)
+		if sliceElem.Kind() != rElem.Kind() {
+			return false, ErrInvalidType
+		}
+		if sliceElem.Interface() == rElem.Interface() {
```
kacf: Isn't this equivalent to comparing pointers? In other words, if we compare "foo" and "foo", but they are pointing to different strings in memory, won't this check fail?

app/mender.go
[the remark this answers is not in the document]
kacf: Ok, but will it then ever be an instance of `[]string`, which is the branch below? I know, I'm playing the devil's advocate here, but I'm also just trying to understand! :-)

utils/compare.go
alfrunes: If the underlying type is a slice of pointer, this is true. Not in the general case though. Do you want me to dereference if the slice is pointer type?

app/mender.go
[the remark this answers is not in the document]
alfrunes: I think the []string case only applies to unit tests as far as I can tell.
BTW: I also did a minimal test example with the reflect package, and it turns out it leaves a smaller footprint than "fmt" standard package. So my guess is it's already built-in the go-runtime.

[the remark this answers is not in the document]
kacf: Ah, makes sense. I guess if it's only used for unit tests it could be removed, but it does no harm either. I'll leave the decision to you.

utils/compare.go
kacf: I don't think we could ever get that from JSON data, so no need for that as far as I can tell.

In general, I thought interfaces were always pointers, but I admit my knowledge of this is a bit fuzzy, so I'll take your word for it!

alfrunes: I'll keep it as is. I think it makes the code clearer in terms of what the expected output should be.

PR #515, opened 2020-03-24
Adding download and apply commands

mullerch:
Changelog : Adding download and apply commands

Signed-off-by: Christian Müller <christian.muller@wifx.net>

### DIFF
--- a/app/standalone.go
+++ b/app/standalone.go
@@ -41,6 +41,82 @@ type standaloneData struct {
 	installers               []installer.PayloadUpdatePerformer
 }
 
+func DoStandaloneDownload(device *dev.DeviceManager, updateURI string,
+	clientConfig client.Config, vKey []byte,
+	stateExec statescript.Executor) error {
+
+	var image io.ReadCloser
+	var imageSize int64
+	var err error
+	var upclient client.Updater
+
+	if strings.HasPrefix(updateURI, "http:") ||
+		strings.HasPrefix(updateURI, "https:") {
+		log.Infof("Performing remote update from: [%s].", updateURI)
+
+		var ac *client.ApiClient
+		// we are having remote update
+		ac, err = client.New(clientConfig)
+		if err != nil {
+			return errors.New("Can not initialize client for performing network update.")
+		}
+		upclient = client.NewUpdate()
+
+		log.Debug("Client initialized. Start downloading image.")
+
+		image, imageSize, err = upclient.FetchUpdate(ac, updateURI, 0)
+		log.Debugf("Image downloaded: %d [%v] [%v]", imageSize, image, err)
+	} else {
+		// perform update from local file
+		log.Infof("Start updating from local image file: [%s]", updateURI)
+		image, imageSize, err = installer.FetchUpdateFromFile(updateURI)
+
+		log.Debugf("Fetching update from file results: [%v], %d, %v", image, imageSize, err)
+	}
+
+	if image == nil || err != nil {
+		return errors.Wrapf(err, "Error while installing Artifact from command line")
+	}
+	defer image.Close()
+
+	fmt.Fprintf(os.Stdout, "Installing Artifact of size %d...\n", imageSize)
+	p := &utils.ProgressWriter{
+		Out: os.Stdout,
+		N:   imageSize,
+	}
+	tr := io.TeeReader(image, p)
+
+	standaloneData, err := doStandaloneInstallStatesDownload(ioutil.NopCloser(tr), vKey, device, stateExec)
+	if err != nil {
+		return err
+	}
+
+	err = storeStandaloneData(device.Store, standaloneData)
+	if err != nil {
+		log.Errorf("Could not update database: %s", err.Error())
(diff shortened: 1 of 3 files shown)


app/standalone.go
```diff
@@ -171,6 +247,65 @@ func doStandaloneInstallStatesDownload(art io.ReadCloser, key []byte,
 	return standaloneData, nil
 }
 
+func doStandaloneInstallStatesApply(standaloneData *standaloneData, device *dev.DeviceManager, stateExec statescript.Executor) error {
```
oleorhagen: Looks good :100:

statescript/executor.go
```diff
@@ -243,9 +243,9 @@ func execute(name string, timeout time.Duration) error {
 
 	if len(bts) > 0 {
 		if len(bts) > 10*1024 {
-			log.Errorf("Collected standard-error while running script %s [%s] (Truncated to 10KB)", name, bts[:10*1024])
+			log.Infof("Collected output (stderr) while running script %s (Truncated to 10KB)\n%s\n---------- end of script output", name, bts[:10*1024])
```
oleorhagen: I think this is originally my handy-work actually, hah.

Anywho. Seems reasonable.

:+1:  from me

app/standalone.go
```diff
@@ -41,6 +41,82 @@ type standaloneData struct {
...
+		N:   imageSize,
+	}
+	tr := io.TeeReader(image, p)
+
+	standaloneData, err := doStandaloneInstallStatesDownload(ioutil.NopCloser(tr), vKey, device, stateExec)
+	if err != nil {
+		return err
+	}
+
+	err = storeStandaloneData(device.Store, standaloneData)
```
mullerch: This the main difference with the existing code. This is done directly after download and no more after the 'install'

mullerch:
Thanks for the review. I realize that I should have added some context to the code.

I thought doing the MR in two steps:
- Add the feature
- Consolidate the code

The goal of this is that I did not know how the MR would be received and that an additional feature not touching existing code is not risky to add, where changing the code covered by existing features is more risky. I did not have time for the second part yet 😮 

So for now the code added is essentially existing code duplicated. The 'download' and 'apply' operation where actually already coded into the 'install' operation. The only difference is the moment when the download data is stored in the DB. I'm not sure about the consequences of this change because and don't understand correctly the context. Here it is : https://github.com/mendersoftware/mender/pull/515/files/a02cf67fbf56d24d7089db742e64c36c21d05662#r398008379 

If this change is fine, I think the install operation could use the new download and apply operations and therefore we can remove the code duplication.

oleorhagen:
> Thanks for the review. I realize that I should have added some context to the code.
> 
> I thought doing the MR in two steps:
> 
>     * Add the feature
> 
>     * Consolidate the code
> 
> 
> The goal of this is that I did not know how the MR would be received and that an additional feature not touching existing code is not risky to add, where changing the code covered by existing features is more risky. I did not have time for the second part yet open_mouth
> 
> So for now the code added is essentially existing code duplicated. The 'download' and 'apply' operation where actually already coded into the 'install' operation. The only difference is the moment when the download data is stored in the DB. I'm not sure about the consequences of this change because and don't understand correctly the context. Here it is : https://github.com/mendersoftware/mender/pull/515/files/a02cf67fbf56d24d7089db742e64c36c21d05662#r398008379
> 
> If this change is fine, I think the install operation could use the new download and apply operations and therefore we can remove the code duplication.

I see. 

Yes, I'm not the one making all the decisions, so I will bring it up during the team meeting tomorrow, and then we will set a course for this, and how (if we want it), to install it in the current code also :)

statescript/executor.go
kacf: Could you submit this as a separate PR? This commit is a very obvious improvement which should be accepted immediately, whereas the other code may require more work/discussion. Good to get it out of the way.

mullerch:
@kacf Well if I understand correctly, the download step does write the new image in the inactive rootfs. This does not change the state of the system, so I don't think rollback does event mean something at this time. To me there is no such thing as "next step", the step you want to do just has some contraints, and those contraints must be fulfilled to go on. The constraint of the 'apply' step is to have an update available in the inactive rootfs. The constraint of the 'commit' step is to have an update installed (or 'applied').

mullerch:
I just see that I've made a mistake by using the master branch. As I need it right now, I'll close this PR and add a new on the correct branch when the code is consolidated and reference this one.
Sorry for this.

kacf:
> @kacf Well if I understand correctly, the download step does write the new image in the inactive rootfs. This does not change the state of the system, so I don't think rollback does event mean something at this time. To me there is no such thing as "next step", the step you want to do just has some contraints, and those contraints must be fulfilled to go on. The constraint of the 'apply' step is to have an update available in the inactive rootfs. The constraint of the 'commit' step is to have an update installed (or 'applied').

Ah yes, you're right, it's not needed for the Download stage, which is by design side effect free. Good, that saves quite a bit of complexity.

PR #520, opened 2020-04-07
QA-153: Add Makefile to instrument the client for cover collection

oleorhagen:
This commit adds the ability to build the client with coverage functionality
enabled. That is it downloads the 'gobinarycoverage' tool, which instruments a
regular main package source code for binary coverage analysis. This adds a
function 'coverReport()' to the main file, which reports coverage for the binary
after execution.

However, in order for this to work, the exit paths in the binary need to call
'coverReport()' before exiting. Hence, the binary is patched with the patch file
found in 'patches/', which makes sure that 'coverReport()' is called before

1) The client calls reboot
2) The client exits normally

The target added is 'instrument-binary' and is a phony target, as it does not
build the binary, it is simply a transformational step, before calling 'go
build' like normal.

Note: This will change the source code in the directory, and hence all source
files are left 'dirty' after the 'instrument-binary' target has been run.

Changelog: None

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -25,7 +25,8 @@ export CGO_CFLAGS
 TOOLS = \
 	github.com/fzipp/gocyclo \
 	gitlab.com/opennota/check/cmd/varcheck \
-	github.com/mendersoftware/deadcode
+	github.com/mendersoftware/deadcode \
+	github.com/mendersoftware/gobinarycoverage
 
 VERSION = $(shell git describe --tags --dirty --exact-match 2>/dev/null || git rev-parse --short HEAD)
 
@@ -212,6 +213,12 @@ coverage:
 	fi
 	rm -f coverage-tmp.txt coverage-missing-subtests.txt
 
+instrument-binary:
+	# Patch the client to make it ready for coverage analysis
+	git apply patches/0001-Instrument-Mender-client-for-coverage-analysis.patch
+	# Then instrument the files with the gobinarycoverage tool
+	gobinarycoverage github.com/mendersoftware/mender
+
 .PHONY: build
 .PHONY: clean
 .PHONY: get-tools
@@ -239,3 +246,4 @@ coverage:
 .PHONY: uninstall-modules-gen
 .PHONY: uninstall-systemd
 .PHONY: uninstall-examples
+.PHONY: instrument-binary
--- a/patches/0001-Instrument-Mender-client-for-coverage-analysis.patch
+++ b/patches/0001-Instrument-Mender-client-for-coverage-analysis.patch
@@ -1,0 +1,118 @@
+From 838f6b3a4554e02088b839ff9e760156381c63b5 Mon Sep 17 00:00:00 2001
+From: Ole Petter <ole.orhagen@northern.tech>
+Date: Tue, 7 Apr 2020 09:02:56 +0200
+Subject: [PATCH] Instrument Mender client for coverage analysis
+
+This change is needed to cover all the exit paths of the client, and calculate
+the test-coverage before exiting the binary. The exit paths covered are the
+reboot call, in Reboot-Enter, and the regular os.Exit in main. The reboot call
+is handled by avoiding the regular call-stack through a panic, which is caught
+in the top-level main file, and the os.Exit is simply done after coverage has
+been calculated.
+
+Also sets the directory for storing the coverage files to '/data/mender/'.
+
+Changelog: None
+
+Signed-off-by: Ole Petter <ole.orhagen@northern.tech>
+---
+ main.go          | 39 ++++++++++++++++++++++++++++++++++++++-
+ system/system.go | 13 +------------
+ 2 files changed, 39 insertions(+), 13 deletions(-)
+
+diff --git a/main.go b/main.go
+index 7f46c64..8c4d540 100644
+--- a/main.go
++++ b/main.go
+@@ -16,12 +16,29 @@ package main
+ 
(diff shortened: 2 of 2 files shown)


oleorhagen:
@kacf @lluiscampos This is the latest version of the coverage instrumentation of the client. 

I think that the commit description should cover pretty much everything, but not that much in depth.

It is probably smart to have a look at the tool itself at (gobinarycoverage)[https://github.com/oleorhagen/gobinarycoverage]

The gist of it is:

1) Modify the source code by running the tool, and applying a minor patch
2) Build just like you normally would.

Hence, this is an added (optional) transformational intermediary step before building the binary, just like normal.

patches/0001-Instrument-Mender-client-for-coverage-analysis.patch
```diff
@@ -0,0 +1,118 @@
...
++			if s, ok := r.(string); ok && s == "Client needs reboot!" {
++				coverReport()
++				rebootClient()
++			} else {
++				panic(r)
++			}
++		}
++	}()
++	// Set the path in which to store the coverage analysis files
++	os.Setenv("COVERAGE_FILEPATH", "/data/mender/")
```
kacf, line 75: For local debugging, would it make sense to set this conditionally? IOW only if not already set?

Makefile
```diff
@@ -25,7 +25,8 @@ export CGO_CFLAGS
 TOOLS = \
 	github.com/fzipp/gocyclo \
 	gitlab.com/opennota/check/cmd/varcheck \
-	github.com/mendersoftware/deadcode
+	github.com/mendersoftware/deadcode \
+	github.com/oleorhagen/gobinarycoverage
```
kacf, line 29: Should change this to `mendersoftware` before merging.

oleorhagen:
> This is impressive! A far more advanced task than we first envisioned!
> 

Thanks :)

> I have to admit that the `gobinarycoverage` tool is a bit beyond me, way too many parsing specifics for me to review in any detail, though the interface looks sane. So instead of that I'm going to ask if you have an example run anywhere? A coverage report which can display how this awesome beast performs? :-)

If you have a look at the tool now: https://github.com/oleorhagen/gobinarycoverage
I have tried to create a pretty comprehensive `Readme`file, on how the tool works on the surface. It is still rough though, and there are plenty of things which should be fixed, but it will work for `Mender` as it is now, and I can't be spending much more time on it.

I still think it turned out pretty cool though, as this should not alter the functionality in any way, and we can actually use this as a straight drop-in for our `Mender` binary. Even in production! (Although I'm sure ur gonna be having some reservations here :p)

patches/0001-Instrument-Mender-client-for-coverage-analysis.patch
oleorhagen: Sure, makes sense.

Makefile
oleorhagen: Alright, that means we should probably at least polish the `Readme` a bit more first.

kacf: Yes, although I thought it was already pretty good. I don't consider this tool a slick, sexy, and user friendly experience. More like the dirty, but trusty tool which won't let you down, by developers, for developers! :-)

oleorhagen: haha, no. Definitely not an Apple product :p

kacf:
> I have tried to create a pretty comprehensive `Readme`file, on how the tool works on the surface. It is still rough though, and there are plenty of things which should be fixed, but it will work for `Mender` as it is now, and I can't be spending much more time on it.

Great, but I was actually referring to a report, as in coverage report. Either on codecov or something manual. Somewhere where we can see that this captures the hard-to-catch code in the snapshotting for example.

> I still think it turned out pretty cool though, as this should not alter the functionality in any way, and we can actually use this as a straight drop-in for our `Mender` binary. Even in production! (Although I'm sure ur gonna be having some reservations here :p)

Hehe, yes, although it's not as intrusive as I first feared. We will have to find a way to run it conditionally for our containers, because this will run on users' computers too, but this is the next iteration, let's leave that for later.

> Then in `meta-mender` I'm suggesting a separate layer for adding this, which simply does a `compile_prepend` which calls this target.

You can consider simply using `meta-mender-ci`. It's already dedicated to CI stuff, so we may not need a new layer.

oleorhagen:
@kacf 
> Great, but I was actually referring to a report, as in coverage report. Either on codecov or something manual. Somewhere where we can see that this captures the hard-to-catch code in the snapshotting for example.
> 

Here is how the coverage file looks from one of the snapshot tests in acceptance, I don't know how much this helps though, and if this is what you had in mind?

https://drive.google.com/open?id=1rgnh4ECAtoTlMgHziUnlpSDJHes8UJjO

oleorhagen:
Test run along with: #meta-mender/952

https://gitlab.com/Northern.tech/Mender/mender-qa/pipelines/135816334

kacf:
> Here is how the coverage file looks from one of the snapshot tests in acceptance, I don't know how much this helps though, and if this is what you had in mind?
> 
> https://drive.google.com/open?id=1rgnh4ECAtoTlMgHziUnlpSDJHes8UJjO
> 
> And here the corresponding html file:
> 
> https://drive.google.com/open?id=1jH2p8Smm2kSjYjMhbdxID2Wp-KQq5qY2

Exactly! That's very encouraging, you can see it hits those hard-to-reach system calls like `FreezeFS`. This will be an awesome help to have!

PR #527, opened 2020-05-07
keystore: use openssl bindings

torbenh3:
Switch the code that signs the servers authentication request
to use openssl. This allows to use ssl_engines, which permit to
use PKCS#11 or TPMs as keystore.

Signed-off-by: Torben Hohn <torben.hohn@linutronix.de>
Reviewed-by: Bastian Germann <bage@linutronix.de>


Makefile
```diff
@@ -25,6 +25,7 @@ export CGO_CFLAGS
 TOOLS = \
 	github.com/fzipp/gocyclo \
 	gitlab.com/opennota/check/cmd/varcheck \
+	github.com/spacemonkeygo/openssl \
```
kacf, line 28: The dependencies should go into the `vendor` folder, as part of the vendored dependencies. I have to admit that since we recently switched to go modules support, I'm not 100% up to date on how this process works anymore. Perhaps @oleorhagen could help with this part.

main_test.go
```diff
@@ -51,7 +51,7 @@ func TestBinarySize(t *testing.T) {
 	// so.
 	//
 	// When increasing, use current binary size on amd64 + 1M.
-	const maxSize int64 = 9103464
+	const maxSize int64 = 11357152
```
kacf, line 54: Doesn't it link dynamically to OpenSSL? Is this size increase really just from the glue code between the two?

Makefile
oleorhagen: Sure thing. These tools mentioned in the Makefile are for use in the Makefile only, not as part of the source tree.

@torbenh3, are you familiar with go modules?

store/keystore_test.go
```diff
@@ -140,7 +134,15 @@ func TestKeystore(t *testing.T) {
 	h.Write(tosigndata)
 	hashed := h.Sum(nil)
 
-	err = rsa.VerifyPKCS1v15(&k.private.PublicKey, crypto.SHA256, hashed, s)
+	//generate pubkey for golang stdlib
+	block, _ := pem.Decode([]byte(expectedaspem))
```
oleorhagen, line 138: Why explicitly ignore the error here?

Makefile
torbenh3: @oleorhagen  i am not familiar with go modules. This is actually my first change ever to go code.

main_test.go
torbenh3: You are correct. It links dynamically. At least the version i have on my armhf target does that.
This size increase worries me.

Makefile
oleorhagen: Alright, then all the more impressive!

If you wan't to have a crack at this yourself, then read https://blog.golang.org/using-go-modules, it should give you most of what you need :)

If not, I can do it for you, or alternatively spell out the commands you need. It is up to you :)

store/keystore_test.go
torbenh3: I copied that from the old code in store/keystore.go loadFromPem() ... will fix it.

Makefile
torbenh3: @oleorhagen i would be happy if you pointed me at the things i need to do.
i have created a softlink to my git checkout at  /home/torbenh/go/src/github.com/mendersoftware/mender

Now the whole "go test" and "go install" does not find logrus...

main.go:22:2: cannot find package "github.com/sirupsen/logrus" in any of:
        /usr/lib/go-1.11/src/github.com/sirupsen/logrus (from $GOROOT)
        /home/torbenh/go/src/github.com/sirupsen/logrus (from $GOPATH)
make: *** [Makefile:185: govet] Error 1

i dont understand, how to make go search in the vendor path.
this is debian buster with:

go version go1.11.6 linux/amd64

store/keystore_test.go
oleorhagen: Thanks :100:

Makefile
oleorhagen: Alright, let's do this. Seems the mender-client repository is not fully migrated to go modules yet though actually, so we'll fall back to govendor.

```bash
# Get the govendor binary
go get github.com/kardianos/govendor
# Download the source for the ssl wrapper (Go) -- Probably not necessary if you've done so already, but it is idempotent, so no harm done
go get github.com/spacemonkeygo/openssl
# Add the new source to the vendor folder
govendor add github.com/spacemonkeygo/openssl
```

And now, commit the sources which have been added to the vendor folder by `govendor`, and the changes done to the `vendor/vendor.json` file.

To test it, use the makefile, as opposed to the regular go build and go test commands.

So - Build:
`make build`
Test:
`make test`

Let's see how this goes.

If you still get the same error as above, report your values of `GOPATH` and `GOROOT`, as well as a description of your source tree, go has some quirks for how it handles the go source files.

torbenh3:
> Very nice work @torbenh3! The code itself looks flawless to me, I just have some comments about the surrounding parts.
> 
> In addition:
> 
>     * You should rebase your work on top of latest master, since there are some conflicts.

yeah. i just wanted to delay the rebase until i had the unit tests sorted out.
> 
>     * It looks like there is still a unit test failure

The last unittest fail is due to openssl not validating the key it loads.
I have not found a function call to make openssl validate the modulo or primes, like
the go stdlib does.

So either we drop TestKeystoreLoadPem() or find some other damage for badPrivKey,
that will make openssl return an error.

store/keystore_test.go
torbenh3: thats actually not the error. its the undecoded rest.

https://golang.org/pkg/encoding/pem/#Decode

i will leave it as it is.

Makefile
torbenh3: Thanks for the advice. This has worked.

The softlink confused go. I had cd into the gopath. when pwd is below the GOPATH it works.

torbenh3:
Whats your policy on the test coverage ? seems like this went down because of the modules
added.

kacf:
I think that's because part of the coverage comes from our acceptance tests, which do not run automatically for external contributors. I'm starting them now, hopefully it should update the count soon.

torbenh3:
I think i am able to get the the binary size down to normal levels again.
It seems like the logging component spacelog that is used by openssl uses reflection.

By removing the logging from openssl, i can drop that component, and binary size goes down
from 13MB to 10.3MB here. 

Please dont merge this yet, i would like to be revert some commits, and cleanup the history.

main_test.go
torbenh3: @kacf i have removed the dependency to spacelog from openssl now, and that allows
us to drop this patch. Things look sane now.

torbenh3:
@kacf can you check the coverage stuff ? I think am happy now. I could drop the size increase patch.

torbenh3:
@oleorhagen i have made a patch to the openssl bindings, which keeps the binary size in check.
For now, i have just patched it in the vendor directory. But this invalidates the contents of vender.json, as the git hash does not relate to the contents of the checkout.

Is there a way to patch a checkout in the vendor framework, or do i have to fork the openssl bindings on github ?

lluiscampos:
> @kacf can you check the coverage stuff ?

The coverage drop is because we have not run the client acceptance tests suite on this PR. @kacf said that he will take care of running them in the upcoming days, so it should be fine.

Nice work @torbenh3 :+1:

oleorhagen:
> @oleorhagen i have made a patch to the openssl bindings, which keeps the binary size in check.
> For now, i have just patched it in the vendor directory. But this invalidates the contents of vender.json, as the git hash does not relate to the contents of the checkout.
> 
> Is there a way to patch a checkout in the vendor framework, or do i have to fork the openssl bindings on github ?

There is no way that I know of to patch this, except to fork it that I know of. One opportunity would be to fork the spacelog library instead, and use our logger underneath.

I can't really decide which I think is better.

@kacf @lluiscampos 

What are your thoughts here?

kacf:
> i have made a patch to the openssl bindings, which keeps the binary size in check.

That's awesome, it makes things much easier!

> For now, i have just patched it in the vendor directory. But this invalidates the contents of vender.json, as the git hash does not relate to the contents of the checkout.
> 
> Is there a way to patch a checkout in the vendor framework, or do i have to fork the openssl bindings on github ?

Things like these should usually be resolved in the upstream repository. I haven't looked at the changes in detail yet, but is it something the upstream repository might accept? Maybe it's something that can be guarded with [Golang build tags](https://golang.org/pkg/go/build/#hdr-Build_Constraints).

The alternative is to fork the repository, but we should prefer the former solution if possible.

> The coverage drop is because we have not run the client acceptance tests suite on this PR.

In general this is something that will happen to all contributors. @lluiscampos: We might want to discuss if there is something we can do here, maybe report two separate values or something. It's quite hard to get feedback for contributors otherwise. I have added it to our next QA Architecture meeting.

> @kacf said that he will take care of running them in the upcoming days, so it should be fine.

Yes, I will take care of this, but the next two days are extremely busy, so I hope to get to it on Friday!

kacf:
@torbenh3: Can you include this change in your PR?

```diff
diff --git a/tests/Dockerfile.daemon b/tests/Dockerfile.daemon
index c794da4..aab84d2 100644
--- a/tests/Dockerfile.daemon
+++ b/tests/Dockerfile.daemon
@@ -3,7 +3,7 @@
 FROM ubuntu:18.04 AS build
 
 RUN apt update
-RUN apt install -y make git build-essential golang liblzma-dev jq
+RUN apt install -y make git build-essential golang liblzma-dev jq libssl-dev
 
 COPY ./ /go/src/github.com/mendersoftware/mender/
 RUN make -C /go/src/github.com/mendersoftware/mender GOPATH=/go clean
```

I could do it myself, but then I need to create a new PR, and it's better to keep it contained here.

kacf:
Thanks, @torbenh3! Merged in #530.

drewwestrick:
I just wanted to chime in here and thank @torbenh3 for putting in this work. This is something I've been keeping an eye on for awhile and I just saw this made it into the Mender 2.5 release.

PR #529, opened 2020-05-14
k8s update module and artifact generator

tranchitella:
k8s update module and artifact generator, compatible with any certified Kubernetes distribution including k3s.io

### DIFF
--- a/support/modules-artifact-gen/k8s-artifact-gen
+++ b/support/modules-artifact-gen/k8s-artifact-gen
@@ -1,0 +1,127 @@
+#!/bin/bash
+
+set -e
+
+show_help() {
+  cat << EOF
+
+Simple tool to generate Mender Artifact suitable for k8s Update Module
+
+Usage: $0 [options] MANIFESTO [MANIFESTOS...] [-- [options-for-mender-artifact] ]
+
+    Options: [ -n|artifact-name -t|--device-type -o|--output_path -h|--help ]
+
+        --artifact-name     - Artifact name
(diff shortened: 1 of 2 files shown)


mirzak:
Good stuff @tranchitella, though I think this should go in https://github.com/mendersoftware/mender-update-modules

tranchitella:
@mirzak shall I launch a PR for https://github.com/mendersoftware/mender-update-modules, both for the module and the artifact gen, following the convention you use there (folders k8s/{module,module-artifact-gen}?

tranchitella:
Just in case: https://github.com/mendersoftware/mender-update-modules/pull/16

mirzak:
> @mirzak shall I launch a PR for https://github.com/mendersoftware/mender-update-modules, both for the module and the artifact gen, following the convention you use there (folders k8s/{module,module-artifact-gen}?

Yeah, that was my intention. Sorry for not being clear.

The intention of the https://github.com/mendersoftware/mender-update-modules repository is to gather all community maintained modules (and generators) in one place, and I do not think that we plan to add any more modules here (part of Mender client source tree) and only reason I think they are here is because we did not have the other repository when they where created.

tranchitella:
@mirzak got it, closing this PR, then.

lluiscampos:
> [...] and only reason I think they are here is because we did not have the other repository when they where created.

@mirzak You are right. Furthermore, we should revisit this decision: I think we could move the deb/rpm there also, leaving here only the ones we consider "first class citizens" kind of Update Modules. I will take a note for the next client team meeting.

PR #530, opened 2020-05-19
keystore: use openssl bindings

kacf:
#527 with extra fixes.


.gitlab-ci.yml
```diff
@@ -16,7 +16,7 @@ test:
     - mkdir -p /go/src/$(dirname $REPO_NAME)/mender /go/src/_/builds
     - cp -r $CI_PROJECT_DIR /go/src/$(dirname $REPO_NAME)
     - cd $GOPATH/src/$REPO_NAME
-    - apt-get update && apt-get install -yyq liblzma-dev
+    - apt-get update && apt-get install -yyq liblzma-dev libssl-dev
```
lluiscampos, line 19: :+1:

lluiscampos:
@kacf FYI I created a separate task to update the Debian package dependencies, so that we don't lose track for it. Hence I think this PR is good to go.

PR #537, opened 2020-06-12
QA-193: Add separate code coverage for unit and acceptance tests.

kacf:
The acceptance tests are covered in mender-qa, so they are not visible
in this repo.

Changelog: None


kacf:
Merged! I also switched both coverage reports to be mandatory before merging. Coverage has dropped too much lately, so I think this is important. If it causes too much trouble we can revert it, but let's try at least.

PR #538, opened 2020-06-14
Check if root mount candidate match rootfsPartA/B

Risca:
When using device-less mounts for ubi rootfs (e.g. root=ubi0:rootfsa), mender
won't find which device is the active device unless I specify them as ubi0_0/1
in mender.conf.

This patch tries to match the root mount candidate, as reported by 'mount',
with what is written in mender.conf.

Example output of `mount`:

        / # mount
        ubi0:rootfsa on / type ubifs (rw,relatime)
        devtmpfs on /dev type devtmpfs (rw,relatime,size=251372k,nr_inodes=62843,mode=755)
        proc on /proc type proc (rw,relatime)
        devpts on /dev/pts type devpts (rw,relatime,gid=5,mode=620)
        tmpfs on /dev/shm type tmpfs (rw,relatime,mode=777)
        tmpfs on /tmp type tmpfs (rw,relatime)
        tmpfs on /run type tmpfs (rw,nosuid,nodev,relatime,mode=755)
        sysfs on /sys type sysfs (rw,relatime)
        ubi0:data on /data type ubifs (rw,relatime)

Example _mender.conf_ file:

        # cat /etc/mender/mender.conf
        {
          "InventoryPollIntervalSeconds": 1800,
          "UpdatePollIntervalSeconds": 1800,
          "RetryPollIntervalSeconds": 300,
          "RootfsPartA": "ubi0:rootfsa",
          "RootfsPartB": "ubi0:rootfsb",
          "ServerCertificate": "/etc/mender/server.crt",
          "ServerURL": "https://REDACTED",
          "TenantToken": "dummy"
        }

Changelog: Title
Signed-off-by: Patrik Dahlström <risca@dalakolonin.se>

### DIFF
--- a/installer/partitions.go
+++ b/installer/partitions.go
@@ -181,6 +181,16 @@ func (p *partitions) getAndCacheActivePartition(rootChecker func(system.StatComm
 			log.Debug("Setting active partition: ", mountCandidate)
 			return p.active, nil
 		}
+		if mountCandidate == p.rootfsPartA {
+			p.active = p.rootfsPartA
+			log.Debugf("Setting active partition from configuration and mount candidate: %s", p.active)
+			return p.active, nil
+		}
+		if mountCandidate == p.rootfsPartB {
+			p.active = p.rootfsPartB
+			log.Debugf("Setting active partition from configuration and mount candidate: %s", p.active)
+			return p.active, nil
+		}
 		// If not see if we are lucky somewhere else
 	}
 


Risca:
Oops, looks like some unit tests failed due to this change. I can take a look at that later.

What's more important though, is that this change breaks the actual update mechanism, because the updater doesn't know where to write an update to - it doesn't know that `ubi0:rootfsa == /dev/ubi0_0`.

This pull request needs a bit more love before it can be merged. Since it expands to a bit more work, I would like to ask:
* does it make sense to have this functionality?

installer/partitions.go
```diff
@@ -181,6 +181,16 @@ func (p *partitions) getAndCacheActivePartition(rootChecker func(system.StatComm
...
+		if mountCandidate == p.rootfsPartA {
+			p.active = p.rootfsPartA
+			log.Debugf("Setting active partition from configuration and mount candidate: %s", p.active)
+			return p.active, nil
+		}
+		if mountCandidate == p.rootfsPartB {
+			p.active = p.rootfsPartB
+			log.Debugf("Setting active partition from configuration and mount candidate: %s", p.active)
+			return p.active, nil
+		}
```
kacf, line 193: I *think* this should be ok, and yeah, the idea makes sense. This has previously proved to be quite a minefield though, so no promises until we run this through our full test suite.

kacf:
FYI, I started a full test run [here](https://gitlab.com/Northern.tech/Mender/mender-qa/-/pipelines/156676335) to see early results, although the unit tests will still have to be fixed. It's a private CI project, so you cannot see it unfortunately, but you can see the statuses for each platform pop up here, and I can let you know if there are any other unforeseen problems.

mirzak:
Hi @Risca.

> it doesn't know that ubi0:rootfsa == /dev/ubi0_0.

I do recall this problem from way back when I added the UBI support and at that time I could not find a cleaner way to implement this.

The reason I opted for `/dev/ubi0_0` style entries, was because this is what is supported in `/etc/fstab`, e.g you can not put `ubi0:rootfsa` in `/etc/fstab`. So some kind of uniformity, and an attempt to replicate block style interfaces, e.g `/dev/mmcblk0p2` etc.

Can you elaborate a bit on your use-case and why supporting `ubi0:rootfsa` is important to you?

Risca:
> The reason I opted for `/dev/ubi0_0` style entries, was because this is what is supported in `/etc/fstab`, e.g you can not put `ubi0:rootfsa` in `/etc/fstab`. So some kind of uniformity, and an attempt to replicate block style interfaces, e.g `/dev/mmcblk0p2` etc.
> 
> Can you elaborate a bit on your use-case and why supporting `ubi0:rootfsa` is important to you?

I have this in my fstab and it seems to work perfectly:
```
ubi0:data       /data           ubifs   defaults        0       0
```

I don't really have a better use case than "it's easier to understand `ubi0:rootfsa` instead of `/dev/ubi0_0`" 🤷 U-boot already support the `rootfsa/rootfsb` naming convention, so it would be nice to be consistent.

I'm busy for the next few days, but I'll take a look at improving this pull request sometime next week

kacf:
All our acceptance tests were green, so only the unit tests need fixing here.

I think the fix is innocent enough that this is acceptable. Even if we can debate whether "ubi:xxx" is the best way to specify a device, it's not really up to us. If the kernel and fstab supports it, then we might as well too.

Risca:
Please note that this pull request is still incomplete. If you would specify "ubiY:xxx" in _mender.conf_, then registration and reporting will work but upgrade *will* fail. Some kind of integration test might be useful here

kacf:
Ok, thanks for the heads up. It's not entirely trivial to make this test, but I think I can dedicate some time to write it, if you can get the implementation sorted out.

Risca:
I've taken a short look on how to fix this but need to learn some more Go before I can do it, and improve my implement+test cycle.
Here's what I believe would work:
* Change `IsUbiBlockDevice()` to return Boolean and a UBI _block device_ (compare with _device name_)
* If it _is_ a UBI block device, use the path returned.

Here's some pseudo code of how I think `IsUbiBlockDevice()` could be adjusted:
```golang
func IsUbiBlockDevice(deviceName string) bool {
    for _, device := range sysfs.Class.Object("ubi").SubObjects() {
        if notdir(device) == deviceName or os.Read(device.attribute("name")) == deviceName {
            return True, "/dev/"+notdir(device)
    return False, nil
}
```
The missing parts to implement/use correct function is `notdir()` and `os.Read()`. After that, _block_device.go_ can be changed to
```golang
typeUBI, ubiDevice := system.IsUbiBlockDevice(device)
...
if typeUBI {
    ...
    device = ubiDevice
    ...
}
```
@mirzak and @kacf, any comments on this approach?

kacf:
To me that looks like a good approach. Just to make sure I understand correctly: Is the key here that the name, `rootfsa`, is stored in `Attribute("name")`, not in `Name()`?

Risca:
There are 2 ways to match:
* _/sys/class/ubi/ubi0_0_ directory name matches "ubi0_0" in config
* _/sys/class/ubi/ubi0_0/name_ contains the text "rootfsa" which matches the config.

I'll probably make it match "ubiX:rootfsa" in config file as well

Risca:
Sorry for the inactivity - summer vacation happened. Unfortunately, I won't get time to work on this until maybe 2 weeks from now.

PR #539, opened 2020-06-15
mender-inventory-geo script to return geo localization data.

merlin-northern:
ChangeLog:none

### DIFF
--- a/support/mender-inventory-geo
+++ b/support/mender-inventory-geo
@@ -1,0 +1,57 @@
+#!/bin/sh
+#
+# Example script called by Mender client to collect inventory data for a
+# particular device. The script needs to be located in $(datadir)/mender and its
+# name shall start with `mender-inventory-` prefix. The script shall exit with
+# non-0 status on errors. In this case the agent will discard any output the
+# script may have produced.
+#
+# The script shall output inventory data in <key>=<value> format, one entry per
+# line. Entries appearing multiple times will be joined in a list under the same
+# key.
+#
+# $ ./mender-inventory-geo
+# country=Poland
+# city=Krakow
+# latitude=50.05750
+# longitude=19.98020
+# ipv4=127.0.0.1
+#
+# The example script collects information on geo localization
+# this can be done in one line:
+# curl https://ipvigilante.com/`curl -4 ifconfig.io 2>/dev/null` 2>/dev/null | jq -r 'select(.status == "success") | "country=" + .data.country_name, "city="+.data.city_name, "latitude="+.data.latitude, "longitude="+.data.longitude, "ipv4="+.data.ipv4 | @text'
+# but in the following we assume busybox version of the utilities.
+# see below for the configuration of the attributes names
+
+function err() {
+ local rc=$1
+
+ shift
+ echo "${0}: $*" >&2
+ exit $rc
+}
+
+# find the ip address
+ip=`wget -qO /dev/stdout --header "Content-Type: text/plain;" \
+     --header 'Host: ifconfig.io' --header 'User-Agent: curl/7.67.0' \
+     --header 'Accept: */*' ifconfig.io | tail -1`
+
+[ -n "${ip}" ] || err 2 "Unable to get the IP address from ifconfig.io"
+
+# get the information, see https://www.ipvigilante.com/api-developer-docs/
+geo=`wget -qO /dev/stdout https://ipvigilante.com/csv/${ip} 2>/dev/null`
+
+[ -n "${geo}" ] || err 3 "Unable to get the geolocalization data from ipvigilante.com"
+
+# the names of the attributes holding the localization data can be configured here
+ATTR_NAME_IP="a-ip"
+ATTR_NAME_CONTINENT="a-continent"
+ATTR_NAME_COUNTRY="a-country"
+ATTR_NAME_CITY="a-city"
+
+echo "${geo}" | awk -v ip="${ATTR_NAME_IP}" -v continent="${ATTR_NAME_CONTINENT}" \
+    -v country="${ATTR_NAME_COUNTRY}" -v city="${ATTR_NAME_CITY}" -F',' \
+    ' length($NF) == 0 { exit(4) }
+      { printf("%s=%s\n%s=%s\n%s=%s\n%s=%s\n",ip,$1,continent,$2,country,$3,city,$6) }
+    '
+
--- a/Makefile
+++ b/Makefile
@@ -54,7 +54,8 @@ INVENTORY_SCRIPTS = \
 	support/mender-inventory-hostinfo \
 	support/mender-inventory-network \
 	support/mender-inventory-os \
-	support/mender-inventory-rootfs-type
+	support/mender-inventory-rootfs-type \
+	support/mender-inventory-geo
 
 MODULES = \
 	support/modules/deb \


kacf:
Nice little script!

Personally I feel like localization data is a bit too intrusive to include in the default install, but we don't really have a community area for submitting scripts that aren't installed by default.

Question is whether we should create a contrib folder in this repository, a new repository, or if I'm being too paranoid, and we should just include this as is. @mirzak, @drewmoseley, @oleorhagen, @lluiscampos: Opinions?

oleorhagen:
> Nice little script!
> 
> Personally I feel like localization data is a bit too intrusive to include in the default install, but we don't really have a community area for submitting scripts that aren't installed by default.
> 
> Question is whether we should create a contrib folder in this repository, a new repository, or if I'm being too paranoid, and we should just include this as is. @mirzak, @drewmoseley, @oleorhagen, @lluiscampos: Opinions?

No strong opinions from my part. A contribution folder for non-standard stuff would be nice though I think.

mirzak:
> Question is whether we should create a contrib folder in this repository, a new repository, or if I'm being too paranoid, and we should just include this as is. @mirzak, @drewmoseley, @oleorhagen, @lluiscampos: Opinions?

I do not think it is to intrusive and I am ok with adding it. We are push ip address of the devices by default, I do not see this being much different as the location information is based on the IP address.

I see a few technical problems with including by default though that I we should address,

1. The script depends on `curl` and `jq`, which are dependencies that we would need to transfer to the top, that is the Mender client would depend on these unconditionally.

   If we can rewrite to `wget` (busybox  version used in Yocto) and json parsing using e.g `sed` or similar that would be better. This would make it portable without requiring additional dependencies

2. I would reduce the values to (added continent, removed longitude/latitude):

   ```
   $ ./mender-inventory-geo
   continent=Europe
   country=Poland
   city=Krakow
   ipv4=127.0.0.1
   ```

   The latitude/longitude I can not imagine being accurate and would be misleading.

support/mender-inventory-geo
```diff
@@ -0,0 +1,26 @@
...
+# latitude=50.05750
+# longitude=19.98020
+# ipv4=127.0.0.1
+#
+# The example script collects information on geo localization
+#
+
+set -ue
+
+curl https://ipvigilante.com/`curl -4 ifconfig.io 2>/dev/null` 2>/dev/null | jq -r 'select(.status == "success") | "country=" + .data.country_name, "city="+.data.city_name, "latitude="+.data.latitude, "longitude="+.data.longitude, "ipv4="+.data.ipv4 | @text'
```
lluiscampos, line 25: For improved readability, I suggest we split this into multiple lines

lluiscampos:
@merlin-northern Nice initiative! :rocket: 

With regards to the high level discussion concerning intrusiveness of the script: I think it is a great bunch of data to have by default, for example it will allow for very fancy dynamic grupos almost OOTB. Any user concern about privacy would review these scripts and possibly remove them from the installation.

But I think it is important to consider @mirzak notes regarding dependencies. We should try to avoid adding these 3rd party dependencies.

merlin-northern:
thats for all the input Lluis, Ole, Kristian, Mirza! actually, I wrote this the moment I was fixing one of Eystein's comments in the docs. The version I had put in the docs looks like that:

```bash
#!/bin/sh
#
# The example script collects geo localization information
#

set -ue

jq --help >/dev/null 2>&1 || exit 1;
curl --help >/dev/null 2>&1 || exit 1;

ip=`curl -4 ifconfig.io 2>/dev/null`

[ "${ip}" == "" ] || exit 1;

curl --fail https://ipvigilante.com/$ip 2>/dev/null | jq -r 'select(.status == "success") | "country=" + .data.country_name, "city="+.data.city_name, "latitude="+.data.latitude, "longitude="+.data.longitude, "ipv4="+.data.ipv4 | @text'
```
I will fix accordingly.

merlin-northern:
> > Question is whether we should create a contrib folder in this repository, a new repository, or if I'm being too paranoid, and we should just include this as is. @mirzak, @drewmoseley, @oleorhagen, @lluiscampos: Opinions?
> 
> I do not think it is to intrusive and I am ok with adding it. We push ip address of the devices by default, I do not see this being much different as the location information is based on the IP address.
> 
> I see a few technical problems with including by default though that I we should address,
> 
> 1. The script depends on `curl` and `jq`, which are dependencies that we would need to transfer to the top, that is the Mender client would depend on these unconditionally.
>    If we can rewrite to `wget` (busybox  version used in Yocto) and json parsing using e.g `sed` or similar that would be better. This would make it portable without requiring additional dependencies
> 2. I would reduce the values to (added continent, removed longitude/latitude):
>    ```
>    $ ./mender-inventory-geo
>    continent=Europe
>    country=Poland
>    city=Krakow
>    ipv4=127.0.0.1
>    ```
>    
>    
>    The latitude/longitude I can not imagine being accurate and would be misleading.

if you do not mind, I will fall back to wget and sed in case jq and curl is not present, ok?

mirzak:
> if you do not mind, I will fall back to wget and sed in case jq and curl is not present, ok?

Can you elaborate on why?

I do not see the point of falling back. If `wget ` and `sed` works then all is good and having additional code for `curl` and `jq` would just be clutter.

Also `curl` and `jq` would rarely be tested as we do not have explicit dependencies on these, so there is a risk of that path not working

merlin-northern:
wow, look at that:
https://www.ipvigilante.com/api-developer-docs/
curl https://ipvigilante.com/csv/`curl -4 ifconfig.io`

so we can make it even simplier.

merlin-northern:
I stopped the pipeline.
ok, @lluiscampos and @mirzak you suggestions are in I believe. busybox does not cut you any breaks, as you can see.

support/mender-inventory-geo
```diff
@@ -0,0 +1,64 @@
...
+
+# get the information, see https://www.ipvigilante.com/api-developer-docs/
+wget -qO "${tmpfile}" https://ipvigilante.com/csv/${ip} 2>/dev/null;
+
+[ -s "${tmpfile}" ] || exit 3;
+
+cat "${tmpfile}" | awk -v ip="a-ip" -v continent="a-continent" \
+     -v country="a-country" -v city="a-city" -F',' \
+     '{ \
+        if (length($NF)<1) { \
```
oleorhagen, line 58: Question:
What exactly does the check guard against?

support/mender-inventory-geo
```diff
@@ -0,0 +1,64 @@
...
+ip=`cat "${tmpfile}" | tail -1`;
+
+[ "${ip}" != "" ] || exit 2;
+
+# get the information, see https://www.ipvigilante.com/api-developer-docs/
+wget -qO "${tmpfile}" https://ipvigilante.com/csv/${ip} 2>/dev/null;
+
+[ -s "${tmpfile}" ] || exit 3;
+
+cat "${tmpfile}" | awk -v ip="a-ip" -v continent="a-continent" \
```
oleorhagen, line 55: Suggestion:
Maybe the variables you define can simply be hard-coded in the `awk` script?

support/mender-inventory-geo
```diff
@@ -0,0 +1,64 @@
...
+trap geo_cleanup SIGINT;
+trap geo_cleanup SIGTERM;
+trap geo_cleanup SIGKILL;
+
+# find the ip address
+wget -qO "${tmpfile}" --header "Content-Type: text/plain;" \
+     --header 'Host: ifconfig.io' --header 'User-Agent: curl/7.67.0' \
+     --header 'Accept: */*' ifconfig.io;
+
+[ -f "${tmpfile}" ] || exit 1;
```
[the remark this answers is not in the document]
merlin-northern: copy-paste error. it was there. thanks. fixed.

support/mender-inventory-geo
```diff
@@ -0,0 +1,64 @@
...
+
+ip=`cat "${tmpfile}" | tail -1`;
```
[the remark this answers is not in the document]
merlin-northern: sure.

support/mender-inventory-geo
```diff
@@ -0,0 +1,64 @@
...
+
+[ "${ip}" != "" ] || exit 2;
```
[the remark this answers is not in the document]
merlin-northern: sure.

support/mender-inventory-geo
```diff
@@ -0,0 +1,64 @@
...
+
+# get the information, see https://www.ipvigilante.com/api-developer-docs/
+wget -qO "${tmpfile}" https://ipvigilante.com/csv/${ip} 2>/dev/null;
```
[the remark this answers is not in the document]
merlin-northern: good practice :>
removed all of them just for you

support/mender-inventory-geo
merlin-northern: this is a guardian in case wget failed and there is no data, so the whole script returns non-success.

support/mender-inventory-geo
merlin-northern: actually I wanted to make them configurable. changed.

merlin-northern:
@oleorhagen instead of arguing (which I would love to do) I decided to finish you off with actually introducing the corrections you requested.

estenberg:
This is extremely cool and would make really powerful demos (e.g. dynamic groups like mentioned)! Might be a bit side-track to the docs project but well worth getting this in I think!

Regarding where / how to put it in I would support having this on a default installation. At minimum by default on a demo mode (we include some extra scripts there in the Yocto receipe at least).

Nice job!

support/mender-inventory-geo
lluiscampos: If there is no data, test `[ -s "${tmpfile}" ]` will fail, right?

support/mender-inventory-geo
```diff
@@ -0,0 +1,68 @@
...
+#
+# The example script collects information on geo localization
+# this can be done in one line:
+# curl https://ipvigilante.com/`curl -4 ifconfig.io 2>/dev/null` 2>/dev/null | jq -r 'select(.status == "success") | "country=" + .data.country_name, "city="+.data.city_name, "latitude="+.data.latitude, "longitude="+.data.longitude, "ipv4="+.data.ipv4 | @text'
+# but in the following we assume busybox version of the utilities.
+
+ATTR_NAME_IP="a-ip"
+ATTR_NAME_CONTINENT="a-continent"
+ATTR_NAME_COUNTRY="a-country"
+ATTR_NAME_CITY="a-city"
```
lluiscampos, line 28: Being an example script, I think that being able to configure the keys of the attributes is adding too much complexity. We should try to get a script as readable as possible.

If you want to keep it, please add a comment explaining what is it, as they are defined here with no comment and used at the end.

support/mender-inventory-geo
```diff
@@ -0,0 +1,68 @@
...
+# get the information, see https://www.ipvigilante.com/api-developer-docs/
+wget -qO "${tmpfile}" https://ipvigilante.com/csv/${ip} 2>/dev/null
+
+[ -s "${tmpfile}" ] || exit 3
+
+cat "${tmpfile}" | awk -v ip="${ATTR_NAME_IP}" -v continent="${ATTR_NAME_CONTINENT}" \
+     -v country="${ATTR_NAME_COUNTRY}" -v city="${ATTR_NAME_CITY}" -F',' \
+     '{ \
+        if (length($NF)<1) { \
+          exit(4); \
```
lluiscampos, line 63: For the sake of simplicity, I would also remove this guard here. Again, my point of view is that this should not need to be production ready by rather demo ready and easy to read, so that users can get inspiration for their advance scripts.

support/mender-inventory-geo
```diff
@@ -0,0 +1,68 @@
...
+
+# this we will use for everything, wget in busybox does not support -O-
+tmpfile=`mktemp`
+[ -f "${tmpfile}" ] || exit 1
+# setup cleanup
+function geo_cleanup() {
+ rm -f "${tmpfile}"
+}
+
+trap geo_cleanup EXIT
```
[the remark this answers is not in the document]
merlin-northern: Ole, if you insist I will change it, but it looks better to me like that.

support/mender-inventory-geo
merlin-northern: I do want to keep it, unless you have strong objections. it is cleaner that way.

merlin-northern: besides: these are the names that will appear in gui and will be used in filters. does putting them at the end of the script make sense to you?

support/mender-inventory-geo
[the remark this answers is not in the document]
oleorhagen: No strong opinions. Was just a suggestion :)

support/mender-inventory-geo
merlin-northern: this is explicit guar for this case:
```echo ,,,,, | awk -F,```
of course I can remove it, but I would like to keep it in, if you do not mind?

File: support/mender-inventory-geo (hunk 8 above)
Comment (lluiscampos) [reply support/mender-inventory-geo:28]: Just add a comment as suggested, then.

Comment (oleorhagen) [reply support/mender-inventory-geo:28]: I think hard-coding them in the awk script is what makes most sense, since they are not configureable.

File: support/mender-inventory-geo (hunk 9 above)
Comment (oleorhagen) [reply support/mender-inventory-geo:63]: :+1:

File: support/mender-inventory-geo [hunk 11]
```diff
@@ -0,0 +1,68 @@
...
+ip=`tail -1 "${tmpfile}"`
+
+[ -n "${ip}" ] || exit 2
+
+# get the information, see https://www.ipvigilante.com/api-developer-docs/
+wget -qO "${tmpfile}" https://ipvigilante.com/csv/${ip} 2>/dev/null
+
+[ -s "${tmpfile}" ] || exit 3
+
+cat "${tmpfile}" | awk -v ip="${ATTR_NAME_IP}" -v continent="${ATTR_NAME_CONTINENT}" \
```
oleorhagen, line 59: Also, no need to `cat` here, `awk` can read files just fine :)

support/mender-inventory-geo
merlin-northern: sure.

support/mender-inventory-geo
```diff
@@ -0,0 +1,65 @@
+#!/bin/sh
+#
+# Example script called by Mender agent to collect inventory data for a
```
oleorhagen, line 3: `s/Mender agent/Mender client/`

support/mender-inventory-geo
```diff
@@ -0,0 +1,65 @@
...
+# particular devce. The script needs to be located in $(datadir)/mender and its
```
oleorhagen, line 4: s/devce/device

support/mender-inventory-geo
merlin-northern: https://github.com/mendersoftware/mender/blob/master/support/mender-device-identity#L3
https://github.com/mendersoftware/mender/blob/master/support/mender-inventory-network#L3
do I correct all other places as well?

oleorhagen: A separate PR would be nice I think :)

merlin-northern: https://github.com/mendersoftware/mender/pull/544

support/mender-inventory-geo
merlin-northern: sure

support/mender-inventory-geo
merlin-northern: sure

merlin-northern:
> Sorry I forgot it, but we need to add the script [in the Makefile](https://github.com/mendersoftware/mender/blob/master/Makefile#L52). We don't need an acceptance test IMO, though.
> 
> (Requesting changes for this detail, the other comments are open for discussion)

oh yes, sure, forgot about that!

support/mender-inventory-geo
```diff
@@ -0,0 +1,65 @@
...
+# curl https://ipvigilante.com/`curl -4 ifconfig.io 2>/dev/null` 2>/dev/null | jq -r 'select(.status == "success") | "country=" + .data.country_name, "city="+.data.city_name, "latitude="+.data.latitude, "longitude="+.data.longitude, "ipv4="+.data.ipv4 | @text'
+# but in the following we assume busybox version of the utilities.
+
+# the names of the attributes holding the localization data can be configured here
+ATTR_NAME_IP="a-ip"
+ATTR_NAME_CONTINENT="a-continent"
+ATTR_NAME_COUNTRY="a-country"
+ATTR_NAME_CITY="a-city"
+
+# this we will use for everything, wget in busybox does not support -O-
```
vpodzime, line 31: Not even `-O/dev/stdout`?

support/mender-inventory-geo
vpodzime: I'd at least move them right next to the `awk` invocation.

support/mender-inventory-geo
```diff
@@ -0,0 +1,65 @@
...
+trap geo_cleanup EXIT
+trap geo_cleanup SIGHUP
+trap geo_cleanup SIGQUIT
+trap geo_cleanup SIGINT
+trap geo_cleanup SIGTERM
+trap geo_cleanup SIGKILL
+
+# find the ip address
+wget -qO "${tmpfile}" --header "Content-Type: text/plain;" \
+     --header 'Host: ifconfig.io' --header 'User-Agent: curl/7.67.0' \
```
vpodzime, line 48: So *wget* or *curl*?

support/mender-inventory-geo
merlin-northern: I thought about it, but maybe lets keep it like this, for absolute maximum pertability?

merlin-northern:
> A few comments from me as well. However, my biggest concern would be adding this "call home" thing to all the devices. I have participated in a similar thing done in the Fedora OS installer (to guess the correct timezone/locale) and we ended up pushing the Fedora Infrastructure to deploy our own instance of GeoIP. People are very sensitive about things like this, they don't want their devices to automatically connect to some third-party server(s) when the devices start.

not my decision I guess. I see your points. most people I guess are reviewing what scripts the are running (someone commented on this here)

support/mender-inventory-geo
merlin-northern: well, this is something I found works with ifconfig.io, does not matter much I guess if wget is pretending to be curl. it si on purpose basically.

merlin-northern: sure. moved before awk.

support/mender-inventory-geo
oleorhagen: With `stdout` it will look more like this: 

Which removes all the traps and the temp-file. A fair trade-off I think.
```bash
#!/bin/sh
#
# Example script called by Mender client to collect inventory data for a
# particular device. The script needs to be located in $(datadir)/mender and its
# name shall start with `mender-inventory-` prefix. The script shall exit with
# non-0 status on errors. In this case the agent will discard any output the
# script may have produced.
#
# The script shall output inventory data in <key>=<value> format, one entry per
# line. Entries appearing multiple times will be joined in a list under the same
# key.
#
# $ ./mender-inventory-geo
# country=Poland
# city=Krakow
# latitude=50.05750
# longitude=19.98020
# ipv4=127.0.0.1
#
# The example script collects information on geo localization
# this can be done in one line:
# curl https://ipvigilante.com/`curl -4 ifconfig.io 2>/dev/null` 2>/dev/null | jq -r 'select(.status == "success") | "country=" + .data.country_name, "city="+.data.city_name, "latitude="+.data.latitude, "longitude="+.data.longitude, "ipv4="+.data.ipv4 | @text'
# but in the following we assume busybox version of the utilities.

IP_ADDR=$(wget -q -O /dev/stdout --header "Content-Type: text/plain;" \
     --header 'Host: ifconfig.io' --header 'User-Agent: curl/7.67.0' \
     --header 'Accept: */*' https://ifconfig.io)

[ -n "${IP_ADDR}" ] || { echo >&2 "Failed to get the IP address from 'https://ifconfig.io'"; exit 1; }

# Get the information, see https://www.ipvigilante.com/api-developer-docs/
GEOLOCATION=$(wget -q -O /dev/stdout https://ipvigilante.com/csv/${IP_ADDR})

[ -n "${GEOLOCATION}" ] || { echo >&2 "Failed to get the Geolocation information from 'https://ipvigilante.com'"; exit 1; }

# the names of the attributes holding the localization data can be configured here
ATTR_NAME_IP="a-ip"
ATTR_NAME_CONTINENT="a-continent"
ATTR_NAME_COUNTRY="a-country"
ATTR_NAME_CITY="a-city"

echo "${GEOLOCATION}" | \
  awk -v ip="${ATTR_NAME_IP}" \
      -v continent="${ATTR_NAME_CONTINENT}" \
      -v country="${ATTR_NAME_COUNTRY}" \
      -v city="${ATTR_NAME_CITY}" -F',' \
      '
       length($NF) == 0 { exit(4) }
      { printf("%s=%s\n%s=%s\n%s=%s\n%s=%s\n",ip,$1,continent,$2,country,$3,city,$6) }
    '
```

vpodzime:
> People are very sensitive about things like this, they don't want their devices to automatically connect to some third-party server(s) when the devices start.

I guess I should have pointed out that this is because it also has security-related consequences.

support/mender-inventory-geo
```diff
@@ -0,0 +1,57 @@
...
+ echo "$0 $* " >&2
+ exit $rc
+}
+
+# find the ip address
+ip=`wget -qO /dev/stdout --header "Content-Type: text/plain;" \
+     --header 'Host: ifconfig.io' --header 'User-Agent: curl/7.67.0' \
+     --header 'Accept: */*' ifconfig.io | tail -1`
+
+[ -n "${ip}" ] || err 2 "Can't get the ip address via GET from ifconfig.io"
```
oleorhagen, line 39: Suggestion:
`Unable to get the IP address..`
Also, I guess it is not necessary to mention the `GET` request verb.

support/mender-inventory-geo
```diff
@@ -0,0 +1,57 @@
...
+
+# get the information, see https://www.ipvigilante.com/api-developer-docs/
+geo=`wget -qO /dev/stdout https://ipvigilante.com/csv/${ip} 2>/dev/null`
+
+[ -n "${geo}" ] || err 3 "Can't get geolocalization data from ipvigilante.com"
```
oleorhagen, line 44: Same

support/mender-inventory-geo
```diff
@@ -0,0 +1,57 @@
...
+# this can be done in one line:
+# curl https://ipvigilante.com/`curl -4 ifconfig.io 2>/dev/null` 2>/dev/null | jq -r 'select(.status == "success") | "country=" + .data.country_name, "city="+.data.city_name, "latitude="+.data.latitude, "longitude="+.data.longitude, "ipv4="+.data.ipv4 | @text'
+# but in the following we assume busybox version of the utilities.
+# see below for the configuration of the attributes names
+
+function err() {
+ local rc=$1
+
+ shift
+ echo "$0 $* " >&2
```
oleorhagen, line 30: Maybe:
echo "$0`:` $*" >&2
Also, there is no need for a space after the `$*` I think.

oleorhagen:
> A few comments from me as well. However, my biggest concern would be adding this "call home" thing to all the devices. I have participated in a similar thing done in the Fedora OS installer (to guess the correct timezone/locale) and we ended up pushing the Fedora Infrastructure to deploy our own instance of GeoIP. People are very sensitive about things like this, they don't want their devices to automatically connect to some third-party server(s) when the devices start.

I guess this in the end will be up to @estenberg

merlin-northern:
> > A few comments from me as well. However, my biggest concern would be adding this "call home" thing to all the devices. I have participated in a similar thing done in the Fedora OS installer (to guess the correct timezone/locale) and we ended up pushing the Fedora Infrastructure to deploy our own instance of GeoIP. People are very sensitive about things like this, they don't want their devices to automatically connect to some third-party server(s) when the devices start.
> 
> I guess this in the end will be up to @estenberg

and I think Eystein already stated his opinion above.

support/mender-inventory-geo
merlin-northern: sure.

support/mender-inventory-geo
merlin-northern: sure

support/mender-inventory-geo
merlin-northern: sure

merlin-northern:
thanks @vpodzime @lluiscampos @oleorhagen @estenberg @kacf @mirzak for comments/reviews. good that it did not get out of hand.

estenberg:
I do understand the potential security/privacy issues however I think it would be OK as long as it's clear that contacting this 3rd party service is done and it is easy to disable it.

Is there a way/place we could document this so that users become aware before going to production?

I also started a post with the community here to see if anyone has feedback on the matter: https://hub.mender.io/t/feedback-on-geolocation-from-3rd-party-as-part-of-default-inventory/2090

mirzak:
> Is there a way/place we could document this so that users become aware before going to production?

We do have this:

https://docs.mender.io/2.3/artifacts/yocto-project/building-for-production

With the new doc arch, we should probably have something more generic, where reviewing/customizing inventory scripts could be one step (this and other inventory scripts that we install by default)

PR #555, opened 2020-06-23
Ignore me

oleorhagen:

### Description

Thank you!

### DIFF
--- a/codecov.yml
+++ b/codecov.yml
@@ -1,11 +1,12 @@
 
 comment:
-  after_n_builds: 8
+  # Only comment after 7 reports uploaded
+  after_n_builds: 7
 
 codecov:
-  require_ci_to_pass: yes
   notify:
-    after_n_builds: 8
+    # Only notify after 7 reports uploaded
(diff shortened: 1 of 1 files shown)


oleorhagen:
@lluiscampos I think this is what we actually want.
1. This will bring coverage back up to where it really is (about 70%)
2. If you look closer at the codecov commits, it seems to me that the two last ones resulted in the exact same results, so I guess the last commit is not really needed.

If you agree, I will squash this, and remove the last commit, and then close all the other PRs.

oleorhagen:
The #549 will be the official follow up of this one.

lluiscampos:
>     2. If you look closer at the codecov commits, it seems to me that the two last ones resulted in the exact same results, so I guess the last commit is not really needed.
> 

Yes, I don't think it should be required, I understand that a report will be generated with the first tag, then just amended when the second part is available.

> 
> If you agree, I will squash this, and remove the last commit, and then close all the other PRs.

:+1:

oleorhagen:
Followd up in #549

PR #556, opened 2020-06-23
[support/modules] Unify shell script style in modules

lluiscampos:
Namely: indent for the break in the case statements.


lluiscampos:
Stopping `mender-qa` pipeline.

oleorhagen:
I think we should change the commit message to not mention bash

PR #557, opened 2020-06-24
MEN-3652: Remove Server config warn on mender setup command

lluiscampos:
Skip the Server configuration verification for mender setup. This
command will override these fields so it is not important to validate
contents of the current config beforehand.

This validation was printing warnings on first time setup for
on-boarding Rasperry Pi install.

Changelog: Title

### DIFF
--- a/conf/config.go
+++ b/conf/config.go
@@ -122,40 +122,49 @@ func LoadConfig(mainConfigFile string, fallbackConfigFile string) (*MenderConfig
 		return nil, loadErr
 	}
 
+	log.Debugf("Loaded %d configuration file(s)", filesLoadedCount)
+
 	if filesLoadedCount == 0 {
 		log.Info("No configuration files present. Using defaults")
 		return config, nil
 	}
 
-	if config.Servers == nil {
-		if config.ServerURL == "" {
+	log.Debugf("Loaded configuration = %#v", config)
+
+	return config, nil
+}
+
+// Validate verifies the Servers fields in the configuration
+func (c *MenderConfig) Validate() error {
+	if c.Servers == nil {
+		if c.ServerURL == "" {
 			log.Warn("No server URL(s) specified in mender configuration.")
 		}
-		config.Servers = make([]client.MenderServer, 1)
-		config.Servers[0].ServerURL = config.ServerURL
-	} else if config.ServerURL != "" {
+		c.Servers = make([]client.MenderServer, 1)
+		c.Servers[0].ServerURL = c.ServerURL
+	} else if c.ServerURL != "" {
 		log.Error("In mender.conf: don't specify both Servers field " +
 			"AND the corresponding fields in base structure (i.e. " +
-			"ServerURL). The first server on the list on the" +
-			"list overwrites these fields.")
-		return nil, errors.New("Both Servers AND ServerURL given in " +
+			"ServerURL). The first server on the list overwrites" +
+			"these fields.")
+		return errors.New("Both Servers AND ServerURL given in " +
 			"mender.conf")
 	}
-	for i := 0; i < len(config.Servers); i++ {
+	for i := 0; i < len(c.Servers); i++ {
 		// Trim possible '/' suffix, which is added back in URL path
-		if strings.HasSuffix(config.Servers[i].ServerURL, "/") {
-			config.Servers[i].ServerURL =
+		if strings.HasSuffix(c.Servers[i].ServerURL, "/") {
+			c.Servers[i].ServerURL =
 				strings.TrimSuffix(
-					config.Servers[i].ServerURL, "/")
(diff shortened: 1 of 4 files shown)


lluiscampos:
@alfrunes I have doubts with this approach. Please share your thoughts.

I tried to follow your advise in the JIRA ticket, but I see now that `LoadConfig` can return a configuration with invalid Server fields. I don't think other parts of the code should be affected (only tests, really) but still feels somewhat incomplete.

An alternative approach to keep the interface backwards compatible could be to just pass a flag to `LoadConfig`to indicate when to skip or not the verify part.

Thanks!

conf/config.go
```diff
@@ -122,40 +122,49 @@ func LoadConfig(mainConfigFile string, fallbackConfigFile string) (*MenderConfig
...
 
-	if config.Servers == nil {
-		if config.ServerURL == "" {
+	log.Debugf("Loaded configuration = %#v", config)
+
+	return config, nil
+}
+
+// VerifyServerConfig verifies the Servers fields in the configuration
+func (c *MenderConfig) VerifyServerConfig() error {
```
alfrunes, line 138: Nitpick: I wouldn't make the public interface so specific; what about simply `Validate() error`? This also leaves room for checking other possible illegal configuration in the future.

conf/config.go
```diff
@@ -122,40 +122,49 @@ func LoadConfig(mainConfigFile string, fallbackConfigFile string) (*MenderConfig
...
 		}
-		config.Servers = make([]client.MenderServer, 1)
-		config.Servers[0].ServerURL = config.ServerURL
-	} else if config.ServerURL != "" {
+		c.Servers = make([]client.MenderServer, 1)
+		c.Servers[0].ServerURL = c.ServerURL
+	} else if c.ServerURL != "" {
 		log.Error("In mender.conf: don't specify both Servers field " +
 			"AND the corresponding fields in base structure (i.e. " +
 			"ServerURL). The first server on the list on the" +
```
alfrunes, line 148: Would be nice if you can correct the typo "... The first server on the list ~~on the list~~ ..."

conf/config.go
```diff
@@ -122,40 +122,49 @@ func LoadConfig(mainConfigFile string, fallbackConfigFile string) (*MenderConfig
...
+func (c *MenderConfig) VerifyServerConfig() error {
+	if c.Servers == nil {
+		if c.ServerURL == "" {
 			log.Warn("No server URL(s) specified in mender configuration.")
 		}
-		config.Servers = make([]client.MenderServer, 1)
-		config.Servers[0].ServerURL = config.ServerURL
-	} else if config.ServerURL != "" {
+		c.Servers = make([]client.MenderServer, 1)
+		c.Servers[0].ServerURL = c.ServerURL
```
alfrunes, line 144: You may also want to move this part to loadConfig. The validation function shouldn't modify the contents but report if their wrong IMO.

conf/config.go
```diff
@@ -122,40 +122,49 @@ func LoadConfig(mainConfigFile string, fallbackConfigFile string) (*MenderConfig
...
-		return nil, errors.New("Both Servers AND ServerURL given in " +
+		return errors.New("Both Servers AND ServerURL given in " +
 			"mender.conf")
 	}
-	for i := 0; i < len(config.Servers); i++ {
+	for i := 0; i < len(c.Servers); i++ {
 		// Trim possible '/' suffix, which is added back in URL path
-		if strings.HasSuffix(config.Servers[i].ServerURL, "/") {
-			config.Servers[i].ServerURL =
+		if strings.HasSuffix(c.Servers[i].ServerURL, "/") {
```
alfrunes, line 155: You may also want to move this part as well to `loadConfig`.

conf/config.go
lluiscampos: Well, that defeats the purpose of doing this work, as the warning would be printed at `Load` time...

conf/config.go
lluiscampos: And this part I guess we could move it also, but I was thinking about an scenario like: the user load something (invalid), does some modifications in memory, then validates. Moving this to load, the possible modification will be invalid (like url's with trailing `/`)

conf/config.go
lluiscampos: This is a good idea, thanks.

conf/config.go
alfrunes: Ahh, you're right. Ignore what I said, I suppose it'll require substantial refactoring to achieve what I was suggesting.

conf/config.go
alfrunes: ahh, I only meant the part trimming the prefix (which only silently corrects the loaded config). But you may ignore this since we can't really do anything with similar correction above.

PR #559, opened 2020-06-24
MEN-3420 Invalidate server device token if tenant token changes

merlin-northern:
Go to unauth state when server URL or tenant token change.

* save tenant token and server URL in the lmdb store
* check if server URL or tenant token changed
* go to unauth state if server URL or tenant token changed

ChangeLog:title

### DIFF
--- a/app/auth.go
+++ b/app/auth.go
@@ -88,16 +88,19 @@ func NewAuthManager(conf AuthManagerConfig) AuthManager {
 }
 
 func (m *MenderAuthManager) IsAuthorized(config conf.MenderConfig) bool {
-	serverURL, err := m.AuthServerURL()
-	if err == nil {
-		return false
-	}
-	if config.ServerURL != serverURL {
-		return false
-	}
-
+	log.Infof("IsAuthorized starting;")
+	//serverURL, err := m.AuthServerURL()
+	//log.Infof("    IsAuthorized config.ServerURL:%s != serverURL:%s; err=%v.", config.Servers[0].ServerURL, serverURL, err)
+	//if err != nil {
+	//	return false
+	//}
+	//if config.Servers[0].ServerURL != serverURL {
+	//	return false
+	//}
+	//
 	tenantToken, err := m.AuthTenantToken()
-	if err == nil {
+	log.Infof("    IsAuthorized config.TenantToken:%s != tenantToken:%s; err=%v.", config.TenantToken, tenantToken, err)
+	if err != nil {
 		return false
 	}
 	if config.TenantToken != tenantToken {
@@ -108,7 +111,7 @@ func (m *MenderAuthManager) IsAuthorized(config conf.MenderConfig) bool {
 	if err != nil {
 		return false
 	}
-
+	log.Infof("    IsAuthorized adata:%s", adata)
 	if adata == noAuthToken {
 		return false
 	}
--- a/app/auth.go
+++ b/app/auth.go
@@ -18,6 +18,7 @@ import (
 	"strings"
 
 	"github.com/mendersoftware/mender/client"
+	"github.com/mendersoftware/mender/conf"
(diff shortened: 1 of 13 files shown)


merlin-northern:
this is just a draft -- please advise if this is a good direction.

datastore/dbkeys.go
```diff
@@ -38,6 +38,12 @@ const (
 	// Key used to store the auth token.
 	AuthTokenName = "authtoken"
 
+	// Key used to store the auth token.
+	AuthTenantTokenName = "tenanttoken"
+
+	// Key used to store the auth token.
+	AuthServerURLName = "serverurl"
```
kacf, line 45: Since these are related to the auth token, I think their values should reflect that too. I suggest `authtoken_tenanttoken` and `authtoken_serverurl`. A bit verbose, but these database values are often long lived, so I think it's good to be explicit.

app/auth.go
```diff
@@ -142,24 +163,54 @@ func (m *MenderAuthManager) MakeAuthRequest() (*client.AuthRequest, error) {
...
 
 	if err := m.store.WriteAll(datastore.AuthTokenName, data); err != nil {
 		return errors.Wrapf(err, "failed to save auth token")
 	}
+	if err := m.store.WriteAll(datastore.AuthTenantTokenName, []byte(tenantToken)); err != nil {
+		return errors.Wrapf(err, "failed to save tenant token")
+	}
+	if err := m.store.WriteAll(datastore.AuthServerURLName, []byte(serverURl)); err != nil {
+		return errors.Wrapf(err, "failed to save serverURL token")
+	}
```
kacf, line 179: Since these values belong together, please use transactions to write them. A good example is how `WriteAll` itself is [implemented](https://github.com/mendersoftware/mender/blob/master/store/dbstore.go#L106-L108). Basically just make sure all the writes are inside one transaction.

app/auth.go
```diff
@@ -142,24 +163,54 @@ func (m *MenderAuthManager) MakeAuthRequest() (*client.AuthRequest, error) {
 	}, nil
 }
 
-func (m *MenderAuthManager) RecvAuthResponse(data []byte) error {
+func (m *MenderAuthManager) RecvAuthResponse(data []byte, tenantToken string, serverURl string) error {
```
lluiscampos, line 166: Nitpick: `serverUrl` or `serverURL`

app/auth.go
```diff
@@ -142,24 +163,54 @@ func (m *MenderAuthManager) MakeAuthRequest() (*client.AuthRequest, error) {
...
 func (m *MenderAuthManager) AuthToken() (client.AuthToken, error) {
-	data, err := m.store.ReadAll(datastore.AuthTokenName)
+	data, err := m.store.ReadAll(datastore.AuthTenantTokenName)
 	if err != nil {
 		if os.IsNotExist(err) {
 			return noAuthToken, nil
 		}
-		return noAuthToken, errors.Wrapf(err, "failed to read auth token data")
+		return noAuthToken, errors.Wrapf(err, "failed to read tenant token data")
 	}
```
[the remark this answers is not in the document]
merlin-northern: wow. thanks. copy-paste mistake.

datastore/dbkeys.go
merlin-northern: thanks for all the comments&pointers will do!

app/auth.go
```diff
@@ -82,7 +87,23 @@ func NewAuthManager(conf AuthManagerConfig) AuthManager {
 	return mgr
 }
 
-func (m *MenderAuthManager) IsAuthorized() bool {
+func (m *MenderAuthManager) IsAuthorized(config conf.MenderConfig) bool {
+	serverURL, err := m.AuthServerURL()
+	if err == nil {
```
merlin-northern, line 92: wow, these two are good as well. I will fail my own review.

PR #564, opened 2020-07-01
Fix TestIgnoreServerConfigVerification

lluiscampos:
This test started failing in CI after the syslog fixes due to a warning
saying when unable to connect to syslog daemon. It passes locally
because it connects to syslog with no problem.

### DIFF
--- a/cli/cli_test.go
+++ b/cli/cli_test.go
@@ -479,6 +479,7 @@ func TestIgnoreServerConfigVerification(t *testing.T) {
 	// Run mender setup command (non-interactive)
 	menderSetupNonInteractive := []string{
 		"mender",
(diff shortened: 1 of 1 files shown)


lluiscampos:
Stopped initegration pipeline, this PR only fixes a unit test.

lluiscampos:
See error in latest master run [here](https://gitlab.com/Northern.tech/Mender/mender/-/jobs/617294073).

lluiscampos:
> Why didn't I see this in Gitlab in the PR...? Weird.

@kacf We were working in parallel with #557 and #558, and we didn't rebase #558 once the other was merged. Hence until both things didn't hit master we could not see the problem...

PR #573, opened 2020-07-14
single-file update module: double quotes added in paths vars referals.

merlin-northern:
ChangeLog:none

### DIFF
--- a/support/modules/single-file
+++ b/support/modules/single-file
@@ -21,39 +21,40 @@ case "$STATE" in
         ;;
 
     ArtifactInstall)
-        dest_dir=$(cat $dest_dir_file)
-        filename=$(cat $filename_file)
+        dest_dir=$(cat "${dest_dir_file}")
+        filename=$(cat "${filename_file}")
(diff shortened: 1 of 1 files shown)


merlin-northern:
the other pipeline was stopped. I changed the misspell in the title and it got trigerred

support/modules/single-file
```diff
@@ -21,39 +21,40 @@ case "$STATE" in
         ;;
 
     ArtifactInstall)
-        dest_dir=$(cat $dest_dir_file)
-        filename=$(cat $filename_file)
+        dest_dir=$(cat "${dest_dir_file}")
+        filename=$(cat "${filename_file}")
```
kacf, line 25: It is more important to put quotes around `$(cat ...)` than around `filename_file`, since we already know those don't have spaces. The content of them may have though.

This goes for all the occurrences of `$(...)`.

kacf: Looks like you're right, it's not needed in the specific case where the `$()` is being assigned to a variable.

merlin-northern:
thanks!

PR #575, opened 2020-07-15
mender.service: update to run after network-online.target

kekiefer:
Mender client startup may happen before the network is ready, so
it can't actually reach the server. It will wait its normal polling
interval to try again, but that's usually on the order of many minutes.

This change adds a weak dependency on the network-online.target and
orders itself after this target. This way, the system's regular checks
for network online will be run before the mender client starts.

Changelog: Title

Signed-off-by: Matt Madison <matt@madison.systems>
Signed-off-by: Kurt Kiefer <kekiefer@gmail.com>

### DIFF
--- a/support/mender-client.service
+++ b/support/mender-client.service
@@ -1,6 +1,7 @@
 [Unit]
 Description=Mender OTA update service
-After=systemd-resolved.service
+Wants=network-online.target
+After=systemd-resolved.service network-online.target
 
 [Service]
 Type=idle


support/mender-client.service
```diff
@@ -1,6 +1,7 @@
 [Unit]
 Description=Mender OTA update service
-After=systemd-resolved.service
+Wants=network-online.target
+After=systemd-resolved.service network-online.target
```
kacf, line 4: I think using `After` should be enough here, no? I would prefer not to use `Wants` if possible to avoid adding dependencies if people have special network setups that don't use `network-online.target`.

kekiefer: The reason this is `Wants` and not `Requires` is exactly so that network-online.target is optional.

This change is meant to provide a sane default. Without the `Wants` here, the network-online.target would only be started by accident unless the user knew they had to include it in some other way; none of the standard Yocto targets include network-online.target directly, it's pulled in by the services that need it (such as autofs.service, where it's also `Wants`).

For the special cases, the user would be doing something knowingly. There, the user would purposefully not include the network-online.target or not put anything in it, and both would work fine.

kacf: Alright, you convinced me! :-)

I will start our full test suite.

PR #576, opened 2020-07-15
Improve message when we detect booting from the wrong partition.

kacf:
Changelog: None

### DIFF
--- a/installer/dual_rootfs_device.go
+++ b/installer/dual_rootfs_device.go
@@ -28,6 +28,14 @@ import (
 	log "github.com/sirupsen/logrus"
 )
 
+const (
+	verifyRebootError = "Reboot to the new update failed. " +
+		"Expected \"upgrade_available\" flag to be true but it was false. " +
+		"Either the switch to the new partition was unsuccessful, or the bootloader rolled back"
+	verifyRollbackRebootError = "Reboot to the old update failed. " +
+		"Expected \"upgrade_available\" flag to be false but it was true"
+)
+
 type DualRootfsDeviceConfig struct {
 	RootfsPartA string
 	RootfsPartB string
@@ -249,7 +257,7 @@ func (d *dualRootfsDeviceImpl) VerifyReboot() error {
 	if err != nil {
 		return err
 	} else if !hasUpdate {
-		return errors.New("Reboot to new update failed. Expected \"upgrade_available\" flag to be true but it was false")
+		return errors.New(verifyRebootError)
 	} else {
 		return nil
 	}
@@ -260,7 +268,7 @@ func (d *dualRootfsDeviceImpl) VerifyRollbackReboot() error {
 	if err != nil {
 		return err
 	} else if hasUpdate {
-		return errors.New("Reboot to old update failed. Expected \"upgrade_available\" flag to be false but it was true")
+		return errors.New(verifyRollbackRebootError)
 	} else {
 		return nil
 	}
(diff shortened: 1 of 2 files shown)


installer/dual_rootfs_device.go
```diff
@@ -28,6 +28,14 @@ import (
 	log "github.com/sirupsen/logrus"
 )
 
+const (
+	verifyRebootError = "Reboot to new update failed. " +
```
merlin-northern, line 32: reword to "_Reboot to the new update failed_"

installer/dual_rootfs_device.go
```diff
@@ -28,6 +28,14 @@ import (
...
+		"Expected \"upgrade_available\" flag to be true but it was false. " +
+		"Either the switch to the new partition was unsuccessful, or the bootloader rolled back"
+	verifyRollbackRebootError = "Reboot to old update failed. " +
```
merlin-northern, line 35: suggested rewording: ".. to *the* old update .."

installer/dual_rootfs_device.go
```diff
@@ -28,6 +28,14 @@ import (
 	log "github.com/sirupsen/logrus"
 )
 
+const (
+	verifyRebootError = "Reboot to new update failed. " +
+		"Expected \"upgrade_available\" flag to be true but it was false. " +
+		"Either the switch to the new partition was unsuccessful, or the bootloader rolled back"
```
merlin-northern, line 34: there is a full-stop ('.') at the end of `verifyRebootError` and there is none at the end of `verifyRollbackRebootError` I remember go once told me not to use '.' in error messages, but take your pick. I guess it would be good to have it uniform.

merlin-northern:
pipeline: [all green](https://gitlab.com/Northern.tech/Mender/mender-qa/-/pipelines/166713904)

kacf:
Wtf... One of your comments disappeared. Anyway, I was gonna say this:

> The reason it is recommended not to end with a period is that errors tend to be "wrapped" and concatenated using ":". So this could for example turn into `Reboot to new update failed. Expected \"upgrade_available\" flag to be true but it was false. Either the switch to the new partition was unsuccessful, or the bootloader rolled back: Unable to upgrade`. Hence what I have done is the best way I think, because each sentence is properly terminated, except the last one, which may wrap a different error.

kacf:
> isnt it a user-facing change?

Strictly speaking, yes. But there is no significant behavior change, bug fix, or new feature here. The key question to myself is always: "Does this change anything for the user?". And this doesn't really change anything, it's just troubleshooting help. So I considered it too small.

But good that you are looking out for this! :-)

installer/dual_rootfs_device.go
kacf: Will change!

installer/dual_rootfs_device.go
kacf: Will change!

merlin-northern:
> Wtf... One of your comments disappeared. Anyway, I was gonna say this:
> 
> > The reason it is recommended not to end with a period is that errors tend to be "wrapped" and concatenated using ":". So this could for example turn into `Reboot to new update failed. Expected \"upgrade_available\" flag to be true but it was false. Either the switch to the new partition was unsuccessful, or the bootloader rolled back: Unable to upgrade`. Hence what I have done is the best way I think, because each sentence is properly terminated, except the last one, which may wrap a different error.

sorry I typed it, and immediately thought it was stupid, and deleted it after 2s. nothing escapes you!

PR #577, opened 2020-07-17
reboot device Update Module

merlin-northern:
ChangeLog:none

### DIFF
--- a/support/modules-artifact-gen/reboot-gen
+++ b/support/modules-artifact-gen/reboot-gen
@@ -1,0 +1,108 @@
+#!/bin/sh
+
+set -e
+
+show_help() {
+  cat << EOF
+
+Simple tool to generate Mender Artifact suitable for reboot Update Module
+
+Usage: $0 [options] [-- [options-for-mender-artifact] ]
+
+    Options: [ -n|artifact-name -t|--device-type -o|--output_path -h|--help ]
+
+        --artifact-name     - Artifact name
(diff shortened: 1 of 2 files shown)


kacf:
I think we should not put this in the main repository, because we know that this is a hacky method, and we would rather implement this in a different way. What we would like is to enable [persistent connections](https://drive.google.com/file/d/15ZyLIB3W5QFWONNEVhmOppUyTNE4dGst/view?usp=sharing) and then adding actionable API that can reboot without going through an update. This could also be used to log in via SSH, without having an open port 22, for example. So I suggest this is added on Mender Hub instead, as a more temporary community hack.

merlin-northern:
> I think we should not put this in the main repository, because we know that this is a hacky method, and we would rather implement this in a different way. What we would like is to enable [persistent connections](https://drive.google.com/file/d/15ZyLIB3W5QFWONNEVhmOppUyTNE4dGst/view?usp=sharing) and then adding actionable API that can reboot without going through an update. This could also be used to log in via SSH, without having an open port 22, for example. So I suggest this is added on Mender Hub instead, as a more temporary community hack.

sure will do, let me wait for the support case this was created for, and them I will move it.

drewmoseley:
@merlin-northern  we do have https://github.com/mendersoftware/mender-update-modules which would be a good place for this.

merlin-northern:
thanks. closing. moved to https://github.com/mendersoftware/mender-update-modules/pull/17

PR #585, opened 2020-08-12
MEN-3114: Client uses certificate private key when mTLS is enabled

alfrunes:

### Description

Thank you!

### DIFF
--- a/app/auth_test.go
+++ b/app/auth_test.go
@@ -97,6 +97,18 @@ func TestAuthManager(t *testing.T) {
 	code, err = am.AuthToken()
 	assert.Equal(t, client.AuthToken("footoken"), code)
 	assert.NoError(t, err)
+
+	am = NewAuthManager(AuthManagerConfig{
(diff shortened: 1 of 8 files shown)


app/auth_test.go
```diff
@@ -97,6 +97,18 @@ func TestAuthManager(t *testing.T) {
...
+		AuthDataStore: ms,
+		IdentitySource: &dev.IdentityDataRunner{
+			Cmdr: cmdr,
+		},
+		KeyStore: store.NewKeystore(ms, "key", true),
+	})
+	err = am.GenerateKey()
+	if assert.Error(t, err) {
+		assert.True(t, store.IsStaticKey(err))
+	}
```
lluiscampos, line 111: I don't think you need to to protect the second assert.

```suggestion
assert.Error(t, err)
assert.True(t, store.IsStaticKey(err))
```

alfrunes: It makes the error message a little cleaner if the first one fails.

PR #586, opened 2020-08-12
Reduce binary size by eliminating golang.org/x/net/http2

alfrunes:
From the [docs](https://pkg.go.dev/golang.org/x/net/http2?tab=doc):
*http2*: This package is low-level and intended to be used directly by very few people. Most users will use it indirectly through the automatic use by the net/http package (from Go 1.6 and later).
This package is responsible for ~0.5MiB of the binary size (on x86-64).

I also cleaned up indirect dependencies from go.mod.


oleorhagen:
LGTM

However, I'm not sure about the indirect dependencies. I thought they were needed?

alfrunes:
> However, I'm not sure about the indirect dependencies. I thought they were needed?

If they are included in go.mod they are fixed/overwritten to the given version which may conflict with the version from the dependency that imported this dependency in the first place (e.g. go-liblzma changed hash in this PR because the mod'ed version in this repo was not the same as in `mender-artifact`).

oleorhagen:
> > However, I'm not sure about the indirect dependencies. I thought they were needed?
> 
> If they are included in go.mod they are fixed/overwritten to the given version which may conflict with the version from the dependency that imported this dependency in the first place (e.g. go-liblzma changed hash in this PR because the mod'ed version in this repo was not the same as in `mender-artifact`).

I think my question is, these indirect deps are added by go mod in the first place. Why were they wrong?

alfrunes:
> I think my question is, these indirect deps are added by go mod in the first place. Why were they wrong?

I'm not sure TBH, perhaps these were leftovers from running `go mod init` with an already existing `govendor` tree.

client/client.go
```diff
@@ -270,10 +269,6 @@ func New(conf Config) (*ApiClient, error) {
 		KeepAlive: connectionKeepaliveTime,
 	}).DialContext
 
-	if err := http2.ConfigureTransport(transport); err != nil {
-		log.Warnf("failed to enable HTTP/2 for client: %v", err)
-	}
```
lluiscampos, line 275: We don't have a way to test that `net/http` will actually use HTTP/2, right?

PR #587, opened 2020-08-12
Move to mender software openssl fork

merlin-northern:
MEN-3870 (related) Moving to mendersoftware/openssl.

Moving from Linutronix fork to our own fork of spacemonkey/openssl.

ChangeLog:none
Signed-off-by: Peter Grzybowski <peter@northern.tech>

PR #588, opened 2020-08-14
MEN-3115: Enable mTLS in the client side transport code 

oleorhagen:
For now, this is two, commits, simply because of the order the work was done in. Will squash into one before merging.

So, in short, enabling the mTLS configuration option in the client. Unit tests it with a mock mTLS server. 

And removes the `SkipVerify` option from the configuration, as it should not be possible to skip verification of client-side TLS.


client/client.go
```diff
@@ -281,21 +281,93 @@ func newHttpClient() *http.Client {
 	return &http.Client{}
 }
 
+func loadServerTrust(ctx *openssl.Ctx, conf *Config) (*openssl.Ctx, error) {
+	// Trust CA's in the standard location (/etc/ssl/certs), and the
+	// configured server certificate when building the certificate chain
+	err := ctx.LoadVerifyLocations(conf.ServerCert, "/etc/ssl/certs/")
```
oleorhagen, line 287: For now, hardcoded `/etc/ssl/certs`, but this should be changed before merging, when the std-certpool loading from @merlin-northern  is in place.

client/client.go
```diff
@@ -327,18 +399,8 @@ func dialOpenSSL(conf Config, network string, addr string) (net.Conn, error) {
...
-
-	if conf.NoVerify {
-		log.Warnf("certificate verification skipped..")
-	}
-	tlsc := tls.Config{
-		RootCAs:            trustedcerts,
-		InsecureSkipVerify: conf.NoVerify,
-	}
 	transport := http.Transport{
-		TLSClientConfig: &tlsc,
```
oleorhagen, line 340: Removed, since I can't see that it is no longer needed, since the `OpenSSL` lib now handles the TLS communication

conf/config.go
```diff
@@ -222,11 +231,30 @@ func SaveConfigFile(config *MenderConfigFromFile, filename string) error {
 	return nil
 }
 
+func maybeHTTPSClient(c *MenderConfig) *client.HttpsClient {
+	if c.HttpsClient.Certificate != "" && c.HttpsClient.Key != "" {
+		return &c.HttpsClient
+	}
+	// TODO -- Duplicate of the above functionality - merge
```
oleorhagen, line 238: Some code duplication. I guess this is a question for @alfrunes 

Is the check and notification needed in the configuration code? For now it's in both places...

client/client.go
```diff
@@ -281,21 +271,93 @@ func newHttpClient() *http.Client {
 	return &http.Client{}
 }
 
+func loadServerTrust(ctx *openssl.Ctx, conf *Config) (*openssl.Ctx, error) {
+	// Trust CA's in the standard location (/etc/ssl/certs), and the
```
oleorhagen, line 275: Also, @merlin-northern, should we possibly load the certificates explicitly here, instead of passively through the `LoadVerifyLocations` call? So that it is similar to `loadClientTrust`? It would leave us with more control I think. At least after you've finished the `SSL_DEFAULT_LOCATION` implementation?

client/client.go
```diff
@@ -350,64 +410,16 @@ func newHttpsClient(conf Config) (*http.Client, error) {
 
 // Client configuration
 
-type Config struct {
-	ServerCert string
-	IsHttps    bool
-	NoVerify   bool
```
alfrunes, line 356: I don't understand, are we removing the option to skip CA validation entirely?

oleorhagen: Perhaps you are right, I removed it because this was set from within the `HttpsClient` struct. 
I guess I can add a new configuration option directly, to set this

client/client.go
merlin-northern: I think we should use LoadVerifyLocations as it calls SSL_CTX_load_verify_locations which is the proper way to load the trust store. I think it can be like that. As soon as we merge this one, then we can move with [pull/587](https://github.com/mendersoftware/mender/pull/587) and then the only modification here, will be: add `SetDefaultVerifyLocations` call.
@oleorhagen do you need to hard code "/etc/ssl/certs/" here for the sake of the tests, or for some other reason? if not, I would drop it, since technically it can break in case the /etc/ssl does not exist. (I am not sure how SSL_CTX_load_verify_locations behaves when the directory does not exist). keep it in if it is needed for this PR.

client/client.go
merlin-northern: thats right. I would like to get this one first, then [pull/587](https://github.com/mendersoftware/mender/pull/587) and then the follow up patch here removing the hardcoded path.

client/client.go
oleorhagen: `/etc/ssl/certs` is only a temporary fix, until #587 and your confirmation on it :)

client/client.go
```diff
@@ -281,21 +281,97 @@ func newHttpClient() *http.Client {
...
+			return ctx, err
+		}
+	}
+
+	keyBytes, err := ioutil.ReadFile(keyFile)
+	if err != nil {
+		return ctx, errors.Wrap(err, "Private key file from the HttpsClient configuration not found")
+	}
+
+	key, err := openssl.LoadPrivateKeyFromPEM(keyBytes)
```
merlin-northern, line 334: just a comment: in here the potential private key loading from an Engine will take place (with EngineLoadPrivateKey).

client/client.go
oleorhagen: See https://github.com/mendersoftware/mender/pull/588#discussion_r470489503

client/client.go
```diff
@@ -281,21 +281,97 @@ func newHttpClient() *http.Client {
...
 
+func loadServerTrust(ctx *openssl.Ctx, conf *Config) (*openssl.Ctx, error) {
+	// Trust CA's in the standard location (/etc/ssl/certs), and the
+	// configured server certificate when building the certificate chain
+	err := ctx.LoadVerifyLocations(conf.ServerCert, "/etc/ssl/certs/")
+	return ctx, err
+}
+
+func loadClientTrust(ctx *openssl.Ctx, conf *Config) (*openssl.Ctx, error) {
+
```
merlin-northern, line 290: do we need this extra line?

client/client.go
```diff
@@ -281,21 +281,97 @@ func newHttpClient() *http.Client {
...
+	if err != nil {
+		return ctx, err
+	}
+
+	for _, pem := range certs {
+		cert, err := openssl.LoadCertificateFromPEM(pem)
+		if err != nil {
+			return ctx, err
+		}
+		err = ctx.AddChainCertificate(cert)
```
merlin-northern, line 323: good call on loading the chain!

client/client.go
```diff
@@ -281,21 +281,97 @@ func newHttpClient() *http.Client {
...
 }
 
+func loadServerTrust(ctx *openssl.Ctx, conf *Config) (*openssl.Ctx, error) {
+	// Trust CA's in the standard location (/etc/ssl/certs), and the
+	// configured server certificate when building the certificate chain
+	err := ctx.LoadVerifyLocations(conf.ServerCert, "/etc/ssl/certs/")
+	return ctx, err
+}
+
+func loadClientTrust(ctx *openssl.Ctx, conf *Config) (*openssl.Ctx, error) {
```
merlin-northern, line 289: we should call here `SSL_CTX_clear_chain_certs` just to be safe (technically we do not know where did ctx come from), but the wrapper currently does not have this function, I will add it to the wrapper, and then come up with a PR.

oleorhagen: Could you expand a little here, I'm not following. We do load the context ourselves, and it is a new ctx everytime?

merlin-northern: yes, it is new every time. cleaning the chain is just a safety precaution. `loadClientTrust` is expected to load the trust, and it does not know anything about previous whereabouts of ctx. not a big deal, and not a blocker here.

client/client.go
oleorhagen: Question:
Why not do it the other way around? First #587 and then this?

oleorhagen: That means we can clean this up also in this same PR: https://github.com/mendersoftware/mender/pull/588/files#r471329025

client/client.go
oleorhagen: Not really, but it is not uncommon for our code to have a little air here

client/client.go
merlin-northern: > Question:
> Why not do it the other way around? First #587 and then this?

the only reason: I figured it would be easier for me to apply #587 on top of this one, than for you to resolve the conflicts. but if you are fine with getting #587 first, then I am all for it!

oleorhagen: Let's do it that way then. I don't mind resolving them, and it would be nice to get this PR and commit clean with most all the `OpenSSL` stuff contained in one nice and clean PR, considering it is a pretty solid functionality shift.

client/client_test.go
```diff
@@ -193,81 +191,116 @@ func TestHttpClientUrl(t *testing.T) {
...
-	var systemOK, oursOK bool
-	subj := certs.Subjects()
-	for i := 0; i < len(subj); i++ {
-		if strings.Contains(string(subj[i]), "VeriSign, Inc.") {
-			systemOK = true
-		}
-		// "Acme Co", just a dummy certificate in this repo.
-		if strings.Contains(string(subj[i]), "Acme Co") {
-			oursOK = true
+func TestCALoading(t *testing.T) {
```
merlin-northern, line 194: shall we consider adding a test for the chain of certificates? (to test `loadClientTrust` logic?)

oleorhagen: I think that is a good idea actually, yeh. Let's!

client/client.go
```diff
@@ -275,6 +280,9 @@ func loadServerTrust(ctx *openssl.Ctx, conf *Config) (*openssl.Ctx, error) {
 	// Trust CA's in the standard location (/etc/ssl/certs), and the
 	// configured server certificate when building the certificate chain
 	err := ctx.LoadVerifyLocations(conf.ServerCert, "/etc/ssl/certs/")
+	if err != nil && strings.Contains(err.Error(), "No such file or directory") {
```
kacf, line 283: Shorter and more robust:

```
if os.IsNotExist(errors.Cause(err)) {
```

client/client.go
```diff
@@ -366,33 +357,50 @@ func dialOpenSSL(conf Config, network string, addr string) (net.Conn, error) {
...
 
 func newHttpsClient(conf Config) (*http.Client, error) {
 	client := newHttpClient()
 
+	ctx, err := openssl.NewCtx()
+	if err != nil {
+		return nil, err
+	}
+
+	ctx, err = loadServerTrust(ctx, &conf)
```
kacf, line 388: Is this needed inside `newHttpsClient`? It seems improper in here.

client/client.go
```diff
@@ -416,6 +416,21 @@ type HttpsClient struct {
...
+	}
+	// Check if both the 'ServerCertificate' and 'PrivateKey' fields are set
+	if h.Certificate != "" || h.Key != "" {
+		if h.Certificate == "" {
+			log.Error("The 'PrivateKey' field is set in the mTLS configuration, but no 'ServerCertificate' is given. Both need to be present in order for mTLS to function")
+		}
+		if h.Key == "" {
+			log.Error("The 'ServerCertificate' field is set in the mTLS configuration, but no 'PrivateKey' is given. Both need to be present in order for mTLS to function")
+		}
+	}
```
kacf: The error messages do not match what is actually being tested, and neither do the comments. I'm guessing this was started in one way, and then the code was later changed, but not the messages and comments. The code appears to be correct, so just changing the texts should be enough.

client/client.go
```diff
@@ -350,64 +419,18 @@ func newHttpsClient(conf Config) (*http.Client, error) {
...
-
-type systemCertPool struct{}
-
-func (systemCertPool) GetSystemCertPool() (*x509.CertPool, error) {
-	return x509.SystemCertPool()
+// HttpsClient holds the configuration for the client side mTLS configuration
+type HttpsClient struct {
+	Certificate string
+	Key         string
+	SkipVerify  bool
```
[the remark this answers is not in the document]
oleorhagen: Yes, this is gone now. thx

client/client.go
merlin-northern: ok, https://github.com/mendersoftware/mender/pull/587 has been merged (with the required meta-mender PR)

client/client.go
merlin-northern: @oleorhagen since https://github.com/mendersoftware/openssl/pull/11 and https://github.com/mendersoftware/mender/pull/587 are in place, maybe lets just add `ctx.SetDefaultVerifyLocations()` here, after rebase? (just a suggestion I can do it afterwards)

client/client_test.go
```diff
@@ -193,81 +191,116 @@ func TestHttpClientUrl(t *testing.T) {
...
-	// Verify that at least one of the certificates belong to us, and one
-	// belongs to a well known certificate authority.
-	var systemOK, oursOK bool
-	subj := certs.Subjects()
-	for i := 0; i < len(subj); i++ {
-		if strings.Contains(string(subj[i]), "VeriSign, Inc.") {
-			systemOK = true
-		}
-		// "Acme Co", just a dummy certificate in this repo.
-		if strings.Contains(string(subj[i]), "Acme Co") {
```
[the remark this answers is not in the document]
merlin-northern: There is no call that can easily be used to get the default location, but we do not need that, it is enough to call `ctx.SetDefaultVerifyLocations()` now, as all is merged (see [above comment](https://github.com/mendersoftware/mender/pull/588#discussion_r472117054))

client/client.go
oleorhagen: Rebased now. Looks good.

client/client.go
oleorhagen: Added

client/client_test.go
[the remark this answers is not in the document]
oleorhagen: Unless you have some strong objections here, @kacf I'm going to resolve this. 

So in short: It is annoyingly difficult to do now, but I'm sure we can do it. Depends on how hard of a requirement it is

client/client_test.go
```diff
@@ -193,81 +191,116 @@ func TestHttpClientUrl(t *testing.T) {
...
+
+		ctx, err = loadServerTrust(ctx, &Config{
+			IsHttps:     true,
+			ServerCert:  "server.crt",
+			HttpsClient: nil,
+			NoVerify:    false,
+		})
+		assert.NoError(t, err)
+	})
+	t.Run("Test loading client trust", func(t *testing.T) {
```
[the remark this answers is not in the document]
oleorhagen: I renamed the test to `TestLoadingTrust` fair compromise?

client/client.go
```diff
@@ -423,14 +416,12 @@ func newHttpsClient(conf Config) (*http.Client, error) {
...
 	Certificate string
 	Key         string
-	SkipVerify  bool
 }
 
 type Config struct {
 	IsHttps    bool
 	ServerCert string
 	*HttpsClient
-	NoVerify bool
```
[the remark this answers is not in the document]
oleorhagen: `NoVerify` is back in favour :partying_face:

client/client.go
```diff
@@ -350,64 +410,31 @@ func newHttpsClient(conf Config) (*http.Client, error) {
...
+		if h.Key == "" {
+			log.Error("The 'ServerCertificate' field is set in the mTLS configuration, but no 'PrivateKey' is given. Both need to be present in order for mTLS to function")
 		}
-
-		syscerts.AppendCertsFromPEM(servcert)
 	}
+}
 
-	if len(syscerts.Subjects()) == 0 {
-		log.Error(errMissingCerts)
```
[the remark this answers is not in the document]
oleorhagen: Let's just keep this in: https://github.com/mendersoftware/mender/pull/588#discussion_r471947929

client/client.go
oleorhagen: It looks correct to me. 
It checks if only one of the two required fields are set
Case-Study:
1) If `Certificate` is set, and not `PrivateKey`
```
if h.Certificate != "" || h.key != "" {
```
is true, since `h.Certificate` is not `""`
```
if h.Certificate == "" {
```
is false, since, `h.Certificate` is set.
```
h.Key == "" {
```
is true, since `h.Key` is not set.
The following error message:
```
The 'ServerCertificate' field is set in the mTLS configuration, but no 'PrivateKey' is given. Both need to be present in order for mTLS to function
```
Then looks correct to me. Other case  follows by symmetry

It'd be fun if I was wrong here though, I had to stare at this for a bit :p

oleorhagen: @kacf

client/client_test.go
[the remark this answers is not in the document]
kacf: It sounds like the effort / payoff ratio is too high here. We will catch errors here eventually, because Hosted Mender will not work if this doesn't work, and testing Hosted Mender is part of our manual tests. It just won't be tested automatically. So I'm ok with resolving this.

client/client_test.go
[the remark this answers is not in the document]
kacf: :+1:

client/client.go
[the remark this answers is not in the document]
kacf: No, sorry, but these are two different things. The comment you linked to is about testing CA certificate loading, and I'm ok with omitting that, but this thread is about warning messages on the device. The reason I think this is important is that installing Mender on a platform which has no CA certificates is a common user mistake, and this helps them (or us on Mender Hub) identify the problem.

client/client.go
```diff
@@ -350,64 +417,32 @@ func newHttpsClient(conf Config) (*http.Client, error) {
...
-
-type systemCertPool struct{}
-
-func (systemCertPool) GetSystemCertPool() (*x509.CertPool, error) {
-	return x509.SystemCertPool()
-}
-
-func loadServerTrust(conf *Config) *x509.CertPool {
-	return loadServerTrustImpl(conf, systemCertPool{})
+// HttpsClient holds the configuration for the client side mTLS configuration
```
kacf, line 420: It's not obvious from the context where this is used. I would consider adding a comment: "Careful when changing this. The struct is directly exposed in the mender.conf file."

client/client.go
kacf: The code is correct, as I said, it's the text which is wrong. The message

> The 'ServerCertificate' field is set in the mTLS configuration, but no 'PrivateKey' is given. Both need to be present in order for mTLS to function

is wrong because neither `ServerCertificate` nor `PrivateKey` are valid fields. They are called `Certificate` and `Key`. Same with the error messages, they are instructing the user to change fields that don't exist.

client/client.go
kacf: Never mind. I misread it, I thought `newHttpsClient` was referring to "client certificate", but it's of course talking about Mender, the client.

client/client.go
oleorhagen: Is changing it to `Certificate` and `Key` sufficient?

kacf (changes requested):

client/client.go
kacf: Github is confused about mixing reviews and comments, apparently...

But yes, that should be good!

client/client.go
oleorhagen: True! Maybe a naming change is better? 
`HttpsClientConfig`?

client/client.go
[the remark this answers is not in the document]
oleorhagen: On it now

client/client.go
kacf: For good style and self containment, I would argue that the struct should live in `conf/config.go`, and either we would use that struct from `client/client.go`, or it can use a different struct altogether.

kacf: Yes, that's fine.

oleorhagen:
Pipeline: https://gitlab.com/Northern.tech/Mender/mender-qa/-/pipelines/181291706

client/client.go
```diff
@@ -281,21 +280,131 @@ func newHttpClient() *http.Client {
 	return &http.Client{}
 }
 
-func dialOpenSSL(conf Config, network string, addr string) (net.Conn, error) {
-	contextSSL, err := openssl.NewCtx()
-	// probably should consider reusing the context, but then we
-	// have to propagate it with every request.
-	err = contextSSL.LoadVerifyLocations(conf.ServerCert, "")
+func listSystemCertsFound(certDir string) (int, error) {
```
oleorhagen, line 283: @kacf This is supposed to mimic the `x509.SystemCertPool()` functionality in the standard `Go` lib. Except, that it only counts valid certificates in the directory.

client/client.go
```diff
@@ -281,21 +280,131 @@ func newHttpClient() *http.Client {
...
+	err = ctx.LoadVerifyLocations(conf.ServerCert, "")
+	if err != nil {
+		if strings.Contains(err.Error(), "No such file or directory") {
+
+			log.Errorf(errMissingServerCertF, conf.ServerCert)
+		} else {
+			log.Errorf("Failed to Load the Server certificate. Err %v", err.Error())
+		}
+		if sysCertsFound == 0 {
+			log.Error(errMissingCerts)
```
oleorhagen, line 335: And therefore we can once again issue the `errMissingCerts` error, like you requested, if no certificates were loaded - neither configuration server cert, or system certificates.

client/client.go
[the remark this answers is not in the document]
oleorhagen: See: https://github.com/mendersoftware/mender/pull/588/files#r475580878

oleorhagen:
@kacf I am sorry for the force-push. But in order to speed things up for fabio, this was rebased on top of the `openSSLv1.0.8` PR. Hence the force push.

client/client.go
```diff
@@ -281,21 +280,131 @@ func newHttpClient() *http.Client {
...
+			continue
+		}
+
+		certs := openssl.SplitPEM(certBytes)
+		if len(certs) == 0 {
+			log.Debugf("No PEM certificate found in '%s'", certFile)
+			continue
+		}
+		first, certs := certs[0], certs[1:]
+		_, err = openssl.LoadCertificateFromPEM(first)
```
kacf, line 302: Should it not load them all? I'm no expert on how System Certs are stored, but I can imagine they could contain concatenated certificates.

oleorhagen: If I've understood this correctly, this is done automatically from the `https://www.openssl.org/docs/man1.1.0/man3/SSL_CTX_set_default_verify_paths.html`, and `http://www.openssl.org/docs/ssl/SSL_CTX_load_verify_locations.html` functions.

client/client.go
kacf: Suggest renaming to `checkSystemCertsFound` to indicate that it doesn't return or store anything. Otherwise looks good!

client/client.go
kacf: :+1:

client/client.go
```diff
@@ -281,21 +280,142 @@ func newHttpClient() *http.Client {
...
+
+	// Set the default system certificate path for this OpenSSL context
+	err = ctx.SetDefaultVerifyPaths()
+	if err != nil {
+		return ctx, fmt.Errorf("Failed to set the default OpenSSL directory. OpenSSL error code: %s", err.Error())
+	}
+	// Load the server certificate into the OpenSSL context
+	err = ctx.LoadVerifyLocations(conf.ServerCert, "")
+	if err != nil {
+		if strings.Contains(err.Error(), "No such file or directory") {
```
kacf, line 337: `os.IsNotExist(errors.Cause(err))`

client/client.go
```diff
@@ -281,21 +280,142 @@ func newHttpClient() *http.Client {
...
+
+			log.Errorf(errMissingServerCertF, conf.ServerCert)
```
kacf, line 339: This was logged at `Warnf` level in the old client, and I think we should keep that, since this is the normal operating procedure for example for Hosted Mender. If we really don't have *any* certificate, we will hit the other error below.

store/dirstore_test.go
```diff
@@ -106,4 +106,10 @@ func TestDirStore(t *testing.T) {
 	// closing is a noop, no errors should be reported
 	err = d.Close()
 	assert.NoError(t, err)
+
+	// Test reading from an absolute path
+	tf, err := ioutil.TempFile(tmppath, "abspathtest")
+	assert.NoError(t, err)
```
kacf, line 112: Nitpick:

`defer os.RemoveAll(tf.Name())`

kacf (changes requested):

client/client.go
oleorhagen: I tried this actually, and it won't work. The error is a string returned from the OpenSSL wrapper, and hence it is not typed :/ so it is not recognized by `IsNotExist`. Sry, I should have mentioned this.

oleorhagen:
Pipeline along with #integration/988:
https://gitlab.com/Northern.tech/Mender/mender-qa/-/pipelines/182508983

oleorhagen:
Latest pipe with #integration/988:
https://gitlab.com/Northern.tech/Mender/mender-qa/-/pipelines/182525902

client/client.go
kacf: Ah. Weird that it doesn't use the errno code, because it is supposed to be supported by `IsNotExist` as well. But oh well, we'll leave it as is then.

kacf (approved):

lluiscampos:
@oleorhagen I think Mr. Codecov just messed up the check here in GitHub. The reason is unknown but I don't want to spend more time on it.

The report of the last commit for this PR ([here](https://codecov.io/gh/mendersoftware/mender/tree/1eb30c410e6b2f53525283154122f336e60e4aed)) shows 70.45% absolute coverage (including acceptance tests); so that should be fine.

Should I merge?

kacf:
:tada:

PR #596, opened 2020-08-24
test: Only include mendersoftware/mender files in coverage 

oleorhagen:
Previously all packages matching the prefix mendersoftware/mender would be
included in the coverage report. This meant that 'mender-artifact' would be
included in the test coverage reports, which is unnecessary, considering it is a
package of its own.

Changelog: None

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -205,7 +205,7 @@ htmlcover: coverage
 
 coverage:
 	rm -f coverage.txt
-	$(GO) test -coverprofile=coverage-tmp.txt -coverpkg=github.com/mendersoftware/... ./...
+	$(GO) test -coverprofile=coverage-tmp.txt -coverpkg=./... ./...
 	if [ -f coverage-missing-subtests.txt ]; then \
 		echo 'mode: set' > coverage.txt; \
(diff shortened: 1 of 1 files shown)


lluiscampos:
@oleorhagen I don't really understand what the problem is here. If I look to the [latest codecov report for master](https://codecov.io/gh/mendersoftware/mender/tree/7f9bd9c1f036a1775e0d956118caa2fffce7b941), there is no trace of mender-artifact.

I guess your point is that mender-artifact coverage is collected by `go test` and then ignored by codecov, and this is just a cleanup with no real effect...

lluiscampos (approved):

oleorhagen:
> I guess your point is that mender-artifact coverage is collected by `go test` and then ignored by codecov, and this is just a cleanup with no real effect...

That is correct.

The difference is noticeable when you run `make htmlcover` locally though. As then the junk shows up.

oleorhagen:
Running it through the entire pipe, just so we're sure we don't mess with `codecov`/

oleorhagen:
No coverage diff, like expected. 
Merging

PR #598, opened 2020-08-28
MEN-3871 Load private key from engine.

merlin-northern:
Ability to load a private key from a given engine.

* if `HttpsClient.Key` contains "pkcs11:" prefix treat it as PCKS#11 URI
* new configuration directive: `HttpsClient.SSLEngine`
* tests
  * done in: https://github.com/mendersoftware/openssl/pull/17
  * integration test coming with [MEN-3905](https://tracker.mender.io/browse/MEN-3905)

ChangeLog:none
(cherry picked from commit b805a45e2d8d64309d370ce75e7cc98c2378aea1)

### DIFF
--- a/client/client.go
+++ b/client/client.go
@@ -364,6 +364,7 @@ func loadPrivateKey(keyFile string, engineId string) (key openssl.PrivateKey, er
 				engineId, err.Error())
 			return nil, err
 		}
+		log.Infof("laoded private key: '%v' from '%s'.", key, engineId)
 	} else {
 		keyBytes, err := ioutil.ReadFile(keyFile)
 		if err != nil {
--- a/client/client.go
+++ b/client/client.go
@@ -41,6 +41,7 @@ const (
 	errMissingCerts = "No trusted certificates. The client will continue running, but will " +
 		"not be able to communicate with the server. Either specify ServerCertificate in " +
 		"mender.conf, or make sure that CA certificates are installed on the system"
+	pkcs11URIPrefix = "pkcs11:"
 )
 
 var (
@@ -348,6 +349,36 @@ func loadServerTrust(ctx *openssl.Ctx, conf *Config) (*openssl.Ctx, error) {
 	return ctx, err
 }
 
+func loadPrivateKey(keyFile string, engineId string) (key openssl.PrivateKey, err error) {
+	if strings.HasPrefix(keyFile, pkcs11URIPrefix) {
+		engine, err := openssl.EngineById(engineId)
+		if err != nil {
+			log.Errorf("Failed to Load '%s' engine. Err %s",
+				engineId, err.Error())
+			return nil, err
+		}
+
+		key, err = openssl.EngineLoadPrivateKey(engine, keyFile)
+		if err != nil {
+			log.Errorf("Failed to Load private key from engine '%s'. Err %s",
+				engineId, err.Error())
+			return nil, err
+		}
+	} else {
+		keyBytes, err := ioutil.ReadFile(keyFile)
+		if err != nil {
+			return nil, errors.Wrap(err, "Private key file from the HttpsClient configuration not found")
+		}
+
+		key, err = openssl.LoadPrivateKeyFromPEM(keyBytes)
+		if err != nil {
+			return nil, err
+		}
+	}
+
+	return key, nil
+}
+
 func loadClientTrust(ctx *openssl.Ctx, conf *Config) (*openssl.Ctx, error) {
 
 	if conf.HttpsClient == nil {
@@ -355,7 +386,6 @@ func loadClientTrust(ctx *openssl.Ctx, conf *Config) (*openssl.Ctx, error) {
 
 	}
 	certFile := conf.HttpsClient.Certificate
-	keyFile := conf.HttpsClient.Key
 
 	certBytes, err := ioutil.ReadFile(certFile)
 	if err != nil {
@@ -388,12 +418,7 @@ func loadClientTrust(ctx *openssl.Ctx, conf *Config) (*openssl.Ctx, error) {
 		}
 	}
 
-	keyBytes, err := ioutil.ReadFile(keyFile)
-	if err != nil {
-		return ctx, errors.Wrap(err, "Private key file from the HttpsClient configuration not found")
-	}
-
-	key, err := openssl.LoadPrivateKeyFromPEM(keyBytes)
+	key, err := loadPrivateKey(conf.HttpsClient.Key, conf.HttpsClient.SSLEngine)
 	if err != nil {
 		return ctx, err
 	}
@@ -485,6 +510,7 @@ func newHttpsClient(conf Config) (*http.Client, error) {
 type HttpsClient struct {
 	Certificate string
 	Key         string
+	SSLEngine   string
 }
 
 func (h *HttpsClient) Validate() {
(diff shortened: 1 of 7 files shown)


client/client.go
```diff
@@ -348,14 +349,38 @@ func loadServerTrust(ctx *openssl.Ctx, conf *Config) (*openssl.Ctx, error) {
 	return ctx, err
 }
 
+func loadPrivateKey(conf *Config) (openssl.PrivateKey, error) {
+	var key openssl.PrivateKey
+	keyFile := conf.HttpsClient.Key
+
+	if strings.HasPrefix(keyFile, pkcs11URIPrefix) {
+		e, err := openssl.EngineById(conf.HttpsClient.SSLEngine)
```
lluiscampos, line 357: Check for `err != nil` here

lluiscampos: Also, I would use `engine` instead of `e`.

client/client.go
```diff
@@ -485,6 +505,7 @@ func newHttpsClient(conf Config) (*http.Client, error) {
 type HttpsClient struct {
 	Certificate string
 	Key         string
+	SSLEngine   string
```
lluiscampos, line 508: SSLEngine is uninitialized when calling `openssl.EngineById`, will openSSL handle this as "the default"? Or what is the logic here?

lluiscampos (changes requested):

client/client.go
```diff
@@ -348,14 +349,38 @@ func loadServerTrust(ctx *openssl.Ctx, conf *Config) (*openssl.Ctx, error) {
 	return ctx, err
 }
 
+func loadPrivateKey(conf *Config) (openssl.PrivateKey, error) {
+	var key openssl.PrivateKey
```
oleorhagen, line 353: Suggestion:
Might as well use the return initializer:
`(key openssl.PrivateKey, err error)`

client/client.go
```diff
@@ -348,14 +349,38 @@ func loadServerTrust(ctx *openssl.Ctx, conf *Config) (*openssl.Ctx, error) {
...
+	keyFile := conf.HttpsClient.Key
+
+	if strings.HasPrefix(keyFile, pkcs11URIPrefix) {
+		e, err := openssl.EngineById(conf.HttpsClient.SSLEngine)
+		key, err = openssl.EngineLoadPrivateKey(e, keyFile)
+		if err != nil {
+			return nil, err
```
oleorhagen, line 360: Pray:
Can I get some more error information as a user here?

client/client.go
```diff
@@ -348,14 +349,38 @@ func loadServerTrust(ctx *openssl.Ctx, conf *Config) (*openssl.Ctx, error) {
 	return ctx, err
 }
 
+func loadPrivateKey(conf *Config) (openssl.PrivateKey, error) {
```
oleorhagen, line 352: Thought:
Should the API surface be narrowed to only the parameters used or not? :thinking: 
Usually I like it that way, but this time, I'm not really sure it's necessary. Hmm

client/client.go
merlin-northern: if the SSLEngine is empty it will just fail. The logic here is: that if you configure `Key` to be a PKCS#11 URI you have to provide some engine. I can check that in `Validate` and output error if engine is not provided. is that ok?

client/client.go
```diff
@@ -41,6 +41,7 @@ const (
 	errMissingCerts = "No trusted certificates. The client will continue running, but will " +
 		"not be able to communicate with the server. Either specify ServerCertificate in " +
 		"mender.conf, or make sure that CA certificates are installed on the system"
+	pkcs11URIPrefix = "pkcs11:"
```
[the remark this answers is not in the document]
merlin-northern: I thought it is nice. I can remove, but if Lluis agrees (see [above](https://github.com/mendersoftware/mender/pull/598#discussion_r479075246)) it will be used in the `Validate` as well.

client/client.go
merlin-northern: you mean passing of conf? exactly what I thought about when I was writing this. maybe since the call is named "loadPrivateKey" it should get privatekey and sslengine.

oleorhagen: That was my initial thought, when looking at this at least

client/client.go
lluiscampos: Yes, I think we should check it for a nicer error to the user.

client/client.go
merlin-northern: discussed privately, decided to keep it in.

merlin-northern:
@lluiscampos @oleorhagen I think I addressed every comment.

merlin-northern:
@lluiscampos @oleorhagen I had to add [c10edc678f592b1bb16ba261bf63137d5dbdf3b2](https://github.com/mendersoftware/mender/pull/598/commits/c10edc678f592b1bb16ba261bf63137d5dbdf3b2) after final manual tests.

client/client.go
```diff
@@ -497,6 +523,9 @@ func (h *HttpsClient) Validate() {
 		}
 		if h.Key == "" {
 			log.Error("The 'Certificate' field is set in the mTLS configuration, but no 'Key' is given. Both need to be present in order for mTLS to function")
+		} else if strings.HasPrefix(h.Key, pkcs11URIPrefix) && len(h.SSLEngine) < 1 {
```
oleorhagen, line 527:
```suggestion
		} else if strings.HasPrefix(h.Key, pkcs11URIPrefix) && len(h.SSLEngine) == 0 {
```

oleorhagen: I guess this could also be `h.SSLEngine == ""` ?

client/client.go
```diff
@@ -497,6 +523,9 @@ func (h *HttpsClient) Validate() {
...
+			log.Errorf("The 'Key' field is set to be loaded from %s, but no 'SSLEngine' is given. Both need to be present in order for loading of the key to function",
```
oleorhagen, line 528: Good that we warn on this :+1:

client/client.go
merlin-northern: sure. changed to the suggestion. (I usually use len()<1 somehow I feel safer ;-))

store/keystore.go
```diff
@@ -47,19 +50,35 @@ func (k *Keystore) GetKeyName() string {
 	return k.keyName
 }
 
-func NewKeystore(store Store, name string, static bool) *Keystore {
+func NewKeystore(store Store, name string, sslEngine string, static bool) *Keystore {
```
lluiscampos, line 53: There are no unit tests to test this function with `sslEngine != ""`, as it will require a setup similar to the one in `mendersoftware/openssl` repo.

As i understand it, it is covered in the unit tests on `mendersoftware/openssl` and in the soon-to-be integration test with the softhsm. So I am okey with this.

lluiscampos (approved):
I had a last overview look, it looks good to me.

store/keystore.go
merlin-northern: yes!

oleorhagen (approved):

merlin-northern:
tha pipeline passed: https://gitlab.com/Northern.tech/Mender/mender-qa/-/pipelines/184097095 and after that the changes were cosmetic. merging.
thanks for reviews @lluiscampos and @oleorhagen !

PR #600, opened 2020-09-08
Add "needs-reboot" command to mender client.

hacpa:
This command allows querying an artifact file to know if
installing this artifact (based on its type) will require a reboot.
To get a clean "yes" or "no" output on the command line,
you may set the log level to "fatal" or "panic" or redirect stderr.

Changelog: Title
Signed-off-by: Pascal Hache <hacpa@touchtunes.com>

### Description

Thank you!

### DIFF
--- a/app/standalone.go
+++ b/app/standalone.go
@@ -332,6 +332,44 @@ func DoStandaloneRollback(device *dev.DeviceManager, stateExec statescript.Execu
 	return firstErr
 }
 
+func DoStandaloneNeedsReboot(device *dev.DeviceManager, updateURI string) error {
+	var image io.ReadCloser
+	var imageSize int64
+	var err error
+
+	log.Infof("Opening local image file: [%s]", updateURI)
+	image, imageSize, err = installer.FetchUpdateFromFile(updateURI)
+	log.Debugf("Fetching update from file results: [%v], %d, %v", image, imageSize, err)
+
+	if image == nil || err != nil {
+		return errors.Wrapf(err, "Error while querying Artifact from command line")
+	}
+
+	_, installers, err := installer.ReadHeaders(image, "", nil,
+		device.StateScriptPath, &device.InstallerFactories)
+	if err != nil {
+		return errors.Wrapf(err, "Reading headers failed")
+	}
+
+	rebootNeeded, err := determineRebootNeeded(installers)
+	if err != nil {
+		return err
+	}
+
+	err = doStandaloneCleanup(device, &standaloneData{installers: installers}, nil)
+	if err != nil {
+		return err
+	}
+
+	if rebootNeeded == true {
+		fmt.Println("yes")
+	} else {
+		fmt.Println("no")
+	}
+
+	return nil
+}
+
 func doStandaloneRollbackState(standaloneData *standaloneData, stateExec statescript.Executor) error {
 	fmt.Println("Rolling back Artifact...")
 
--- a/cli/cli.go
+++ b/cli/cli.go
@@ -24,13 +24,14 @@ import (
 	"runtime"
 	"strings"
 
+	"log/syslog"
+
 	"github.com/mendersoftware/mender/app"
 	"github.com/mendersoftware/mender/conf"
 	"github.com/mendersoftware/mender/installer"
 	mender_syslog "github.com/mendersoftware/mender/log/syslog"
 	"github.com/mendersoftware/mender/system"
 	log "github.com/sirupsen/logrus"
-	"log/syslog"
 
 	"github.com/pkg/errors"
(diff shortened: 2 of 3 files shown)


hacpa:
I tried to make this change as small as possible with as little code duplication as possible. It ignores the device type and signature key, allowing to get the data from an unsigned artifact as well as one that is not installable on the current device. This does cause 2 undesirable prints (short version here):
- WARNING: verification key is missing.
- ERROR: unknown device_type.

These messages do not appear if, as proposed in the commit message, mender is invoked with a low log level or with stderr redirected.

If, however, one of both of these must be fixed, there are 2 ways that I can see:
- Provide the key and/or device type to the "installer.ReadHeaders" function. Con: it is no longer possible to query an unsigned/wrong device type artifact.
- Reimplement part of the "installer.ReadHeaders" function as well as the internal "registerHandlers" function from the same file, but do not check the device type and/or signature. Con: code duplication.

I welcome any feedback and/or improvements.

hacpa:
Noticed that I forgot to cleanup after opening the artifact, which would leave stuff in /data/mender/ if using an update module... added cleanup, amended commit and force-pushed!

kacf (changes requested):
Thanks, @hacpa! I know I said on Mender Hub that `needs-reboot` would solve the issue, but having looked at this some more, I'm not sure if this is the best way of doing things. Take a look at the paragraph right above [this one](https://github.com/mendersoftware/mender/blob/master/Documentation/update-modules-v3-file-api.md#artifactverifyreboot-state) (the last paragraph of the previous header). This is an important property in the Update Module specification: `NeedsArtifactReboot` should not run before `ArtifactInstall` has run, as all the information may not be available then.

My suggestion is to do this in a different way: We need to query for `NeedsArtifactReboot` during the installation. However, for an Artifact without rollback support, the Artifact is committed immediately, so it is difficult to provide it as a separate command; by the time one can execute it, the installation is already over. I instead propose to introduce the `--reboot-exit-code` (or a better name if you can think of it), which returns a special error code when calling `mender install` and a reboot is required. This solves the problem that the query has to be done midway in the installation, and should also be quite simple to add, since the code for doing the query already exists. And it maintains backwards compatibility.

I propose error code 4, since 0, 1 and 2 [are used already](https://github.com/mendersoftware/mender/blob/fecb925fab76b1bad4b9ad654f1485136dde6f82/main.go#L29-L35). I think 4 is better than 3, because I suspect we might need to OR together error codes at some point if we need to report more than one.

Long explanation, but does that make sense?

kacf:
Pinging @drewmoseley, @mirzak and @oleorhagen as well, in case they have input on the functionality.

oleorhagen:
First, just a thank you to @hacpa for a nice and clean PR! Always appreciated!

> My suggestion is to do this in a different way: We need to query for `NeedsArtifactReboot` during the installation. However, for an Artifact without rollback support, the Artifact is committed immediately, so it is difficult to provide it as a separate command; by the time one can execute it, the installation is already over. I instead propose to introduce the `--reboot-exit-code` (or a better name if you can think of it), which returns a special error code when calling `mender install` and a reboot is required. This solves the problem that the query has to be done midway in the installation, and should also be quite simple to add, since the code for doing the query already exists. And it maintains backwards compatibility.
> 

And I have to vote with @kacf on this one.

drewmoseley:
Looks reasonable to me. I don't have anything more to add.

hacpa:
@kacf To be honest, I had already spotted that mention in the update module documentation, and thought this change would deviate from that specification. I tested with update modules and rootfs updates and they all seemed to work properly, including one update module that is basically a reimplementation of the rootfs updater with some more features, including a different response to the "NeedsArtifactReboot" query if the installed rootfs is the same as the one we are attempting to install (it answers "No" in that case, or "Automatic" otherwise). Everything worked fine in all my tests, so I didn't make a fuss of it. I see now that I should have.

I do have one small issue with utilizing a return code: my automated update script runs with set -e. So, if any command (including "mender -install") returns a non-zero return code, that is perceived as an error and the script exits. While there might be ways to work around the script exit issue, this is the reason why I would prefer a standalone command that preserves the 0 return code.

Now, I understand that design-wise it makes a bit more sense to use a return code for this purpose. Tell you what: I'll create a new branch and experiment with the return code approach. I will create a separate PR for this version if I am successful.

hacpa:
@kacf And here it is: https://github.com/mendersoftware/mender/pull/601

I tested it and it works well. Please have a look at the change!

I was also able to adapt my script to catch the return value instead of failing, so that part that was problematic for me has been resolved.

kacf:
Thanks @hacpa! I'm closing this in favor of #601 then.

PR #602, opened 2020-09-10
[README] Add libssl-dev as a dependency to Mender

lluiscampos:


lluiscampos:
@hacpa Can you verify that installing this package you can compile in `master`? Thanks.

hacpa:
> 
> 
> @hacpa Can you verify that installing this package you can compile in `master`? Thanks.

Yes! The libssl-dev package contains both missing .pc files. I was then able to compile and run mender.

PR #605, opened 2020-09-15
Fix rootfs-image-v2 commit in standalone mode when upgrade fails

hacpa:
This changes the behavior of the ArtifactCommit function. Now, the variable
"upgrade_available" is verified, just like the built-in "rootfs-image",
and the commit fails if no upgrade is in progress. So, if an upgrade fails and
the user tries to commit after returning to the old OS, the commit will fail
and an appropriate message will be printed.

Changelog: Title

Signed-off-by: Pascal Hache <hacpa@touchtunes.com>

### Description

Thank you!

### DIFF
--- a/tests/rootfs-image-v2
+++ b/tests/rootfs-image-v2
@@ -63,7 +63,14 @@ EOF
         ;;
 
     ArtifactCommit)
-        fw_setenv upgrade_available 0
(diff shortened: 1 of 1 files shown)


kacf (approved):
Thanks @hacpa!

Just ignore the codecov tests, we are in the middle of reconfiguring that.

PR #606, opened 2020-09-15
Gracefully stop on SIGTERM.

merlin-northern:
ChangeLog:none

### DIFF
--- a/cli/commands.go
+++ b/cli/commands.go
@@ -228,6 +228,7 @@ func runDaemon(d *app.MenderDaemon) error {
 		c := make(chan os.Signal, 2)
 		signal.Notify(c, syscall.SIGUSR1) // SIGUSR1 forces an update check.
 		signal.Notify(c, syscall.SIGUSR2) // SIGUSR2 forces an inventory update.
+		signal.Notify(c, syscall.SIGTERM) // SIGTERM marks the exit.
 		defer signal.Stop(c)
 
 		for {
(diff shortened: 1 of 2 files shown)


oleorhagen (approved):
Makes sense to me!

Could we just add a log message here also?

oleorhagen:
Maybe a small changelog be nice also?

cli/commands.go
```diff
@@ -238,6 +239,8 @@ func runDaemon(d *app.MenderDaemon) error {
 			} else if s == syscall.SIGUSR2 {
 				log.Debug("SIGUSR2 signal received.")
 				d.ForceToState <- app.States.InventoryUpdate
+			} else if s == syscall.SIGTERM {
+				d.StopDaemon()
```
kacf, line 243: Have you tested this? It doesn't look like this breaks out of any sleep, it just sets a flag.

oleorhagen: It breaks out of sleep on the following line, no?

kacf: Oh, lol! You're right. The codecov box hid it from me!

merlin-northern:
> Maybe a small changelog be nice also?

added!

merlin-northern:
> Makes sense to me!
> 
> Could we just add a log message here also?

added!

merlin-northern:
merging as the pipeline [passed](https://gitlab.com/Northern.tech/Mender/mender-qa/-/pipelines/190240469)

PR #608, opened 2020-09-21
Call PATCH inventory/device/attributes only on inventory changes

lluiscampos:
Make the client save a checksum of the current inventory attributes in
memory, and only call the API endpoint if the newly collected inventory
attributes differ from the already sent ones. It will always send the
inventory the first time after a boot.

Changelog: Title

### DIFF
--- a/app/mender.go
+++ b/app/mender.go
@@ -14,10 +14,14 @@
 package app
 
 import (
+	"crypto/sha256"
+	"encoding/json"
 	"fmt"
 	"io"
 	"os"
 	"path"
+	"sort"
+	"strings"
 	"time"
 
 	log "github.com/sirupsen/logrus"
@@ -666,7 +670,7 @@ func (m *Mender) InventoryRefresh() error {
 		return errors.Wrap(errNoArtifactName, errstr)
 	}
 
-	idata, err := idg.Get()
+	iData, err := idg.Get()
 	if err != nil {
 		// at least report device type
 		log.Errorf("Failed to obtain inventory data: %s", err.Error())
@@ -682,20 +686,36 @@ func (m *Mender) InventoryRefresh() error {
 		{Name: "mender_client_version", Value: conf.VersionString()},
 	}
 
-	if idata == nil {
-		idata = make(client.InventoryData, 0, len(reqAttr))
+	if iData == nil {
+		iData = make(client.InventoryData, 0, len(reqAttr))
 	}
-	_ = idata.ReplaceAttributes(reqAttr)
+	_ = iData.ReplaceAttributes(reqAttr)
 
-	if idata == nil {
-		log.Infof("No inventory data to submit")
+	if iData == nil {
+		log.Info("No inventory data to submit")
 		return nil
 	}
 
-	err = ic.Submit(m.api.Request(m.authToken, nextServerIterator(m), reauthorize(m)), m.Config.Servers[0].ServerURL, idata)
+	// sort slice to have a reproducible hash
+	sort.Slice(iData, func(i, j int) bool { return strings.Compare(iData[i].Name, iData[j].Name) < 0 })
+
+	iDataBuf, err := json.Marshal(iData)
 	if err != nil {
-		return errors.Wrapf(err, "failed to submit inventory data")
+		return err
 	}
+	iDataSha := sha256.Sum256(iDataBuf)
+	if err == nil && iDataSha == m.LastInventoryHash {
+		log.Info("The inventory data has not changed since the last time it was collected. Skipping server inventory update call")
+		return nil
+	}
+
+	err = ic.Submit(m.api.Request(m.authToken, nextServerIterator(m), reauthorize(m)), m.Config.Servers[0].ServerURL, iData)
+	if err != nil {
+		return errors.Wrap(err, "failed to submit inventory data")
+	}
+
+	log.Debugf("New inventory data, updating database with hash: %064x, inventory: %s", iDataSha, iData)
+	copy(m.LastInventoryHash[:], iDataSha[:])
 
 	return nil
 }
--- a/app/mender_test.go
+++ b/app/mender_test.go
@@ -737,6 +737,7 @@ echo foo=bar`),
 	srv.Auth.Token = []byte("tokendata")
 	err = mender.InventoryRefresh()
 	assert.Nil(t, err)
+	assert.True(t, srv.Inventory.Called)
 	exp = []client.InventoryAttribute{
 		{Name: "device_type", Value: "foo-bar"},
 		{Name: "artifact_name", Value: "fake-id"},
@@ -747,6 +748,27 @@ echo foo=bar`),
 		assert.Contains(t, srv.Inventory.Attrs, a)
 	}
 
+	// 2a. call with same inventory shall skip the actual send to the server
+	srv.Reset()
+	srv.Auth.Verify = true
+	srv.Auth.Token = []byte("tokendata")
+	err = mender.InventoryRefresh()
+	assert.Nil(t, err)
+	assert.False(t, srv.Inventory.Called)
+
+	// 2b. new call changing inventory shall be sent to the server
+	srv.Reset()
+	srv.Auth.Verify = true
+	srv.Auth.Token = []byte("tokendata")
+	err = ioutil.WriteFile(path.Join(invpath, "mender-inventory-bar"),
+		[]byte(`#!/bin/sh
+echo bar=foo`),
+		os.FileMode(syscall.S_IRWXU))
+	assert.NoError(t, err)
+	err = mender.InventoryRefresh()
+	assert.Nil(t, err)
+	assert.True(t, srv.Inventory.Called)
+
 	// no artifact name should error
 	ioutil.WriteFile(artifactInfo, []byte(""), 0600)
 	err = mender.InventoryRefresh()
@@ -756,7 +778,7 @@ echo foo=bar`),
 	// 3. pretend client is no longer authorized
 	srv.Auth.Token = []byte("footoken")
 	err = mender.InventoryRefresh()
-	assert.NotNil(t, err)
+	assert.Error(t, err)
 }
 
 func MakeFakeUpdate(data string) (string, error) {
--- a/device/device.go
+++ b/device/device.go
@@ -35,6 +35,7 @@ type DeviceManager struct {
 	DeviceTypeFile      string
 	Installers          []installer.PayloadUpdatePerformer
 	InstallerFactories  installer.AllModules
+	LastInventoryHash   [32]byte
 	StateScriptExecutor statescript.Executor
 	StateScriptPath     string
 	Store               store.Store


app/mender.go
```diff
@@ -682,19 +685,39 @@ func (m *Mender) InventoryRefresh() error {
...
-		idata = make(client.InventoryData, 0, len(reqAttr))
+	if iData == nil {
+		iData = make(client.InventoryData, 0, len(reqAttr))
 	}
-	_ = idata.ReplaceAttributes(reqAttr)
+	_ = iData.ReplaceAttributes(reqAttr)
 
-	if idata == nil {
-		log.Infof("No inventory data to submit")
+	if iData == nil {
```
oleorhagen, line 693: Nitpick:
I usually like this in a separate commit, but no hard requirement

app/mender_test.go
```diff
@@ -756,7 +778,7 @@ echo foo=bar`),
 	// 3. pretend client is no longer authorized
 	srv.Auth.Token = []byte("footoken")
 	err = mender.InventoryRefresh()
-	assert.NotNil(t, err)
```
oleorhagen, line 759: Question:
Was this wrong previously?

client/inventory.go
```diff
@@ -41,3 +41,11 @@ func (id *InventoryData) ReplaceAttributes(attr []InventoryAttribute) error {
 	}
 	return nil
 }
+
+// ByName implements sort.Interface based on the Name field, so that an
```
oleorhagen, line 45:
```suggestion
// ByName implements sort.Interface based on the Name field, so that a
```

client/inventory.go
```diff
@@ -41,3 +41,11 @@ func (id *InventoryData) ReplaceAttributes(attr []InventoryAttribute) error {
...
+// slice of InventoryAttribute can be sorted and compared.
+type ByName []InventoryAttribute
```
oleorhagen, line 47: 👍

app/mender.go
```diff
@@ -682,19 +685,39 @@ func (m *Mender) InventoryRefresh() error {
...
+	sort.Sort(client.ByName(iData))
+
+	iDataBuf, err := json.Marshal(iData)
+	if err != nil {
+		return err
+	}
+	iDataSha := sha256.Sum256(iDataBuf)
+	lastInvSha, err := m.GetCurrentInventoryHash()
+	if err == nil && iDataSha == lastInvSha {
+		log.Info("Inventory data has not changed, skip submit")
```
oleorhagen, line 708: 🤔 
Logging is always hard. I don't particularly like this message, but I'm also struggling with finding something "better":
`The inventory data has not changed since the last time it was collected. Nothing to submit. Skipping server inventory update call`. 
Or something along these lines?

app/mender.go
```diff
@@ -682,19 +685,39 @@ func (m *Mender) InventoryRefresh() error {
...
+		return nil
+	}
+
+	log.Debugf("New inventory data, updating database with hash: %v , inventory: %v", iDataSha, iData)
```
oleorhagen, line 712: Question:
Cant the hash be formatted with:
`%x` ? And the inventory be stringified?

oleorhagen (commented):
I only had time to do a superficial lookover now. But I think it looks good! No need to panic on spending time away from the client ;)

Comments are mostly questions, I think this is fine. Feel free to cherry pick at will. Will do a closer look tomorrow morning :)

app/mender.go
lluiscampos: Generally speaking I agree. But this time I am "using" idata/iData in my commit, so in that sense I think is justified to rename it :)

app/mender.go
alfrunes: ... or even `%064x`?

app/mender_test.go
lluiscampos: No, is just that checking `err` is not nil covers the same as checking that `err` is an `Error` (while IMO the later is more meaningful)

app/mender.go
lluiscampos: Hehe, `%v` is the lazy way...

I'll rework it a bit (at the end of the day, the main goal was to practice golang!)

kacf: Yes, `%v` is generally discouraged because it silences all warnings about type mismatch, even useful ones. We have plenty of examples where we have made mistakes here.

app/mender.go
```diff
@@ -682,19 +685,39 @@ func (m *Mender) InventoryRefresh() error {
...
-	if idata == nil {
-		log.Infof("No inventory data to submit")
+	if iData == nil {
+		log.Info("No inventory data to submit")
 		return nil
 	}
 
-	err = ic.Submit(m.api.Request(m.authToken, nextServerIterator(m), reauthorize(m)), m.Config.Servers[0].ServerURL, idata)
+	// sort slice to have a reproducible hash
+	sort.Sort(client.ByName(iData))
```
alfrunes, line 699:
```suggestion
	sort.Slice(func(i, j) bool { return strings.Compare(iData[i], iData[j]) < 0 })
```
With this you don't have to introduce the extra `ByName` type.

kacf (approved):
Apart from the things already mentioned, looks good to me!

app/mender_test.go
alfrunes: The assertion is identical, it just makes it more obvious if it fails due to an error.

alfrunes (approved):
Well done!
I just have a couple of semantic suggestions.

tranchitella:
@lluiscampos @mzedel heads-up: this PR is going to change the semantic of the "updated_ts" inventory attribute, which contains the current date/time of the latest upsert. In the UI, this field is reported as "Last check-in", and some of our customers are using it as a sort-of monitoring tool. When this PR is merged, or the equivalent in the backend (https://github.com/mendersoftware/inventory/pull/211) is merged, we won't update anymore this attribute and the "Last check-in" in the UI will have a different meaning (latest inventory update).

mzedel:
Good point @tranchitella! This will then also break the "active devices" widget or the "liveness" check for devices - I'm not sure if that's acceptable... as long as we don't have a persistent connection between device and backend or a different way of tracking the check-in, I'm quite sure this would cause trouble...

lluiscampos:
@tranchitella @mzedel Thanks for bringing this up. Could the UI get this information from the API call to deployments `/next`?

app/mender.go
lluiscampos: Thanks, I'll go with your suggestion.

app/mender.go
lluiscampos: You suggestion was a bit wordy, I cut the middle part:
`The inventory data has not changed since the last time it was collected. Skipping server inventory update call`

app/mender.go
lluiscampos: Fixed now.

lluiscampos:
I reworked now following @oleorhagen and  @alfrunes suggestion. I won't merge this until 1) I demo it and 2) we have a clear follow-up with the frontend team in a JIRA ticket.

tranchitella:
> @tranchitella @mzedel Thanks for bringing this up. Could the UI get this information from the API call to deployments `/next`?

No, because deployments/next doesn't write anywhere the latest check-in.
By the way, writing something to mongodb each time we get an API call is exactly what we want to avoid...

mzedel:
if we imagine a device that has a stable inventory and no new updates, removing the timestamp functionality would lead to this being seen as a potentially offline device - besides the widget, this was supposed to be shown in the device list at some point if I remember correctly...
We might at least be able to ease the writing frequency, as the resolution of the check-ins does not necessarily have to be high... so it might be possible to go the same way etcd goes e.g. writing the checkins in batches once enough checkins have accumulated or on schedules and accepting lost writes in a single node - either through redundancy (as long as you get a quorum) or as long as the update interval is small enough to not care about some evicted nodes...

app/mender.go
oleorhagen: Nice! A little back and forth usually does the trick with these

app/mender.go
oleorhagen: Should it not be `%32x` ? Or am I being schtoopid here?

app/mender.go
```diff
@@ -682,19 +686,39 @@ func (m *Mender) InventoryRefresh() error {
...
+	}
+	iDataSha := sha256.Sum256(iDataBuf)
+	lastInvSha, err := m.GetCurrentInventoryHash()
+	if err == nil && iDataSha == lastInvSha {
+		log.Info("The inventory data has not changed since the last time it was collected. Skipping server inventory update call")
+		return nil
+	}
+
+	log.Debugf("New inventory data, updating database with hash: %064x, inventory: %s", iDataSha, iData)
+	err = m.Store.WriteAll(datastore.InventoryHashKey, iDataSha[:])
```
lluiscampos, line 714: This should be done after `ic.Submit`, as pointed today in Sprint Review by @pasinskim

mirzak:
> Make the client save a checksum of the current inventory attributes its local store

A couple of questions regarding "local store".

Is "local store" the db that resides in the data partition?

Also does this really need to be stored in the db (if that is the case)? I would imagine that volatile memory should be enough, and it is ok for the device to send inventory once on each reboot/restart? Using volatile memory is preferred to avoid disk writes, which could increase significantly for the Mender client if the inventory values changes.

Could using "volatile memory" also help with the "last check-in" problem?

oleorhagen:
> Could using "volatile memory" also help with the "last check-in" problem?

This sounds to me like a very good idea.

lluiscampos:
> > Make the client save a checksum of the current inventory attributes its local store
> 
> A couple of questions regarding "local store".
> 
> Is "local store" the db that resides in the data partition?

Yes

> Also does this really need to be stored in the db (if that is the case)? I would imagine that volatile memory should be enough, and it is ok for the device to send inventory once on each reboot/restart? Using volatile memory is preferred to avoid disk writes, which could increase significantly for the Mender client if the inventory values changes.

This is something I considered during development. The pros and cons are exactly what you pointed out. As my goal was to optimize on API calls, I decided to save on disk. But re-thinking now after your comment I think it makes sense to just save it in memory (and hence forcing an inventory PATCH after reboot).

> 
> Could using "volatile memory" also help with the "last check-in" problem?

Well, in this case "last check-in" will correspond really to "last boot time", helps a bit but still won't work as today.

Thanks for your input @mirzak, I'll rework :)

app/mender.go
lluiscampos: Fixed now

lluiscampos: See https://github.com/mendersoftware/mender/compare/822e0fc90fa98f45e37eecb32dee62d7eaafca3b..2a77a52b5ebd787a73788d5b68f55e1663283669

lluiscampos:
> Thanks for your input @mirzak, I'll rework :)

Fixed now (see rework [here](https://github.com/mendersoftware/mender/compare/2a77a52b5ebd787a73788d5b68f55e1663283669..daf691de4e3e0227445749b4d4ea765f1c4b491e)) and also edited the commit message a bit.

tranchitella:
@lluiscampos after discussion with @estenberg, his feedback is to NOT merge this PR and wait for an alternative implementation of the latest check-in in the UI: too many users are relying on this (non-)feature.

tranchitella (changes requested):
As long as Mender does not offer another mean to provide the "latest check-in" feature, we cannot merge this PR.

lluiscampos:
Closing for now, see [MEN-3981](https://tracker.mender.io/browse/MEN-3981).

PR #609, opened 2020-09-23
Use new command syntax on the service ExecStart

lluiscampos:
At the moment both syntax are supported, but we will remove the -command
one on Mender 3.0, so better to start spreading the word ;)

Changelog: None


alfrunes (approved):
:+1:

kacf (approved):

PR #611, opened 2020-09-24
Revert "MEN-3684: Switch to fw_setenv script syntax which libubootenv supports."

oleorhagen:
Reverts mendersoftware/mender#554


kacf:
I guess if the patch against libubootenv was rejected, we need to use the backup approach of trying both methods?

oleorhagen:
Closed in favour of #616

PR #612, opened 2020-09-27
MEN-3074: implement "show-provides" command on client

tranchitella:
Changelog: title


kacf (approved):
Nice

PR #616, opened 2020-09-29
MEN-3970: Add backwards compatibility to write to old U-Boot envs

oleorhagen:
This change is needed to support older U-Boot versions, where the firmware tools
only support the 'key value', syntax, whereas the newer tools, and libubootenv
especially only support the '=' separator, giving 'key=value' as the only option.

This is solved through first writing to the environment using the legacy whitespace separator syntax, and then reading the variable back, to see if the operation applied a change to the environment. If it did, assume the separator to use is the legacy space separator.

Changelog: Title

### DIFF
--- a/installer/bootenv.go
+++ b/installer/bootenv.go
@@ -25,6 +25,13 @@ import (
 	log "github.com/sirupsen/logrus"
 )
 
+const (
+	menderUBootSeparatorProbe = "mender_uboot_separator"
+
+	uBootEnvStandardSeparator = "="
+	uBootEnvLegacySeparator   = " "
+)
+
 type UBootEnv struct {
 	system.Commander
 }
@@ -91,10 +98,55 @@ func (e *UBootEnv) ReadEnv(names ...string) (BootVars, error) {
 	return vars, err
 }
 
+// ProbeSeparator tries writing 'mender_uboot_separator' to the U-Boot
+// environment using whitespace as a separator. If this write is succesful it
+// means that we are on a legacy U-Boot implementation, which supports the
+// 'key<whitespace>value' syntax. If not, then use the new '=' separator.
+func (e *UBootEnv) probeSeparator() (string, error) {
+	log.Info("Probing the Bootloader environment for which separator to use")
+	// Try writing using the old separator syntax
+	err := e.writeEnvImpl(BootVars{
+		menderUBootSeparatorProbe: "1",
+	}, uBootEnvLegacySeparator)
+	if err != nil {
+		return uBootEnvStandardSeparator, err
+	}
+	v, err := e.ReadEnv(menderUBootSeparatorProbe)
+	if err != nil {
+		return uBootEnvStandardSeparator, err
+	}
+	if v[menderUBootSeparatorProbe] == "1" {
+		// Clean variable from the environment
+		err = e.writeEnvImpl(BootVars{
+			menderUBootSeparatorProbe: "",
+		}, uBootEnvLegacySeparator)
+		return uBootEnvLegacySeparator, err
+	}
+	return uBootEnvStandardSeparator, nil
+}
+
+// WriteEnv attempts to write the given 'BootVars' to the bootloader
+// environment.
 func (e *UBootEnv) WriteEnv(vars BootVars) error {
 	if err := e.checkEnvCanary(); err != nil {
 		return err
 	}
+	// Probe for the separator used by U-Boot. Newer versions support '=',
+	// and libubootenv only supports '=' as the separator, older versions
+	// only support ' '
+	separator, err := e.probeSeparator()
+	if err != nil {
+		log.Errorf("Failed to probe the U-Boot environment for which separator to use. Got error: %s", err.Error())
+		return err
+	}
+	log.Debugf("Using (%s) as the bootloader environment separator", separator)
+	return e.writeEnvImpl(vars, separator)
+}
+
+func (e *UBootEnv) writeEnvImpl(vars BootVars, separator string) error {
+
+	log.Infof("Writing %v to the U-Boot environment, using separator: %s",
+		vars, separator)
 
 	// Make environment update atomic by using fw_setenv "-s" option.
 	setEnvCmd := e.Command("fw_setenv", "-s", "-")
@@ -113,7 +165,7 @@ func (e *UBootEnv) WriteEnv(vars BootVars) error {
 		return err
 	}
 	for k, v := range vars {
-		_, err = fmt.Fprintf(pipe, "%s=%s\n", k, v)
+		_, err = fmt.Fprintf(pipe, "%s%s%s\n", k, separator, v)
 		if err != nil {
 			log.Error("Error while setting U-Boot variable: ", err)
 			pipe.Close()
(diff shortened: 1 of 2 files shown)


installer/bootenv.go
```diff
@@ -25,8 +25,14 @@ import (
...
 )
 
+var (
+	unexpectedUBootEnvironmentError = errors.New("The U-Boot environment on the device does not match the expected environment")
+	uBootWriteEnvironmentError      = errors.New("Failed to write the U-Boot environment")
+)
+
 type UBootEnv struct {
 	system.Commander
+	SkipValidate bool
```
lluiscampos, line 35: I would add a comment here indicating that it is meant only for unit tests.

installer/bootenv.go
```diff
@@ -129,6 +154,30 @@ func (e *UBootEnv) WriteEnv(vars BootVars) error {
...
 		return err
 	}
+	return e.validate(vars)
+}
+
+// validate validates the given UBoot environment variables in 'vars', by
+// returning the UBoot environment and comparing them with the expected
+// environment.
+func (e *UBootEnv) validate(expected BootVars) error {
+	if e.SkipValidate {
```
oleorhagen, line 164: I know this is not the prettiest solution, but it is certainly the least intrusive in terms of how much code that needs to change in order to support overriding this one method in the tests (composition over inheritance isn't always a treat).

mirzak:
> This is solved through first writing to the environment using the modern syntax,
and then reading it back, and verifiying that the new environment variables set
are correct. Then, if that fails, it falls back to the old syntax, with the
space separator, and reads back the environment once again to verify

This is not atomic right? There is a chance that the device is bricked if the first operation fails and you are interrupted before you are able to run the second? I guess this only applies to older U-Boot versions but still.

Would it be better to do an "probe" to see which method is supported, e.g writing `mender_dummy_1` & `mender_dummy_2`  variables using both methods and reading it back to see which worked. 

Maybe it is also worth to store the method used once we figure it out, to avoid doing these dummy writes on each iteration?

oleorhagen:
> > This is solved through first writing to the environment using the modern syntax,
> > and then reading it back, and verifiying that the new environment variables set
> > are correct. Then, if that fails, it falls back to the old syntax, with the
> > space separator, and reads back the environment once again to verify
> 
> This is not atomic right? There is a chance that the device is bricked if the first operation fails and you are interrupted before you are able to run the second? I guess this only applies to older U-Boot versions but still.
> 

You're thinking that the failed write wipes the environment clean, right?  I don't think it'll happen, as then the variable it cleans is 'key=value', and not 'key', right?

> Would it be better to do an "probe" to see which method is supported, e.g writing `mender_dummy_1` & `mender_dummy_2` variables using both methods and reading it back to see which worked.
> 

I like this! Implementing

> Maybe it is also worth to store the method used once we figure it out, to avoid doing these dummy writes on each iteration?

Sure. Good idea.

mirzak:
> You're thinking that the failed write wipes the environment clean, right? 

No. If you use `mender_boot_part=1` syntax on older U-Boot versions, the result will be that you remove the `mender_boot_part` variable (it will be empty/non-existent) = bricked :D

oleorhagen:
@kacf you can have a look now

installer/bootenv.go
```diff
@@ -91,7 +98,44 @@ func (e *UBootEnv) ReadEnv(names ...string) (BootVars, error) {
...
+// environment.
 func (e *UBootEnv) WriteEnv(vars BootVars) error {
+	// Probe for the separator used by U-Boot. Newer versions support '=',
+	// and libubootenv only supports '=' as the separator, older versions
+	// only support ' '
+	separator, err := e.ProbeSeparator()
+	if err != nil {
+		log.Errorf("Failed to probe the U-Boot environment for which separator to use. Got error: %v, Going to try writing the U-Boot environment using the separator: %q", err, separator)
+		log.Info("Trying to write the U-Boot environment despite error received when probing the environment for which environment separator to use")
+	}
```
kacf, line 134: I think we should instead return error here. It's quite unlikely that we will get here, but if we do, continuing is not a good idea, IMHO.

installer/bootenv.go
```diff
@@ -91,7 +98,44 @@ func (e *UBootEnv) ReadEnv(names ...string) (BootVars, error) {
...
+
+// WriteEnv attempts to write the given 'BootVars' to the bootloader
+// environment.
 func (e *UBootEnv) WriteEnv(vars BootVars) error {
+	// Probe for the separator used by U-Boot. Newer versions support '=',
+	// and libubootenv only supports '=' as the separator, older versions
+	// only support ' '
+	separator, err := e.ProbeSeparator()
+	if err != nil {
+		log.Errorf("Failed to probe the U-Boot environment for which separator to use. Got error: %v, Going to try writing the U-Boot environment using the separator: %q", err, separator)
```
kacf, line 132: In general I think we should use `%s` and `err.Error()`. We've had so many cases of `%v` printing the wrong thing because there is no type checking.

kacf (changes requested):
Overall good solution with regards to the code. But the commit message no longer matches what you do, it should be adjusted.

Also there are two conflicting changelog entries.

installer/bootenv.go
```diff
@@ -91,7 +98,44 @@ func (e *UBootEnv) ReadEnv(names ...string) (BootVars, error) {
...
+
+// WriteEnv attempts to write the given 'BootVars' to the bootloader
+// environment.
 func (e *UBootEnv) WriteEnv(vars BootVars) error {
+	// Probe for the separator used by U-Boot. Newer versions support '=',
+	// and libubootenv only supports '=' as the separator, older versions
+	// only support ' '
+	separator, err := e.ProbeSeparator()
+	if err != nil {
+		log.Errorf("Failed to probe the U-Boot environment for which separator to use. Got error: %s, Going to try writing the U-Boot environment using the separator: %q", err.Error(), separator)
```
kacf, line 132: Good, but you should drop the final part now, we are not going to write anything! :-)

kacf (changes requested):

installer/bootenv.go
oleorhagen: Lol, thx. sry. I'm going to let it run though, as I had some U-Boot acceptance test errors on `vexpress-qemu` :/
Having a dig locally also, ref:
https://gitlab.com/Northern.tech/Mender/mender-qa/-/pipelines/195878436/test_report

kacf (approved):

oleorhagen:
https://gitlab.com/Northern.tech/Mender/mender-qa/-/pipelines/200293795

oleorhagen:
@kacf One spurious failure in `mtls` tests, like we have been seeing lately. Good to go?

installer/bootenv.go
```diff
@@ -91,11 +98,56 @@ func (e *UBootEnv) ReadEnv(names ...string) (BootVars, error) {
...
+	if err != nil {
+		log.Errorf("Failed to probe the U-Boot environment for which separator to use. Got error: %s", err.Error())
+		return err
+	}
+	log.Debugf("Using (%s) as the bootloader environment separator", separator)
+	return e.writeEnvImpl(vars, separator)
+}
+
+func (e *UBootEnv) writeEnvImpl(vars BootVars, separator string) error {
 	if err := e.checkEnvCanary(); err != nil {
```
kacf, line 144: This check should be kept at the beginning of `WriteEnv`. Not only is it being carried out multiple times here, but it should be before any writing happens to protect the environment.

kacf (changes requested):

kacf (approved):

PR #622, opened 2020-10-03
Fixing misspell in log line and hiding the key URI.

merlin-northern:
ChangeLog:none

### DIFF
--- a/client/client.go
+++ b/client/client.go
@@ -364,7 +364,7 @@ func loadPrivateKey(keyFile string, engineId string) (key openssl.PrivateKey, er
 				engineId, err.Error())
 			return nil, err
 		}
-		log.Infof("laoded private key: '%v' from '%s'.", key, engineId)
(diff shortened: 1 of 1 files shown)


merlin-northern:
@lluiscampos @kacf this is related to https://github.com/mendersoftware/integration/blob/master/tests/tests/test_mtls.py#L283 it also removed the logging of PKCS#11 URI which may contain `pin-value`
(for that reason I am not stopping the integration test)

tranchitella (approved):

kacf (approved):

lluiscampos (approved):
Thanks :+1: 

I think we should cherry-pick to 2.4.x, in case we modify the test in the future at some point while still using client 2.4.x

merlin-northern:
> I think we should cherry-pick to 2.4.x, in case we modify the test in the future at some point while still using client 2.4.x

[right](https://github.com/mendersoftware/mender/pull/624)!

PR #625, opened 2020-10-08
MEN-4032: Add Glib's GIO dependency and demo integration

lluiscampos:
Added a new package `dbus` in the source code defining an interface and
a default implementation for it (using C bindings for libgio).

Implement wrappers for the two easiest calls and add unit tests for
them.

Changelog: Add Glib's GIO dependency for D-Bus interface. It can be
opt-out using `nodbus` at compile time.

### DIFF
--- a/README.md
+++ b/README.md
@@ -71,7 +71,7 @@ issue. We thank you in advance for your cooperation.
 
 * C compiler
 * [Go compiler](https://golang.org/dl/)
-* liblzma-dev and libssl-dev packages
+* liblzma-dev, libssl-dev and libglib2.0-dev packages
 
 #### LZMA support opt-out
 
@@ -81,6 +81,14 @@ dependency and substitute the `make` commands in the instructions below for:
 ```
 make TAGS=nolzma
 ```
+
+#### D-Bus support opt-out
+
+If no D-Bus support if desired, you can ignore the `libglib2.0-dev` package dependency and substitute
+the `make` commands in the instructions below for:
+
+```
+make TAGS=nodbus
 
 ### Steps
 
--- a/dbus/dbus.go
+++ b/dbus/dbus.go
@@ -1,0 +1,31 @@
+// Copyright 2020 Northern.tech AS
+//
+//    Licensed under the Apache License, Version 2.0 (the "License");
+//    you may not use this file except in compliance with the License.
+//    You may obtain a copy of the License at
+//
+//        http://www.apache.org/licenses/LICENSE-2.0
+//
+//    Unless required by applicable law or agreed to in writing, software
+//    distributed under the License is distributed on an "AS IS" BASIS,
+//    WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+//    See the License for the specific language governing permissions and
+//    limitations under the License.
+
+package dbus
+
+import "github.com/pkg/errors"
+
+var dbusAPI DBusAPI = nil
+
+type DBusAPI interface {
+	GenerateGUID() string
+	IsGUID(string) bool
+}
+
+func GetDBusAPI() (DBusAPI, error) {
+	if dbusAPI != nil {
+		return dbusAPI, nil
+	}
+	return nil, errors.New("no D-Bus interface available")
+}
--- a/dbus/dbus_libgio.go
+++ b/dbus/dbus_libgio.go
@@ -1,0 +1,41 @@
+// Copyright 2020 Northern.tech AS
+//
+//    Licensed under the Apache License, Version 2.0 (the "License");
+//    you may not use this file except in compliance with the License.
+//    You may obtain a copy of the License at
+//
+//        http://www.apache.org/licenses/LICENSE-2.0
+//
+//    Unless required by applicable law or agreed to in writing, software
+//    distributed under the License is distributed on an "AS IS" BASIS,
+//    WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+//    See the License for the specific language governing permissions and
+//    limitations under the License.
+// +build !nodbus,cgo
+
+package dbus
+
+// #cgo pkg-config: gio-2.0
+// #include <gio/gio.h>
+import "C"
+
+type dbusAPILibGio struct{}
+
+// GenerateGUID generates a D-Bus GUID that can be used with e.g. g_dbus_connection_new().
+// https://developer.gnome.org/gio/stable/gio-D-Bus-Utilities.html#g-dbus-generate-guid
+func (d *dbusAPILibGio) GenerateGUID() string {
+	guid := C.g_dbus_generate_guid()
+	defer C.g_free(C.gpointer(guid))
+	return goString(guid)
+}
+
+// IsGUID Checks if string is a D-Bus GUID.
+// https://developer.gnome.org/gio/stable/gio-D-Bus-Utilities.html#g-dbus-is-guid
+func (d *dbusAPILibGio) IsGUID(str string) bool {
+	cstr := C.CString(str)
+	return goBool(C.g_dbus_is_guid(cstr))
+}
+
+func init() {
+	dbusAPI = &dbusAPILibGio{}
+}
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -26,7 +26,7 @@ test:
     - mkdir -p /go/src/$(dirname $REPO_NAME)/mender /go/src/_/builds
     - cp -r $CI_PROJECT_DIR /go/src/$(dirname $REPO_NAME)
     - cd $GOPATH/src/$REPO_NAME
-    - apt-get update && apt-get install -yyq liblzma-dev libssl-dev
+    - apt-get update && apt-get install -yyq liblzma-dev libssl-dev libglib2.0-dev
     - make get-tools
   script:
     - make extracheck
--- a/dbus/dbus_libgio_helpers.go
+++ b/dbus/dbus_libgio_helpers.go
@@ -1,0 +1,29 @@
+// Copyright 2020 Northern.tech AS
+//
+//    Licensed under the Apache License, Version 2.0 (the "License");
+//    you may not use this file except in compliance with the License.
+//    You may obtain a copy of the License at
+//
+//        http://www.apache.org/licenses/LICENSE-2.0
+//
+//    Unless required by applicable law or agreed to in writing, software
+//    distributed under the License is distributed on an "AS IS" BASIS,
+//    WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+//    See the License for the specific language governing permissions and
+//    limitations under the License.
+// +build !nolibgio,cgo
+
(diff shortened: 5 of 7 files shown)


README.md
```diff
@@ -82,6 +81,14 @@ dependency and substitute the `make` commands in the instructions below for:
 make TAGS=nolzma
 ```
 
+#### Glib's GIO support opt-out
+
+If no Glib's GIO library support if desired, you can ignore the `glib2-dev` package
```
kacf, line 87: I would rather say "If no client DBus support is desired, you can...".

README.md
```diff
@@ -82,6 +81,14 @@ dependency and substitute the `make` commands in the instructions below for:
...
+dependency and substitute the `make` commands in the instructions below for:
+
+```
+make TAGS=nolibgio
```
kacf, line 91: Same here, even though it's relevant to mention glib, I think the tag itself should be feature focused, not library focused, so I would call it `nodbus`.

README.md
```diff
@@ -71,8 +71,7 @@ issue. We thank you in advance for your cooperation.
 
 * C compiler
 * [Go compiler](https://golang.org/dl/)
-* liblzma-dev and libssl-dev packages
-
```
tranchitella, line 75: Did you remove the extra new line on purpose?

tranchitella (approved):

README.md
lluiscampos: No, by mistake. Thanks

lluiscampos: Fixed now.

dbus/dbus.go
```diff
@@ -0,0 +1,35 @@
...
+//    distributed under the License is distributed on an "AS IS" BASIS,
+//    WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+//    See the License for the specific language governing permissions and
+//    limitations under the License.
+
+package dbus
+
+import "github.com/pkg/errors"
+
+var dbusInterface DBusInterface = nil
```
kacf, line 19: Isn't it better to return this from a function instead and allow callers to decide whether it should be global or not?

dbus/dbus_libgio.go
```diff
@@ -0,0 +1,43 @@
...
+
+// IsGUID Checks if string is a D-Bus GUID.
+// https://developer.gnome.org/gio/stable/gio-D-Bus-Utilities.html#g-dbus-is-guid
+func (d *DBusInterfaceLibGio) IsGUID(str string) bool {
+	cstr := C.CString(str)
+	return goBool(C.g_dbus_is_guid(cstr))
+}
+
+func init() {
+	RegisterDBusInterface(&DBusInterfaceLibGio{})
```
kacf, line 42: I don't understand this. Why do you need a register-function if the struct is just empty?

dbus/dbus.go
lluiscampos: Sorry I don't understand your question. It is returned from `GetDBusInterface()`, it is only on the same package that is accessible.

dbus/dbus_libgio.go
lluiscampos: I added the register-function to mimic the `Compressor/CompressorLzma` approach. I can remove the function and do here:
```suggestion
	dbusInterface = &DBusInterfaceLibGio{}
```

dbus/dbus_libgio.go
```diff
@@ -0,0 +1,43 @@
...
+//    Unless required by applicable law or agreed to in writing, software
+//    distributed under the License is distributed on an "AS IS" BASIS,
+//    WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+//    See the License for the specific language governing permissions and
+//    limitations under the License.
+// +build !nolibgio,cgo
+
+package dbus
+
+// #cgo pkg-config: gio-2.0
```
oleorhagen, line 18: ninja :kaaba:

dbus/dbus_libgio.go
lluiscampos: My last force-push reworks as I suggested.

dbus/dbus.go
```diff
@@ -0,0 +1,31 @@
...
+
+type DBusInterface interface {
```
oleorhagen, line 21: Question:
I don't fully understand why the interface is needed here. Can't the package just define the interface?
i.e:
```
dbus.GenerateGUID()
...
```

oleorhagen (commented):
Cool ticket! I think the compile flags are a good way of solving this. Just some comments to stoke a little discussion :)

Nit: 

s/out-out/opt-out/
in commit message

dbus/dbus.go
lluiscampos: @oleorhagen If the package defines the interface here (in the always compiled-in file), were do we put the implementation (i.e. the maybe opt-out code)?

dbus/dbus.go
kacf: Alright, I got it, just rename to DBusAPI to avoid confusion around the concept which is "DBus interface".

README.md
```diff
@@ -82,6 +82,14 @@ dependency and substitute the `make` commands in the instructions below for:
 make TAGS=nolzma
 ```
 
+#### D-Bus support opt-out
+
+If no D-Bus support if desired, you can ignore the `glib2-dev` package dependency and substitut
```
oleorhagen, line 87:
```suggestion
If no D-Bus support if desired, you can ignore the `glib2-dev` package dependency and substitute
```

dbus/dbus_libgio.go
kacf: Yep, all good!

dbus/dbus.go
```diff
@@ -0,0 +1,31 @@
...
+	GenerateGUID() string
+	IsGUID(string) bool
+}
+
+func GetDBusInterface() (DBusInterface, error) {
+	if dbusInterface != nil {
+		return dbusInterface, nil
+	}
+	return nil, errors.New("no DBus interface available")
```
oleorhagen, line 30:
```suggestion
	return nil, errors.New("no D-Bus interface available")
```

dbus/dbus.go
lluiscampos: :+1:

lluiscampos:
@kacf @oleorhagen Thanks for the feedback. Does it look better now?

dbus/dbus.go
oleorhagen: Just playing with ideas here, but something along the lines of:

```
diff --git a/dbus/dbus.go b/dbus/dbus.go
index 41bdbf7..de4b020 100644
--- a/dbus/dbus.go
+++ b/dbus/dbus.go
@@ -16,16 +16,3 @@ package dbus
 
 import "github.com/pkg/errors"
 
-var dbusInterface DBusInterface = nil
-
-type DBusInterface interface {
-	GenerateGUID() string
-	IsGUID(string) bool
-}
-
-func GetDBusInterface() (DBusInterface, error) {
-	if dbusInterface != nil {
-		return dbusInterface, nil
-	}
-	return nil, errors.New("no DBus interface available")
-}
diff --git a/dbus/dbus_libgio.go b/dbus/dbus_libgio.go
index 7ab000e..44efbc3 100644
--- a/dbus/dbus_libgio.go
+++ b/dbus/dbus_libgio.go
@@ -19,8 +19,8 @@ package dbus
 // #include <gio/gio.h>
 import "C"
 
-type DBusInterfaceLibGio struct {
-	d DBusInterface
+func Available() error {
+	return nil
 }
 
 // GenerateGUID generates a D-Bus GUID that can be used with e.g. g_dbus_connection_new().
diff --git a/dbus/nodbus_libgio.go b/dbus/nodbus_libgio.go
new file mode 100644
index 0000000..2cddaf0
--- /dev/null
+++ b/dbus/nodbus_libgio.go
@@ -0,0 +1,16 @@
+
+// +build nodbus,cgo
+
+package dbus
+
+func Available() error () {
+	return errors.New("No D-Bus available")
+}
+
+func GenerateGUID() string {
+	return ""
+}
+
+func IsGUID(str string) bool {
+	return false
+}
```

kacf: I prefer Lluís' method, otherwise we have to define every function we don't use.

dbus/dbus.go
```diff
@@ -0,0 +1,31 @@
...
+type DBusAPI interface {
+	GenerateGUID() string
+	IsGUID(string) bool
+}
+
+func GetDBusAPI() (DBusAPI, error) {
+	if dbusAPI != nil {
+		return dbusAPI, nil
+	}
+	return nil, errors.New("no D-Bus interface available")
```
kacf, line 30: Btw, this is not important at all, but you could define this function twice, once with `nodbus` and once with `!nodbus`. Then you could return the answer directly and you don't need the global.

dbus/dbus_libgio.go
```diff
@@ -0,0 +1,43 @@
...
+// +build !nodbus,cgo
+
+package dbus
+
+// #cgo pkg-config: gio-2.0
+// #include <gio/gio.h>
+import "C"
+
+type DBusAPILibGio struct {
+	d DBusAPI
```
kacf, line 23: Does this actually need to be here? It is implicitly fulfilled, right?

kacf (approved):
Nice going! Only nitpicks, not even sure if they're appropriate.

dbus/dbus.go
oleorhagen: Yes, it is just a suggestion :)

You're right, they have to be defined. However, the readability is pretty good :) It's a trade-off.

lluiscampos: @oleorhagen Now I understand what you meant.

But as Kristian also said I think we would need a lot of boiler plate code. Besides, these are the trivial methods, but we most likely require others with custom structs as parameters and it will get too ugly to mock all of that.

dbus/dbus.go
oleorhagen: haha, where have I seen this idea before

dbus/dbus_libgio.go
lluiscampos: Right!

dbus/dbus.go
lluiscampos: Yes, that is also a valid approach. I wanted to avoid an extra file with a single function.

If during implementation turns out that there are more methods that need to be available for both cases, then we it would make sense to rework as you said.

lluiscampos:
Testing together with https://github.com/mendersoftware/meta-mender/pull/1111 @ [Pipeline 199818145](https://gitlab.com/Northern.tech/Mender/mender-qa/-/pipelines/199818145)

dbus/dbus.go
kacf: > haha, where have I seen this idea before

Not sure if I understood this, it's not the same approach as you suggested. Though you could say it's inspired by it. This would still only include one function call which returns error, the API getter, and all the other functions calls would only have real implementations, no nil implementations.

But yes, it's fine the way it is as well.

oleorhagen: > Not sure if I understood this, it's not the same approach as you suggested. Though you could say it's inspired by it. 

That is what I meant :)

PR #626, opened 2020-10-12
log: Add a log message explicitly stating authorization success

oleorhagen:
Just a nice to have imo

### DIFF
--- a/app/state.go
+++ b/app/state.go
@@ -437,6 +437,7 @@ func (a *authorizeState) Handle(ctx *StateContext, c Controller) (State, bool) {
 	}
 	// if everything is OK we should let Mender figure out what to do
 	// in MenderStateCheckWait state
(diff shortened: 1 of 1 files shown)


app/state.go
```diff
@@ -437,6 +437,7 @@ func (a *authorizeState) Handle(ctx *StateContext, c Controller) (State, bool) {
 	}
 	// if everything is OK we should let Mender figure out what to do
 	// in MenderStateCheckWait state
+	log.Info("Server authorization succesful")
```
kacf, line 440: `s/succesful/successful/` :smile:

kacf (approved):

app/state.go
oleorhagen: omg, I even was careful with this one.

lluiscampos (approved):

PR #632, opened 2020-10-19
(cherry) testfix: Set the default config directory to a temporary directory

oleorhagen:
This fixes the test issue where this will fail locally if your permissions are
set to exclude access by the test runner.

Changelog: None
(cherry picked from commit 696d59dc17439eed0cb112a4c8e4396164b52ccc)

### DIFF
--- a/cli/cli_test.go
+++ b/cli/cli_test.go
@@ -494,6 +494,7 @@ func TestIgnoreServerConfigVerification(t *testing.T) {
 		"--retry-poll",
 		"30",
 	}
+	conf.DefaultDataStore = tdir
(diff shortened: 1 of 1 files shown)


lluiscampos (approved):

lluiscampos:
@oleorhagen You will need to cherry-pick the gocyclo fix to here (or independently to the release branch, but I think here is fine).

Also, the integration pipeline has no effect on verifying a unit test fix, so I suggest we stop the pipeline and later ping me or other admin for the merge.

oleorhagen:
Incorporating this into #633

PR #634, opened 2020-10-19
Coveralls 2.4.x

oleorhagen:


oleorhagen:
@lluiscampos I guess the required `codecov` check should be disabled also?

lluiscampos (approved):

lluiscampos:
> @lluiscampos I guess the required `codecov` check should be disabled also?

Done. I added a requirement for the `coveralls` one, even though it will be missing in old branches.

PR #642, opened 2020-10-22
fix: Do not close the response body before reading error message

oleorhagen:
This shows up in the client logs as:

2020-10-22 12:14:24 +0000 UTC error: Update fetch failed: (request_id: ): error receiving scheduled update information server error message: failed to parse server response: http: read on closed response body

Due to the prematurely closed response body.

The fix is to simply extract the error message before closing the response body.

Changelog: Correctly log the error message from the server on failed update
download attempts.


lluiscampos:
@oleorhagen Sorry I am late. LGTM.

PR #643, opened 2020-10-22
Some small fixes for dbus.

kacf:
```
commit f8a47b301201916389db5d7191a76531dbfc3ed0
Date:   Thu Oct 22 15:11:34 2020

    Fix incorrect build tag, nolibgio should be nodbus.
    
    Changelog: None
    
commit b93297cde3a288afce6617a81b9db348db94bcd3
Date:   Thu Oct 22 14:37:32 2020

    Make sure dbus is linked to if nodbus tag is off.
    
    Changelog: None
    
```


lluiscampos (approved):

lluiscampos:
We are missing [libglib2.0-dev](https://github.com/mendersoftware/mender/blob/feature-client_auth_api/README.md#requirements) at integration pipeline, but I was not expecting it to fail in `build_servers`... why is this job building the client?

kacf:
> We are missing [libglib2.0-dev](https://github.com/mendersoftware/mender/blob/feature-client_auth_api/README.md#requirements) at integration pipeline, but I was not expecting it to fail in `build_servers`... why is this job building the client?

This is not the QEMU client image, it's the Docker client image. I think this goes all the way back to Jenkins actually. Strictly speaking it is wrong to build it in this job, but this part is quite entangled, so I think I prefer not to rock the boat. It has not caused any issues so far.

As far as the failure is concerned, it is a missing dependency in the Docker file. This should be easy to fix.

lluiscampos:
> This is not the QEMU client image, it's the Docker client image. (...)

Ah, of course. Thanks.

kacf:
@lluiscampos: Do you agree with the very last commit? I can't think of any bad effects it would have.

lluiscampos:
> @lluiscampos: Do you agree with the very last commit? I can't think of any bad effects it would have.

Yes, I also think it should be fine.

lluiscampos (approved):

PR #645, opened 2020-10-27
QA-214: Publish the 'latest' debian client package on new tags

oleorhagen:
This is done through comparing the latest MENDER_CLIENT_VERSION from the
mender-client source, and then comparing it to the CI_COMMIT_TAG gitlab
variable, which is set to the tag of the current build.

Changelog: None

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -89,11 +89,16 @@ trigger:mender-dist-packages:
   stage: trigger
   before_script:
     - apk add --no-cache curl
+    - MENDER_CLIENT_VERSION=$(git tag | egrep -e '^[0-9]+\.[0-9]+\.[0-9]+$' | sort -V -r | head -n1)
+    # This will only be true in the case that we have a new newest client tag
+    - PUBLISH_LATEST_CLIENT_PACKAGE="false"
+    - test "${MENDER_CLIENT_VERSION}" = "${CI_COMMIT_REF_NAME}" && export PUBLISH_LATEST_CLIENT_PACKAGE="true" || true
   script:
     - curl -v -f -X POST
       -F token=$MENDER_DIST_PACKAGES_TRIGGER_TOKEN
       -F ref=master
       -F variables[MENDER_VERSION]=$CI_COMMIT_REF_NAME
+      -F variables[PUBLISH_LATEST_CLIENT_PACKAGE]=$PUBLISH_LATEST_CLIENT_PACKAGE
       https://gitlab.com/api/v4/projects/14968223/trigger/pipeline
   only:
     - tags


oleorhagen:
This then also needs to be cherry-picked to:
* `2.4.x` if we are to publish a latest package this time around.

kacf (approved):

oleorhagen:
@lluiscampos can you look at this in the morning :) ?

.gitlab-ci.yml
```diff
@@ -89,11 +89,15 @@ trigger:mender-dist-packages:
   stage: trigger
   before_script:
     - apk add --no-cache curl
+    - export MENDER_CLIENT_VERSION=$(git tag | egrep -e '^[0-9]+\.[0-9]+\.[0-9]+$' | sort -V -r | head -n1)
+    # This will only be true in the case that we have a new newest client tag
+    - test ${MENDER_CLIENT_VERSION} = ${CI_COMMIT_TAG} && export PUBLISH_LATEST_CLIENT_PACKAGE="true" || true
```
lluiscampos, line 94: I would quote the variable in case it is empty for some reason
```suggestion
    - test "${MENDER_CLIENT_VERSION}" = ${CI_COMMIT_TAG} && export PUBLISH_LATEST_CLIENT_PACKAGE="true" || true
```

lluiscampos, line 94: I would set a default value of "false", adding:
```
    - PUBLISH_LATEST_CLIENT_PACKAGE="false"
```

I am not sure if this is important, will curl fail parsing a parameter like `-F variables[something]=` ?

.gitlab-ci.yml
```diff
@@ -89,11 +89,15 @@ trigger:mender-dist-packages:
   stage: trigger
   before_script:
     - apk add --no-cache curl
+    - export MENDER_CLIENT_VERSION=$(git tag | egrep -e '^[0-9]+\.[0-9]+\.[0-9]+$' | sort -V -r | head -n1)
```
lluiscampos, line 92: Not important neither, but for future reference. You don't need `export` here: GitLab just appends `before_script` and `script` and runs it on a shell, so the variables are visible.

lluiscampos (changes requested):
Some minor comments for a bit more robustness in the script. But the design looks good!

lluiscampos:
> @lluiscampos can you look at this in the morning :) ?

@oleorhagen Which morning? Did I made the deadline? :)

oleorhagen:
> > @lluiscampos can you look at this in the morning :) ?
> 
> @oleorhagen Which morning? Did I made the deadline? :)

Just in the nick of time :rocket:

.gitlab-ci.yml
oleorhagen: How about, just to be sure, we do:
```
test "${MENDER_CLIENT_VERSION}" = "${CI_COMMIT_TAG:-false}"
```
just to be sure?

oleorhagen: Better safe than sorry. Adding.

.gitlab-ci.yml
oleorhagen: Thx, noted and removed.

.gitlab-ci.yml
lluiscampos: If we quote, the default value "-false" is not needed.

Actually, now that I see it with fresh eyes, we should use `CI_COMMIT_REF_NAME` (same as we pass to the pipeline below) instead of the TAG one. For tags they will be the same, but for any other build the TAG one will be undefined (hence your default value above).

With regards to the test syntax quoting both is the safest, but if we use CI_COMMIT_REF_NAME we are guaranteed to have a value from GitLab.

oleorhagen: > With regards to the test syntax quoting both is the safest, but if we use CI_COMMIT_REF_NAME we are guaranteed to have a value from GitLab.

Yes, this is much nicer!

oleorhagen: One more question:
`    - test "${MENDER_CLIENT_VERSION}" = "${CI_COMMIT_REF_NAME}" && export PUBLISH_LATEST_CLIENT_PACKAGE="true" || true`
Is it more correct to actually let the job fail here? i.e., remove the `||true` suffix?
I think so

lluiscampos (approved):

oleorhagen:
Validated the syntax with: https://gitlab.com/Northern.tech/Mender/mender/-/ci/lint

Should be good to go now

.gitlab-ci.yml
```diff
@@ -89,11 +89,16 @@ trigger:mender-dist-packages:
   stage: trigger
   before_script:
     - apk add --no-cache curl
+    - MENDER_CLIENT_VERSION=$(git tag | egrep -e '^[0-9]+\.[0-9]+\.[0-9]+$' | sort -V -r | head -n1)
+    # This will only be true in the case that we have a new newest client tag
+    - PUBLISH_LATEST_CLIENT_PACKAGE="false"
```
mzedel, line 94: shouldn't these be `export`ed?

oleorhagen: Hmm, why? They're not needed in any sub-processes, I think?

lluiscampos: They are not needed, as `before_script` and `script` are just concatenated and executed in the same shell.

If any, we could also remove `export` from line below

oleorhagen: I'll create a tidy-up PR

oleorhagen: https://github.com/mendersoftware/mender/pull/661

mzedel: the main reason I'm asking is https://gitlab.com/Northern.tech/Mender/mender/-/jobs/840156620 - trying it locally the comparison failed unless I went the `export` way...

oleorhagen: Hmm, really :thinking:  How can that be?

lluiscampos: @oleorhagen  After some investigation, the problem seems to be in the subshell, and my current hypothesis is that `git tag` returns an empty list because the GitLab checkout is a shallow clone. I asked @mzedel to follow-up :+1:

mzedel: great... now after debugging the git behavior in #662 and the `git tag`s being present already, I retried the failed job and => https://gitlab.com/Northern.tech/Mender/mender/-/jobs/840632341 succeeded... without changes...

PR #646, opened 2020-10-27
MEN-4009, MEN-4016, MEN-4017: Implement DBus API

tranchitella:
* **MEN-4009**: Implement DBus API: GetJwtToken
* **MEN-4016**: Implement Dbus API: FetchJwtToken
* **MEN-4017**: Implement DBus signal: ValidJwtTokenAvailable


dbus/dbus_libgio.go
```diff
@@ -36,6 +48,34 @@ func (d *dbusAPILibGio) IsGUID(str string) bool {
 	return goBool(C.g_dbus_is_guid(cstr))
 }
 
+// ServerNew creates a new D-Bus server that listens on the specified address
+// https://developer.gnome.org/gio/stable/GDBusServer.html#g-dbus-server-new-sync
+func (d *dbusAPILibGio) ServerNew(address string, flags int) (unsafe.Pointer, error) {
```
kacf, line 63: As discussed in chat, if this is not going to be used in the client itself, but is useful in tests, it would be good to move it into test files or its own package, to avoid importing it in the client and wasting space.

dbus/dbus_connection_test.go
```diff
@@ -0,0 +1,89 @@
...
+
+	for name, tc := range testCases {
+		t.Run(name, func(t *testing.T) {
+			api := &dmocks.DBusAPI{}
+			defer api.AssertExpectations(t)
+
+			api.On("ConnectionNewForAddress",
+				mock.AnythingOfType("string"),
+				mock.AnythingOfType("int"),
+			).Return(unsafe.Pointer(nil), tc.connNewErr)
```
kacf, line 59: It would be great, and fairly easy, to test this using a real DBus server. The key difference will be to connect to the user bus instead of the system bus (avoids root), and perhaps we could also switch interface prefix for testing purposes, to make sure we are not messing with anything real.

For connecting to the user bus, use the address in the `DBUS_SESSION_BUS_ADDRESS` environment variable. For connecting to the system bus, use `unix:path=/var/run/dbus/system_bus_socket` (defined in `/usr/share/dbus-1/system.conf`, but in practice I think it's universal).

dbus-server/Dockerfile
```diff
@@ -0,0 +1,7 @@
+FROM debian:buster
+
+RUN apt-get update && apt-get install -y dbus && rm -rf /var/lib/apt/lists/*
+
+CMD ["/usr/bin/dbus-daemon", "--system", "--address=tcp:host=0.0.0.0,port=12434", "--nofork", "--nopidfile", "--systemd-activation", "--syslog-only"]
```
kacf, line 5: I'm not against doing it this way, but I would investigate the possibility of calling this directly inside `.gitlab-ci.yml` instead, so that we can avoid an extra Docker instance in local unit tests, and let them use the already available DBus server instead.

kacf (changes requested):
I will do a bit more thorough review later, but the general direction is good.

dbus/dbus.go
```diff
@@ -14,15 +14,33 @@
...
 	IsGUID(string) bool
+
+	// ServerNew creates a new D-Bus server that listens on the specified address
+	ServerNew(string, int) (unsafe.Pointer, error)
+
+	// ServerStart starts the server
+	ServerStart(unsafe.Pointer)
+
+	// ServerStop stops the server
+	ServerStop(unsafe.Pointer)
```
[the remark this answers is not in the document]
tranchitella: I've Introduced a type to make it prettier

dbus/dbus_connection_test.go
tranchitella: To be honest, for unit testing in the client and server packages, I really prefer using mocks, as it allows us to test edge cases (e.g. failure paths). Said that, I'll follow your suggestion and use a local dbus when testing it in action.

dbus-server/Dockerfile
tranchitella: Ack, in gitlab we'll use a local dbus daemon.

dbus/dbus_libgio.go
tranchitella: moved to dbus/server

tranchitella:
@kacf @oleorhagen this PR is ready for review, I could successfully run the tests applying my commit on top of my previous PR which introduces channel communication (see: https://gitlab.com/Northern.tech/Mender/mender/-/pipelines/210249813), but I fail to run tests with the current `feature-client_auth_api` branch for reasons unrelated to my changes.

I tried to run tests for master and they pass, as expected. I also tried to run tests for the `feature-client_auth_api` branch and they fail, because of other changes we already merged there. I wanted to bisect the changes and find the culprit, but I believe you have a better overview of what is happening here. The tests simply stop (e.g. exit/panic) without any error message.

conf/config.go
```diff
@@ -88,6 +90,10 @@ type MenderConfig struct {
 	RootfsScriptsPath   string
 }
 
+type DbusConfig struct {
```
kacf, line 93: Nitpick: Inconsistent naming, `DbusConfig` should be `DBusConfig` to follow the style of the rest.

Discovered while regex searching with case sensitivity! :-)

conf/config_test.go
```diff
@@ -68,6 +68,13 @@ var testTooManyServerDefsConfig = `{
   "Servers": [{"ServerURL": "mender.io"}]
 }`
 
+var testDBusConfig = `{
+  "ServerURL": "mender.io",
+  "DBus": {
+	"Enabled": true
+  }
```
kacf, line 75: Indentation.

dbus/dbus_libgio.go.h
```diff
@@ -0,0 +1,92 @@
...
+static GMainLoop*
+to_gmainloop(void* ptr)
+{
+	return (GMainLoop*)ptr;
+}
+
+// create a new GVariant from a string value
+static GVariant *g_variant_new_from_string(gchar *value)
+{
+    return g_variant_new("(s)", value);
```
kacf, line 24: Inconsistent indentation in several places in this file. I would argue we should use tabs, as in the Go code.

dbus/dbus_libgio_test.go
```diff
@@ -35,3 +48,295 @@ func TestIsGUID(t *testing.T) {
 	guid := libgio.GenerateGUID()
 	assert.True(t, libgio.IsGUID(guid))
 }
+
+func TestBusGet(t *testing.T) {
+	if testing.Short() {
+		t.Skip("skipping test in short mode.")
```
kacf, line 54: Does it take that long?

dbus/dbus_libgio_test.go
```diff
@@ -35,3 +48,295 @@ func TestIsGUID(t *testing.T) {
...
+			gid, err = libgio.BusRegisterInterface(conn, tc.objectPath, xml)
+			assert.NoError(t, err)
+			assert.Greater(t, gid, uint(0))
+
+			loop := libgio.MainLoopNew()
+			go libgio.MainLoopRun(loop)
+			defer libgio.MainLoopQuit(loop)
+
+			// let the dbus-daemon to set up
+			time.Sleep(500 * time.Millisecond)
```
kacf, line 329: I'm no fan of sleeps. This is what typically causes strange failures six months down the road. Isn't there a better way to check this? Maybe using [`assert.Eventually`](https://godoc.org/github.com/stretchr/testify/assert#Eventually) with a longer timeout? Or perhaps you could get the progress of setup and teardown through channels?

dbus/pointer.go
```diff
@@ -0,0 +1,21 @@
...
+//    See the License for the specific language governing permissions and
+//    limitations under the License.
+// +build !nodbus,cgo
+
+package dbus
+
+import "unsafe"
+
+// Pointer is an unsafe pointer, only prettier :)
+type Pointer unsafe.Pointer
```
kacf, line 21: Hehe, I was actually thinking of something like `Handle` or something like that. But no strong preference.

dbus/dbus_libgio_test.go
```diff
@@ -35,3 +48,295 @@ func TestIsGUID(t *testing.T) {
...
+			assert.Greater(t, gid, uint(0))
+
+			libgio.RegisterMethodCallCallback(tc.path, tc.interfaceName, tc.methodName, tc.callback)
+
+			loop := libgio.MainLoopNew()
+			go libgio.MainLoopRun(loop)
+			defer libgio.MainLoopQuit(loop)
+
+			// let the dbus-daemon to set up
+			time.Sleep(500 * time.Millisecond)
```
kacf, line 245: Same here with regards to sleep.

kacf (changes requested):
Sorry for the late review, I had a lot of interruptions today.

I have to admit that a lot of the C interfacing goes over my head, but I'm sure you had fun learning it! :-)

app/auth.go
```diff
@@ -40,6 +43,27 @@ const (
 	EventAuthTokenAvailable = "AUTH_TOKEN_AVAILABLE"
 )
 
+// Constants for the auth manager DBus interface
+const (
+	AuthManagerDBusPath                         = "/io/mender/AuthenticationManager"
+	AuthManagerDBusObjectName                   = "io.mender.AuthenticationManager"
+	AuthManagerDBusInterfacetName               = "io.mender.AuthenticationManager1"
```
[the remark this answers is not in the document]
tranchitella: typo, fixed

app/auth.go
```diff
@@ -156,8 +189,55 @@ func (m *MenderAuthManager) GetBroadcastMessageChan(name string) <-chan AuthMana
...
+			return message.Event == EventFetchAuthToken, message.Error
+		case <-time.After(5 * time.Second):
+		}
+		return false, errors.New("timeout when calling GetJwtToken")
+	})
+}
+
 // Run is the main routine of the Mender authorization manager
 func (m *MenderAuthManager) Run() error {
+	// run the DBus interface, if availablle
```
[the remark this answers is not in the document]
tranchitella: typo, fixed

conf/config.go
tranchitella: Fixed

conf/config_test.go
tranchitella: Fixed, two spaces instead of tab

dbus/dbus_libgio.go.h
tranchitella: I used spaces and tried to make it consistent, don't know if I can use some sort-of formatter.
I also added an editorconfig file to enforce it.
What do you think?

dbus/dbus_libgio_test.go
tranchitella: No, it was a way to disable tests that need a dbus daemon, we do that way in the backend (e.g. mongo-related tests).
Never mind, I'm removing the switch, as it doesn't make too much sense.

dbus/pointer.go
tranchitella: Do you mean, calling it "Handle" instead of Pointer?

app/auth.go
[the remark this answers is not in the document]
tranchitella: Ack, you are right, I renamed the interface.

app/auth.go
```diff
@@ -40,6 +43,27 @@ const (
...
+	AuthManagerDBusInterface                    = `<node>
+	<interface name="io.mender.AuthenticationManager1">
+		<method name="GetJwtToken">
+			<arg type="s" name="token" direction="out"/>
+		</method>
+		<method name="FetchJwtToken">
+			<arg type="b" name="success" direction="out"/>
+		</method>
+		<signal name="ValidJwtTokenAvailable">
+			<arg type="b" name="success" direction="out"/>
```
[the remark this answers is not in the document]
tranchitella: It is not needed, I removed it, it is now a bare signal

dbus/dbus_libgio_test.go
tranchitella: This is not needed anymore, it was part of my "debugging" because I thought we have issues here but the test failed in any case. The tests bass without any sleep, thus I'm removing them.

tranchitella:
@kacf ready for another round, fixes are in independent commits, I'll squash them later on

.editorconfig
```diff
@@ -0,0 +1,17 @@
...
+end_of_line = lf
+insert_final_newline = true
+
+# Tab indentation (no size specified)
+[Makefile]
+indent_style = tab
+
+# cgo code
+[*.go.h]
+indent_size = 2
```
kacf, line 17: From the code it doesn't look like this actually gave the correct result. 

However, I talked about this with @oleorhagen, and we decided that maybe it's better to just standardize the C code template to something we use in the rest of the company, instead of something that matches Golang. So my suggestion is, use [this one](https://github.com/mendersoftware/mender-binary-delta/blob/master/.clang-format) instead. It's originally taken from CFEngine, and used in mender-binary-delta.

Together with that, you should add [a clang install](https://github.com/mendersoftware/mender-binary-delta/blob/master/.gitlab-ci.yml#L85) and [a check](https://github.com/mendersoftware/mender-binary-delta/blob/master/.gitlab-ci.yml#L105).

Does that sound good?

kacf (changes requested):
Thanks @tranchitella! Only one issue left which I think should be handled, the rest can be followed up later.

app/auth.go
```diff
@@ -40,6 +43,25 @@ const (
...
+	<interface name="io.mender.Authentication1">
+		<method name="GetJwtToken">
+			<arg type="s" name="token" direction="out"/>
+		</method>
+		<method name="FetchJwtToken">
+			<arg type="b" name="success" direction="out"/>
+		</method>
+		<signal name="ValidJwtTokenAvailable"></signal>
+	</interface>
+</node>`
```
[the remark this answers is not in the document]
tranchitella: Yes, we can load it from a file, but it makes sense only if we can do it at build time, and I'm not sure it is doable.
Loading it at run time means the file should be installed somewhere in the device, and this is something we don't really need.

.editorconfig
tranchitella: Let me try :)

tranchitella: Done!

app/auth.go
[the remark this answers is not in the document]
kacf: Ok, it's not the end of the world if we need to have two. But we can try it separately when doing MEN-4104.

.editorconfig
kacf: This looks better, thanks!

kacf (approved):

.editorconfig
tranchitella: I also added the check in the pipeline, let's see if it passes.

PR #651, opened 2020-10-30
[inventory-geo-script]: Add a separate install target

oleorhagen:
This way, it can later on be made optional, whether or not to include the
'inventory-geo-script'.

This is followed by a separate target in meta-mender, called:
mender-inventory-geo-script, which can be added ad-hoc by a superuser which
whishes full control over his install.

This follows from the situation where we have had some issues with the
third-party provider we rely on for the geolocation data.

Changelog: The 'inventory-geo-script' now has a separate install target: 'install-inventory-geo-script'

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -56,8 +56,10 @@ INVENTORY_SCRIPTS = \
 	support/mender-inventory-os \
 	support/mender-inventory-provides \
 	support/mender-inventory-rootfs-type \
-	support/mender-inventory-geo \
 	support/mender-inventory-update-modules
+
+INVENTORY_NETWORK_SCRIPTS = \
+	support/mender-inventory-geo
 
 MODULES = \
 	support/modules/deb \
@@ -94,9 +96,15 @@ install-identity-scripts: install-datadir
 	install -m 755 -d $(prefix)$(datadir)/mender/identity
 	install -m 755 $(IDENTITY_SCRIPTS) $(prefix)$(datadir)/mender/identity/
 
-install-inventory-scripts: install-datadir
+install-inventory-scripts: install-inventory-local-scripts install-inventory-network-scripts
+
+install-inventory-local-scripts: install-datadir
 	install -m 755 -d $(prefix)$(datadir)/mender/inventory
 	install -m 755 $(INVENTORY_SCRIPTS) $(prefix)$(datadir)/mender/inventory/
+
+install-inventory-network-scripts: install-datadir
+	install -m 755 -d $(prefix)$(datadir)/mender/inventory
+	install -m 755 $(INVENTORY_NETWORK_SCRIPTS) $(prefix)$(datadir)/mender/inventory/
 
 install-modules:
 	install -m 755 -d $(prefix)$(datadir)/mender/modules/v3
@@ -131,8 +139,17 @@ uninstall-identity-scripts:
 	done
 	-rmdir -p $(prefix)$(datadir)/mender/identity
 
-uninstall-inventory-scripts:
+uninstall-inventory-scripts: uninstall-inventory-local-scripts uninstall-inventory-network-scripts
+	-rmdir -p $(prefix)$(datadir)/mender/inventory
+
+uninstall-inventory-local-scripts:
 	for script in $(INVENTORY_SCRIPTS); do \
+		rm -f $(prefix)$(datadir)/mender/inventory/$$(basename $$script); \
+	done
+	-rmdir -p $(prefix)$(datadir)/mender/inventory
+
+uninstall-inventory-network-scripts:
+	for script in $(INVENTORY_NETWORK_SCRIPTS); do \
 		rm -f $(prefix)$(datadir)/mender/inventory/$$(basename $$script); \
 	done
 	-rmdir -p $(prefix)$(datadir)/mender/inventory
@@ -236,6 +253,8 @@ instrument-binary:
 .PHONY: install-datadir
 .PHONY: install-identity-scripts
 .PHONY: install-inventory-scripts
+.PHONY: install-inventory-local-scripts
+.PHONY: install-inventory-network-scripts
 .PHONY: install-modules
 .PHONY: install-modules-gen
 .PHONY: install-systemd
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -96,6 +98,10 @@ install-inventory-scripts: install-datadir
 	install -m 755 -d $(prefix)$(datadir)/mender/inventory
 	install -m 755 $(INVENTORY_SCRIPTS) $(prefix)$(datadir)/mender/inventory/
 
+install-inventory-geo-script:
```
merlin-northern, line 101: so, perhaps we could use it and install it by default?

oleorhagen: The issue is that we'd like to install it conditionally here: https://github.com/mendersoftware/meta-mender/pull/1161. And I don't know how to do this otherwise. Here it is installed by default (as it is now), only, it is possible for more experienced users to leave it out, and in the future, maybe have this as opt in.

Good ideas are always welcome though

kacf: I think I would split this a bit differently. I agree with @merlin-northern that we should install it by default when running `make`. Here's my proposal:

* `install-inventory-scripts` - Installs all scripts
* `install-local-inventory-scripts` - Installs all scripts except geo
* `install-network-inventory-scripts` - Installs geo
* `install` - Depends on `install-inventory-scripts` and hence installs all inventory scripts.

oleorhagen: Alright, sounds smrt, I'll do this.

lluiscampos: :+1:

Makefile
```diff
@@ -55,7 +55,9 @@ INVENTORY_SCRIPTS = \
 	support/mender-inventory-network \
 	support/mender-inventory-os \
 	support/mender-inventory-rootfs-type \
-	support/mender-inventory-geo
+
+INVENTORY_NETWORK_SCRIPTS = \
+  support/mender-inventory-geo
```
lluiscampos, line 60: Before Kristian comments: wrong indentation :p

Makefile
```diff
@@ -92,10 +94,18 @@ install-identity-scripts: install-datadir
 	install -m 755 -d $(prefix)$(datadir)/mender/identity
 	install -m 755 $(IDENTITY_SCRIPTS) $(prefix)$(datadir)/mender/identity/
 
-install-inventory-scripts: install-datadir
+install-inventory-scripts: install-datadir install-inventory-network-scripts
```
lluiscampos, line 97: This should be:
```suggestion
install-inventory-scripts: install-local-inventory-scripts install-inventory-network-scripts
```
(sic)

Makefile
```diff
@@ -92,10 +94,18 @@ install-identity-scripts: install-datadir
...
+	install -m 755 -d $(prefix)$(datadir)/mender/inventory
+	install -m 755 $(INVENTORY_SCRIPTS) $(prefix)$(datadir)/mender/inventory/
+
+install-local-inventory-scripts: install-datadir
```
lluiscampos, line 101: For consistency, I suggest:
```
install-local-inventory-scripts
install-network-inventory-scripts
```
or
 ```
install-inventory-local-scripts
install-inventory-network-scripts
```

File: Makefile [hunk 5]
```diff
@@ -135,6 +145,9 @@ uninstall-inventory-scripts:
 	done
 	-rmdir -p $(prefix)$(datadir)/mender/inventory
 
+uninstall-inventory-network-scripts:
+	rm -f $(prefix)$(datadir)/mender/inventory/$(INVENTORY_NETWORK_SCRIPTS)
```
lluiscampos, line 149: This won't work for more than one file, see targets above for inspiration on how to remove many.

Makefile
```diff
@@ -135,6 +145,9 @@ uninstall-inventory-scripts:
...
+
```
lluiscampos, line 150: Also, not very important, but I am missing an equivalent `uninstall-inventory-local-scripts` (sic).

lluiscampos (changes requested):
Looking good. I produced a bunch of nit-pick comments:

Makefile
oleorhagen: Hmm, yeah. Good idea. Then `install-inventory-scripts` is just a shim target, tying together the `local` and `network` targets.

Makefile
oleorhagen: Yes, true. Let's unify this. I would vote for the second option ?

Makefile
oleorhagen: Shite, leftover from the `Geo-script` singular.

Makefile
oleorhagen: This is consistent with the other indentation :)

It's the old that was wrong ;)

lluiscampos (approved):
Very nice :climbing_man:

Makefile
```diff
@@ -129,11 +137,18 @@ uninstall-identity-scripts:
...
 
-uninstall-inventory-scripts:
+uninstall-inventory-scripts: uninstall-inventory-local-scripts uninstall-inventory-network-scripts
+	-rmdir -p $(prefix)$(datadir)/mender/inventory
+
+uninstall-inventory-local-scripts:
 	for script in $(INVENTORY_SCRIPTS); do \
 		rm -f $(prefix)$(datadir)/mender/inventory/$$(basename $$script); \
 	done
-	-rmdir -p $(prefix)$(datadir)/mender/inventory
```
kacf, line 136: This should stay, no? And it should be added below too.

kacf (changes requested):

Makefile
oleorhagen: The thinking went like:
```
uninstall-inventory-scripts: uninstall-inventory-local-scripts uninstall-inventory-network-scripts
	-rmdir -p $(prefix)$(datadir)/mender/inventory
```
This uninstalls all the inventory scripts, like it says.

The `uninstall-inventory-local-scripts` and `uninstall-inventory-network-scripts` are a subset of the inventory scripts installed. And hence, uninstalling only one of them, may actually leave scripts behind. Thus, this should not remove the directory, as then it would remove more scripts, than what the target actually installs, and would hence not be symmetric.

kacf: Yes, but `rmdir` is "soft", it will not remove the directory if it's not empty.

oleorhagen: Ahh, I see. Kk. Adding.

PR #652, opened 2020-10-31
Add: inventory-script for reporting installed update-modules

oleorhagen:
This simple scripts lists all the executable files found in the inventory-script
path:

/usr/share/mender/modules/*/*

Changelog: Add an inventory script for reporting the update-modules currently
installed on a device.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -55,7 +55,8 @@ INVENTORY_SCRIPTS = \
 	support/mender-inventory-network \
 	support/mender-inventory-os \
 	support/mender-inventory-rootfs-type \
-	support/mender-inventory-geo
+	support/mender-inventory-geo \
+	support/mender-inventory-update-modules
 
 MODULES = \
 	support/modules/deb \
--- a/support/mender-inventory-update-modules
+++ b/support/mender-inventory-update-modules
@@ -1,0 +1,27 @@
+#!/bin/sh
+#
+# Returns the update-modules currently installed on the system
+# For more information on update-modules see:
+# https://hub.mender.io/c/update-modules/13
+#
+
+set -e
+
+RESET_INVENTORY="true"
+
+if [ ! -d /usr/share/mender/modules/ ]; then
+    echo update_modules=""
+    exit 0
+fi
+
+for file in /usr/share/mender/modules/*/*; do
+    if [ ! -x "${file}" ]; then
+      continue
+    fi
+    echo update_modules=$(basename "${file}")
+    RESET_INVENTORY="false"
+done
+
+if [ "${RESET_INVENTORY}" = "true" ]; then
+    echo update_modules=""
+fi


support/mender-inventory-update-modules
```diff
@@ -0,0 +1,28 @@
+#!/bin/sh
+#
+# Returns an the update-modules currently installed on the system
+# For more information on update-modules see:
+# https://hub.mender.io/c/update-modules/13
+#
+
+set -e
+
+FOUND=0
```
oleorhagen, line 10: This I'm not sure if we need or not?
Like, do we reset the inventory on a fresh rootfs install?

Makefile
```diff
@@ -55,7 +55,8 @@ INVENTORY_SCRIPTS = \
 	support/mender-inventory-network \
 	support/mender-inventory-os \
 	support/mender-inventory-rootfs-type \
-	support/mender-inventory-geo
+	support/mender-inventory-geo \
+  support/mender-inventory-update-modules
```
kacf, line 59: Wrong indentation.

support/mender-inventory-update-modules
```diff
@@ -0,0 +1,28 @@
...
+
+if [ ! -d /usr/share/mender/modules/ ]; then
+  echo update_module'(s)'=""
```
kacf, line 13: Aren't we just asking for shell parsing trouble here? I'm not against it per se, but I just reckon `update_modules` is just as clear and correct, less likely to be misunderstood as some shell magic, and easier to handle if you ever want to query the inventory from a script.

support/mender-inventory-update-modules
kacf: No, we don't. So I think it's a good idea to have this.

support/mender-inventory-update-modules
```diff
@@ -0,0 +1,28 @@
...
+  exit 0
+fi
+
+for file in /usr/share/mender/modules/*/*; do
+  if [ ! -x "${file}" ]; then
```
kacf, line 18: Standard indentation in the other inventory scripts is 4 spaces.

kacf (changes requested):

support/mender-inventory-update-modules
oleorhagen: I guess that then, maybe we should bring this idea to the rest of the scripts?
@lluiscampos what do you think?

support/mender-inventory-update-modules
oleorhagen: Agreed!

support/mender-inventory-update-modules
```diff
@@ -0,0 +1,28 @@
...
+    continue
+  fi
+  echo update_module'(s)'=$(basename "${file}")
```
lluiscampos, line 21: I don't get it: won't this override the same `update_module(s)` key for every module? I will produce:

```
update_module(s)=deb
update_module(s)=directory
update_module(s)=docker
update_module(s)=rpm
update_module(s)=script
update_module(s)=single-file
```

support/mender-inventory-update-modules
lluiscampos: Good idea. I agree :+1:

support/mender-inventory-update-modules
oleorhagen: Yes, but according to our docs:
https://docs.mender.io/client-installation/inventory#lists-of-values
this is correct for a list of values I think.

lluiscampos: Thanks. Shame on me.

oleorhagen: https://giphy.com/gifs/vX9WcCiWwUF7G/html5

Feature request for the `mender-test-bot`

oleorhagen: And Github appareantly

Makefile
oleorhagen: Shame on me!

support/mender-inventory-update-modules
oleorhagen: I wen't through the other `inventory` scripts, and they actually seem quite good. The all report some zero value, in case the expected values are not found. This means that they behave like we expect them to.

kacf (approved):

lluiscampos (approved):

PR #658, opened 2020-11-05
MEN-4030: Add busconfig file for DBus API to install steps. 

kacf:


kacf:
Hmm, the failure reveals a bit bigger problem: This needs to be runnable locally without running as root. I'll see if it's possible to adapt the tests to use the session bus during testing.

Makefile
```diff
@@ -123,6 +141,10 @@ uninstall-conf:
 	rm -f $(prefix)$(sysconfdir)/mender/artifact_info
 	-rmdir -p $(prefix)$(sysconfdir)/mender
 
+uninstall-dbus:
+	rm -f $(prefix)$(datadir)/dbus-1/system.d/io.mender.conf
```
tranchitella, line 145: Are you sure about the file name convention (from what I see, it should be the name of the object we register, not its prefix) and the path? On a Debian stretch machine, the conf files are here: /etc/dbus-1/system.d/

tranchitella: > Hmm, the failure reveals a bit bigger problem: This needs to be runnable locally without running as root. I'll see if it's possible to adapt the tests to use the session bus during testing.

Unfortunately, I failed to use the session dbus because it requires X11 running. I could have done something wrong, I don't know. But the error I see in CI/CD is different:

```
$ cp dbus-1/system.d/mender.conf /etc/dbus-1/system.d/mender.conf
cp: cannot stat 'dbus-1/system.d/mender.conf': No such file or directory
```

kacf: Yes, this error is easy to fix. But it's the fact that it copies into a root-owned directory which is the problem. I want to understand this a bit better, so I will spend a bit more time looking for ways to avoid that. If nothing else, I will learn about DBus. :-)

kacf: > Are you sure about the file name convention (from what I see, it should be the name of the object we register, not its prefix) and the path?

I think you're right, it should be `io.mender.AuthenticationManager.conf`, right?

> On a Debian stretch machine, the conf files are here: /etc/dbus-1/system.d/

Hmm, I don't know. There is a mix on my machine. Intuitively, it seems more correct to put them in `/usr/share`, so that you can override them from `/etc`. I think what is written [here](https://manpages.debian.org/experimental/dbus/dbus-daemon.1.en.html#CONFIGURATION_FILE) supports this view. What do you think?

kacf:
@tranchitella: Fixed it so the test can be run without a running DBus server, and without running as root. Also fixed the conf filename, but I left the file path alone. I still suspect my path is preferable, but I'd like your input on it.

Makefile
tranchitella: Man page of dbus-deamon says:
```
       This is intended to allow extension of the system bus by particular packages. For example,
       if CUPS wants to be able to send out notification of printer queue changes, it could
       install a file to /usr/share/dbus-1/system.d or /etc/dbus-1/system.d that allowed all apps
       to receive this message and allowed the printer daemon user to send it.
```

I believe your path is good, let's stick to it.

tranchitella (approved):

kacf:
@tranchitella: This is finally ready for a new review. Sorry this got kind of messy, I bumped into a lot of issues on the way and I had to merge with master to make it pass. However, none of the old commits have changed, so it should be relatively straightforward to review the new commits one by one. Everything south of [this comment](https://github.com/mendersoftware/mender/pull/658#event-3979206226) should be new.

dbus/dbus_libgio.go
```diff
@@ -55,6 +55,12 @@ func (d *dbusAPILibGio) BusGet(busType uint) (Handle, error) {
 	if Handle(gerror) != nil {
 		return Handle(nil), ErrorFromNative(Handle(gerror))
 	}
+
+	// For most applications it makes sense to close when the connection to
+	// the session ends. But Mender should keep running so that a broken
+	// DBus setup does not prevent the device from being updated.
+	C.g_dbus_connection_set_exit_on_close(conn, C.gboolean(0))
```
tranchitella, line 62: Oh, good catch, this is the one which I could not understand: the Mender client quit when I stopped the dbus service!

app/auth.go
```diff
@@ -251,10 +275,6 @@ func (m *MenderAuthManager) Run() error {
 			m.fetchAuthToken(msg.ResponseChannel)
 		}
 	}
-	// stop the DBus interface, if available
-	if dbusConn != dbus.Handle(nil) && dbusLoop != dbus.MainLoop(nil) {
-		m.dbus.MainLoopQuit(dbusLoop)
-	}
```
tranchitella: Only for my learning, we did you remove the MainLoopQuit here?

tranchitella (approved):

app/auth.go
kacf: I didn't remove it, I just moved it to a `defer` in the DBus initialization section above. It may or may not execute depending on the outcome of the initialization steps, so rechecking the pointers here is redundant.

dbus/dbus_libgio.go
kacf: Exactly what happened to me as well! :-)

kacf:
Merging despite failures. Apparently they are not introduced by this PR, so I will continue the investigation on the branch.

PR #662, opened 2020-11-10
fix/mender-client-version-detection

mzedel:


.gitlab-ci.yml
```diff
@@ -89,17 +89,20 @@ trigger:mender-dist-packages:
   stage: trigger
   before_script:
     - apk add --no-cache curl git
+    - git fetch --unshallow
```
lluiscampos, line 92: Debug tip: just add `git tag` here too to print all tags.

mzedel: thanks for the tip! unfortunately https://gitlab.com/Northern.tech/Mender/mender/-/jobs/840590671 says the tags are available ☹️

lluiscampos (approved):

PR #664, opened 2020-11-16
Fix regression: Cause of error not reported when authorization fails.

kacf:
This was causing some integration tests which look for the error to
fail.

Changelog: None


kacf:
This fixes two of the currently four failing integration tests.

tranchitella (approved):
LGTM

kacf:
There are still issues, but it solves the problems I expected it to solve. Merging, and continuing with other failures independently.

PR #665, opened 2020-11-19
MEN-4177: Implement fallback to PATCH endpoint for inventory updates.

kacf:
Otherwise a new client cannot submit inventory to an old server.

Changelog: None

### DIFF
--- a/client/client_inventory.go
+++ b/client/client_inventory.go
@@ -35,29 +35,43 @@ func NewInventory() InventorySubmitter {
 
 // Submit reports status information to the backend
 func (i *InventoryClient) Submit(api ApiRequester, url string, data interface{}) error {
-	req, err := makeInventorySubmitRequest(url, data)
+	// PATCH used to be the only method available in Mender Product 2.5, so
+	// fall back to that if PUT fails.
+	r, err := doSubmitInventory(api, http.MethodPut, url, data)
+	if r != nil && r.StatusCode == http.StatusMethodNotAllowed {
+		r, err = doSubmitInventory(api, http.MethodPatch, url, data)
+	}
+
+	log.Debugf("Inventory update sent, response %v", r)
+
 	if err != nil {
-		return errors.Wrapf(err, "failed to prepare inventory submit request")
+		log.Error(err.Error())
+	}
+
+	return err
+}
+
+func doSubmitInventory(api ApiRequester, method, url string, data interface{}) (*http.Response, error) {
+	req, err := makeInventorySubmitRequest(method, url, data)
+	if err != nil {
+		return nil, errors.Wrapf(err, "failed to prepare inventory submit request")
 	}
 
 	r, err := api.Do(req)
 	if err != nil {
 		log.Error("Failed to submit inventory data: ", err)
-		return errors.Wrapf(err, "inventory submit failed")
+		return r, errors.Wrapf(err, "inventory submit failed")
 	}
 
 	defer r.Body.Close()
 
 	if r.StatusCode != http.StatusOK {
-		log.Errorf("Got unexpected HTTP status when submitting to inventory: %v", r.StatusCode)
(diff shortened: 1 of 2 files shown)


tranchitella (approved):

client/client_inventory.go
```diff
@@ -35,29 +35,37 @@ func NewInventory() InventorySubmitter {
...
 // Submit reports status information to the backend
 func (i *InventoryClient) Submit(api ApiRequester, url string, data interface{}) error {
-	req, err := makeInventorySubmitRequest(url, data)
-	if err != nil {
-		return errors.Wrapf(err, "failed to prepare inventory submit request")
-	}
+	// PATCH used to be the only method available in Mender Product 2.5, so
+	// fall back to that if PUT fails.
+	methods := []string{http.MethodPut, http.MethodPatch}
+	for _, method := range methods {
```
oleorhagen, line 41:
```suggestion
	for _, method := range []string{http.MethodPut, http.MethodPatch}
```

oleorhagen: Suggestion: :)
Don't really see the need for a reference to this array

client/client_inventory.go
```diff
@@ -35,29 +35,37 @@ func NewInventory() InventorySubmitter {
...
-		return errors.Wrapf(err, "inventory submit failed")
-	}
+		r, err := api.Do(req)
+		if err != nil {
+			log.Error("Failed to submit inventory data: ", err)
+			return errors.Wrapf(err, "inventory submit failed")
+		}
 
-	defer r.Body.Close()
+		r.Body.Close()
```
oleorhagen, line 53: This you cannot do actually. `NewAPIError` unmarshals the error message from the response body.

oleorhagen (changes requested):
Just one thing I see you need to fix :)

Otherwise looks good

client/client_inventory.go
kacf: Oh, good catch! I did it because `defer` doesn't scope correctly in loops, but maybe I'll just move the whole thing inside a function then.

kacf:
Updated! Moving to a separate function substantially changed the patch, so I decided to clean up some of the logging as well:

* Remove "inventory submit failed, bad status", we are essentially saying the same thing twice.
* Print the debug log always. Previously it was only printed on success, but it's more likely to be useful on failure.

oleorhagen:
> Updated! Moving to a separate function substantially changed the patch, so I decided to clean up some of the logging as well:
> 
>     * Remove "inventory submit failed, bad status", we are essentially saying the same thing twice.
Thanks!
> 
>     * Print the debug log always. Previously it was only printed on success, but it's more likely to be useful on failure.

haha, not going to complain on this point.

Now, to the code :)

client/client_inventory.go
```diff
@@ -35,29 +35,43 @@ func NewInventory() InventorySubmitter {
 
 // Submit reports status information to the backend
 func (i *InventoryClient) Submit(api ApiRequester, url string, data interface{}) error {
-	req, err := makeInventorySubmitRequest(url, data)
+	// PATCH used to be the only method available in Mender Product 2.5, so
+	// fall back to that if PUT fails.
+	r, err := doSubmitInventory(api, http.MethodPut, url, data)
+	if r != nil && r.StatusCode == http.StatusMethodNotAllowed {
+		r, err = doSubmitInventory(api, http.MethodPatch, url, data)
```
oleorhagen, line 42: Nit:
```suggestion
		_, err = doSubmitInventory(api, http.MethodPatch, url, data)
```
I don't see this used anywhere :eyes:

oleorhagen (approved):
Much better on the eyes too ^-^

client/client_inventory.go
kacf: I think this should stay. If more code is added after this, `r` might be assigned the wrong result if `_` is used in the second call.

PR #666, opened 2020-11-19
tests: Work around race condition when detecting bad certificate.

kacf:
If a bad certificate is detected, which is what we are testing, the
Golang request processing (http.Client.Do) will close the connection.
Sometimes, but not always, this will lead to a "connection closed"
error instead of a "bad certificate" error, the latter of which we are
looking for in the string. It appears that this is outside of our
control, since our server is not torn down until after the check has
been made.

Work around this by repeating the test multiple times using
assert.Eventually.

Changelog: None

### DIFF
--- a/app/mender_test.go
+++ b/app/mender_test.go
@@ -18,6 +18,7 @@ import (
 	"crypto/rand"
 	"crypto/tls"
 	"crypto/x509"
+	"fmt"
 	"io"
 	"io/ioutil"
 	"os"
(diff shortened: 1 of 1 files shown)


kacf:
Cancelling mender-qa pipeline. This only affects unit tests.

app/mender_test.go
```diff
@@ -1243,15 +1254,24 @@ func TestMutualTLSClientConnection(t *testing.T) {
...
+				err = mender.UploadLog(
+					&datastore.UpdateInfo{
+						ID: "foobar",
+					},
+					logs,
+				)
+				t2.failed = false
+				test.assertFunc(t2, err, srv.Log.Logs)
+				return !t2.failed
+			}, 10*time.Second, 500*time.Millisecond)
```
merlin-northern, line 1269: maybe we could define those timeouts somewhere above as vars or in the test case.

merlin-northern (approved):
lgtm

app/mender_test.go
kacf: Done!

oleorhagen (approved):

app/mender_test.go
merlin-northern: lovely!

PR #668, opened 2020-11-19
Add passphrase-file global option.

Styne13:
### Description

Add passphrase-file global option for decrypting encrypted private keys.

### DIFF
--- a/cli/cli.go
+++ b/cli/cli.go
@@ -179,7 +179,11 @@ func SetupCLI(args []string) error {
 					Destination: &runOptions.rebootExitCode,
 					Usage: "Return exit code 4 if a manual reboot " +
 						"is required after the Artifact installation.",
-				},
+				}, &cli.StringFlag{
+					Name:        "passphrase-file",
+					Usage:       "Passphrase file for decrypting an encrypted private key. '-' loads passphrase from stdin.",
+					Value:       "",
+					Destination: &runOptions.keyPassphrase},
 			},
 		},
 		{
@@ -387,6 +391,11 @@ func SetupCLI(args []string) error {
 			Name:        "skipverify",
 			Usage:       "Skip certificate verification.",
 			Destination: &runOptions.Config.NoVerify},
+		&cli.StringFlag{
+			Name:        "passphrase-file",
+			Usage:       "Passphrase file for decrypting an encrypted private key. '-' loads passphrase from stdin.",
+			Value:       "",
+			Destination: &runOptions.keyPassphrase},
 	}
 	cli.HelpPrinter = upgradeHelpPrinter(cli.HelpPrinter)
 	cli.VersionPrinter = func(c *cli.Context) {
--- a/store/keystore.go
+++ b/store/keystore.go
@@ -35,11 +35,12 @@ var (
 )
 
 type Keystore struct {
-	store     Store
-	private   openssl.PrivateKey
-	keyName   string
-	sslEngine string
-	staticKey bool
+	store         Store
+	private       openssl.PrivateKey
+	keyName       string
+	keyPassphrase string
+	sslEngine     string
+	staticKey     bool
 }
 
 func (k *Keystore) GetStore() Store {
@@ -50,16 +51,17 @@ func (k *Keystore) GetKeyName() string {
 	return k.keyName
 }
 
-func NewKeystore(store Store, name string, sslEngine string, static bool) *Keystore {
+func NewKeystore(store Store, name string, sslEngine string, static bool, passphrase string) *Keystore {
 	if store == nil {
 		return nil
 	}
 
 	return &Keystore{
-		store:     store,
-		keyName:   name,
-		sslEngine: sslEngine,
-		staticKey: static,
+		store:         store,
+		keyName:       name,
+		sslEngine:     sslEngine,
+		staticKey:     static,
+		keyPassphrase: passphrase,
 	}
 }
 
@@ -89,7 +91,13 @@ func (k *Keystore) Load() error {
 	}
 	defer inf.Close()
 
-	k.private, err = loadFromPem(inf)
+	passphrase, err := loadPassphrase(k.keyPassphrase)
+	if err != nil {
+		log.Errorf("Failed to load passphrase-file parameter: %s", err)
+		return err
+	}
+
+	k.private, err = loadFromPem(inf, passphrase)
 	if err != nil {
 		log.Errorf("Failed to load key: %s", err)
 		return err
@@ -174,13 +182,17 @@ func IsStaticKey(err error) bool {
 	return err == errStaticKey
 }
 
-func loadFromPem(in io.Reader) (openssl.PrivateKey, error) {
+func loadFromPem(in io.Reader, keyPassphrase string) (openssl.PrivateKey, error) {
 	data, err := ioutil.ReadAll(in)
+	var key openssl.PrivateKey
 	if err != nil {
 		return nil, err
 	}
-
-	key, err := openssl.LoadPrivateKeyFromPEM(data)
+	if keyPassphrase != "" {
+		key, err = openssl.LoadPrivateKeyFromPEMWithPassword(data, keyPassphrase)
+	} else {
+		key, err = openssl.LoadPrivateKeyFromPEM(data)
+	}
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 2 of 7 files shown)


oleorhagen:
Hi @Styne13 and welcome to the Mender project :raised_hands: 

Just let me know when you would like this reviewed :)

Styne13:
Hi @oleorhagen,

thanks! 👍 

Had this feature running in an earlyer version of mender (2.2.0) and would like to have it in the new client once it is updated.

The change shall be complete but I have to get the tests running again. Maybe you can assist with this.

cli/cli.go
```diff
@@ -179,7 +179,11 @@ func SetupCLI(args []string) error {
 					Destination: &runOptions.rebootExitCode,
 					Usage: "Return exit code 4 if a manual reboot " +
 						"is required after the Artifact installation.",
-				},
+				},&cli.StringFlag{
+			Name:        "key-passphrase",
+			Usage:       "Passphrase for decrypting an encrypted private key",
+			Value:       "",
+			Destination: &runOptions.keyPassphrase},
```
kacf, line 186: Note that passing in a passphrase on the command line is not secure, it can be seen by all users on the system through the process table. It should either be passed in via a file or via stdin. May I suggest a filename argument, where `-` means opening stdin?

kacf (changes requested):

oleorhagen:
> The change shall be complete but I have to get the tests running again. Maybe you can assist with this.

This I can, just follow up on the change request first :)

store/keystore_test.go
```diff
@@ -205,9 +246,27 @@ func TestSignED25519(t *testing.T) {
...
+func TestKeystoreLoadPemWithSecretFail(t *testing.T) {
+        // this should fail
+        nk, err := loadFromPem(bytes.NewBufferString(badPrivKey), "verysecure")
+        assert.Nil(t, nk)
+        assert.Error(t, err)
+}
+func TestKeystoreLoadPemWithWrongSecret(t *testing.T) {
+        nk, err := loadFromPem(bytes.NewBufferString(encryptedPrivKey), "wrongsecret")
+        assert.Nil(t, nk)
+        assert.Error(t, err)
```
kacf, line 271: Good tests! :+1:

store/keystore.go
```diff
@@ -199,3 +208,41 @@ func saveToPem(out io.Writer, key openssl.PrivateKey) error {
...
+				var output []rune
+				for {
+					input, _, err := reader.ReadRune()
+					if err != nil && err == io.EOF {
+						break
+					}
+					output = append(output, input)
+				}
+				passphrase = string(output[:len(output)-1])
+			}
```
kacf, line 233: Why do we need to do it differently here? Can't we just open the right source (stdin or file) and then use `ioutil.ReadAll` for both?

store/keystore.go
```diff
@@ -199,3 +208,41 @@ func saveToPem(out io.Writer, key openssl.PrivateKey) error {
...
+		} else {
+			log.Debugf("Read passphrase from file.")
+			dat, err := ioutil.ReadFile(passphrase_file)
+			if err != nil {
+				log.Fatal(err)
+				return "", err
+		    }
+			passphrase = string(dat[:len(dat)-1])
```
kacf, line 241: I assume you did this because of the trailing newline many files have. However, I believe it's better to do:

```
passphrase = strings.TrimRight(string(dat), "\n")
```

This will work also if the file does not have a trailing newline.

kacf (changes requested):
Nice addition!

In addition to the comments below, you will also have to address:

* Our CI checks for correct code format. Use `go fmt ./...` in your repository before committing, and it will adjust the codestyle correctly.
* The commits should be squashed to one commit, and you will need to add a few things to the message:
  * A changelog tag which describes to users what has been added. Probably in your case you can just use the commit title, so add `Changelog: Title`, and it will use the text from there.
  * You will need to sign off your commit using the `-s` argument when running `git commit` (legal requirement).

store/keystore_test.go
Styne13: Thx :)

store/keystore.go
Styne13: Checked that and your version also works. Will change this in the next commit.

store/keystore.go
Styne13: I'm not sure about that.
This is my first software in go and I only found this version.
Will try to unite this.

store/keystore.go
```diff
@@ -199,3 +207,24 @@ func saveToPem(out io.Writer, key openssl.PrivateKey) error {
...
+	var dat []byte
+	if passphrase_file != "" {
+		if passphrase_file == "-" {
+			log.Debugf("Read passphrase from stdin.")
+			fi, err = os.Open("/dev/stdin")
+		} else {
+			log.Debugf("Read passphrase from file.")
+			fi, err = os.Open(passphrase_file)
+		}
+		dat, err = ioutil.ReadAll(fi)
```
kacf, line 222: Several missing error checks. Each `err` assignment should check for `err != nil` and return error.

store/keystore.go
```diff
@@ -89,7 +91,9 @@ func (k *Keystore) Load() error {
 	}
 	defer inf.Close()
 
-	k.private, err = loadFromPem(inf)
+	passphrase, err := loadPassphrase(k.keyPassphrase)
```
kacf, line 94: Missing error check.

kacf (changes requested):
Thanks @Styne13! It's starting to look good, the only problem I can spot is a a few missing error checks. If you fix those we can start a full test suite run afterwards.

store/keystore.go
```diff
@@ -199,3 +211,36 @@ func saveToPem(out io.Writer, key openssl.PrivateKey) error {
...
+				return "", err
+			}
+		}
+		dat, err = ioutil.ReadAll(fi)
+		if err != nil {
+			log.Errorf("Failed to read passphrase: %s", err)
+			return "", err
+		}
+		passphrase = strings.TrimRight(string(dat), "\n")
+		fi.Close()
```
kacf, line 240: With the new error handling, this should be moved. Instead of calling it here, call it right before `ReadAll`, like this:

```go
defer fi.Close()
dat, err = ioutil.ReadAll(fi)
```

Then it will be closed when the function exits, regardless of where it exits. This way it will also be called when exiting through the error return.

store/keystore.go
```diff
@@ -199,3 +211,36 @@ func saveToPem(out io.Writer, key openssl.PrivateKey) error {
 
 	return err
 }
+
+func loadPassphrase(passphrase_file string) (passphrase string, err error) {
```
kacf, line 215: I see now that the test suite is complaining about code coverage. Could you also add a unit test for this function? It's not well covered at the moment.

The stdin case is probably going to be tricky to test, so we can skip that, but at least the "from file" codepath should be tested, with some dummy file created in the test code.

kacf (changes requested):
Thanks! Some new things were revealed in the test results, see below.

store/keystore.go
Styne13: Also saw the coverage error and already working on this.

kacf (approved):
Excellent! Very nice job!

kacf:
@Styne13: I was demoing this during our bi-weekly "new contributions" session, and something popped up, which may or may not be a bug: If the passphrase for the key file is wrong, either because it is typed in wrong, or because the passphrase file has the wrong content, then the client deletes the key and generates a new one. This means that you lose the private key unless you have a backup of it somewhere else. If you are using Mender in a way where the passphrase is typed in, this is clearly a bug, since an innocent keyboard mistake can lead to the key being deleted.

However, from an automation perspective, when the passphrase is obtained automatically, it is less clear whether this is a bug or not. If the passphrase has somehow gotten destroyed, and the client doesn't generate a new key, then Mender is essentially dead, and will never connect to the server again. At least by generating a new key it has a chance to recover, although obviously in a less secure fashion, by accepting the new, passphrase-free key on the server.

This feature is very contributor-driven, we don't use it ourselves or promote it, so I'm interested in your opinion about what would be the right behavior. How are you using it?

If you have time to look into the code and fix it, that would also be great.

Styne13:
@kacf : The usecase for this feature in our case is the following:
Initial a self-signed certificate is generated, the private key is encrypted by a secred which is related to a hardware element on the device (read only, not modifiable) and the files are stored in a read-only partition.
On startup of the device a symlink to the private key is generated in /etc/mender.

We use a modified systemd startup routine to read out the secret from hardware element and hand it over to mender.
Normaly this secret should not be wrong because it is not modifyable in the hardware element.

But I will also keep in mind to do a fix, e.g log message and abort in case a wrong secret is handed over to mender, if I have time.

PR #669, opened 2020-11-19
MEN-4170: Shutdown the client on SIGTERM

oleorhagen:
A signal handler for SIGTERM is added to the daemon run, but this is only able
to interrupt the client on the state-machine changing state. Thus operations
taking a long time, such as writing an update, are never in fact interrupted on
SIGTERM. Also, the Reboot state is stopped, due to this addition.

This can cause long shutdown times on clients in the process of writing an
update, as the client will not terminate until it has either switched state, or
worse, if it is already in the Reboot state, a SIGKILL signal needs to be sent
from the outside in order to shut it down.

This fix simply exits with Exit code 0 upon a received SIGTERM. The client is
already safe from interruptions, so this call is fine.

Changelog: Exit with code 0 on a received SIGTERM signal.

### DIFF
--- a/cli/commands.go
+++ b/cli/commands.go
@@ -59,6 +59,7 @@ var out io.Writer = os.Stdout
 
 var (
 	errArtifactNameEmpty = errors.New("The Artifact name is empty. Please set a valid name for the Artifact!")
+	ErrSIGTERM           = errors.New("Daemon terminated with SIGTERM")
 )
 
 func commonInit(config *conf.MenderConfig, opts *runOptionsType) (*app.MenderPieces, error) {
@@ -266,6 +267,7 @@ func PrintProvides(device *dev.DeviceManager) error {
 
 func runDaemon(d *app.MenderDaemon) error {
 	// Handle user forcing update check.
+	daemonExit := make(chan error)
 	go func() {
 		c := make(chan os.Signal, 2)
 		signal.Notify(c, syscall.SIGUSR1) // SIGUSR1 forces an update check.
@@ -282,13 +284,16 @@ func runDaemon(d *app.MenderDaemon) error {
 				log.Debug("SIGUSR2 signal received.")
 				d.ForceToState <- app.States.InventoryUpdate
 			} else if s == syscall.SIGTERM {
-				d.StopDaemon()
+				daemonExit <- ErrSIGTERM
 			}
 			d.Sctx.WakeupChan <- true
 			log.Debug("Sent wake up!")
 		}
 	}()
-	return d.Run()
+	go func() {
+		daemonExit <- d.Run()
+	}()
+	return <-daemonExit
 }
 
 // updateCheck sends a SIGUSR{1,2} signal to the running mender daemon.
--- a/main.go
+++ b/main.go
@@ -30,6 +30,8 @@ func doMain() int {
 		} else if err == installer.ErrorNothingToCommit {
 			log.Warnln(err.Error())
 			return 2
+		} else if err == cli.ErrSIGTERM {
+			return 0
 		} else {
 			log.Errorln(err.Error())
 			return 1
--- a/main.go
+++ b/main.go
@@ -25,14 +25,16 @@ import (
 
 func doMain() int {
 	if err := cli.SetupCLI(os.Args); err != nil {
-		if err == app.ErrorManualRebootRequired {
+		switch err {
+		case cli.ErrSIGTERM:
+			log.Infoln(err.Error())
+			return 0
+		case app.ErrorManualRebootRequired:
 			return 4
-		} else if err == installer.ErrorNothingToCommit {
+		case installer.ErrorNothingToCommit:
 			log.Warnln(err.Error())
 			return 2
-		} else if err == cli.ErrSIGTERM {
-			return 0
-		} else {
+		default:
 			log.Errorln(err.Error())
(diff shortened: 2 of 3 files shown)


oleorhagen:
I'll let you decide how/if you want this tested more extensively

cli/commands.go
```diff
@@ -283,6 +284,10 @@ func runDaemon(d *app.MenderDaemon) error {
 				d.ForceToState <- app.States.InventoryUpdate
 			} else if s == syscall.SIGTERM {
 				d.StopDaemon()
+				time.AfterFunc(time.Second * 5, func() {
```
merlin-northern, line 287: this at least should be a global variable or const and maybe even consider configuration option?
have we considered sending INT, HUP, STOP, or else to interrupt the operations?
does it mainly concern writing an update?
is it possible that we brick if we os.Exit in the middle of something?
in my book it is a bad practice to do anything else than set an atomic flag in the signal handler.

kacf: > this at least should be a global variable or const and maybe even consider configuration option?

Const I can agree with, but I think this is too implementation specific to be a configuration option.

> have we considered sending INT, HUP, STOP, or else to interrupt the operations?

The commit message is actually slightly misleading: The primary problem is that this happens while rebooting, and there is no operation going on then, the client is just waiting to be killed from the outside. But since it is waiting inside the reboot state, it is not returning to the state loop where it would exit. Even if we could fix this, we can't fix it for Update Modules, because we don't know what they do inside their scripts.

The hanging *also* happens during Download, as the commit message mentions, but this is not the primary problem.

> does it mainly concern writing an update?
> is it possible that we brick if we os.Exit in the middle of something?

The client is prepared to handle sudden exits (actually, sudden reboots), so exiting in the middle of something is no problem.

> in my book it is a bad practice to do anything else than set an atomic flag in the signal handler.

This isn't a signal handler in the same way as in C, so we don't need to worry about all the atomic stuff.

kacf: I'm unsure about the timer though. It's a good idea in principle, and for other daemons it would clearly have been the best way. But my concern is that this will happen on absolutely all updates. In fact, we are *guaranteed* to hit this code in a normal update because the client "hangs" inside the reboot state until it is killed from the outside. So essentially the "clean path" will almost never be taken, and I'm wondering if we're just delaying things for no reason. I think my vote would be in favor of no delay, just straight exit, but I'm open to both solutions. What do you think @merlin-northern and @oleorhagen?

merlin-northern: ok I can understand that, but when I do `systemctl restart mender-client` I would expect it to gracefully handle SIGTERM and exit if possible. I am just not a fan of an idea that we just do exit on SIGTERM regardless of anything. is it possible here to detect somehow that we have to call exit?

kacf: But which part about calling `os.Exit()` is not graceful? As I said, the client is prepared to handle this scenario (we only use ACID transactions in the database, it reloads state from where it left off, etc).

merlin-northern: probably only in the part in my head :) just my feeling. if you are ok with os.Exit then I will not argue (too much).

kacf: In general it's a good feeling to have, and I agree, but this particular daemon is essentially *made for* sudden exits!

oleorhagen:
Hmm, @kacf worrying that the coverage has gone down so much for the acceptance test coverage. Looking at it I can't really make it out. Like `client_auth.go` has no coverage anymore basically. How can this be?

kacf:
Could it be that the coverage collector depends on some `defer` function running? These probably won't run if `os.Exit()` is called, so it's possible it's "lost" somehow. This might be tricky to get around by just calling `os.Exit()`.

Maybe we have to use your first proposal then, with a delay which allows graceful exit if possible. As long as it's a short delay, I think it's a worthwhile compromise to keep the coverage reports sane.

oleorhagen:
> Could it be that the coverage collector depends on some `defer` function running? These probably won't run if `os.Exit()` is called, so it's possible it's "lost" somehow. This might be tricky to get around by just calling `os.Exit()`.
> 
Yes, you are absolutely right of course! Thank you! I was just not in the mood, and decided coveralls is stupid, and I don't like using my brain this very day.

> Maybe we have to use your first proposal then, with a delay which allows graceful exit if possible. As long as it's a short delay, I think it's a worthwhile compromise to keep the coverage reports sane.

I will think about this some little more.

cli/commands.go
```diff
@@ -282,13 +284,16 @@ func runDaemon(d *app.MenderDaemon) error {
...
 			}
 			d.Sctx.WakeupChan <- true
 			log.Debug("Sent wake up!")
 		}
 	}()
-	return d.Run()
+	go func() {
+		daemonExit <- d.Run()
+	}()
+	return <-daemonExit
```
kacf, line 296: Clever!

kacf (approved):

oleorhagen:
I am going to go ahead and merge this despite the quite drastic fall in coverage:
Reason is I do not trust the numbers. Example:
[latest master coverage (x86-64-uefi-grub) ](https://coveralls.io/jobs/71237299)
[pr coverage (-7.8 according to coveralls)](https://coveralls.io/jobs/71475132)

If you put them side by side and look at them, they are essentially identical (as expected).

PR #671, opened 2020-11-23
MEN-4169: Remove auth token caching + several other fixes.

kacf:
Best reviewed commit by commit.


app/auth.go
```diff
@@ -295,15 +322,23 @@ mainloop:
 }
 
 // Stops the running MenderAuthManager. Must not be called in the same go
-// routine as Run().
+// routine as run(). This is idempotent, it is safe to call on a stopped
```
oleorhagen, line 325: Nit:
`This is idempotent, _and_ is safe to call on a stopped service`. right?

app/auth_test.go
```diff
@@ -86,26 +88,26 @@ func TestAuthManager(t *testing.T) {
 			Cmdr: cmdr,
 		},
 		KeyStore: store.NewKeystore(ms, "key", "", false),
-	})
+	}).(*MenderAuthManager)
```
oleorhagen, line 91: Question:
Why not return the struct from `NewAuthManager`, as opposed to an interface. It returns the same type anyways, and the interface is still implemented. If you need the interface, just receive it with an interface variable.

client/client.go
```diff
@@ -217,15 +217,28 @@ func (ar *ApiRequest) Do(req *http.Request) (*http.Response, error) {
 		req.URL.Host = host
 		req.Host = host
 
+		log.Debugf("Connecting to server %s", host)
+
+		var body io.ReadCloser
+		if req.GetBody != nil {
```
oleorhagen, line 223: Question:
It's late, and I'm leaving now, and was not fully able to grasp this. But could another workaround have been buffering this outside of the loop in a bytes buffer, and resetting it on every invocation?

oleorhagen (approved):
Nice cleanups!

Sorry it took this long to go through, I didn't mean to keep you waiting. Just slight `org/lisp` mishap on my end!

First three are real nice. But I have to admit that when I saw the `runtime` hack, my stomach dropped, and I was ready to kill.

However, you've done it nicely, and I won't be difficult, even though on principle I don't think this is the right way of doing it, but you know that too.

All the other stuff are just minor questions and musings, and I was more asking for you to explain things to me, than have anything changed!

Very well done, sir :medal_military:

app/mender.go
```diff
@@ -179,15 +179,6 @@ func (m *Mender) IsAuthorized() bool {
 }
 
 func (m *Mender) Authorize() menderError {
```
tranchitella, line 176: I'm not 100% sure you remember, but in my first PR we've had ForceAuthorize exactly for this reason... your implementation is way more robust, though, and it makes a lot of sense to always fetch a new JWT token when calling this function. Well done.

tranchitella (approved):
Great work, LGTM

app/auth.go
kacf: Being idempotent means that it is safe to call on a stopped service, so I think the original sentence is better. They are not two different properties, just the same property listed, and then explained.

app/auth_test.go
kacf: I can try this, it might be simpler. Returning an interface was preexisting, so I didn't change it, but it might be better style in this case.

client/client.go
kacf: Possibly, but we would be "swimming against the current", so to speak. The Golang http library states that it may also call `GetBody` (for example when following redirects), so it has to be kept working. So we might as well call it as well, to be consistent, and to test it.

app/auth.go
oleorhagen: No, being idempotent means that it doesn't change when applied multiple times no? It's not specific to a state of the environment the function is called on.

oleorhagen: But it's not important.

kacf: Hehe true, I didn't mean that this is the definition of idempotent, but that this is the effect of applying it in this particular case. But yes, not really important.

lluiscampos (approved):
The fixes look good to me. Thanks for very descriptive commit messages :top:

app/auth_test.go
kacf: Turns out this is tricky because of the `WithDBus` function. It returns `AuthManager`, which means it can no longer be put in the variable which was declared to hold `*MenderAuthManager` (default when using `:=`). And we also cannot change `WithDBus` to return `*MenderAuthManager`, because then the struct would be part of the interface. The only workaround is to declare the variable explicitly to hold `AuthManager`, using `var`, but I think this makes the API worse, and it's better that this slight casting ugliness is kept in the tests only.

It was worth a shot!

app/mender.go
kacf: Ok, it's possible I did not fully grasp the consequences at the time. I have also learned a lot about this as I have worked on it.

kacf:
Thanks for reviews, I believe I have found the reason for the test failures, and will launch what is hopefully a successful build.

app/auth_test.go
kacf: So @oleorhagen convinced me that it is possible to do this, only the implementation's return type needs to be changed, the interface can still use an interface return type. A nice feature I didn't know about which solves the problems I encountered (thanks for sticking with it @oleorhagen!)

However, due to the number of people depending on this PR right now, I'm going to merge it as is, since it passed, and will followup with a new PR to fix it.

kacf:
As far as I can tell, the failed coverage is bogus, since the total is still the same. Merging.

PR #673, opened 2020-11-26
Use getter/setter-pattern instead of return pattern to avoid casting.

kacf:
This means we can get rid of the returned interface in the function
definition, which again allows us return the type in the constructor,
instead of the interface. Thus we can test this without a lot of
casting.

Changelog: None


kacf:
@oleorhagen: As requested!

oleorhagen:
:100:

tranchitella (approved):

PR #674, opened 2020-11-26
Emit signals only on valid AuthToken and with valid AuthToken

merlin-northern:
ChangeLog:none

### DIFF
--- a/app/auth.go
+++ b/app/auth.go
@@ -356,17 +356,23 @@ func (m *menderAuthManagerService) broadcast(message AuthManagerResponse) {
 		}
 	}
 	// emit signal on dbus, if available
-	if m.dbus != nil {
+	if m.dbus != nil && message.AuthToken != noAuthToken {
 		m.dbus.EmitSignal(m.dbusConn, "", AuthManagerDBusPath,
 			AuthManagerDBusInterfaceName, AuthManagerDBusSignalValidJwtTokenAvailable)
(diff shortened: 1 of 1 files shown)


app/auth.go
```diff
@@ -356,17 +356,23 @@ func (m *menderAuthManagerService) broadcast(message AuthManagerResponse) {
...
 func (m *menderAuthManagerService) broadcastAuthTokenAvailable() {
 	authToken, err := m.authToken()
+	log.Debugf("broadcastAuthTokenAvailable: running after m.authToken:%s err:%v", authToken, err)
 	if err == nil && authToken != noAuthToken {
-		m.broadcast(AuthManagerResponse{Event: EventAuthTokenAvailable})
+		m.broadcast(AuthManagerResponse{
+			AuthToken: authToken,
+			Event:     EventAuthTokenAvailable,
+			Error:     nil,
+		})
```
merlin-northern, line 375: @kacf @tranchitella @lluiscampos so I am dropping this part, and keep the rest, right? (after the discussion a couple of minutes ago
)

kacf: If we do [MEN-4191](https://tracker.mender.io/browse/MEN-4191), then the original is correct, we should always emit the signal. So I think this PR is not needed anymore.

tranchitella:
Superseded by https://github.com/mendersoftware/mender/pull/678

PR #679, opened 2020-12-06
QA-238: check_commits: move from unit tests to template

tranchitella:
Changelog: none


tranchitella:
Cancel the mender-qa pipeline, as it doesn't bring any value

kacf (approved):

license_test.go
```diff
@@ -1,24 +0,0 @@
...
-//    limitations under the License.
-
-package main
-
-import "testing"
-import "os"
-import mt "github.com/mendersoftware/mendertesting"
-
-func TestLicenses(t *testing.T) {
-	os.Setenv("CHKSUM_FILE", "vendor/LIC_FILES_CHKSUM.sha256")
```
lluiscampos, line 22: I think this env variable was completely ignored by the script... The file does not exist anyway.

lluiscampos (approved):

PR #687, opened 2020-12-11
MEN-4245: Fix error parsing response for getting tenant token on setup

alfrunes:
The problem is that the endpoint now returns an embedded JSON object
with info about the api limits, which fails to unmarshal into an object of type
map[string]string.

### DIFF
--- a/cli/setup.go
+++ b/cli/setup.go
@@ -484,6 +484,10 @@ func (opts *setupOptionsType) askServerCert(ctx *cli.Context,
 
 func (opts *setupOptionsType) getTenantToken(
 	client *http.Client, userToken []byte) error {
+	type tenantTokenResponse struct {
+		Token string `json:"tenant_token"`
+	}
+
 	tokReq, err := http.NewRequest(
 		"GET",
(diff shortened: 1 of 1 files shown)


cli/setup.go
```diff
@@ -509,13 +509,13 @@ func (opts *setupOptionsType) getTenantToken(
...
 	}
-	dataJson := make(map[string]string)
+	dataJson := make(map[string]interface{})
 	err = json.Unmarshal(data, &dataJson)
 	if err != nil {
 		return errors.Wrap(err,
 			"Error parsing JSON response.")
 	}
-	opts.tenantToken = dataJson["tenant_token"]
+	opts.tenantToken = dataJson["tenant_token"].(string)
```
oleorhagen, line 518: Question:
For future proofing this, perhaps we should check the typecast here, so that changing the format on the server, will not crash the client?

alfrunes: You're right, this is too lazy. I updated the PR with an implicit type check in the Unmarshal call.

oleorhagen (approved):

PR #690, opened 2020-12-16
probeSeparator: clean-up return values

lluiscampos:
Namely, return empty strings on errors. This is not very important, as
the user will discard the return on error != nil, but helps humans to
understand the logic ;)

### DIFF
--- a/installer/bootenv.go
+++ b/installer/bootenv.go
@@ -98,8 +98,8 @@ func (e *UBootEnv) ReadEnv(names ...string) (BootVars, error) {
 	return vars, err
 }
 
-// ProbeSeparator tries writing 'mender_uboot_separator' to the U-Boot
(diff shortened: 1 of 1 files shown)


kacf (approved):

installer/bootenv.go
```diff
@@ -98,8 +98,8 @@ func (e *UBootEnv) ReadEnv(names ...string) (BootVars, error) {
 	return vars, err
 }
 
-// ProbeSeparator tries writing 'mender_uboot_separator' to the U-Boot
-// environment using whitespace as a separator. If this write is succesful it
+// probeSeparator tries writing 'mender_uboot_separator' to the U-Boot
+// environment using whitespace as a separator. If this write is successful it
```
oleorhagen, line 102: ouch

oleorhagen (approved):
Thanks

PR #696, opened 2021-01-13
MEN-4294: Fix FetchJwtToken call stalling the whole DBus API. 

kacf:
The DBus API routine is single threaded, and hence a call to
FetchJwtToken can stall the whole API if this takes a long time. For
slow connections, this is not uncommon.

Fix by moving the FetchJwtToken workload into its own Go routine.

Changelog: None


kacf:
Run together with https://github.com/mendersoftware/meta-mender/pull/1268.

tranchitella (approved):
Nice catch!

lluiscampos (approved):

PR #703, opened 2021-01-21
Improvements to Docker image

lluiscampos:
* [Dockerfile.daemon] Update build image to Ubuntu 20.04
So that we have golang 1.13
Note that we keep the final image on Ubuntu 18.04 to avoid a very
strange issue with `COPY --from build` which seems to make /
non accessible. We'll follow up independently.

* [pipeline] Add smoke test to build Docker image
So that we can catch regressions faster.

* [tests/build-docker] Require sh instead of bash

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -46,6 +46,15 @@ test:
     paths:
       - unit-coverage.tar
 
+test:docker:
+  image: docker
+  needs: []
+  services:
+    - docker:19.03.5-dind
+  stage: test
+  script:
+    - ./tests/build-docker
+
 publish:tests:
   stage: publish
   image: golang:1.14-alpine3.11
--- a/tests/Dockerfile.daemon
+++ b/tests/Dockerfile.daemon
@@ -1,7 +1,8 @@
 # Creates a container which acts as a bare bones non-VM based Mender
 # installation, for use in tests.
-FROM ubuntu:18.04 AS build
+FROM ubuntu:20.04 AS build
 
+ENV DEBIAN_FRONTEND=noninteractive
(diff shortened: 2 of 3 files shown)


lluiscampos:
The motivation is to prevent failures like #700, where the error is only reported via `mender-qa` pipeline, on `build_servers` job

mzedel (approved):
The changes on their own look ok, but that a faulty client build is surfaced in a `build_servers` job on the other hand 🧐...

considering that this client is used in the tests - could we not add a publishing step to pick that client up or at least pull before building it to use layer caching?

kacf (approved):

kacf:
> The changes on their own look ok, but that a faulty client build is surfaced in a `build_servers` job on the other hand monocle_face...

I think this is a relic from when it was all executed as a huge script in Jenkins, and related to how that script was organized. I think it could be moved to `build_client` with no major problems.

lluiscampos:
> > The changes on their own look ok, but that a faulty client build is surfaced in a `build_servers` job on the other hand monocle_face...
> 
> I think this is a relic from when it was all executed as a huge script in Jenkins, and related to how that script was organized. 

Yes.

>I think it could be moved to `build_client` with no major problems.

We could, and it will also help for this custom pipeline that publishes the client *images*, which actually need to run both `build_client` and `build_servers` for this specific reason :) I'll follow-up independently, though.

lluiscampos:
I have some errors in the pipes, so I'll force-push shortly.

lluiscampos:
I partially reverted the image upgrade, keeping only the builder on 20.04. For a reason I don't understand, having the second image on 20.04 triggers a `no such file or directory` error on `ls /`.

PR #706, opened 2021-02-02
Change provider in inventory-geo script to ipinfo.io

alfrunes:
The current script fails due to expired certificate from
'ipvigilante.com'.
There are some differences in the formatting of the
information received from the new API:
 - Country now holds the shortened landcode (e.g. US for the
   United States)
 - There's no continent information, instead I replaced it with
   geo-timezone.

ipinfo.io only supports JSON encoding in API so I had to
get creative parsing the payload using tools compatible
with Busybox. (Tested using Alpine docker image)

changelog: title

### DIFF
--- a/support/mender-inventory-geo
+++ b/support/mender-inventory-geo
@@ -26,9 +26,14 @@ err() {
  exit $rc
 }
 
-# Timeout value for connections to ipvigilante.
-# This value is in seconds
-TIMEOUT=10
+TIMEOUT=10 # Request timeout
+TEMPFILE="/tmp/mender/inventory-geo"
+
+# the names of the attributes holding the localization data can be configured here
+ATTR_NAME_IP="geo-ip"
+ATTR_NAME_CITY="geo-city"
+ATTR_NAME_COUNTRY="geo-country"
+ATTR_NAME_TIMEZONE="geo-timezone"
 
 ipinfo=$(wget --timeout=${TIMEOUT} -q -O /dev/stdout \
               --header 'Accept: application/json' \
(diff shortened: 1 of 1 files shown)


support/mender-inventory-geo
```diff
@@ -11,17 +11,12 @@
 # key.
 #
 # $ ./mender-inventory-geo
-# country=Poland
-# city=Krakow
-# latitude=50.05750
-# longitude=19.98020
-# ipv4=127.0.0.1
+# geo-ip=8.8.8.8
```
oleorhagen, line 14: :+1:

support/mender-inventory-geo
```diff
@@ -35,27 +30,33 @@ err() {
...
-[ -n "${geo}" ] || err 3 "Unable to get the geolocalization data from ipvigilante.com"
+ipinfo=$(wget --timeout=${TIMEOUT} -q -O /dev/stdout \
+              --header 'Accept: application/json' \
+              https://ipinfo.io)
+if [ $? != 0 ] || [ -z "${ipinfo}" ]; then
+    err 2 "Unable to get IP info from ipinfo.io"
+fi
+# Convert JSON (Object{string -> string}) to key=value format.
+ipinfo=$(echo "${ipinfo}" | \
+             tr -d "\n" | \
```
oleorhagen, line 41: This one I had to stare at for a while :p 
And concluded that it is just fine, still I couldn't help myself, so I tried a little with my favourite `awk`, which gave me a PoC:
```
cat test.json | awk -F : '
length($1) > 1 {gsub("\"|,| ", "", $1); \
gsub("\"|,| ", "", $2); printf ("%s=%s\n", $1, $2)}'

Which gives me out:
```
ip=195.159.235.214
hostname=195-159-235-214.customer.powertech.no
city=Oslo
region=Oslo
country=NO
loc=59.912710.7461
org=AS2116GLOBALCONNECTAS
postal=0001
timezone=Europe/Oslo
readme=https
```
Which is also not robust, if you look at the `readme`. But still fun!

Tldr; this is just fine, I just couldn't help myself.
```

support/mender-inventory-geo
```diff
@@ -35,27 +30,33 @@ err() {
...
-
+ATTR_NAME_COUNTRY="geo-country"
+ATTR_NAME_TIMEZONE="geo-timezone"
+
+echo "${ipinfo}" | awk -F'=' -v attr_ip="${ATTR_NAME_IP}" \
+                       -v attr_city="${ATTR_NAME_CITY}" \
+                       -v attr_country="${ATTR_NAME_COUNTRY}" \
+                       -v attr_zone="${ATTR_NAME_TIMEZONE}" '
+$1 ~ /^ip/{printf "%s=%s\n", attr_ip, $2};
+$1 ~ /^city/{printf "%s=%s\n", attr_city, $2};
```
oleorhagen, line 59: Suggestion:
A little bit of air between the `pattern` and the `expression` would be nice actually, so maybe something like:
```
echo "${ipinfo}" | awk -F'=' -v attr_ip="${ATTR_NAME_IP}" \
                       -v attr_city="${ATTR_NAME_CITY}" \
                       -v attr_country="${ATTR_NAME_COUNTRY}" \
                       -v attr_zone="${ATTR_NAME_TIMEZONE}" '

$1 ~ /^ip/       {printf "%s=%s\n", attr_ip, $2}
$1 ~ /^city/     {printf "%s=%s\n", attr_city, $2}
$1 ~ /^country/  {printf "%s=%s\n", attr_country, $2}
$1 ~ /^timezone/ {printf "%s=%s\n", attr_zone, $2}
'
```

oleorhagen (approved):
All musings, and brain farts. Feel free to merge as is!

oleorhagen:
Oh, nearly forgot:

`s/countinent/continent/` in the commit msgs

alfrunes:
@oleorhagen I also realized that our demo intervals will hit the monthly rate limits for this API, so I think it's only fair to run this query once per power cycle so I'm caching the results in a tempfile.

support/mender-inventory-geo
```diff
@@ -35,27 +30,47 @@ err() {
...
+fi
 
-[ -n "${ip}" ] || err 2 "Unable to get the IP address from ifconfig.io"
+if [ ! -f /tmp/mender/inventory-geo ]; then
+    # Fetch and cache geo location data
 
-# get the information, see https://www.ipvigilante.com/api-developer-docs/
-geo=`wget --timeout=${TIMEOUT} -qO /dev/stdout https://ipvigilante.com/csv/${ip} 2>/dev/null`
+    mkdir -p /tmp/mender
+    if [ $? != 0 ]; then
```
oleorhagen, line 44: Suggestion:
```
if mkdir -p /tmp/mender ...
```

support/mender-inventory-geo
```diff
@@ -35,27 +30,47 @@ err() {
...
-     --header 'Accept: */*' ifconfig.io | tail -1`
+ipinfo=$(wget --timeout=${TIMEOUT} -q -O /dev/stdout \
+              --header 'Accept: application/json' \
+              https://ipinfo.io)
+if [ $? != 0 ] || [ -z "${ipinfo}" ]; then
+    err 2 "Unable to get IP info from ipinfo.io"
+fi
 
-[ -n "${ip}" ] || err 2 "Unable to get the IP address from ifconfig.io"
+if [ ! -f /tmp/mender/inventory-geo ]; then
```
oleorhagen, line 40: Suggestion:
Screams for a constant, no?

oleorhagen (approved):
Yes, I agree!
Smrt!
Let's cache it.

oleorhagen:
Also, I think the second commit might warrant a `Changelog` too :thinking:

lluiscampos:
> Also, I think the second commit might warrant a `Changelog` too thinking

Yes, please

oleorhagen:
@alfrunes Perhaps it should re-send on a network change? :thinking:

alfrunes:
> @alfrunes Perhaps it should re-send on a network change? :thinking:

I'm not sure how to do this in a compliant way without adding more dependencies. Suggestions?

@lluiscampos I will need your help for merging this PR as I'm blocked by the faulty coverage status.

merlin-northern:
> > @alfrunes Perhaps it should re-send on a network change? 🤔
> 
> I'm not sure how to do this in a compliant way without adding more dependencies. Suggestions?

if you assume that you have NetworkManager running, then you can listen to DBus events.
otherwise the only way to detect a network change is to store the outbound src IP somewhere. I guess the above means that you want to resend on the ip address change. that is a larger part of the story of sending inventory only if it changed, I wonder if we should also trigger the send when it changes.

seems a bit out of scope though.

> 
> @lluiscampos I will need your help for merging this PR as I'm blocked by the faulty coverage status.

oleorhagen:
> seems a bit out of scope though.

Yes, definitely, it just struck me since this PR introduced one form of caching already

PR #707, opened 2021-02-02
Dockerfile beautification + upgrade

lluiscampos:
* [Dockerfile.daemon] Minor beautifications and optimizations
* [Dockerfile.daemon] Upgrade base image to ubuntu:20.04

### DIFF
--- a/tests/Dockerfile.daemon
+++ b/tests/Dockerfile.daemon
@@ -13,7 +13,7 @@ RUN jq ".ServerCertificate=\"/usr/share/doc/mender-client/examples/demo.crt\" | 
     < examples/mender.conf.demo > /mender-install/etc/mender/mender.conf
 RUN echo artifact_name=original > /mender-install/etc/mender/artifact_info
 
-FROM ubuntu:18.04
+FROM ubuntu:20.04
 
 RUN apt update && apt install -y openssh-server
 
--- a/tests/Dockerfile.daemon
+++ b/tests/Dockerfile.daemon
@@ -5,11 +5,13 @@ FROM ubuntu:20.04 AS build
 ENV DEBIAN_FRONTEND=noninteractive
 RUN apt update && apt install -y make git build-essential golang liblzma-dev jq libssl-dev libglib2.0-dev
 
-COPY ./ /go/src/github.com/mendersoftware/mender/
-RUN make -C /go/src/github.com/mendersoftware/mender GOPATH=/go clean
-RUN make -C /go/src/github.com/mendersoftware/mender GOPATH=/go prefix=/mender-install install
-RUN sh -c 'jq ".ServerCertificate=\"/usr/share/doc/mender-client/examples/demo.crt\" | .ServerURL=\"https://docker.mender.io/\"" \
-    < /go/src/github.com/mendersoftware/mender/examples/mender.conf.demo > /mender-install/etc/mender/mender.conf'
+WORKDIR /go/src/github.com/mendersoftware/mender
+COPY ./ .
+RUN make GOPATH=/go clean
+RUN make GOPATH=/go prefix=/mender-install install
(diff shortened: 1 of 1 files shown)


lluiscampos:
@mzedel Except for a tiny modification, I committed the file as you pasted it to me in Slack. Thanks!

tests/Dockerfile.daemon
```diff
@@ -5,29 +5,30 @@ FROM ubuntu:20.04 AS build
...
-RUN sh -c 'echo Port 8822 >> /etc/ssh/sshd_config'
-
-COPY --from=build /mender-install/ /
-
-RUN sh -c 'mkdir -p /var/lib/mender && echo device_type=docker-client > /var/lib/mender/device_type'
-RUN sh -c 'echo artifact_name=original > /etc/mender/artifact_info'
+RUN echo 'PermitEmptyPasswords yes\n\
+PermitRootLogin yes\n\
+Port 22\n\
+Port 8822\n' >> /etc/ssh/sshd_config
```
kacf, line 26: Are you sure that this works? `echo` does not translate `\n` sequences, so you end up with a literal backslash and 'n'. That is, unless Docker translates it.

kacf (commented):
Good cleanup!

tests/Dockerfile.daemon
lluiscampos: This part actually works - I just verified that the lines as correctly appended to `sshd_config`. But something else is broken: the three integration tests that use this image are all failing :)

Investigating...

lluiscampos:
>[...] But something else is broken: the three integration tests that use this image are all failing :)

Some paths were missing, like `/usr/share/mender`. Fixed now :+1:

tests/Dockerfile.daemon
```diff
@@ -18,16 +20,18 @@ RUN apt update && apt install -y openssh-server
...
+
+COPY --from=build /mender-install/usr/bin/mender /usr/bin/mender
+COPY --from=build /mender-install/etc/mender /etc/mender
+COPY --from=build /mender-install/usr/share/mender /usr/share/mender
+COPY --from=build /mender-install/usr/share/doc/mender-client /usr/share/doc/mender-client
+COPY --from=build /mender-install/lib/systemd/system/mender-client.service /lib/systemd/system/mender-client.service
+COPY --from=build /mender-install/usr/share/dbus-1/system.d/io.mender.AuthenticationManager.conf /usr/share/dbus-1/system.d/io.mender.AuthenticationManager.conf
```
kacf, line 33: I guess you have a reason, but this is very cumbersome to maintain. Why doesn't the original `COPY --from=build /mender-install/ /` cut it?

kacf (approved):

tests/Dockerfile.daemon
mzedel: that will just copy the changed files from the build stage - so everything else (e.g. `ls`, `sh`...) is gone and I have never seen the same behaviour for another image...
from what I found on looking into the ubuntu dockerfile, I _think_ the reason is a change in the ubuntu image build process (they at least planned to change to `buildah`), but couldn't track down the build steps and thus settled for the `COPY` approach...

kacf: Ah, I see. It's less like `cp` and more like `rsync -r --delete`. :-)

mzedel: unfortunately yes... but it should be more like `cp` (and it works that way in other repos + was with the `18.04` image) 🤷‍♂️

mzedel (approved):

tests/Dockerfile.daemon
lluiscampos: @kacf Exactly. IOW, the real motivation to do this was to be able to upgrade to 20.04 image... I hope this does not bit us back in the near future when adding new directories or files to the Mender installation.

PR #709, opened 2021-02-16
Cherry-pick MEN-4085 to 2.5.x

alfrunes:

### DIFF
--- a/support/mender-inventory-geo
+++ b/support/mender-inventory-geo
@@ -26,9 +26,14 @@ err() {
  exit $rc
 }
 
-# Timeout value for connections to ipvigilante.
-# This value is in seconds
-TIMEOUT=10
+TIMEOUT=10 # Request timeout
+TEMPFILE="/tmp/mender/inventory-geo"
+
+# the names of the attributes holding the localization data can be configured here
+ATTR_NAME_IP="geo-ip"
+ATTR_NAME_CITY="geo-city"
+ATTR_NAME_COUNTRY="geo-country"
+ATTR_NAME_TIMEZONE="geo-timezone"
 
 ipinfo=$(wget --timeout=${TIMEOUT} -q -O /dev/stdout \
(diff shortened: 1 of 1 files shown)


lluiscampos:
IMO, these do not qualify as a bugfixes. It is too bad that the script in 2.5.x is not working, but changing the provider is a bit intruisife.

Do you have an opinion @kacf ?

oleorhagen:
> IMO, these do not qualify as a bugfixes. It is too bad that the script in 2.5.x is not working, but changing the provider is a bit intrusive.

I think this could be open for discussion. The thinking being that as soon as we started distributing this as  a part of our default setup, we are kinda responsible for it working OOTB, and as such, the endpoint not working I would consider a bug.

The second commit I don't think belongs here however.

kacf:
Normally I'd agree that this is too intrusive, but given that the original is completely non-functional, I'll side with @oleorhagen in that this has to be considered a bug. Since it's already broken, the net risk is not higher than any other bug fix.

But I agree I would leave out the second commit, this is just optimization.

lluiscampos (changes requested):
@oleorhagen @kacf Thanks for sharing your point of view: you convinced me :)

@alfrunes Let's drop the second commit, then (here and in #710)

alfrunes:
@oleorhagen @lluiscampos @kacf; the second commit is not just an optimization - it's also keeping the device within the [rate-limits](https://ipinfo.io/developers#rate-limits) of 50,000 requests per month. Using our demo configuration the device makes up to 518,400 inventory calls per month. Besides, I don't think it's fair usage to apply our demo polling intervals to third party services.

kacf:
Ah, so it might actually break without it, if they rate limit the clients... Ok, I guess it should be included then. Again, I guess the net risk is not that high, since right now nothing is working.

lluiscampos (approved):

PR #712, opened 2021-03-01
single-file: Use atomic file operations.

kacf:
Changelog: Title

### DIFF
--- a/support/modules/single-file
+++ b/support/modules/single-file
@@ -10,6 +10,17 @@ dest_dir_file="$FILES"/files/dest_dir
 filename_file="$FILES"/files/filename
 permissions_file="$FILES"/files/permissions
 
+safe_copy() {
+    if [ $# -gt 2 ]; then
+        echo "safe_copy can only handle one file copy at a time" >&2
+        exit 2
+    fi
+    cp -a "$1" "$2".tmp || return $?
+    sync "$2".tmp || return $?
+    mv "$2".tmp "$2" || return $?
+    sync "$(dirname "$2")" || return $?
+}
+
 case "$STATE" in
 
     NeedsArtifactReboot)
@@ -29,7 +40,7 @@ case "$STATE" in
         mkdir -p "${tmp_dest_dir}"
         if test -f "${dest_dir}/${filename}"
         then
-            if ! cp -a "${dest_dir}/${filename}" "${tmp_dest_dir}"
+            if ! safe_copy "${dest_dir}/${filename}" "${tmp_dest_dir}"
             then
                 ret=$?
                 # Make sure there is no half-backup lying around.
@@ -37,14 +48,14 @@ case "$STATE" in
                 exit $ret
             fi
         fi
-        cp "$FILES"/files/"${filename}" "${dest_dir}/${filename}"
         # Previous revisions of this update module did not use the
         # permissions_file, so it might not exist.
         if test -f "${permissions_file}"
         then
             mode=$(cat "${permissions_file}")
-            chmod "${mode}" "${dest_dir}/${filename}"
+            chmod "${mode}" "$FILES"/files/"${filename}"
         fi
+        safe_copy "$FILES"/files/"${filename}" "${dest_dir}/${filename}"
         ;;
 
     ArtifactRollback)
@@ -53,8 +64,7 @@ case "$STATE" in
         dest_dir=$(cat "${dest_dir_file}")
         test -z "$dest_dir" -o -z "$filename" && \
             echo "Fatal error: dest_dir or filename are undefined." && exit 1
-        rm "${dest_dir}/${filename}"
-        cp -a "${tmp_dest_dir}/${filename}" "${dest_dir}/${filename}"
+        safe_copy "${tmp_dest_dir}/${filename}" "${dest_dir}/${filename}"
         ;;
 esac
 
--- a/support/modules/single-file
+++ b/support/modules/single-file
@@ -35,7 +35,7 @@ case "$STATE" in
         dest_dir=$(cat "${dest_dir_file}")
         filename=$(cat "${filename_file}")
         test -z "$dest_dir" -o -z "$filename" && \
-            echo "Fatal error: dest_dir or filename are undefined." && exit 1
+            echo "Fatal error: dest_dir or filename are undefined." >&2 && exit 1
         mkdir -p "${dest_dir}"
         mkdir -p "${tmp_dest_dir}"
         if test -f "${dest_dir}/${filename}"


kacf:
@mender-test-bot: start pipeline

kacf:
@mender-test-bot: start pipeline

support/modules/single-file
```diff
@@ -10,6 +10,17 @@ dest_dir_file="$FILES"/files/dest_dir
 filename_file="$FILES"/files/filename
 permissions_file="$FILES"/files/permissions
 
+safe_copy() {
+    if [ -n "$3" ]; then
+        echo "safe_copy can only handle one file copy at a time" 1>&2
```
oleorhagen, line 15: Nit:
Is the `1` in `1>&2` reallly necessary?

support/modules/single-file
```diff
@@ -10,6 +10,17 @@ dest_dir_file="$FILES"/files/dest_dir
 filename_file="$FILES"/files/filename
 permissions_file="$FILES"/files/permissions
 
+safe_copy() {
+    if [ -n "$3" ]; then
```
oleorhagen, line 14: :thinking: Is this the best way of checking nr args

support/modules/single-file
```diff
@@ -53,8 +64,7 @@ case "$STATE" in
         dest_dir=$(cat "${dest_dir_file}")
         test -z "$dest_dir" -o -z "$filename" && \
             echo "Fatal error: dest_dir or filename are undefined." && exit 1
```
oleorhagen, line 66: I know it's not on you, but could you add an extra commit with echo to `stderr` here? :pray:

support/modules/single-file
```diff
@@ -53,8 +64,7 @@ case "$STATE" in
...
-        rm "${dest_dir}/${filename}"
```
oleorhagen, line 56: Question:
How does removing this relate to `safe_copy` of the files?

support/modules/single-file
```diff
@@ -29,22 +40,22 @@ case "$STATE" in
...
             fi
         fi
-        cp "$FILES"/files/"${filename}" "${dest_dir}/${filename}"
         # Previous revisions of this update module did not use the
         # permissions_file, so it might not exist.
         if test -f "${permissions_file}"
         then
             mode=$(cat "${permissions_file}")
-            chmod "${mode}" "${dest_dir}/${filename}"
+            chmod "${mode}" "$FILES"/files/"${filename}"
```
oleorhagen, line 56: Question:
Was this wrong previously?

lluiscampos (approved):
Also curious about the couple of things that @oleorhagen has asked.

Regardless, the safe_copy part looks good to me

TheMeaningfulEngineer:
I would ask for some help it trying to define what the "most intuitive expected" behavior for this might be if that file is something which is being used during the time of the update.
This is testing the limits of my understanding of the Linux file handles.

As an example
I just tried copying a random file over `mender-store`. `mender-store` was just a random pick of a file which is "consumed" by a process. In lack of a better word "consumed" means it's listed with `ls -la /proc/$(pidof mender)/fd | grep mender-store`.

In a recent discussion there was a case with the previous version of the module where the overwrite couldn't happen. `cp: cannot create regular file '/path/to/file'. Text file busy`. I was hoping to see something like this also for my `mender-store` experiment, but didn't.

@kacf @lluiscampos @olehermanse 
Do you have any thoughts on this?

support/modules/single-file
```diff
@@ -10,6 +10,17 @@ dest_dir_file="$FILES"/files/dest_dir
 filename_file="$FILES"/files/filename
 permissions_file="$FILES"/files/permissions
 
+safe_copy() {
+    if [ -n "$3" ]; then
+        echo "safe_copy can only handle one file copy at a time" 1>&2
+        exit 2
+    fi
+    cp -a "$1" "$2".tmp || return $?
```
oleorhagen, line 18: Also, could we use the long option in scripts? It'd just be a good convention to follow I think

support/modules/single-file
kacf: I thought it was, but apparently not. I guess `>&2` is better since `1>&2` looks more similar to `2>&1` (easier to misunderstand). Will change it.

support/modules/single-file
kacf: It's essentially the same as `$# -ge 3`, both are saying the same thing. But now that you point it out, it occurred to me that the `$#` style is more resilient if `set -u` is enabled, so I'll change it!

support/modules/single-file
kacf: I will!

support/modules/single-file
kacf: Removing it first means that it's not atomic. There is a window in time where the file does not exist. This is why we use `mv` inside `safe_copy` as the update step, because it is atomic.

support/modules/single-file
kacf: Same reason: It is not atomic. There is a window in time where the file has been replaced, but it is not executable. It would be possible to do the `chmod` inside `safe_copy` instead of in advance here, but this method avoids having a special case for executables inside `safe_copy`, and does the same job.

support/modules/single-file
kacf: I somewhat agree, but not for very common options. In this particular case I think the original is actually better, because the full name `--archive`, does not really reveal what it does, and is even a bit misleading (a better name would be `--preserve-attributes` or something like that). I think in this case it's better to look it up.

Definitely for our various options in mender-artifact I agree though, long option over short option.

kacf:
@TheMeaningfulEngineer: The key difference lies in using `mv` instead of `cp`. `cp` opens the file, and then writes into it, but this is not allowed for executables which are being executed. `mv` on the other hand *replaces* the file with a new file in one step. The old file, which may still be executed, is deleted from the file system, but still occupy space, and the filesystem blocks essentially become "garbage collected" once the program terminates.

So what happens is that the existing program continues to run with the old binary, whereas all new invocations of the program use the new binary. There is never a point in time where the program is half updated, and there is also no point where "Text file busy" can be returned, because the old file is never opened.

Did that answer the question?

support/modules/single-file
oleorhagen: I think I'd prefer the special case handling, but all yours

oleorhagen (approved):

kacf:
> In a recent discussion there was a case with the previous version of the module where the overwrite couldn't happen. `cp: cannot create regular file '/path/to/file'. Text file busy`. I was hoping to see something like this also for my `mender-store` experiment, but didn't.

And to answer this question: The reason it does not happen here is because the database is not being executed. "Text file busy" is a very misleading message, but AFAIK it is reserved exclusively for executables (at least for Linux, some other Unixes may treat it differently, but who uses them nowadays? :smile: )

support/modules/single-file
kacf: I guess we agree to disagree then! ;-)
It's partially because `safe_copy` actually comes from mender-configure, and it's nice if it's reusable.

kacf:
@mender-test-bot: start pipeline

TheMeaningfulEngineer:
@kacf 
Thanks, it does.
I didn't really know that the OS consumes the file differently depending if its "executed" or just "opened by another process".
For the "replacing an executed file" the mv makes total sense.

As for replacing say a configuration file which is being used, I guess "doing the same as Linux would" which is it would overwrite the config file which is opened by an ongoing process is the most intuitive thing to do.

Thanks for the clarification :)

TheMeaningfulEngineer (approved):

oleorhagen:
> And to answer this question: The reason it does not happen here is because the database is not being executed. "Text file busy" is a very misleading message, but AFAIK it is reserved exclusively for executables (at least for Linux, some other Unixes may treat it differently, but who uses them nowadays? smile )

`Text` here refers to the `.Text` (Which is the actual instructions) section in the executable. So this is why it happens only for executables.

PR #717, opened 2021-03-10
MEN-4518: Always send inventory after a successful deployment

oleorhagen:
Previously a reboot was needed in order to force-resend the inventory. This was
never discovered, as the client up until quite recently has almost always
rebooted.

The fix is dead simple, reset the in-memory storage of the last
inventory-update, which is equivalent to a reboot.

Changelog: Send the inventory after a successful deployment, even though the
device has not rebooted.

### DIFF
--- a/app/state.go
+++ b/app/state.go
@@ -1,4 +1,4 @@
-// Copyright 2020 Northern.tech AS
+// Copyright 2021 Northern.tech AS
 //
 //    Licensed under the Apache License, Version 2.0 (the "License");
 //    you may not use this file except in compliance with the License.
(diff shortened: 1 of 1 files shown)


oleorhagen:
Simple fix, with no tests. I don't trust unit tests for this, so either we bring out the big guns, or manual verification is OK?

kacf:
I agree that unit tests can't be trusted for this kind of thing. I think an integration test for this should be pretty easy and fast to execute though, since you only need a `single-file` update. I can recommend the `standard_setup_one_docker_client_bootstrapped` fixture for speed as well.

kacf (approved):

oleorhagen:
Integration test: https://github.com/mendersoftware/integration/pull/1254

oleorhagen:
https://gitlab.com/Northern.tech/Mender/mender-qa/-/pipelines/268223372

oleorhagen:
Cherries:
* 2.5.x - https://github.com/mendersoftware/mender/pull/721

PR #718, opened 2021-03-10
2.3.x single-file: Use atomic file operations.

kacf:


kacf:

PR #736, opened 2021-04-07
MEN-4580: Split demo certificate into singular pieces.

kacf:
This is required for programs that look up certificates by their hash,
such as curl. Concatenated certificates are not supported when
creating hashed symlinks and are simply ignored by
`update-ca-certificates`.

Changelog: None

### DIFF
--- a/cli/setup.go
+++ b/cli/setup.go
@@ -15,6 +15,7 @@ package cli
 
 import (
 	"bufio"
+	"bytes"
 	"encoding/json"
 	"fmt"
 	"io"
@@ -23,7 +24,6 @@ import (
 	"os"
 	"os/exec"
 	"path"
-	"path/filepath"
 	"regexp"
 	"strconv"
 	"strings"
@@ -70,16 +70,14 @@ const ( // state enum
 
 var (
 	// needed so that we can override it when testing
-	DefaultMenderDemoCertDir   = "/usr/share/doc/mender-client/examples"
-	DefaultLocalTrustMenderDir = "/usr/local/share/ca-certificates/mender"
+	DefaultMenderDemoCertDir      = "/usr/share/doc/mender-client/examples"
+	DefaultLocalTrustMenderDir    = "/usr/local/share/ca-certificates/mender"
+	DefaultLocalTrustMenderPrefix = "mender-demo-"
+	DefaultLocalTrustMenderFormat = "mender-demo-%d.crt"
 )
 
 func getMenderDemoCertPath() string {
 	return path.Join(DefaultMenderDemoCertDir, "demo.crt")
-}
-
-func getLocalTrustMenderCertPath() string {
-	return path.Join(DefaultLocalTrustMenderDir, "server.crt")
 }
 
 const (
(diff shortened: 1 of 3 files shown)


kacf:
I also would like to fix this at the Yocto level. Although seemingly not necessary under Yocto (different curl version? different openssl setup?), this could break at any moment, and the warning mechanism inside the client won't work without it either.

kacf:

mzedel:
Just out of curiosity, would this still be necessary once https://github.com/mendersoftware/integration/pull/1253 gets in? since then only a single cert would be used that references the different domains...

cli/setup.go
```diff
@@ -70,18 +70,16 @@ const ( // state enum
 
 var (
 	// needed so that we can override it when testing
-	DefaultMenderDemoCertDir   = "/usr/share/doc/mender-client/examples"
-	DefaultLocalTrustMenderDir = "/usr/local/share/ca-certificates/mender"
+	DefaultMenderDemoCertDir      = "/usr/share/doc/mender-client/examples"
+	DefaultLocalTrustMenderDir    = "/usr/local/share/ca-certificates/mender"
+	DefaultLocalTrustMenderPrefix = "mender-demo-"
+	DefaultLocalTrustMenderFormat = "mender-demo-%d.crt"
```
lluiscampos, line 76: Good idea :+1:

cli/setup_test.go
```diff
@@ -282,12 +284,22 @@ func TestInstallDemoCertificateLocalTrust(t *testing.T) {
...
+		checkCrtInstall(t, path.Join(DefaultLocalTrustMenderDir, entry.Name()), crtSource)
+	}
+}
+
+func checkCrtInstall(t *testing.T, cert string, demoCertContent []byte) {
+	assert.True(t, strings.HasPrefix(path.Base(cert), DefaultLocalTrustMenderPrefix))
+	content, err := ioutil.ReadFile(cert)
+	require.NoError(t, err)
+	assert.Greater(t, len(content), 10)
+	assert.True(t, bytes.Contains(demoCertContent, content))
```
lluiscampos, line 304: We could also verify that the first line has "BEGIN CERTIFICATE" and the last one "END CERTIFICATE"

lluiscampos (approved):
Looks good!

kacf:
> Just out of curiosity, would this still be necessary once [mendersoftware/integration#1253](https://github.com/mendersoftware/integration/pull/1253) gets in? since then only a single cert would be used that references the different domains...

No, it wouldn't. @alfrunes and I discussed this yesterday as a possible solution. However, that PR has a huge number of ripple effects, and a lot of work still to do, so we decided we can't include it at this time.

cli/setup_test.go
kacf: Good idea!

lluiscampos (approved):

kacf:
Cherry-picked directly to 2.6.x (didn't do a mender-qa build, since I will do a full release build right after this anyway).

PR #738, opened 2021-04-08
Make: Ignore the 'stringop-overflow' compiler warning

oleorhagen:
This is out of our control, and the warning is annoying.

Changelog: None

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -19,7 +19,7 @@ CGO_ENABLED=1
 export CGO_ENABLED
 
 # Get rid of useless warning in lmdb
-CGO_CFLAGS ?= -Wno-implicit-fallthrough
+CGO_CFLAGS ?= -Wno-implicit-fallthrough -Wno-stringop-overflow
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -19,7 +19,7 @@ CGO_ENABLED=1
 export CGO_ENABLED
 
 # Get rid of useless warning in lmdb
```
lluiscampos, line 21: Please update the comment too

lluiscampos (commented):
Thanks for doing this! It annoys me too :)

kacf (approved):
:+1:

Makefile
oleorhagen: The warning is still in `lmdb`, so it's still right, but I can remove it still if you like

lluiscampos: No, if that is the case it is fine. I thought it came from somewhere else.

oleorhagen:
@lluiscampos will you merge this?

PR #740, opened 2021-04-09
MEN-4104: Add D-Bus introspection spec file

lluiscampos:


Documentation/io.mender.Authentication1.xml
```diff
@@ -0,0 +1,52 @@
...
+
+<node>
+  <!--
+    io.mender.Authentication1:
+    @short_description: Mender Authentication API v1
+
+    This interface lets applications authenticate with the Mender server. The
+    Mender client will handle the authentication and provide to the user with a
+    JSON Web Token (JWT) and the server URL, which which the user can do its own
+    API calls to the Mender server.
```
tranchitella, line 12: This interface lets applications authenticate with the Mender server. The
Mender client will handle the authentication and provide the user with a
JSON Web Token (JWT) and the server URL, which the user can use to
API calls to the Mender server on his own.

Documentation/io.mender.Authentication1.xml
```diff
@@ -0,0 +1,52 @@
...
+  -->
+  <interface name="io.mender.Authentication1">
+
+    <!--
+      GetJwtToken:
+      @token: Valid JWT token
+      @server_url: Server URL
+
+      Gets current JWT token and server URL. If no JWT token is available, error
+      is returned.
```
tranchitella, line 22: Gets current JWT token and server URL. If no JWT token is available, an error
is returned.

Documentation/io.mender.Authentication1.xml
```diff
@@ -0,0 +1,52 @@
...
+      <arg type="b" name="success" direction="out"/>
+    </method>
+
+    <!--
+      JwtTokenStateChange:
+      @token: Current JWT token
+      @server_url: Server URL
+
+      Emitted whenever a valid JWT is available in the Mender client. The new
+      token is passed with the event, together with the server URL.
```
tranchitella, line 45: Emitted whenever a valid JWT is available in the Mender client. The event includes the new
token and the server URL.

Documentation/io.mender.Authentication1.xml
```diff
@@ -0,0 +1,52 @@
...
+      GetJwtToken:
+      @token: Valid JWT token
+      @server_url: Server URL
+
+      Gets current JWT token and server URL. If no JWT token is available, error
+      is returned.
+    -->
+    <method name="GetJwtToken">
+      <arg type="s" name="token" direction="out"/>
+			<arg type="s" name="server_url" direction="out"/>
```
tranchitella, line 26: Why is the second arg indented with tabs?

Documentation/io.mender.Authentication1.xml
```diff
@@ -0,0 +1,52 @@
...
+      JwtTokenStateChange:
+      @token: Current JWT token
+      @server_url: Server URL
+
+      Emitted whenever a valid JWT is available in the Mender client. The new
+      token is passed with the event, together with the server URL.
+    -->
+    <signal name="JwtTokenStateChange">
+      <arg type="s" name="token"/>
+			<arg type="s" name="server_url"/>
```
tranchitella, line 49: Why is the second arg indented with tabs?

tranchitella (commented):
A few questions and suggestions

oleorhagen (approved):
No too familiar with the area, but had a read through.

Only thing which struct me is:
Why do we refer to `JWT` tokens in all comments, but the methods all spell it `Jwt`?

Documentation/io.mender.Authentication1.xml
kacf: Gets current JWT token and server URL. If no JWT token is available, **an empty token**
is returned.

Documentation/io.mender.Authentication1.xml
```diff
@@ -0,0 +1,52 @@
...
+    <method name="GetJwtToken">
+      <arg type="s" name="token" direction="out"/>
+			<arg type="s" name="server_url" direction="out"/>
+    </method>
+
+    <!--
+      FetchJwtToken:
+      @success: false on errors
+
+      Instructs the Mender client to fetch the JWT token from the server.
```
kacf, line 33: Append: "When the token is ready, a FetchJwtToken signal will be sent".

Documentation/io.mender.Authentication1.xml
lluiscampos: My editor playing tricks on me

lluiscampos:
> Only thing which struct me is:
> Why do we refer to `JWT` tokens in all comments, but the methods all spell it `Jwt`?

We did name our methods `Jwt`, we cannot do anything about that now, while `JWT` is the generally used name (just Google-ing a bit).

lluiscampos:
@tranchitella @kacf Thank you for your feedback. I rewrote it following your suggestions.

tranchitella (approved):

PR #742, opened 2021-04-22
MEN-4535: DBus interface for SetUpdateControlMap

oleorhagen:


dbus/dbus_libgio.go.h
```diff
@@ -26,12 +29,46 @@ static GVariant *g_variant_new_from_two_strings(gchar *value1, gchar *value2)
...
     return g_variant_new("(b)", value);
 }
 
+// create a new GVariant from a int value
+static GVariant *g_variant_new_from_int(gint value)
+{
+    return g_variant_new("(i)", value);
+}
+
+static const gchar* extract_parameter(GVariant * parameters) {
```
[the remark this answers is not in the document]
oleorhagen: All in all, this whole function, either returns the `JSON` string, or NULL.

app/updatemanager.go
```diff
@@ -0,0 +1,163 @@
...
+	UpdateManagerDBusObjectName    = "io.mender.UpdateManager"
+	UpdateManagerDBusInterfaceName = "io.mender.Update1"
+	UpdateManagerDBusInterface     = `
+        <node>
+            <interface name="io.mender.Update1">
+                <method name="SetUpdateControlMap">
+                    <arg type="s" name="update_control_map" direction="in"/>
+                    <arg type="i" name="refresh_timeout" direction="out"/>
+                 </method>
+	    </interface>
```
kacf, line 40: Nitpick: Indentation.

app/updatemanager.go
```diff
@@ -0,0 +1,163 @@
...
+		})
+	defer u.dbus.UnregisterMethodCallCallback(
+		UpdateManagerDBusPath,
+		UpdateManagerDBusInterfaceName,
+		"SetUpdateControlMap")
+	select {
+	case <-ctx.Done():
+		return nil
+	}
+	return nil
```
kacf, line 130: Since there is an unconditional wait, and only one possible path, I think this can be shortened to:

```go
	<-ctx.Done()
	return nil
```

Whether it's clearer, well that's a different discussion. You can decide!

app/updatemanager.go
```diff
@@ -0,0 +1,163 @@
...
+func (c *ControlMap) Set(cm *UpdateControlMap) {
+	log.Error("Setting..")
+	c.mutex.Lock()
+	defer c.mutex.Unlock()
+	c.controlMap[cm.key()] = cm
+}
+
+func (c *ControlMap) Get(key string) *UpdateControlMap {
+	log.Errorf("Key: %s: %v", key, c.controlMap[key])
+	return c.controlMap[key]
```
kacf, line 162: `Get` needs to use the mutex too. See [here](https://golang.org/doc/faq#atomic_maps).

app/updatemanager.go
```diff
@@ -0,0 +1,163 @@
...
+	return &ControlMap{
+		controlMap: make(map[string]*UpdateControlMap),
+	}
+}
+
+func (c *ControlMap) Set(cm *UpdateControlMap) {
+	log.Error("Setting..")
+	c.mutex.Lock()
+	defer c.mutex.Unlock()
+	c.controlMap[cm.key()] = cm
```
kacf, line 157: I see a potential problem here. Here we are storing the map with a key which is combination of the ID and priority. This is fine in itself, but we need to be able to look up the map with the ID only. The interface would be more akin to this:

```go
Set(cm *UpdateControlMap)
Get(id string) []*UpdateControlMap
```

So it would return a list of all the maps that match, with possibly several different priorities. You could make this an extra function, in addition to the original `Get`, but the client will always call the getter which takes an ID only.

app/updatemanager_test.go
```diff
@@ -0,0 +1,123 @@
...
+	}()
+	go func() {
+		cm.Set(&UpdateControlMap{
+			ID:       "foobarbaz",
+			Priority: 3,
+		})
+		wg.Done()
+	}()
+	wg.Wait()
+	assert.Equal(t, cm.Get("foo0"), &UpdateControlMap{ID: "foo", Priority: 0}, cm.controlMap)
```
kacf, line 56: If you really want to verify thread safety, this is kind of thing that should be repeated 10000 times. :-)

app/updatemanager_test.go
```diff
@@ -0,0 +1,123 @@
...
+}
+
+// TestControlMapDBus relies on the DBus server set up in TestMain for the dbus package
+func TestControlMapDBus(t *testing.T) {
+	return
```
kacf, line 61: I didn't get the relation between the comment and the empty test function...

support/dbus/io.mender.AuthenticationManager.conf
```diff
@@ -6,11 +6,15 @@
...
+    <allow own="io.mender.UpdateManager"/>
   </policy>
 
   <!-- Allow root to invoke methods on Mender -->
   <policy user="root">
     <allow send_destination="io.mender.AuthenticationManager"/>
     <allow receive_sender="io.mender.AuthenticationManager"/>
+
+    <allow send_destination="io.mender.UpdateManager"/>
+    <allow receive_sender="io.mender.UpdateManager"/>
```
kacf, line 18: These should be in their own file (see filename).

app/updatemanager.go
oleorhagen: I like it

app/updatemanager_test.go
oleorhagen: Removing

app/updatemanager.go
```diff
@@ -0,0 +1,155 @@
...
+func NewControlMap() *ControlMap {
+	return &ControlMap{
+		controlMap: make(map[string][]*UpdateControlMap),
+	}
+}
+
+func (c *ControlMap) Set(cm *UpdateControlMap) {
+	c.mutex.Lock()
+	defer c.mutex.Unlock()
+	c.controlMap[cm.ID] = append(c.controlMap[cm.ID], cm)
```
kacf, line 148: Maybe you misunderstood my original comment. When storing the map, all pairs of ID and priority must be unique. The way you stored it earlier did do this correctly (though lookup was tricky). However, here you have lost the uniqueness: If I update using a map with the same ID and priority, it will be appended to the end of the list instead of being replaced.

Here's my suggestion for nailing this in a test: Use my exact example from [MEN-4535](https://tracker.mender.io/browse/MEN-4535). For the first three cases, verify that inserting the two maps returns two maps as well. For the last case, verify that it returns only one map, the last one.

What data structure to use, hmm, you could use your original one containing both ID and priotity. You'd have to do a linear search to look up by ID, but this is fine, we are not expecting many maps, so performance is not a concern.

app/updatemanager.go
kacf: This was resolved, but not fixed. Not sure if you meant that you had it locally or not, but just in case you forgot.

support/dbus/io.mender.AuthenticationManager.conf
kacf: Same here... I guess it's me misunderstanding, but just in case!

kacf (approved):
Pre-approving! This is a feature branch anyway, so we can anyway fix minor details when I get back.

app/updatemanager.go
oleorhagen: I stared at this thing for the longest time, and found one indent which was 3 spaces, instead of 4. I guess that was not what you meant then? 
I'm re-indenting the whole thing with four spaces to indent it away from the variable declarations, if this is not what you mean still, let's just fix it when you're back :)

oleorhagen:

lluiscampos (approved):
Looks very good overall.

app/updatemanager.go
```diff
@@ -0,0 +1,160 @@
...
+
+var UpdateManagerMap *ControlMap
+
+func init() {
+	UpdateManagerMap = NewControlMap()
+}
+
+type UpdateManager struct {
+	dbus                        dbus.DBusAPI
+	dbusConn                    dbus.Handle
```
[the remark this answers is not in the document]
oleorhagen: Good catch!

app/updatemanager.go
```diff
@@ -0,0 +1,160 @@
...
+	}
+	defer u.dbus.BusUnregisterInterface(dbusConn, intGID)
+
+	u.dbus.RegisterMethodCallCallback(
+		UpdateManagerDBusPath,
+		UpdateManagerDBusInterfaceName,
+		"SetUpdateControlMap",
+		func(objectPath string, interfaceName string, methodName string, updateControlMap string) (interface{}, error) {
+			controlMap := UpdateControlMap{}
+			err := json.Unmarshal([]byte(updateControlMap), &controlMap)
```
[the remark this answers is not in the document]
oleorhagen: I disagree actually. `updateControlMap` is way more descriptive. _This_ callback requires exactly what it says. The `updateControlMap`, not a generic `methodParameter`.

app/updatemanager.go
```diff
@@ -0,0 +1,160 @@
...
+			if err != nil {
+				log.Errorf("Failed to unmarshal the JSON in the string received via SetControlMap on D-Bus: %s", err)
```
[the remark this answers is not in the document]
oleorhagen: Agreed :+1:

oleorhagen:
Simply squashed the review commits, nothing has changed, thus will not rerun the QA pipeline

PR #747, opened 2021-04-27
MEN-4546: Validate and sanitize Update Control Maps passed via DBus

lluiscampos:
If the validation fails, an error is returned to the caller via DBus
endpoint.

If the sanitization of a particular state results in default values, it
is removed from the map. Then if the sanitized map ends up with 0
states, it is not added to the pool (but no warning is emitted to the
caller).


lluiscampos:
> @lluiscampos, Let me know if you want to start the integration pipeline by mentioning me and the command "start pipeline".

@mender-test-bot :ok_man:

app/updatemanager.go
```diff
@@ -125,9 +138,137 @@ func (u *UpdateManager) run(ctx context.Context) error {
...
+
+func (m UpdateControlMap) Validate() error {
+	// ID is mandatory
+	if m.ID == "" {
+		return errors.New("ID cannot be empty")
+	}
+
+	// Priority must be 0 or higher
+	if m.Priority < 0 {
+		return fmt.Errorf("invalid Priority %q, value must be a positive ingeger", m.Priority)
```
oleorhagen, line 239: s/ingeger/integer

oleorhagen, line 239: 0 is not a positive integer

app/updatemanager.go
```diff
@@ -125,9 +138,137 @@ func (u *UpdateManager) run(ctx context.Context) error {
...
+		if !found {
+			return fmt.Errorf("invalid value %q", s.OnActionExecuted)
+		}
+	}
+
+	return nil
+}
+
+func (s *UpdateControlMapState) Sanitize() {
+	if s.OnMapExpire == UpdateControlMapStateOnMapExpireDefault(s.Action) {
```
oleorhagen, line 217: I'm a little unsure about these, would we not want to keep the value in, even if it is the defaults? :thinking: 
Like how do I know afterwards what the default is?

app/updatemanager.go
```diff
@@ -125,9 +138,137 @@ func (u *UpdateManager) run(ctx context.Context) error {
...
+func UpdateControlMapStateOnMapExpireDefault(action string) string {
+	if action == "pause" {
+		return "fail"
+	}
+	return action
+}
+
+var UpdateControlMapStateOnActionExecutedDefault = UpdateControlMapStateActionDefault
+
+func (s UpdateControlMapState) Validate() error {
```
oleorhagen, line 182: Just thinking:
Maybe we should also: https://golang.org/pkg/encoding/json/#Decoder.DisallowUnknownFields
In case of spelling errors, etc? :thinking:

app/updatemanager.go
```diff
@@ -125,9 +138,137 @@ func (u *UpdateManager) run(ctx context.Context) error {
...
+
 type UpdateControlMap struct {
-	ID       string `json:"id"`
-	Priority int    `json:"priority"`
+	ID       string                           `json:"id"`
+	Priority int                              `json:"priority"`
+	States   map[string]UpdateControlMapState `json:"states"`
+}
+
+var UpdateControlMapStateKeyValid []string = []string{
```
oleorhagen, line 153: Nit:
const

Here and below

app/updatemanager.go
```diff
@@ -125,9 +138,137 @@ func (u *UpdateManager) run(ctx context.Context) error {
 	return nil
 }
 
+type UpdateControlMapState struct {
+	Action           string `json:"action"`
```
oleorhagen, line 142: One idea could be to have these as `types`, and then implement custom JSON unmarshaling on them, but how you've done it is just as fine :)

oleorhagen (approved):
Beautiful code, senor!

I don't have proper understanding of this epic yet, so we are just going to have to crowd-source consensus on this one between the two of us :)

Left some questions, musings thoughts that we can discuss, but I don't see anything wrong with the code as is, in fact I think it looks beautiful! 

Good job!

app/updatemanager.go
lluiscampos: This was a bit unclear to me from the acceptance criteria and I think it is an implementation detail that can work both ways.

Either:
1) (implemented in this PR) We remove all fields that are default, in the business logic we need to consider empty values as default.
2) We set all non-provided fields to default values (except for ID everything is optional, so it is likely that a user passes incomplete maps). Then the business logic later on does not need to check for empty values.

My reason for going for 1) is that we are going to eliminate whole maps when they are all defaults (that is explicit in the acceptance criteria) so it seemed more coherent to also eliminate (or blank) fields that were default.

I don't have an strong opinion, and if you think 2) above will result in more readable code I can change it.

app/updatemanager.go
lluiscampos: You cannot have const arrays/slice in golang

app/updatemanager.go
lluiscampos: We are relying on default unmarshaling for most other JSON objects like `AuthReqData`, `LogData`, etc, so I'll keep it simple here too.

app/updatemanager.go
lluiscampos: Thanks. I didn't know about this. I am reading up and adding it to the PR, it is a good idea

app/updatemanager.go
oleorhagen: :exploding_head: TIL

app/updatemanager.go
oleorhagen: I'm leaning toward 2) I think, but feel free to do what you feel will work best here

lluiscampos: Reworked for 2)

@oleorhagen please have another look.

app/updatemanager.go
```diff
@@ -125,9 +142,143 @@ func (u *UpdateManager) run(ctx context.Context) error {
...
+}
+
+func (m UpdateControlMap) Validate() error {
+	// ID is mandatory
+	if m.ID == "" {
+		return errors.New("ID cannot be empty")
+	}
+
+	// Priority must be 0 or higher
+	if m.Priority < 0 {
```
kacf, line 244: Priority can be any number, and there are valid reasons for it to be negative. Since we don't expose priority in the UI, it will always use 0. Therefore, using negative priorities for a local UI process allows a setup where the UI/backend always has the last word. IOW, the device operator has some control over the update progress, unless the central UI/backend operator overrides it.

We could consider limiting its range though. I can hardly imagine why anyone would exceed -10 and 10, so that could be a possibility.

kacf: @lluiscampos: You haven't forgotten this, have you? :-)

lluiscampos: No, I haven't!

lluiscampos: See https://github.com/mendersoftware/mender/pull/767

PR #748, opened 2021-04-29
MEN-4547: Implement the UpdateControlMap expiration logic

oleorhagen:
Read the ticket: https://tracker.mender.io/browse/MEN-4547 *thoroughly*

Some points which is explicitly written, but from reading this and the other ticket's I don't fully understand why is needed are the complete separation of the expired maps, and the once still valid. Like why can they not simply be triggered by a boolean?

This would also help with this point:
* The expired pool must have the same primary keys and thread safety as the regular pool. 

* When UpdateControlMapExpirationTimeSeconds elapses without an update, the map is considered expired, and the map must be moved to a separate pool of expired maps.

This essentially requires monitoring of the maps in a separate goroutine, if they are required to _at all time_ be correct.
The way I've done it now is to simply check expiration on the `_Get_` call, which means it will (as far as I can tell), behave correctly (this is the only way to extract a map from either pool), but I will need some consensus here I think :)

### DIFF
--- a/app/updatemanager.go
+++ b/app/updatemanager.go
@@ -20,6 +20,7 @@ import (
 	"fmt"
 	"strings"
 	"sync"
+	"time"
 
 	"github.com/mendersoftware/mender/dbus"
 	"github.com/mendersoftware/mender/utils"
@@ -44,10 +45,10 @@ const (
 	    </node>`
 )
 
-var UpdateManagerMap *ControlMap
+var UpdateManagerMapPool *ControlMapPool
 
 func init() {
-	UpdateManagerMap = NewControlMap()
+	UpdateManagerMapPool = NewControlMap()
 }
 
 type UpdateManager struct {
@@ -129,7 +130,7 @@ func (u *UpdateManager) run(ctx context.Context) error {
 				log.Debugf("Ignoring UpdateControlMap %s with no non-default States", controlMap.ID)
 			} else {
 				log.Debugf("Setting the control map parameter to: %s", controlMap.ID)
-				UpdateManagerMap.Set(&controlMap)
+				UpdateManagerMapPool.Insert(controlMap.Stamp(u.updateControlTimeoutSeconds))
 			}
 
 			return u.updateControlTimeoutSeconds / 2, nil
@@ -149,9 +150,22 @@ type UpdateControlMapState struct {
 }
 
 type UpdateControlMap struct {
-	ID       string                           `json:"id"`
-	Priority int                              `json:"priority"`
-	States   map[string]UpdateControlMapState `json:"states"`
+	ID         string                           `json:"id"`
+	Priority   int                              `json:"priority"`
+	States     map[string]UpdateControlMapState `json:"states"`
+	expiryTime time.Time
+	expired    bool
+}
+
+func (u *UpdateControlMap) Stamp(updateControlTimeoutSeconds int) *UpdateControlMap {
+	u.expiryTime = time.Now().Add(
+		time.Duration(updateControlTimeoutSeconds) * time.Second)
+	// expiration monitor
+	go func() {
+		<-time.After(time.Duration(updateControlTimeoutSeconds) * time.Second)
+		u.expired = true
+	}()
+	return u
 }
 
 var UpdateControlMapStateKeyValid []string = []string{
@@ -281,31 +295,101 @@ func (m UpdateControlMap) Sanitize() {
 	}
 }
 
-type ControlMap struct {
-	controlMap map[string][]*UpdateControlMap
-	mutex      sync.Mutex
+func (u *UpdateControlMap) Expired() bool {
+	return u.expired
 }
 
-func NewControlMap() *ControlMap {
-	return &ControlMap{
-		controlMap: make(map[string][]*UpdateControlMap),
+func (u *UpdateControlMap) Equal(other *UpdateControlMap) bool {
+	if u.ID != other.ID {
+		return false
+	}
+	if u.Priority != other.Priority {
+		return false
+	}
+	return true
+}
+
+func (u *UpdateControlMap) String() string {
+	return fmt.Sprintf("ID: %s Priority: %d ", u.ID, u.Priority)
+}
+
+type ControlMapPool struct {
+	Pool  []*UpdateControlMap
+	mutex sync.Mutex
+}
+
+func NewControlMap() *ControlMapPool {
+	return &ControlMapPool{
+		Pool: []*UpdateControlMap{},
 	}
 }
 
-func (c *ControlMap) Set(cm *UpdateControlMap) {
+func (c *ControlMapPool) Insert(cm *UpdateControlMap) {
 	c.mutex.Lock()
 	defer c.mutex.Unlock()
-	// Map ID and Priority must be unique
-	for _, element := range c.controlMap[cm.ID] {
-		if element.ID == cm.ID && element.Priority == cm.Priority {
-			return
+	nm := []*UpdateControlMap{}
+	query(
+		c.Pool,
+		// Collector
+		func(u *UpdateControlMap) {
+			nm = append(nm, u)
+		},
+		// Predicates
+		func(u *UpdateControlMap) bool { return !cm.Equal(u) },
+	)
+	c.Pool = append(nm, cm)
+}
+
+func (c *ControlMapPool) Get(ID string) (active []*UpdateControlMap, expired []*UpdateControlMap) {
+	c.mutex.Lock()
+	defer c.mutex.Unlock()
+	query(
+		c.Pool,
+		// Collector
+		func(u *UpdateControlMap) {
+			if u.expired {
+				expired = append(expired, u)
+				return
+			}
+			active = append(active, u)
+		},
+		// Predicates
+		func(u *UpdateControlMap) bool {
+			return u.ID == ID
+		},
+	)
+	return active, expired
+}
+
+func (c *ControlMapPool) ClearExpired() {
+	c.mutex.Lock()
+	defer c.mutex.Unlock()
+	newPool := []*UpdateControlMap{}
+	query(
+		c.Pool,
+		// Collector
+		func(u *UpdateControlMap) {
+			newPool = append(newPool, u)
+		},
+		// Predicates
+		func(u *UpdateControlMap) bool {
+			return !u.Expired()
+		},
+	)
+	c.Pool = newPool
+}
+
+// query is a utility function to run a 'closure' on all values of
+// the list 'm' matching the predicates in 'predicates...'
+func query(m []*UpdateControlMap, f func(*UpdateControlMap), predicates ...func(*UpdateControlMap) bool) {
+	for _, e := range m {
+		// Sentinels
+		for _, predicate := range predicates {
+			if !predicate(e) {
+				goto nextElement
+			}
 		}
+		f(e)
+	nextElement:
 	}
-	c.controlMap[cm.ID] = append(c.controlMap[cm.ID], cm)
-}
-
-func (c *ControlMap) Get(ID string) []*UpdateControlMap {
-	c.mutex.Lock()
-	defer c.mutex.Unlock()
-	return c.controlMap[ID]
 }
--- a/app/updatemanager_test.go
+++ b/app/updatemanager_test.go
@@ -25,33 +25,32 @@ import (
 	"github.com/mendersoftware/mender/dbus/mocks"
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/mock"
+	"github.com/stretchr/testify/require"
 )
 
 // TestControlMap tests the ControlMap structure, and verifies the thread-safety
 // of the data access, and writes.
 func TestControlMap(t *testing.T) {
 	cm := NewControlMap()
-	cm.Set(&UpdateControlMap{
+	cm.Insert(&UpdateControlMap{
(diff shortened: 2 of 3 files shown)


app/updatemanager.go
```diff
@@ -301,11 +319,66 @@ func (c *ControlMap) Set(cm *UpdateControlMap) {
...
+	UpdateManagerMapExpired.InsertKey(cm.ID)
 	c.controlMap[cm.ID] = append(c.controlMap[cm.ID], cm)
 }
 
 func (c *ControlMap) Get(ID string) []*UpdateControlMap {
 	c.mutex.Lock()
 	defer c.mutex.Unlock()
-	return c.controlMap[ID]
+	m := c.controlMap[ID]
+	// Prune the control map based on the UpdateCOntrolMapTimeout expiry
```
oleorhagen, line 336: This is the only place I currently expire maps

app/updatemanager.go
```diff
@@ -281,6 +289,10 @@ func (m UpdateControlMap) Sanitize() {
 	}
 }
 
+func (u *UpdateControlMap) Expired() bool {
+	return time.Now().After(u.expiry)
```
oleorhagen, line 293: The ticket says to explicitly store the monotonic time, but the way I'm doing it now, I'm not really respecting it...
However, I see nothing wrong with this approach as is...

app/updatemanager.go
```diff
@@ -301,11 +319,66 @@ func (c *ControlMap) Set(cm *UpdateControlMap) {
...
 
 func (c *ControlMap) Get(ID string) []*UpdateControlMap {
 	c.mutex.Lock()
 	defer c.mutex.Unlock()
-	return c.controlMap[ID]
+	m := c.controlMap[ID]
+	// Prune the control map based on the UpdateCOntrolMapTimeout expiry
+	// FIXME - This should probably be handled some other way, like a
+	// separate goroutine (?)
+	for i, e := range m {
```
oleorhagen, line 339: This probably needs to be separated out, so that the expired map does not check it's own expiries

app/updatemanager.go
lluiscampos: I think you are actually using monotonic time.

From the [docs](https://golang.org/pkg/time/#hdr-Monotonic_Clocks):
"If Times t and u both contain monotonic clock readings, the operations t.After(u), t.Before(u), t.Equal(u), and t.Sub(u) are carried out using the monotonic clock readings", and in our case both of these are based on `time.Now()` which have monotonic clock reading.

app/updatemanager.go
```diff
@@ -132,6 +135,9 @@ func (u *UpdateManager) run(ctx context.Context) error {
 				UpdateManagerMap.Set(&controlMap)
 			}
 
+			log.Debugf("Setting the control map parameter to: %s", controlMap.ID)
+			expiry := time.Now().Add(time.Duration(u.updateControlTimeoutSeconds) * time.Second)
+			UpdateManagerMap.SetAndStamp(&controlMap, expiry)
```
lluiscampos, line 140: This does not seem correct to me, we should `Set` in once. Is it WIP or am I miss-understanding?

app/updatemanager.go
```diff
@@ -281,6 +289,10 @@ func (m UpdateControlMap) Sanitize() {
 	}
 }
 
+func (u *UpdateControlMap) Expired() bool {
+	return time.Now().After(u.expiry)
+}
+
 type ControlMap struct {
```
lluiscampos, line 296: Reading the code now, I think this should be `ControlMapPool` or something like that. Methods like `Delete` or `Contains` would read better.

app/updatemanager.go
```diff
@@ -301,11 +319,66 @@ func (c *ControlMap) Set(cm *UpdateControlMap) {
...
+	if !ok {
+		return c
+	}
+	for i, e := range m {
+		if e == cm {
+			log.Debugf("Deleting %#v from the expired map", cm)
+			m = append(m[:i], m[i+1:]...)
+		}
+	}
+	return c
```
lluiscampos, line 365: With regards to our conversation on copy vs mutation.

This one is very strange to me: we are actually mutating (pointer receiver method) and then returning itself? Can you explain me the benefit?

lluiscampos:
I have dropped some comments (minor things), but I cannot figure out the main problem you are asking me here.

> Read the ticket: https://tracker.mender.io/browse/MEN-4547 _thoroughly_
> 
> Some points which is explicitly written, but from reading this and the other ticket's I don't fully understand why is needed are the complete separation of the expired maps, and the once still valid. Like why can they not simply be triggered by a boolean?

In MEN-4548 seems that we need to take a decision based on both maps. But as the maps seem to need to be unique (ie when a new map is set we would delete the matching one in the expired pool) I also see your point that a simple boolean should do. Are we overlooking something?

> 
> This would also help with this point:
> 
>     * The expired pool must have the same primary keys and thread safety as the regular pool.
> 
>     * When UpdateControlMapExpirationTimeSeconds elapses without an update, the map is considered expired, and the map must be moved to a separate pool of expired maps.
> 
> 
> This essentially requires monitoring of the maps in a separate goroutine, if they are required to _at all time_ be correct.
> The way I've done it now is to simply check expiration on the `_Get_` call, which means it will (as far as I can tell), behave correctly (this is the only way to extract a map from either pool), but I will need some consensus here I think :)

I have similarly "useless" input on this point: I don't see why we need to expire them at real time if there is a single point of entry for getting a map (the `Get` method). So again, I share your point of view...

app/updatemanager.go
oleorhagen: Nope, it's supposed to be a copy, fixing

app/updatemanager.go
oleorhagen: > `we should Set in once`
I'm sorry, I don't fully understand what you mean here? :thinking:

app/updatemanager.go
oleorhagen: Yes, agree

app/updatemanager.go
oleorhagen: It was just a quick hack to get it to compile at the time :)

app/updatemanager.go
lluiscampos: That we use `Set` in line 135, and then `SetAndStamp` on line 140. This is probably the result of a wrong conflict resolution when pulling in my changes...

oleorhagen: Oh, I see :)

No it is not:
* The time is updated to the current time each time the map is updated by an outside source (not by Mender itself).
So only maps from the DBus inlet should be stamped

oleorhagen:
Note: Went for the double map implementation, it turned out cleaner imo

oleorhagen:
@kacf is this sort of what you had in mind?

app/updatemanager.go
```diff
@@ -149,9 +150,18 @@ type UpdateControlMapState struct {
...
+	Priority         int                              `json:"priority"`
+	States           map[string]UpdateControlMapState `json:"states"`
+	lastModifiedTime time.Time
+	expiryTime       time.Time
+}
+
+func (u *UpdateControlMap) Stamp(updateControlTimeoutSeconds int) *UpdateControlMap {
+	u.lastModifiedTime = time.Now()
+	u.expiryTime = u.lastModifiedTime.Add(
+		time.Duration(updateControlTimeoutSeconds) * time.Second)
```
lluiscampos, line 163: If both `lastModifiedTime` and `expiryTime` are set only here, we should be able to remove one of them. :thinking:

oleorhagen: Yes, I've been thinking the same, depends a bit on what is actually needed down the line

app/updatemanager.go
```diff
@@ -281,31 +291,92 @@ func (m UpdateControlMap) Sanitize() {
...
-func (c *ControlMap) Get(ID string) []*UpdateControlMap {
+func (c *ControlMapPool) Get(ID string) (active []*UpdateControlMap, expired []*UpdateControlMap) {
 	c.mutex.Lock()
 	defer c.mutex.Unlock()
-	return c.controlMap[ID]
+	c.Active[ID], c.Expired[ID] = expireActiveMap(c.Active[ID])
+	return c.Active[ID], c.Expired[ID]
+}
+
+func (c *ControlMapPool) Clear() {
```
kacf, line 355: Ambiguous name, I think it should be called `ClearExpired`.

app/updatemanager_test.go
```diff
@@ -363,3 +366,203 @@ func TestUpdateManager(t *testing.T) {
...
+			"ArtifactInstall": UpdateControlMapState{
+				Action: "continue",
+			},
+		},
+	}
+	testMapPool.Insert(cm.Stamp(0))
+	testMapPool.Insert(cmn.Stamp(0))
+	testMapPool.Clear()
+	assert.Equal(t, 0, len(testMapPool.Active["foo"]), testMapPool.Active["foo"])
+	assert.Equal(t, 0, len(testMapPool.Expired["foo"]))
```
kacf, line 505: Maybe mix in one active map too, so that we know it doesn't just clear all maps?

app/updatemanager_test.go
```diff
@@ -363,3 +366,203 @@ func TestUpdateManager(t *testing.T) {
...
+		States: map[string]UpdateControlMapState{
+			"ArtifactReboot": UpdateControlMapState{
+				Action: "continue",
+			},
+		},
+	}).Stamp(3))
+	testMapPool.Get("foo")
+	assert.Contains(t, testMapPool.Expired, "foo")
+	assert.Contains(t, testMapPool.Expired, "bar")
+	assert.Contains(t, testMapPool.Expired, "baz")
```
kacf, line 567: Here also, maybe mix in one active map so we can test that too?

app/updatemanager.go
```diff
@@ -281,31 +291,92 @@ func (m UpdateControlMap) Sanitize() {
...
+	for key := range c.Active {
+		active, _ := expireActiveMap(c.Active[key])
+		c.Active[key] = active
+	}
+	c.Expired = make(map[string][]*UpdateControlMap)
```
kacf, line 360: You need a lock around this.

kacf (changes requested):
Nice work in my absence!

> > Some points which is explicitly written, but from reading this and the other ticket's I don't fully understand why is needed are the complete separation of the expired maps, and the once still valid. Like why can they not simply be triggered by a boolean?
> 
> In MEN-4548 seems that we need to take a decision based on both maps. But as the maps seem to need to be unique (ie when a new map is set we would delete the matching one in the expired pool) I also see your point that a simple boolean should do. Are we overlooking something?

You are not! Indeed a bool is enough. Sorry, it is not always easy to know how the criteria will be interpreted. *Semantically*, they are two different pools, but separating them with a bool is perfectly fine, and would probably simplify several constructs in this PR.

> > This essentially requires monitoring of the maps in a separate goroutine, if they are required to _at all time_ be correct.
> > The way I've done it now is to simply check expiration on the `_Get_` call, which means it will (as far as I can tell), behave correctly (this is the only way to extract a map from either pool), but I will need some consensus here I think :)
> 
> I have similarly "useless" input on this point: I don't see why we need to expire them at real time if there is a single point of entry for getting a map (the `Get` method). So again, I share your point of view...

Unfortunately the event loop will need to be spun when a map is expiring, so this needs to happen in real time. This was somewhat described in [MEN-4549](https://tracker.mender.io/browse/MEN-4549), but I added an explicit point about it just now. Strictly speaking you don't *have* to address it in this PR, since it's not part of the MEN-4547 task, but it is going to be needed. I'll leave it to you decide what's easiest.

app/updatemanager.go
```diff
@@ -281,31 +291,92 @@ func (m UpdateControlMap) Sanitize() {
...
+		return false
+	}
+	if u.Priority != other.Priority {
+		return false
+	}
+	return true
+}
+
+func (u *UpdateControlMap) String() string {
+	return fmt.Sprintf("ID: %s: %d Priority ", u.ID, u.Priority)
```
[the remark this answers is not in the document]
oleorhagen: haha, yes.
Football score style? :p

oleorhagen:
@kacf new implementation, based around the query function, you will find this a lot easier to read and understand I think (hope) :)

app/updatemanager.go
```diff
@@ -149,9 +150,22 @@ type UpdateControlMapState struct {
...
+	expired    bool
+}
+
+func (u *UpdateControlMap) Stamp(updateControlTimeoutSeconds int) *UpdateControlMap {
+	u.expiryTime = time.Now().Add(
+		time.Duration(updateControlTimeoutSeconds) * time.Second)
+	// expiration monitor
+	go func() {
+		<-time.After(time.Duration(updateControlTimeoutSeconds) * time.Second)
+		u.expired = true
```
kacf, line 166: This doesn't really solve the problem, unfortunately. The issue is that the event loop needs to know when an expiration has happened, and then reevaluate whether it should stay in the current state or not.

But I think I would recommend that you just leave this alone for now. What you have works for the current acceptance criteria, and it should be clearer exactly what you need once you start working on MEN-4549.

kacf (approved):

app/updatemanager.go
oleorhagen: Postponing it till `MEN-4549` :)

Then it'll be either mine, or urs problem :p

oleorhagen:
Merged, the failing `single-file-update-module` test is unrelated.

PR #749, opened 2021-04-30
MEN-4669: Fixed issue where device_type file not consistent with -d flag

NilsOlavKJohansen:
Changelog: Now the device_type file is always consistent with --data flag options

Signed-off-by: Nils Olav Kvelvane Johansen <nils.olav@northern.tech>

### Description

Thank you!

### DIFF
--- a/cli/cli.go
+++ b/cli/cli.go
@@ -421,6 +421,23 @@ func (runOptions *runOptionsType) commonCLIHandler(
 		return nil, nil, err
 	}
 
+	// Make sure that paths that are not configurable via the config file is conconsistent with --data flag
+	config.ArtifactScriptsPath = path.Join(runOptions.dataStore, "scripts")
+	config.ModulesWorkPath = path.Join(runOptions.dataStore, "modules", "v3")
+
+	// Checks if the DeviceTypeFile is defined in config file.
+	if config.MenderConfigFromFile.DeviceTypeFile != "" {
+		// Sets the config.DeviceTypeFile to the value in config file.
+		config.DeviceTypeFile = config.MenderConfigFromFile.DeviceTypeFile
+
+	} else {
+		// If --data flag is not used then dataStore is /var/lib/mender
+		config.MenderConfigFromFile.DeviceTypeFile = path.Join(
+			runOptions.dataStore, "device_type")
+		config.DeviceTypeFile = path.Join(
+			runOptions.dataStore, "device_type")
+	}
+
 	// Skip verify for setup, as the configuration will be overridden
 	if ctx.Command.Name != "setup" {
 		err := config.Validate()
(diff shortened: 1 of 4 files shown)


kacf (approved):

NilsOlavKJohansen:

conf/config.go
```diff
@@ -95,14 +95,12 @@ type DBusConfig struct {
 
 func NewMenderConfig() *MenderConfig {
 	return &MenderConfig{
-		MenderConfigFromFile: MenderConfigFromFile{
-			DeviceTypeFile: DefaultDeviceTypeFile,
```
kacf, line 99: The reason this is removed is because we cannot provide a static default, we have to consult the `runOptions.dataStore` path, correct? If so, I think we should also remove the `DefaultDeviceTypeFile` constant.

cli/cli.go
```diff
@@ -421,6 +421,23 @@ func (runOptions *runOptionsType) commonCLIHandler(
 		return nil, nil, err
 	}
 
+	// Make sure that paths that are not configurable via the config file is conconsistent with --data flag
+	config.ArtifactScriptsPath = path.Join(runOptions.dataStore, "scripts")
+	config.ModulesWorkPath = path.Join(runOptions.dataStore, "modules", "v3")
```
kacf, line 426: Good catch with these also!

kacf (changes requested):

cli/cli.go
NilsOlavKJohansen: :+1:

conf/config.go
NilsOlavKJohansen: That's right, but DefaultDeviceTypeFile is still used in cli/setup.go.
Then maybe I should make some changes here then?
```
// Make sure devicetypefile and serverURL is set
	if config.DeviceTypeFile == "" {
		// Default devicetype file as defined in device.go
		config.DeviceTypeFile = conf.DefaultDeviceTypeFile
	}

```

kacf: Oh, right! Hmm, I think it would not be appropriate to use the same construct here, as `setup` is supposed to be run completely from scratch, with no prior configuration.

What I suggest is this: *Move* the `DefaultDeviceTypeFile` to `setup.go` instead, so that it is confined there, and change the code to use that reference instead. Does that make sense?

NilsOlavKJohansen: Good idea, i'll do that!

kacf (approved):

NilsOlavKJohansen:

kacf:
You're right that it's a bit risky, but it's also a somewhat serious bug (telling it to write in one place, and then it writes in another, is quite bad, IMHO). We test the data directory writing rather well in our tests, so hopefully we should avoid regressions. You have my approval to go all the way! :slightly_smiling_face:

oleorhagen:
PR(s):
* 2.6.x: https://github.com/mendersoftware/mender/pull/777
* 2.5.x: https://github.com/mendersoftware/mender/pull/778
* 2.4.x: https://github.com/mendersoftware/mender/pull/779 (Not really needed, but, yeh)

PR #751, opened 2021-05-03
DBus docs: Elaborate a bit more and document on_action_executed. 

kacf:
Also merge with master. Only the last commit is new.


lluiscampos (approved):
Makes sense

PR #758, opened 2021-05-10
MEN-4619: Add saving of Update Control Maps to local database.

kacf:
Based on #755. The main commit is the title, the rest are support commits, best reviewed individually.

### DIFF
--- a/app/updatemanager.go
+++ b/app/updatemanager.go
@@ -46,19 +46,15 @@ const (
 	    </node>`
 )
 
-var UpdateManagerMapPool *ControlMapPool
-
-func init() {
-	UpdateManagerMapPool = NewControlMap()
-}
-
 type UpdateManager struct {
 	dbus                        dbus.DBusAPI
+	controlMapPool              *ControlMapPool
 	updateControlTimeoutSeconds int
 }
 
-func NewUpdateManager(updateControlTimeoutSeconds int) *UpdateManager {
+func NewUpdateManager(controlMapPool *ControlMapPool, updateControlTimeoutSeconds int) *UpdateManager {
 	return &UpdateManager{
+		controlMapPool:              controlMapPool,
 		updateControlTimeoutSeconds: updateControlTimeoutSeconds,
 	}
 }
@@ -131,7 +127,7 @@ func (u *UpdateManager) run(ctx context.Context) error {
 				log.Debugf("Ignoring UpdateControlMap %s with no non-default States", controlMap.ID)
 			} else {
 				log.Debugf("Setting the control map parameter to: %s", controlMap.ID)
-				UpdateManagerMapPool.Insert(controlMap.Stamp(u.updateControlTimeoutSeconds))
+				u.controlMapPool.Insert(controlMap.Stamp(u.updateControlTimeoutSeconds))
 			}
 
 			return u.updateControlTimeoutSeconds / 2, nil
--- a/app/updatemanager.go
+++ b/app/updatemanager.go
@@ -331,10 +331,17 @@ type ControlMapPool struct {
 	store store.Store
 }
 
-func NewControlMap() *ControlMapPool {
-	return &ControlMapPool{
-		Pool: []*UpdateControlMap{},
+// loadTimeout is how far in the future to set the map expiry when loading from
+// the store.
+func NewControlMap(store store.Store, loadTimeout int) *ControlMapPool {
+	pool := &ControlMapPool{
+		Pool:  []*UpdateControlMap{},
+		store: store,
 	}
+
+	pool.loadFromStore(loadTimeout)
+
+	return pool
 }
 
 func (c *ControlMapPool) SetStore(store store.Store) {
@@ -450,7 +457,7 @@ func (c *ControlMapPool) saveToStore() {
 
 // `timeout` is the timeout that should be given to expired maps that we load
 // from disk.
-func (c *ControlMapPool) LoadFromStore(timeout int) {
+func (c *ControlMapPool) loadFromStore(timeout int) {
 	c.mutex.Lock()
 	defer c.mutex.Unlock()
 
--- a/app/updatemanager.go
+++ b/app/updatemanager.go
@@ -18,12 +18,15 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
+	"os"
 	"sort"
 	"strings"
 	"sync"
 	"time"
 
+	"github.com/mendersoftware/mender/datastore"
 	"github.com/mendersoftware/mender/dbus"
+	"github.com/mendersoftware/mender/store"
 	"github.com/mendersoftware/mender/utils"
 
 	"github.com/pkg/errors"
@@ -325,12 +328,17 @@ func (u *UpdateControlMap) String() string {
 type ControlMapPool struct {
 	Pool  []*UpdateControlMap
 	mutex sync.Mutex
+	store store.Store
 }
 
 func NewControlMap() *ControlMapPool {
 	return &ControlMapPool{
 		Pool: []*UpdateControlMap{},
 	}
+}
+
+func (c *ControlMapPool) SetStore(store store.Store) {
+	c.store = store
 }
 
 func (c *ControlMapPool) Insert(cm *UpdateControlMap) {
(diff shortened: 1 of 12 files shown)


kacf:

lluiscampos (approved):
Maybe I am out of training, but I couldn't spot any error... Nice work

app/updatemanager_test.go
```diff
@@ -931,3 +932,63 @@ func TestQueryLogic(t *testing.T) {
...
+	expectedTime := time.Now().Add(time.Duration(boot) * time.Second)
+
+	expiredMap := map[string]bool{
+		"foo": true,
+		"bar": false,
+		"baz": false,
+	}
+
+	assert.Equal(t, 3, len(loadedPool.Pool))
+	for _, m := range loadedPool.Pool {
```
oleorhagen, line 990: Nice test!

app/updatemanager.go
```diff
@@ -333,6 +337,10 @@ func NewControlMap() *ControlMapPool {
 	}
 }
 
+func (c *ControlMapPool) SetStore(store store.Store) {
```
oleorhagen, line 340: Suggestion:
Since this should always be backed by a DB, can this not be part of the initializer?

app/updatemanager.go
```diff
@@ -386,6 +409,81 @@ func (c *ControlMapPool) ClearExpired() {
...
+	active, expired := c.getAll()
+
+	toSave := controlMapPoolDbFormat{
+		// Intentionally not using label assignment here. We want to
+		// know if we're missing any fields.
+		active,
+		expired,
+	}
+
+	data, err := json.Marshal(&toSave)
```
oleorhagen, line 437: Idea:
Why not add a method to the `dbstore`, which looks something like:
```
Write(name string, m json.Marshaler)
```
and then add a `MarshalJSON` to the `controlMapDbFormat`, and writing can be:

```
db.Write(datastore.UpdateControlMaps, maps)
```

app/updatemanager.go
```diff
@@ -386,6 +409,81 @@ func (c *ControlMapPool) ClearExpired() {
...
+	if err != nil {
+		log.Errorf("Could not save Update Control Maps to database: %s", err.Error())
+		// There isn't much we can do if it fails.
+		return
+	}
+}
+
+// `timeout` is the timeout that should be given to expired maps that we load
+// from disk.
+func (c *ControlMapPool) LoadFromStore(timeout int) {
```
oleorhagen, line 456: I think I might have made a mistake with the `stamp` interface. Perhaps it would have been better if it took a `time.Duration` parameter all along?

conf/config.go
```diff
@@ -189,6 +191,15 @@ func (c *MenderConfig) Validate() error {
 	return nil
 }
 
+func applyConfigDefaults(config *MenderConfig) {
+	if config.MenderConfigFromFile.UpdateControlMapExpirationTimeSeconds == 0 {
```
oleorhagen, line 201: Thx :smile_cat:

cli/commands.go
```diff
@@ -66,7 +67,26 @@ var (
 	ErrSIGTERM           = errors.New("Daemon terminated with SIGTERM")
 )
 
-func commonInit(config *conf.MenderConfig, opts *runOptionsType) (*app.MenderPieces, error) {
+func initDualRootfsDevice(config *conf.MenderConfig) installer.DualRootfsDevice {
```
oleorhagen, line 70: I like it!

oleorhagen (approved):
Nice work

I've only left nitpicks, and points for discussion/potential polishing.

Feel free to ignore them as you see fit :smile_cat:

app/updatemanager.go
```diff
@@ -386,6 +409,81 @@ func (c *ControlMapPool) ClearExpired() {
...
+
+type controlMapPoolDbFormat struct {
+	Active  []*UpdateControlMap `json:"active"`
+	Expired []*UpdateControlMap `json:"expired"`
+}
+
+// Must only be called from functions that have already locked the mutex.
+func (c *ControlMapPool) saveToStore() {
+	if c.store == nil {
+		// Saving not enabled. This should not happen outside tests.
```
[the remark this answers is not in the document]
kacf: I considered it, but decided not to. If `c.store == nil`, then the program will crash because this check does not exist in `LoadFromStore`, which is part of the normal execution flow (but not part of most tests). So we will get to know about it either way, and a log message here would only clutter the test output.

app/updatemanager.go
```diff
@@ -386,6 +409,81 @@ func (c *ControlMapPool) ClearExpired() {
...
+
+// `timeout` is the timeout that should be given to expired maps that we load
+// from disk.
+func (c *ControlMapPool) LoadFromStore(timeout int) {
+	c.mutex.Lock()
+	defer c.mutex.Unlock()
+
+	data, err := c.store.ReadAll(datastore.UpdateControlMaps)
+	if errors.Is(err, os.ErrNotExist) {
+		return
```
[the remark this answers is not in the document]
kacf: You're right, we should log at debug level. Will add it!

app/updatemanager.go
kacf: The primary motivation was to avoid needing this in every test. I can agree that this is perhaps less clean than it could be though, so maybe we should not let test code dictate this. If fixed, this would also remove the need for the `c.store == nil` check that you pointed out in your first comment.

I'll see if I can fix it without too much trouble.

app/updatemanager.go
```diff
@@ -386,6 +409,81 @@ func (c *ControlMapPool) ClearExpired() {
...
+// Must only be called from functions that have already locked the mutex.
+func (c *ControlMapPool) saveToStore() {
+	if c.store == nil {
+		// Saving not enabled. This should not happen outside tests.
+		return
+	}
+
+	active, expired := c.getAll()
+
+	toSave := controlMapPoolDbFormat{
```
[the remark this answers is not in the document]
oleorhagen: It seems we're using `DB` other places in the source :)

app/updatemanager.go
```diff
@@ -386,6 +409,81 @@ func (c *ControlMapPool) ClearExpired() {
...
+		// Intentionally not using label assignment here. We want to
+		// know if we're missing any fields.
```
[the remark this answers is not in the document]
oleorhagen: Then I'm afraid I've misunderstood, you want it to warn you when you're missing a field in the struct?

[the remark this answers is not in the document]
kacf: Exactly. Since this is saved in the database, it might potentially be long lived and bugs in this area might "survive" from one mender version to the next even if fixed in the new version. That's why I want to be really pedantic in checking everything.

app/updatemanager.go
kacf: I like the abstraction, but I think it will not save a lot, at least not as long as we're only marshalling pure JSON, and only once. It's just moving the code to a `MarshallJSON` function, but it's still the same code. Since I also have to add it to the other store types then, I think I will save this for another day.

app/updatemanager.go
kacf: Instead of `int`, you mean? Maybe, but I think it's fine as it is. I already made the mistake that `time.Duration(1)` is one nanosecond, not one second, which is what we use as the unit in the config file. So using the int, which forces conversion, is a nice reminder to multiply it with `time.Second`.

app/updatemanager.go
oleorhagen: Yeh, it's a little out of scope :)

I have a feeling we're doing this a lot of place throughout our codebase though, like don't we always store JSON basically in the DB? :p

app/updatemanager.go
oleorhagen: Makes sense. Let's keep it

app/updatemanager_test.go
```diff
@@ -518,8 +518,8 @@ func TestInsertMatching(t *testing.T) {
 			},
 		},
 	}
-	testMapPool.Insert(cm.Stamp(0))
-	testMapPool.Insert(cmn.Stamp(0))
+	testMapPool.Insert(cm.Stamp(60))
```
oleorhagen, line 522: Oh... nice catch

app/updatemanager_test.go
```diff
@@ -954,6 +954,8 @@ func TestSaveAndLoadToDb(t *testing.T) {
 			},
 		},
 	}).Stamp(expired))
+	// Give it a second to expire.
```
oleorhagen, line 957: Ref the commit message:
This is the TestSaveAndLoadToDb test, no?
Also: ToDB?

oleorhagen (approved):
:star2: 
Looking good!

kacf:
All remaning errors are spurious. Merging.

PR #768, opened 2021-05-12
MEN-4549: Implement the continue/pause/fail state machine logic

oleorhagen:


app/mender.go
```diff
@@ -466,6 +466,34 @@ func (m *Mender) TransitionState(to State, ctx *StateContext) (State, bool) {
 	return transitionState(to, ctx, m)
 }
 
+func spinEventLoop(c *ControlMapPool, ctx *StateContext, to State) (State, bool) {
```
oleorhagen, line 469: So far untested, but I'd want some feedback on the approach, is this sort of correct?
i.e., passing in the `Transition` name from all transitions to `QueryAndUpdate`?
and returning the `NewUpdateReportState` in case of error?

app/mender.go
```diff
@@ -483,6 +511,11 @@ func transitionState(to State, ctx *StateContext, c Controller) (State, bool) {
 		}
 	}
 
+	spin := true
```
oleorhagen, line 514: sort of horrid code, looking at it now, this might as well be in the `spinEventLoop` function...

app/updatemanager.go
```diff
@@ -326,17 +342,19 @@ func (u *UpdateControlMap) String() string {
...
 
 // loadTimeout is how far in the future to set the map expiry when loading from
 // the store.
 func NewControlMap(store store.Store, loadTimeout int) *ControlMapPool {
 	pool := &ControlMapPool{
-		Pool:  []*UpdateControlMap{},
-		store: store,
+		Pool:    []*UpdateControlMap{},
+		store:   store,
+		Updates: make(chan struct{}, 1),
```
oleorhagen, line 357: Give the channel a 1-size buffer, and simply ignore more notifications filling up. Hence the non-blocking:
`select{
<- write
}`
calls

app/updatemanager.go
```diff
@@ -373,7 +398,7 @@ func (c *ControlMapPool) Get(ID string) (active []*UpdateControlMap, expired []*
 		c.Pool,
 		// Collector
 		func(u *UpdateControlMap) {
-			if u.expired {
+			if u.Expired() {
```
oleorhagen, line 401: `Expirations` are now guarded by a separate `mutex`, which should resolve the issue we've been discussing

app/updatemanager_test.go
```diff
@@ -935,8 +935,6 @@ func TestQueryLogic(t *testing.T) {
 }
 
 func TestSaveAndLoadToDB(t *testing.T) {
-	// MEN-4549: Reenable this after expiry mechanics have been fixed.
-	return
```
oleorhagen, line 939: Works now :smile:

app/mender.go
kacf: The principle is very much in line with how I envisioned it, though there are some details that need to be worked out. See individual comments.

app/mender.go
```diff
@@ -466,6 +466,34 @@ func (m *Mender) TransitionState(to State, ctx *StateContext) (State, bool) {
 	return transitionState(to, ctx, m)
 }
 
+func spinEventLoop(c *ControlMapPool, ctx *StateContext, to State) (State, bool) {
+	for {
+		switch to.Transition() {
+		case ToArtifactReboot_Enter, ToArtifactCommit_Enter, ToArtifactInstall:
```
kacf, line 472: It's important that this handling only happens at `ArtifactInstall_Enter`. Does this happen at every internal transition within `ArtifactInstall`, including Leave? I don't remember exactly how these mechanics work, but this should be checked. It's possible we have to create a new const for the individual `Enter` and `Leave` states.

app/mender.go
```diff
@@ -466,6 +466,34 @@ func (m *Mender) TransitionState(to State, ctx *StateContext) (State, bool) {
...
+			case "continue":
+				return to, false
+			case "pause":
+				// wait until further notice
+				<-c.Updates
+			case "fail":
+				// Fail (the update) immediately
+				// TODO/FIXME
+				// Simply set the to state to failed ? Hmmm... not sure if this is right
+				sd, err := datastore.LoadStateData(ctx.Store)
```
kacf, line 484: Yeah, it feels like the wrong place to do this. But each state has a `HandleError` function which is called whenever a state script fails. In principle this is the same kind of failure, so could we reuse that? See [this part](https://github.com/mendersoftware/mender/pull/768/commits/a578cefe4abe503915908405d5562c9eb3702f0f#diff-31fdb7c71693c702fb30629bde922bafc9da308305ac0330597849769d7e1dedR533-R537).

app/mender.go
kacf: Yes, and it looks like you've already done it, so this can be removed. I agree it's better to spin inside the function instead of here.

app/updatemanager.go
kacf: Makes sense!

app/updatemanager.go
```diff
@@ -150,22 +151,35 @@ type UpdateControlMapState struct {
...
+}
+
+func (u *UpdateControlMap) expire(in time.Duration) {
+	// TODO - Add a way to remove the timer upon object deletion from the pool
 	go func() {
-		<-time.After(time.Duration(updateControlTimeoutSeconds) * time.Second)
+		<-time.After(in)
+		u.mutex.Lock()
+		defer u.mutex.Unlock()
 		u.expired = true
```
kacf, line 177: I haven't found any definitive answer online, but from what I can piece together, the lock triggers a "synchronization event" which flushes the processor cache and registers. So *hopefully* this should be safe. And mutexes would be pretty useless if it wasn't, so let's hope this common sense saves us! :-)

app/updatemanager.go
```diff
@@ -296,6 +310,8 @@ func (m UpdateControlMap) Sanitize() {
 }
 
 func (u *UpdateControlMap) Expired() bool {
+	u.mutex.Lock()
+	defer u.mutex.Unlock()
 	return u.expired
```
kacf, line 315: :+1:

app/updatemanager_test.go
```diff
@@ -993,3 +993,50 @@ func TestSaveAndLoadToDB(t *testing.T) {
...
+		testMapPool.Insert((&UpdateControlMap{
+			ID:       "foo",
+			Priority: 2,
+			States: map[string]UpdateControlMapState{
+				"ArtifactInstall": UpdateControlMapState{
+					Action: "force_continue",
+				},
+			},
+		}).Stamp(1))
+		assert.Eventually(t,
```
kacf, line 1007: I think `Eventually` won't work correctly unless you `select` with a `default` case. If the announcement doesn't work, then this will just hang forever.

app/updatemanager_test.go
```diff
@@ -993,3 +993,50 @@ func TestSaveAndLoadToDB(t *testing.T) {
...
+					Action: "force_continue",
+				},
+			},
+			expirationChannel: testMapPool.Updates,
+		}
+		m.Stamp(1)
+		assert.Eventually(t,
+			func() bool {
+				<-testMapPool.Updates
+				return true
```
kacf, line 1033: Same here.

app/updatemanager_test.go
oleorhagen: No, it will time out after `2 seconds`, tested it

kacf: Ok, then there is something I don't understand about `Eventually()`. :-)

app/mender.go
oleorhagen: It looks like `ToArtifactInstall` is enough, considering this only accepts the `to State` transition :)

oleorhagen:
@kacf ready for first round review.

Notable things left out, the timers are not cancelled, even after the `ControlMap` is removed from the pool. 
But feel free to do the review as is, I will fix this in parallell, should not touch the existing functionality at all.

kacf (approved):
Beautiful!

oleorhagen:

oleorhagen:
All passed, only success reports missing. Merging

PR #770, opened 2021-05-18
MEN-3978: Implement support for non-U-Boot tool names.

kacf:
The tools still have to be command line compatible with the U-Boot
tools (either u-boot-fw-utils or libubootenv), but the names can be
different. This allows having U-Boot tools installed alongside
grub-mender-grubenv tools, whose new names are
`grub-mender-grubenv-set` and `grub-mender-grubenv-print`, instead of
`fw_setenv` and `fw_printenv`.

The two new configuration settings `BootUtilitiesSetActivePart` and
`BootUtilitiesGetNextActivePart` have been introduced to configure the
names. If no names are set, then the default is to try the
grub-mender-grubenv tools first, followed by the "fw_" tools if the
former are not found.

Changelog: Commit


kacf:
This goes together with https://github.com/mendersoftware/grub-mender-grubenv/pull/16.

I'm still not sure about the configuration names, but there is a discussion going on in [MEN-3978](https://tracker.mender.io/browse/MEN-3978).

kacf:

installer/bootenv.go
```diff
@@ -151,22 +167,43 @@ func (e *UBootEnv) writeEnvImpl(vars BootVars, separator string) error {
...
+	var setEnvCmd *exec.Cmd
+	var pipe io.WriteCloser
+	var err error
+	found := false
+	for _, executable := range e.setCommand {
+		// Make environment update atomic by using "-s" option.
+		setEnvCmd = e.Command(executable, "-s", "-")
+		pipe, err = setEnvCmd.StdinPipe()
+		if err != nil {
+			log.Errorln("Could not set up pipe to fw_setenv command: ", err)
```
lluiscampos, line 179: Something like:

```suggestion
			log.Errorf("Could not set up pipe to %s command: %s", executable, err)
```

lluiscampos (approved):
Looks good!

installer/bootenv.go
```diff
@@ -151,22 +167,43 @@ func (e *UBootEnv) writeEnvImpl(vars BootVars, separator string) error {
...
+			return err
+		}
+		err = setEnvCmd.Start()
+		if errors.Is(err, exec.ErrNotFound) {
+			log.Debugf("Tried command %q, but it was not found", executable)
+			continue
+		} else if err != nil {
+			log.Errorln("Could not execute fw_setenv: ", err)
+			pipe.Close()
+			if os.Geteuid() != 0 {
```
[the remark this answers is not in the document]
kacf: No, it's recommended practice to always try first, and check conditions afterwards. This is because you don't always know what goes on in kernel space. Special arrangements may have taken place to ensure that it works even without root rights, such CAP settings or setuid binaries.

kacf:
@MuchoLucho: While I was working on this I decided to add some of the root filesystem checks myself, check [this one out](https://github.com/mendersoftware/mender/pull/770/commits/f82e1b2a7709206744e115beb16b13144eaec991). It won't match your use case exactly, but it may help you understand what I was talking about. You would need to determine the `active` variable from the slot number.

kacf:
@TheMeaningfulEngineer may also be interested ^

tests/rootfs-image-v2
```diff
@@ -21,24 +21,35 @@ fi
...
     passive=$MENDER_ROOTFS_PART_B
     passive_num=$MENDER_ROOTFS_PART_B_NUMBER
 else
+    active=$MENDER_ROOTFS_PART_B
     passive=$MENDER_ROOTFS_PART_A
     passive_num=$MENDER_ROOTFS_PART_A_NUMBER
 fi
 
+ensure_correct_root_mounted() {
+    if [ "$(stat -L -c %02t%02T $active)" != "$(stat -L -c %04D /)" ]; then
```
TheMeaningfulEngineer, line 34: What's happening here (the logic with the `stat-s`)? :)

TheMeaningfulEngineer:
Thanks @kacf 

I gave it a shot just for how the fully generic commands could be named.
So instead of setenv/getenv these actions would be used.

```
GET_ACTIVE_PARTITION_NUMBER
# When executed this returns the number of the currently running partition
# For uboot/grub this is "$(${PRINTENV} mender_boot_part)"

FORCE_BOOT_FROM_INACTIVE_ON_NEXT_REBOOT
# When executed on the active partition it will make sure that during next restart the inactive partition will be booted
# For uboot/grub 
#
# ${SETENV} -s - <<EOF
# mender_boot_part $passive_num
# upgrade_available 1
# bootcount 0
# EOF
```

It's just a suggestion predominately to help me understand the logic better. 
The two above are simple to conceptually define.

What I'm struggling to explain in simple language is the conditions involving checks for `upgrade_available` in `ArtifactVerifyReboot`, `ArtifactVerifyRollbackReboot` and `ArtifactCommit`.

i.e. How would you call a function which does this `test "$(${PRINTENV} upgrade_available)" != "upgrade_available=1"`?
Care to give it a try? :)

kacf:
> Thanks @kacf
> 
> I gave it a shot just for how the fully generic commands could be named.
> So instead of setenv/getenv these actions would be used.
> 
> ```
> GET_ACTIVE_PARTITION_NUMBER
> # When executed this returns the number of the currently running partition
> # For uboot/grub this is "$(${PRINTENV} mender_boot_part)"
> 
> FORCE_BOOT_FROM_INACTIVE_ON_NEXT_REBOOT
> # When executed on the active partition it will make sure that during next restart the inactive partition will be booted
> # For uboot/grub 
> #
> # ${SETENV} -s - <<EOF
> # mender_boot_part $passive_num
> # upgrade_available 1
> # bootcount 0
> # EOF
> ```
> 
> It's just a suggestion predominately to help me understand the logic better.
> The two above are simple to conceptually define.

Correct!

> What I'm struggling to explain in simple language is the conditions involving checks for `upgrade_available` in `ArtifactVerifyReboot`, `ArtifactVerifyRollbackReboot` and `ArtifactCommit`.
> 
> i.e. How would you call a function which does this `test "$(${PRINTENV} upgrade_available)" != "upgrade_available=1"`?
> Care to give it a try? :)

That specific test checks whether the boot loader is currently set to boot an **uncommitted** partition. The opposite tests whether we are on a **committed** partition. An uncommitted partition will be automatically rolled back by the boot loader if we reboot more than once, whereas a committed one is permanent.

After the first reboot, we expect to be on the uncommitted partition (still being "tested", so to speak). If we roll back, there will be a second reboot, after which we should be on the committed partition.

So to break it down, here's what we expect of each state:

* `ArtifactVerifyReboot` - Uncommitted
* `ArtifactCommit` - Uncommitted (it will be committed immediately after, by the `SETENV`)
* `ArtifactRollback` - Uncommitted
* `ArtifactVerifyRollbackReboot` - Committed (the old partition)

TheMeaningfulEngineer:
I always find this `upgrade_available` confusing me in constructing models on how things work. :)
Let me try to think out loud.

The [definition of upgrade_availabe from the docs](https://github.com/ARM-software/u-boot/blob/master/drivers/bootcount/Kconfig#L43) it is:

> If "upgrade_available" is
> 	  0, "bootcount" is always 0, if "upgrade_available" is
> 	  1 "bootcount" is incremented in the environment.
> 	  So the Userspace Application must set the "upgrade_available"
> 	  and "bootcount" variable to 0, if a boot was successfully.

So `upgrade_availabe` is used for multiple things:
- start counting boots in uboot once `upgrade_available=1`
- get info that an update is taking place when reading `upgrade_available=1`
- commit an update by setting `upgrade_available=0`

---

@kacf 
Ok, things are clearer now.
Let me try again.
Just using these function examples as communication aids :).
Is this correct?

```
function get_bootloader_active_partition_number() {
    # When executed this returns the number of the currently running partition
    ${PRINTENV} mender_boot_part
}

function get_real_active_partition_number() {
    # We never do this directly. It's indirectly used in the check.
    # https://github.com/mendersoftware/mender/blob/f82e1b2a7709206744e115beb16b13144eaec991/tests/rootfs-image-v2#L34 
    # I'm not sure we need it, just put it here to clearly define "real active partition number" and "bootloader active partition number"
   # Assuming this was skipped because of the error prone bash parsing needed to get the number?
    

function set_partition_to_boot_after_restart () {
    local partition_number=$1
    ${SETENV} -s - <<EOF
mender_boot_part $partition_number
upgrade_available 1
bootcount 0
EOF
}

function commit_success_after_restart () {
    ${SETENV} upgrade_available 0
}

function is_update_ongoing () {
    if test "$(${PRINTENV} upgrade_available)" != "upgrade_available=1"; then
        return 1
    else
        return 0
    fi
}
```

kacf:
> Is this correct?

@TheMeaningfulEngineer: Yes! :slightly_smiling_face:

PR #775, opened 2021-05-21
MEN-4558: Consume the UpdateControlMap from the update response

oleorhagen:


app/mender.go
```diff
@@ -275,6 +276,35 @@ func verifyArtifactDependencies(
 	return nil
 }
 
+type UpdateResponse struct {
```
oleorhagen, line 279: Update response parsing has now been moved to `mender.go`. Which is really our only option AFAICT.

app/state.go
```diff
@@ -645,6 +645,10 @@ func (u *updateCheckState) Handle(ctx *StateContext, c Controller) (State, bool)
 
 	update, err := c.CheckUpdate()
 
+	// Insert the UpdateControlMap into the pool if present
```
oleorhagen, line 648: The other alternative is doing it here. I vied for `mender.go`, but all ears here :ear:

app/updatemanager.go
```diff
@@ -387,22 +387,40 @@ func (c *ControlMapPool) Get(ID string) (active []*UpdateControlMap, expired []*
...
+		},
+	)
+	c.Pool = filteredPool
+}
+
 func (c *ControlMapPool) ClearExpired() {
 	c.mutex.Lock()
 	defer c.mutex.Unlock()
-	newPool := []*UpdateControlMap{}
+	filteredPool := []*UpdateControlMap{}
```
oleorhagen, line 411: Cleaning up out of band

client/client_update.go
```diff
@@ -66,6 +66,7 @@ func (u *CurrentUpdate) MarshalJSON() ([]byte, error) {
 	}
 	u.Provides["artifact_name"] = u.Artifact
 	u.Provides["device_type"] = u.DeviceType
+	u.UpdateControlMap = true
```
oleorhagen, line 69: Always set to true for newer clients henceforth. A little too magic this way maybe? :thinking:

client/client_update_test.go
```diff
@@ -148,6 +125,28 @@ func (d *testReadCloser) Close() error {
...
+		{200, []byte(correctUpdateResponseMultipleDevices), false, true, http.StatusOK},
+		// {200, []byte(updateResponseEmptyDevices), true, false, 0},
+		{204, []byte(""), false, true, http.StatusNoContent},
+		{404, []byte(`{
+		 "error": "Not found"
+		 }`), true, true, http.StatusNotFound},
+		{500, []byte(`{
+		"error": "Invalid request"
+		}`), true, false, 0},
+		// {200, []byte(malformedUpdateResponse), true, false, 0},
```
oleorhagen, line 145: These are obviously not valid anylonger, as the validation has been moved to `mender.go`. If you are happy with the move, then I will re-implement the tests there.

client/client_update_test.go
```diff
@@ -226,34 +225,34 @@ func Test_GetScheduledUpdate_responseMissingParameters_UpdateFailing(t *testing.
 	assert.NoError(t, err)
 }
 
-func Test_GetScheduledUpdate_ParsingResponseOK_updateSuccess(t *testing.T) {
```
oleorhagen, line 229: Same here

client/client_update.go
kacf: This is what is stored in the database, correct?

I think it's not necessary to store this information actually, because it is already stored as part of the active maps. And you are not using it anyway, AFAICS..?

client/test/server.go
```diff
@@ -37,6 +37,7 @@ type updateType struct {
 	Unauthorized bool
 	Called       bool
 	Current      *client.CurrentUpdate
+	ControlMap   interface{}
```
kacf, line 40: I sense there is something I don't understand about why `interface{}` is being used, but I'll ask anyway: Why `interface{}` and not a proper type?

datastore/statedata.go
```diff
@@ -269,6 +270,9 @@ type UpdateInfo struct {
 	// data and discover that it is a different version. See also the
 	// StateDataKeyUncommitted key.
 	HasDBSchemaUpdate bool
+
+	// UpdateControl map from the server FIXME (write proper)
+	UpdateControlMap interface{} `json:"update_control_map"` // TODO - should this be the updatemanager control map (?)
```
kacf, line 275: As mentioned in an earlier comment, I don't think we need to store this.

kacf (changes requested):

datastore/statedata.go
oleorhagen: Yeh, agree. Removed.

client/test/server.go
oleorhagen: Because the 'real' type `UpdateResponse` is now defined in `app/mender.go`, and importing it will cause a cycle :(

client/client_update.go
oleorhagen: No, this is what is sent by the client, when it reports it's own current update.
The point being, this way, newer clients will always have the `update_control_map` field  set to `true` in the request.
(see `client_update.go` for usage). 
But like I said, it is probably a little over the top magic wise. Maybe it should be explicitly set in the initialization of the struct instead :thinking:

app/mender.go
```diff
@@ -325,18 +355,45 @@ func (m *Mender) CheckUpdate() (*datastore.UpdateInfo, menderError) {
...
 		return nil, nil
 	}
-	update, ok := haveUpdate.(datastore.UpdateInfo)
-	if !ok {
-		return nil, NewTransientError(errors.Errorf("not an update response?"))
-	}
 
-	log.Debugf("Received update response: %v", update)
+	var data *UpdateResponse
+	if err := json.Unmarshal(haveUpdate.([]byte), &data); err != nil {
```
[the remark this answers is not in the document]
oleorhagen: I'll do number two, and be defensive :)

client/test/server.go
kacf: Oh right, it's what the talked about, I see it now.

But seeing these pure interfaces everywhere, this really isn't pretty... Maybe we should just bite the bullet and make these two different structs. It's not super nice either, but IMHO the pure `interface{}`s are worse. They're hard to read and understand, and they open up for very easy mistakes later on.

client/client_update_test.go
kacf: Now that I understand it better, I think I vote against this move. The validation of the response really should be done in `client`, that's where it belongs.

I think that validating the response, and validating the Update Control map do not need to happen in the same place. Validating the response is just making sure that it follows the REST API. Whether it's actually a valid map can be checked afterwards in `mender.go`, as part of processing the response it got.

client/client_update.go
kacf: Ah, I completely misread this. Yes, the functionality is correct, but this seems a bit too magical, to do it "behind your back" in `MarshalJSON`. I agree it should be moved to struct initialization.

kacf (commented):
@oleorhagen: Some more comments. We could also do some face to face time, talking about this, if you want. This is difficult stuff, and I don't claim to fully understand everything either.

app/updatemanager.go
```diff
@@ -387,22 +387,41 @@ func (c *ControlMapPool) Get(ID string) (active []*UpdateControlMap, expired []*
...
+			filteredPool = append(filteredPool, u)
+		},
+		// Predicates
+		func(u *UpdateControlMap) bool {
+			return u.ID != ID
+		},
+	)
+	c.Pool = filteredPool
+	c.saveToStore()
+}
```
kacf, line 407: FYI, I "imported" this function to #783 with an extra call to `announceUpdate`, which was missed here.

app/mender_test.go
```diff
@@ -179,7 +180,8 @@ func Test_CheckUpdateSimple(t *testing.T) {
 	mender = newTestMender(nil,
 		conf.MenderConfig{
 			MenderConfigFromFile: conf.MenderConfigFromFile{
-				Servers: []client.MenderServer{{ServerURL: srv.URL}},
+				Servers:                               []client.MenderServer{{ServerURL: srv.URL}},
```
oleorhagen, line 183: Odd formatting :thinking:

oleorhagen:
@kacf splitting the `UpdateControlMap` out into a separate package did the trick it seems

client/client_update_test.go
```diff
@@ -171,6 +171,101 @@ func TestParseUpdateResponse(t *testing.T) {
...
+	"id": "malformed-update-control-map",
+	"artifact": {
+		"source": {
+			"uri": "https://menderupdate.com",
+			"expire": "2016-03-11T13:03:17.063+0000"
+		},
+		"device_types_compatible": ["BBB"],
+		"artifact_name": "myapp-release-z-build-123"
+	},
+        "update_control_map": {
```
kacf, line 244: Suggestion: Also test one which is wrong type, like `"update_control_map": 3`.

kacf (approved):
Nice work! Moving out `UpdateControlMap` made everything much easier!

oleorhagen:
> Nice work! Moving out `UpdateControlMap` made everything much easier!

Not just my idea! ;)

And great, added the unit test. Running the pipeline!

oleorhagen:
Ignoring the known failures in `auditlogs` enterprise, as they're unrelated. Merging

PR #780, opened 2021-05-25
Fix Update Control Maps having no effect.

kacf:
The heart of the issue is that the control map pool was being created
twice, and then the state logic and the update logic were working on
different pools. To remedy this, move the initialization of pool into
NewDaemon, which is really where it belongs, since it cannot be used
in non-daemon mode [1]. This allows it to be created once and used
everywhere.

One of the unit tests which tests a nil argument is invalid after this
change, so just remove it.

[1] Actually the same thing can be said for the auth manager, but
    moving it has a lot more ripple effects which I don't want to deal
    with right now.

Changelog: None

### DIFF
--- a/app/daemon_test.go
+++ b/app/daemon_test.go
@@ -29,6 +29,7 @@ import (
 	"github.com/mendersoftware/mender/installer"
 	"github.com/mendersoftware/mender/store"
 	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
 )
 
 type FakeDevice struct {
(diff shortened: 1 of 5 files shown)


kacf:

app/daemon_test.go
```diff
@@ -168,26 +169,23 @@ func TestDaemon(t *testing.T) {
...
 	d.Cleanup()
 	mstore.AssertExpectations(t)
-
-	mstore = &store.MockStore{}
-	mstore.On("Close").Return(errors.New("foo"))
-	assert.NotPanics(t, func() {
-		d := NewDaemon(nil, mstore, nil, nil)
-		d.Cleanup()
-	})
-	mstore.AssertExpectations(t)
```
lluiscampos, line 190: I didn't get why this test is not relevant anymore (neither I completely see the value of the original test...). But can you elaborate what have we changed in this PR to invalidate the test?

lluiscampos (approved):
It makes sense. From a high level logical point of view, everything that has to do with DBus APIs should only be initialized on daemon mode :+1:

oleorhagen:
Another fix: https://github.com/mendersoftware/mender/pull/781

I decided not to push on top of here, since it seemed you were running tests. Feel free to cherry as you see fit though.

oleorhagen:
> Another fix: #781
> 
> I decided not to push on top of here, since it seemed you were running tests. Feel free to cherry as you see fit though.

On second thought, I'll just merge this fix independently

app/daemon_test.go
kacf: I understood the mechanics of the test, but not why it was there. It was testing that passing a nil argument would not panic, which is understandable for a publicly facing function. But this is an internal function, only called from a few tightly controlled places. Now that we need to pass in the Controller, the test just doesn't make sense anymore, it can't be called with a `nil` argument.

kacf: As for the exact reason, it is [this line](https://github.com/mendersoftware/mender/pull/780/files#diff-ef0bcd55c1d2292c666ecebf69ede458d884fc5bb514a4c9bf512f55cecd69f7R44). `mender` would be nil here if this test ran.

lluiscampos: Thanks

PR #781, opened 2021-05-26
Fix: Sanitized maps are not updated in the control struct

oleorhagen:
This is due to the copy semantics of the Golang for loop.

Fix is real easy, just re-assign the state after modification.


lluiscampos (approved):
Classic gotcha. Sorry

oleorhagen:

kacf (approved):

kacf:
Change seems spurious. I will merge this, since this is a feature branch and I have upcoming changes depending on this, and Ole has already gone home for the day.

PR #783, opened 2021-05-31
Fixes for Feature dbus update control

kacf:
Various small fixes for issues discovered while making acceptance tests.

### DIFF
--- a/app/mender.go
+++ b/app/mender.go
@@ -472,28 +472,33 @@ func (m *Mender) TransitionState(to State, ctx *StateContext) (State, bool) {
 func spinEventLoop(c *ControlMapPool, to State, ctx *StateContext, controller Controller) State {
 	for {
 		log.Debugf("Spinning the event loop for:  %s", to.Transition())
-		switch to.Transition() {
-		case ToArtifactReboot_Enter, ToArtifactCommit_Enter, ToArtifactInstall:
-			action := c.QueryAndUpdate(to.Transition().String())
-			log.Debugf("Event loop Action: %s", action)
-			switch action {
-			case "continue":
-				return to
-			case "pause":
-				// wait until further notice
-				log.Debug("Pausing the event loop")
-				<-c.Updates
-			case "fail":
-				log.Debug("Failing due to Update Control Map")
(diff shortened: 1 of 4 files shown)


app/updatemanager.go
```diff
@@ -209,7 +210,9 @@ func UpdateControlMapStateOnMapExpireDefault(action string) string {
 	return action
 }
 
-var UpdateControlMapStateOnActionExecutedDefault = UpdateControlMapStateActionDefault
+func UpdateControlMapStateOnActionExecutedDefault(action string) string {
+	return action
+}
```
lluiscampos, line 215: Ahá! You were right on Friday, I misunderstood this (and your comment, too, as I thought you were talking about `on_map_expire`).

lluiscampos (approved):

app/mender.go
```diff
@@ -472,28 +472,33 @@ func (m *Mender) TransitionState(to State, ctx *StateContext) (State, bool) {
...
-			case "fail":
-				log.Debug("Failing due to Update Control Map")
-				next, _ := to.HandleError(ctx, controller,
-					NewTransientError(errors.New("Forced a failed update")))
-				return next
-			default:
-				log.Debugf("Unknown Action: %s", action)
-			}
+		var mapState string
+		switch t := to.Transition(); t {
```
oleorhagen, line 476: Although the fix bugs me a little, this seems the way to go.

How come the transition naming is so incoherent?

oleorhagen (approved):
All look good to me.

Good job!

app/mender.go
kacf: I was only involved in some of the development of the transitions, but I believe the reason is that some states require special treatment of the Enter and Leave states, and have both, and some others don't need it, and therefore have only one state with the default behavior for all three sub states (Enter, actual state, and Leave). Reboot requires it because it needs to recover from a program restart during reboot, and Commit needs it because Leave scripts are supposed to be re-executed after a spontaneous reboot, unlike scripts in other states.

IIRC, each state here needs to reflected by a state struct in `state.go`, so it is unfortunately quite a lot of work to "fill out" this list with Enter/Leave states.

PR #784, opened 2021-06-01
Fix channels not being assigned correctly when loading from storage.

kacf:
Changelog: None


oleorhagen (approved):

kacf:
Merging this. I have verified that this is not the cause of the failures.

PR #785, opened 2021-06-01
Update and fix feature-update_control_maps branch.

kacf:
Merge `feature-dbus_update_control` into `feature-update_control_maps` and then fix a few problems arising from that. Only the commits after the last merge commit need to be reviewed.


oleorhagen (approved):
Nice!

This is gonna be a fun rebase tomorrow!

PR #788, opened 2021-06-03
MEN-4107: trigger rebuild of the docs site on changes

tranchitella:
Changelog: none

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -120,6 +120,25 @@ trigger:mender-dist-packages:
     - tags
     - master
 
+trigger:mender-docs-site:
+  image: alpine
+  stage: trigger
+  before_script:
+    - apk add --no-cache curl
+  script:
+    - curl -v -f -X POST
+      -F token=$MENDER_DOCS_SITE_TRIGGER_TOKEN
+      -F ref=master
+      https://gitlab.com/api/v4/projects/16842368/trigger/pipeline
+    - curl -v -f -X POST
+      -F token=$MENDER_DOCS_SITE_TRIGGER_TOKEN
+      -F ref=production
+      https://gitlab.com/api/v4/projects/16842368/trigger/pipeline
+  rules:
+    - if: '$CI_COMMIT_BRANCH =~ /^(master|[0-9]+\.[0-9]+\.x)$/'
+      changes:
+      - Documentation/*
+
 trigger:integration:
   image: alpine
   stage: trigger


.gitlab-ci.yml
```diff
@@ -120,6 +120,26 @@ trigger:mender-dist-packages:
...
+      https://gitlab.com/api/v4/projects/16842368/trigger/pipeline
+    - curl -v -f -X POST
+      -F token=$MENDER_DOCS_SITE_TRIGGER_TOKEN
+      -F ref=production
+      https://gitlab.com/api/v4/projects/16842368/trigger/pipeline
+  rules:
+    - if: '$CI_COMMIT_REF_NAME !~ /^(master|[0-9]+\.[0-9]+\.x)$/'
+      when: never
+    - changes:
+      - Documentation/*
```
lluiscampos, line 141: It looks good. But I think it is completely equivalent to:
```suggestion
    - if: '$CI_COMMIT_REF_NAME =~ /^(master|[0-9]+\.[0-9]+\.x)$/'
      changes:
      - Documentation/*
```

which to me reads better. Not a strong opinion, though.

On top of that, as we are not catching (nor should we catch) tags, I think is better if we use `CI_COMMIT_BRANCH` instead. For consistency with practices on the template (the `mender` pipeline is old, so we can't really compare here). For this I have an stronger opinion :)

lluiscampos (approved):
One suggestion and one weak suggestion. But looks good as-is too!

tranchitella:
@lluiscampos I won't run the full integration pipeline for this PR

.gitlab-ci.yml
```diff
@@ -120,6 +120,25 @@ trigger:mender-dist-packages:
...
+      -F ref=master
+      https://gitlab.com/api/v4/projects/16842368/trigger/pipeline
+    - curl -v -f -X POST
+      -F token=$MENDER_DOCS_SITE_TRIGGER_TOKEN
+      -F ref=production
+      https://gitlab.com/api/v4/projects/16842368/trigger/pipeline
+  rules:
+    - if: '$CI_COMMIT_BRANCH =~ /^(master|[0-9]+\.[0-9]+\.x)$/'
+      changes:
+      - Documentation/*
```
mzedel, line 140: since triggers are now available in the free tier, we might also be able to do: 
```suggestion
trigger:mender-docs-site:master:
  except:
  - pipelines
  rules:
    - if: '$CI_COMMIT_BRANCH !~ /^(master|[0-9]+\.[0-9]+\.x)$/'
      when: never
    - changes:
      - Documentation/*
  stage: trigger
  trigger:
    branch: master
    project: Northern.tech/Mender/mender-docs-site
trigger:mender-docs-site:production:
  except:
  - pipelines
  rules:
    - if: '$CI_COMMIT_BRANCH !~ /^(master|[0-9]+\.[0-9]+\.x)$/'
      when: never
    - changes:
      - Documentation/*
  stage: trigger
  trigger:
    branch: production
    project: Northern.tech/Mender/mender-docs-site
```
and get the easier navigation to the pipeline + don't need to take care of a token

lluiscampos: Thanks @mzedel for jumping in.

But I think we should wait until we conclude on the task you are working on (PoC to my eyes as per today) to learn how reliable and useful the feature is. Whenever we are confident with the `trigger` functionality, we should then migrate a full pipeline at once (for example this repo has 3 or 4 triggers, it won't be cool to have one with `trigger` and the rest with `curl` method...)

lluiscampos (approved):

.gitlab-ci.yml
mzedel: based on what I've seen in use so far, I doubt we will be able to get rid of all the curl triggers - case in point: [meta-mender](https://github.com/mendersoftware/meta-mender/blob/master/.gitlab-ci.yml#L18-L30), that needs additional logic to get the trigger working... there is a way to work around that as I'm doing [in mender-connect](https://github.com/mendersoftware/mender-connect/pull/32/files), but it's not always really appropriate... but in this case it's that much clearer, the recommended, supported way to do, does not rely on any variables + the navigation to the pipeline is even in development a huge benefit... applying the boy scout rule, it is very appropriate to use it wherever possible.

PR #789, opened 2021-06-03
MEN-4704: Conditionally update the control maps from the server during an update

oleorhagen:
Now aligned with the ideas in: 

https://github.com/mendersoftware/mender/pull/789#discussion_r646396680


app/state.go
```diff
@@ -657,11 +667,50 @@ func (u *updateCheckState) Handle(ctx *StateContext, c Controller) (State, bool)
...
+			maps, _ := pool.Get(ID)
+			for _, m := range maps {
+				ht := m.GetHalfTime()
+				if ht < nextCheck {
+					nextCheck = ht
+				}
+			}
+		case <-updateCtx.Done():
+			log.Info("Update is done, the maps are no longer updated from the server")
+		}
```
kacf, line 710: Did you mean to `break` here?

kacf (changes requested):

app/updatemanager.go
```diff
@@ -396,6 +397,33 @@ func (c *ControlMapPool) QueryAndUpdate(state string) (action string) {
 	return "continue"
 }
 
+// TODO - rename to NextUpdateControlMapTime / or similar
+func (c *ControlMapPool) NextControlMapHalfTime(ID string) time.Time {
```
kacf, line 401: Will you rename? I think the currently used name is better, actually...

app/updatemanager.go
```diff
@@ -396,6 +397,33 @@ func (c *ControlMapPool) QueryAndUpdate(state string) (action string) {
...
+			m = append(m, u)
+		},
+		// Predicates
+		func(u *updatecontrolmap.UpdateControlMap) bool {
+			return u.ID == ID
+		},
+	)
+	// Return the time of the map with the closest time to halfway expired
+	if len(m) == 0 {
+		return time.Now() // TODO - fhwat to return (?)
```
kacf, line 415: I would just return `UpdateControlMapExpirationTimeSeconds / 2`. This is guaranteed to be as long as, or shorter than, any map which is inserted later on.

kacf (changes requested):

app/mender.go
```diff
@@ -555,13 +571,53 @@ func spinEventLoop(c *ControlMapPool, to State, ctx *StateContext, controller Co
...
+		}
+		time.Sleep(wait)
+		_, err = controller.CheckUpdate()
+		attempts += 1
+		if err == nil {
+			return
+		}
+		log.Errorf("Attempt: %d Failed to update the control map from the server: %s", attempts, err.Error())
+	}
+	log.Errorf("Failed to refetch the update control map in 5 attempts, giving up")
```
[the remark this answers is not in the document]
oleorhagen: Hmm, good catch. How is this not part of our static checks? :thinking: 
I thought it was...

app/updatemanager.go
oleorhagen: Oh, really? haha, OK, fixed!

app/mender.go
```diff
@@ -555,13 +571,53 @@ func spinEventLoop(c *ControlMapPool, to State, ctx *StateContext, controller Co
...
+		if err == client.MaxRetriesExceededError {
+			log.Errorf("Failed to refetch the update control map in %d attempts, giving up", attempts)
+			return
+		} else if err != nil {
+			log.Errorf("Unexpected error: %s", err.Error())
+			return
+		}
+		time.Sleep(wait)
+		_, err = controller.CheckUpdate()
+		attempts += 1
```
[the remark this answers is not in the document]
oleorhagen: Simply started the count at 1 instead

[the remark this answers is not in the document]
kacf: I was about to write that then the log statement below would be wrong then, before realizing that I'm actually wrong! The original is correct. In my (very contrived) example above, the `MaxRetriesExceededError` will only happen on the second iteration, if there is one retry available, in which case it will write `1 attempts`, IOW correct. Sorry for the confusion.

app/mender_test.go
```diff
@@ -1398,3 +1399,117 @@ func TestUploadPauseStatus(t *testing.T) {
...
+	}
+
+	for name, test := range tests {
+		t.Run(name, func(t *testing.T) {
+			ms := store.NewMemStore()
+			pool := NewControlMap(
+				ms,
+				conf.DefaultUpdateControlMapBootExpirationTimeSeconds)
+			pool.Insert(&updatecontrolmap.UpdateControlMap{
+				ID:       "foo",
```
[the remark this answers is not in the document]
oleorhagen: I will hold off on rebasing until we got this one where we want it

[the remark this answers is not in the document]
oleorhagen: Rebased and fixed

oleorhagen:
Passing integration test: https://gitlab.com/Northern.tech/Mender/mender-qa/-/pipelines/323481353

app/state.go
```diff
@@ -656,7 +657,7 @@ func (u *updateCheckState) Handle(ctx *StateContext, c Controller) (State, bool)
 			// Just report successful update and return to normal operations.
 			return NewUpdateStatusReportState(update, client.StatusAlreadyInstalled), false
 		}
-		if err.Cause() == client.ErrNoDeploymentAvailable {
+		if errors.Cause(err) == client.ErrNoDeploymentAvailable {
```
kacf, line 660: I've read that the best way to test whether an error is of a certain kind is to use [`errors.Is()`](https://golang.org/pkg/errors/#Is), since it will catch any error in the chain, not just top one (`err`) or bottom one (`errors.Cause(err)`).

client/client_update.go
```diff
@@ -21,7 +21,6 @@ import (
 	"io/ioutil"
 	"net/http"
 	"net/url"
-	"strconv"
```
kacf: Squash this with the previous commit?

app/mender.go
```diff
@@ -525,6 +525,7 @@ func spinEventLoop(c *ControlMapPool, to State, ctx *StateContext, controller Co
 		case "continue":
 			return to
 		case "pause":
+			log.Infof("Update Control: Pausing the update in %s state", mapState)
```
kacf, line 528: The original message said "Update Control: Pausing before entering %s state", which is technically more correct, it pauses between the states, before state scripts and such, not in the state.

app/mender.go
```diff
@@ -554,13 +548,107 @@ func spinEventLoop(c *ControlMapPool, to State, ctx *StateContext, controller Co
...
 
+	// Check if we need to update the control map from the server
+	updateState, ok := to.(Updater)
+	if ok {
+		ctx.NextUpdateMapTime = c.GetControlMapPool().
+			NextControlMapHalfTime(updateState.Update().ID)
+		if time.Now().Sub(ctx.NextUpdateMapTime) < 0 {
+			// Maps are magically updated in CheckUpdate
+			_, err := c.CheckUpdate()
+			if err != nil {
```
kacf, line 647: Should you check for `ErrNoDeploymentAvailable` here as well?

app/mender_test.go
```diff
@@ -1356,31 +1359,37 @@ func TestUploadPauseStatus(t *testing.T) {
...
+				conf.DefaultUpdateControlMapBootExpirationTimeSeconds,
+			)
 			pool.Insert(&updatecontrolmap.UpdateControlMap{
 				ID:       "foo",
 				Priority: 1,
 				States: map[string]updatecontrolmap.UpdateControlMapState{
 					test.state: updatecontrolmap.UpdateControlMapState{
 						Action:           "pause",
-						OnActionExecuted: "pause",
+						OnActionExecuted: "continue",
```
kacf, line 1375: Do these actually work well together? One weakness I discovered (at least in earlier revisions) is if `Action: pause` and `OnActionExecuted: continue`, then it might somewhat randomly decide to continue simply because the loop spins and the next action gets set. I concluded that it's not very important to fix because the combination of those two is quite weird and not too well defined.

But it raises the question: What is the reason for switching it here, and does it really do what you think it does?

app/state.go
```diff
@@ -651,6 +656,9 @@ func (u *updateCheckState) Handle(ctx *StateContext, c Controller) (State, bool)
 			// Just report successful update and return to normal operations.
 			return NewUpdateStatusReportState(update, client.StatusAlreadyInstalled), false
 		}
+		if err.Cause() == client.ErrNoDeploymentAvailable {
```
kacf: Same comment about `errors.Is()`.

app/updatemanager.go
```diff
@@ -401,6 +404,32 @@ func (c *ControlMapPool) QueryAndUpdate(state string) (action string) {
...
+			m = append(m, u)
+		},
+		// Predicates
+		func(u *updatecontrolmap.UpdateControlMap) bool {
+			return u.ID == ID
+		},
+	)
+	// Return the time of the map with the closest time to halfway expired
+	if len(m) == 0 {
+		return time.Now().Add(time.Duration(c.updateControlMapExpirationTimeSeconds) / 2)
```
kacf, line 421: Why do you need a time at all in this case? Can't you just return some kind of "empty" error instead? If I understand it correctly, this could save several things:

* You don't need to introduce the `updateControlMapExpirationTimeSeconds` variable in the `ControlMap` struct.
* If you check for the error in the caller, you can avoid a call to the backend if we don't have any update maps.

app/updatemanager_test.go
```diff
@@ -761,3 +829,44 @@ func TestMapUpdates(t *testing.T) {
 			100*time.Millisecond)
 	})
 }
+
+func TestUpdateControlMapRequired(t *testing.T) {
```
kacf, line 833: Doesn't this test the half-time..? A bit confused about the name.

client/client_update.go
```diff
@@ -245,7 +246,7 @@ func processUpdateResponse(response *http.Response) (interface{}, error) {
 
 	case http.StatusNoContent:
 		log.Debug("No update available")
-		return nil, nil
+		return nil, ErrNoDeploymentAvailable
```
kacf, line 250: Without having investigated anything at all, I have a hunch that if you're getting test failures, this could be a quite likely source of changes in behavior.

kacf (commented):
Nice, a few nitpicks and a few questions.

app/state.go
oleorhagen: Nice! TIL

app/mender.go
oleorhagen: I agree, fixing

app/mender.go
oleorhagen: What do you suggest we do in the case of an aborted deployment here?

app/updatemanager.go
oleorhagen: Good idea!

app/mender_test.go
oleorhagen: I decided to simply remove it completely.
The reason it was changed, was to reign in the goroutines resulting from the `go spinEventLoop`. I don't see how changing this should have any effect on the test itself.

app/mender.go
kacf: I guess the same as for other types of failures, `return from.HandleError(...)`, with some log message printed to say why the failure happened.

kacf: Oh, and should check the return code from `refetch` as well.

client/client.go
```diff
@@ -124,6 +124,10 @@ func (a *APIError) Cause() error {
 	return a.error
 }
 
+func (a *APIError) Unwrap() error {
+	return a.error
+}
```
kacf, line 129: Lol, you too!

app/state.go
```diff
@@ -1843,3 +1843,230 @@ func setBrokenArtifactFlag(store store.Store, artName string) {
...
+type Wrap struct {
+	state UpdateState
+}
+
+type controlMapState struct {
+	wrappedState UpdateState
+	baseState
+}
+
+func NewControlMapState(wrapsState Wrap) State {
```
kacf, line 1860: Isn't `NewControlMapState(wrapsState State)` essentially the same thing?

kacf (commented):
Looking very good at this point, I don't really have anything to add!

app/state.go
oleorhagen: Yes, it's exactly the same, I was just experimenting with the interface a little

oleorhagen: Changing it back to the default (sane option).
Just wanted to see how it read :)

oleorhagen:

oleorhagen:
Link: https://gitlab.com/Northern.tech/Mender/mender-qa/-/pipelines/326868986

app/state.go
```diff
@@ -1835,3 +1847,207 @@ func setBrokenArtifactFlag(store store.Store, artName string) {
...
+	}
+
+	f.retries++
+
+	log.Infof("Wait %v before next update control map fetch/update attempt", intvl)
+	return f.Wait(
+		NewFetchControlMapState(f.wrappedState),
+		f,
+		intvl,
+		ctx.WakeupChan,
```
oleorhagen, line 1992: One option here, is to wake on updates to the control maps, and return to the `QueryAndUpdate` loop

app/state.go
```diff
@@ -1835,3 +1847,207 @@ func setBrokenArtifactFlag(store store.Store, artName string) {
...
+
+	if errors.Is(err, NoUpdateMapsErr) {
+		log.Error("No control maps no longer present, continuing")
+		nextMapRefresh = time.Now().Add(30 * time.Second)
+	}
+
+	updateMapFromServerIn := nextMapRefresh.Sub(time.Now())
+	log.Infof("Next update refresh from the server in: %s", updateMapFromServerIn)
+
+	if updateMapFromServerIn < 0 {
```
oleorhagen, line 2035: This happens occasionally when the maps are close to expiring, I feel this is a safe fallback, but a discussion here be nice.

app/state.go
```diff
@@ -448,7 +450,7 @@ type updateCommitState struct {
 	reportTries int
 }
 
-func NewUpdateCommitState(update *datastore.UpdateInfo) State {
+func NewUpdateCommitState(update *datastore.UpdateInfo) UpdateState {
```
oleorhagen, line 453: Return `UpdateState` as then the `controlMap` wrapper states, don't need to store the update themselves.
This should be strictly transparent, as `UpdateState` is a super state of `State`, and hence, can be passed in as `State` anywhere it is required, so this change will not change any existing logic.

app/state.go
```diff
@@ -1835,3 +1843,207 @@ func setBrokenArtifactFlag(store store.Store, artName string) {
...
+}
+
+func (c *controlMapState) mapStateToName(t Transition) string {
+	switch t {
+	case ToArtifactReboot_Enter, ToArtifactCommit_Enter:
+		return c.wrappedState.Transition().String()
+	case ToArtifactInstall:
+		return "ArtifactInstall_Enter"
+	}
+	panic(fmt.Sprintf("No state name mapping for: %s. This is a logic error in the state machine code", t))
```
oleorhagen, line 1881: I think this is fair, as it should catch logic errors during testing, but feel free to object.

app/state.go
```diff
@@ -1835,3 +1843,207 @@ func setBrokenArtifactFlag(store store.Store, artName string) {
 		// already in an error path.
 	}
 }
+
+type controlMapState struct {
+	wrappedState UpdateState
+	baseState
+}
+
+func NewControlMapState(wrapsState UpdateState) State {
```
oleorhagen, line 1852: From looking at your new state machine failsafe PR, these should now implement `UpdateState` right?

PR #791, opened 2021-06-04
Fix several race conditions in Dbus Go routines.

kacf:


kacf:
@merlin-northern: Do you have a chance to try this out and see if solves the problem you had with mender-connect and token signals when restarting the Mender client?

merlin-northern:
> @merlin-northern: Do you have a chance to try this out and see if solves the problem you had with mender-connect and token signals when restarting the Mender client?

thanks for the ping. I will try and report here. (although later in the afternoon)

lluiscampos (approved):
Good finds

tranchitella (approved):

kacf:
Failures are due to an unrelated test problem. This passed on the platforms where it was able to run. Merging, since I have a followup PR which will re-check on all platforms anyway.

PR #792, opened 2021-06-07
Fix: Correctly log the error response message from server errors

oleorhagen:
In some cases the server does not return JSON data, as is advertised in the API
upon errors. Instead it simply returns an error string.

Therefore, in the cases in which the error cannot be unmarshalled to JSON, fall
back to returning the response body as a string.

An example of the error can be seen here:

```
level=error msg="Error receiving scheduled update data: (request_id: ): Invalid response received from server server error message: failed to parse server response: json: cannot unmarshal object into Go struct field .error of type string"
```

The error stems from the migration to Traefic, which does some magic rewriting
of our error responses.

Changelog: Title

### DIFF
--- a/client/client.go
+++ b/client/client.go
@@ -621,8 +621,12 @@ func unmarshalErrorMessage(r io.Reader) string {
 	e := new(struct {
 		Error string `json:"error"`
 	})
-	if err := json.NewDecoder(r).Decode(e); err != nil {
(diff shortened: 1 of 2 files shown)


oleorhagen:

kacf (approved):

oleorhagen:
Cherrie(s):
* [2.5.x]: #816
* [2.6.x]: #815

PR #794, opened 2021-06-07
Fix use-after-free issue when receiving signal.

kacf:
When runDaemon() returns, the first thing that happens is that it
cleans up the resources used by the MenderDaemon structure. But if we
got a signal, we escaped runDaemon without finishing Run(), which
means this memory is still in use. This is worse than simply exiting,
because operations can be corrupted, as opposed to simply
truncated. There is no way to make this safe with the current design.

Therefore split the signal handling, so that SIGTERM is handled in
doMain, directly where we want to exit from. This means that runDaemon
can keep running without its data being destroyed.

An alternative would have been to simply call os.Exit() from the old
location, but it seemed better to keep the exit codes in one place.

Changelog: Fix occasional crash when exiting using SIGTERM.

### DIFF
--- a/main.go
+++ b/main.go
@@ -25,29 +25,43 @@ import (
 	log "github.com/sirupsen/logrus"
 )
 
+var termSignalChan = make(chan os.Signal, 1)
+
 func init() {
 	// SIGUSR1 forces an update check.
 	// SIGUSR2 forces an inventory update.
 	// SIGTERM marks the exit.
-	signal.Notify(cli.SignalHandlerChan, syscall.SIGUSR1, syscall.SIGUSR2, syscall.SIGTERM)
+	signal.Notify(cli.SignalHandlerChan, syscall.SIGUSR1, syscall.SIGUSR2)
+	signal.Notify(termSignalChan, syscall.SIGTERM)
 }
 
 func doMain() int {
-	if err := cli.SetupCLI(os.Args); err != nil {
-		switch err {
-		case cli.ErrSIGTERM:
-			log.Infoln(err.Error())
-			return 0
-		case app.ErrorManualRebootRequired:
-			return 4
-		case installer.ErrorNothingToCommit:
-			log.Warnln(err.Error())
-			return 2
-		default:
-			log.Errorln(err.Error())
-			return 1
+	cliResultChan := make(chan error, 1)
+	go func() {
+		cliResultChan <- cli.SetupCLI(os.Args)
+	}()
+
+	// Wait for result from either the program, or a signal.
+	select {
+	case err := <-cliResultChan:
+		if err != nil {
+			switch err {
+			case app.ErrorManualRebootRequired:
+				return 4
+			case installer.ErrorNothingToCommit:
+				log.Warnln(err.Error())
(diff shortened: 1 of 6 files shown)


oleorhagen (approved):
Makes sense

main.go
```diff
@@ -25,29 +25,43 @@ import (
 	log "github.com/sirupsen/logrus"
 )
 
+var termSignalChan = make(chan os.Signal, 1)
+
 func init() {
 	// SIGUSR1 forces an update check.
 	// SIGUSR2 forces an inventory update.
 	// SIGTERM marks the exit.
 	signal.Notify(cli.SignalHandlerChan, syscall.SIGUSR1, syscall.SIGUSR2, syscall.SIGTERM)
```
lluiscampos, line 34: We should remove `syscall.SIGTERM` from here. I guess it is harmless (not covered in if/else if block) but clearer if removed.

lluiscampos: (if we do this, the patch below would require adjustment)

lluiscampos (changes requested):
See my comment below. Am I wrong?

main.go
kacf: Well spotted, you're right, I'll fix it!

kacf:
@lluiscampos: Updated!

patches/0001-Instrument-Mender-client-for-coverage-analysis.patch
```diff
@@ -23,8 +23,8 @@ index 98e138a..ecb5730 100644
  
  	"github.com/mendersoftware/mender/app"
  	"github.com/mendersoftware/mender/cli"
-@@ -32,6 +34,21 @@ func init() {
- 	signal.Notify(cli.SignalHandlerChan, syscall.SIGUSR1, syscall.SIGUSR2, syscall.SIGTERM)
+@@ -35,6 +37,21 @@ func init() {
+ 	signal.Notify(termSignalChan, syscall.SIGTERM)
```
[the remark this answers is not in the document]
kacf: I think you are reading it wrong. The new patch contains only the `signal.Notify(termSignalChan, syscall.SIGTERM)` line, which is still there. The long line doesn't exist in the patch anymore because it is outside the diff distance.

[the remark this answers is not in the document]
lluiscampos: Oh, you are right :see_no_evil:

kacf:
* @TheMeaningfulEngineer: FYI, https://github.com/mendersoftware/mender/pull/794/commits/560ba66ca40f251af12b50930597fc65b346db88
* @lluiscampos, @tranchitella: FYI, https://github.com/mendersoftware/mender/pull/794/commits/cabd404a43b0616a8674fce6b8c9cf5b4effb5ea

The [last commit](https://github.com/mendersoftware/mender/pull/794/commits/1b7f56559225484fbe2ccdd36583f908446a0b23) comes from #795, and I probably will not submit it as part of this PR, I'm just using it for debugging.

TheMeaningfulEngineer (approved):
Thanks

PR #795, opened 2021-06-09
Do not close stdout/stderr on child process execution by default.

kacf:
Golang has a somewhat unusual default in that it closes all output
file descriptors for child processes unless you keep them open
actively. Make this the default across the Mender client, since output
from various processes we call may be helpful when reading user logs.

We have to create our own type for this unfortunately, because the
existing exec.Cmd type does not tolerate Stdout and Stderr being
non-nil when calling functions that change them (like StdoutPipe()).

Changelog: Title

### DIFF
--- a/cli/cli.go
+++ b/cli/cli.go
@@ -19,7 +19,6 @@ import (
 	"io"
 	"io/ioutil"
 	"os"
-	"os/exec"
 	"path"
 	"runtime"
 	"strings"
@@ -144,8 +143,8 @@ func SetupCLI(args []string) error {
 			Usage: "Force update check.",
 			Action: func(_ *cli.Context) error {
 				return sendSignalToProcess(
-					exec.Command("kill", "-USR1"),
-					exec.Command("systemctl",
+					system.Command("kill", "-USR1"),
+					system.Command("systemctl",
 						"show", "-p",
 						"MainPID", "mender-client"))
 			},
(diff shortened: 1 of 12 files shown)


kacf:
@lluiscampos, @oleorhagen: What do you think about this? This is inspired by threads like [this one](https://hub.mender.io/t/getting-error-fw-setenv-returned-failure-exit-status-2-after-committing-update-ota-update/3669/30), where we simply cannot see the output from external tools, and therefore have a hard time debugging what goes wrong.

kacf:

kacf:
@oleorhagen: To answer the question you asked in the standup: We could call something like `CombinedOutput()` and simply print it afterwards, but it's easy to forget to do this, and it requires extra code in all places where we call commands. The idea was to make it a bit more transparent and standardized within the binary. But as I mentioned, it's also more intrusive, and I'm not married to this method, so opinions are welcome.

kacf:

oleorhagen (approved):

oleorhagen:
I think I would make a slightly less technical `Changelog` entry though, but up to you

kacf:
@oleorhagen How about: "Make sure output from child processes is visible"?

oleorhagen:
> @oleorhagen How about: "Make sure output from child processes is visible"?

I like it. But maybe even more detail?

Maybe something like:

```
Previously the Mender client would hide output from the child processes it ran, leading to errors from tools such as `fw_printenv`, not showing up in the logs.

Now default to showing all output from the child processes called in the client logs.
```

kacf:
Thanks, used that and merged manually (it already passed).

PR #796, opened 2021-06-10
Prevent conflicting mender services from running simultaneously.

kacf:
Changelog: Fix: `mender-client.service` and the old `mender.service`
services can no longer run at the same time. If anyone has both, then
`mender.service` should be removed from the system.

### DIFF
--- a/support/mender-client.service
+++ b/support/mender-client.service
@@ -1,7 +1,8 @@
 [Unit]
 Description=Mender OTA update service
 Wants=network-online.target
-After=systemd-resolved.service network-online.target
+After=systemd-resolved.service network-online.target mender.service
+Conflicts=mender.service
 
 [Service]
 Type=idle


kacf:
@lluiscampos: This was inspired by [this post](https://hub.mender.io/t/high-rate-of-debian-failures/3687/10). However, it may not be quite that simple, because older client releases are still using the old name, and may need to patch this out of the file. However, this caused quite severe breakage, so I reckon it's still worth it. What do you think?

Let's never rename the client again...

kacf:

lluiscampos:
> @lluiscampos: This was inspired by [this post](https://hub.mender.io/t/high-rate-of-debian-failures/3687/10). However, it may not be quite that simple, because older client releases are still using the old name, and may need to patch this out of the file. However, this caused quite severe breakage, so I reckon it's still worth it. What do you think?

While the change you submit here is good, I don't really understand how this prevents another user to end up in the same situation. All that it will do is to block `mender-client` service, but the `mender` one will still be running right? So to my eyes "hides" the problem that the service file still will be wrong.

> Let's never rename the client again...

Roger that!

lluiscampos (approved):

kacf:
> While the change you submit here is good, I don't really understand how this prevents another user to end up in the same situation. All that it will do is to block `mender-client` service, but the `mender` one will still be running right? So to my eyes "hides" the problem that the service file still will be wrong.

What it does it prevent them from running at the same time. Either one can run in isolation. However, if both are enabled at boot, then both will fail to start, and it counts as an error from systemd's side. So this is the only thing that this patch changes, you cannot run them at the same time anymore. It does not fix the fact that there may be an old service file in the system, but that was the case before this patch as well.

I can see what you mean with "hiding" the issue, but the original issue was quite hidden as well, and the effect was much more severe than it will be after this patch. So we should still be on the lookout for the problem, but at least we have some protection from the worst effects.

lluiscampos:
Okey, let's merge it then. Thanks for the discussion.

kacf:
Thanks!

PR #799, opened 2021-06-10
Fix: use ampersand characters in D-Bus documentation

tranchitella:
Changelog: none


tranchitella:
I believe we don't need the QA pipeline for this PR...

kacf (approved):
Neither do I!

PR #801, opened 2021-06-16
Make command output less ugly in the log.

kacf:
This is a followup to 26f2ccbd523932. After that commit, the output is
printed correctly, but completely without context, making it hard to
follow in logs. Since the primary purpose of having this output is
debugging, wrap it in a writer to prettify the output.

No changelog, since this is just augmenting the mentioned commit.

Changelog: None

### DIFF
--- a/system/system.go
+++ b/system/system.go
@@ -18,9 +18,11 @@ import (
 	"io"
 	"os"
 	"os/exec"
+	"strings"
 	"time"
 
 	"github.com/pkg/errors"
+	log "github.com/sirupsen/logrus"
 )
 
 type SystemRebootCmd struct {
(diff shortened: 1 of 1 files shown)


kacf:

system/system.go
```diff
@@ -80,11 +82,27 @@ func (c *Cmd) StdoutPipe() (io.ReadCloser, error) {
...
+type cmdLogger struct {
+	cmdName string
+	stream  string
+}
+
+func (c *cmdLogger) Write(buf []byte) (int, error) {
+	lines := strings.Split(string(buf), "\n")
+	for _, line := range lines {
+		if len(line) > 0 {
+			log.Infof("Output (%s) from command %q: %s", c.stream, c.cmdName, line)
```
oleorhagen, line 94: Actually, I was envisioning something more along the lines of:
```
Collected output from command...:
line
line
.
.
.
line
```
Like, this is really one log.

What do you think?

kacf: We should not collect all the output from the command, because this can grow without bound, and we lose both timestamps and ordering. We could perhaps do it as you say for each block that is written, but then it will be:

```
Collected output from command...:
line
line
.
.
.
line
Some other Mender log output...
Collected output from command...:
line
line
.
.
.
line
```

because `Write()` will be called multiple times. So maybe a bit inconsistent. Do you think it's better? Is so, I should at least check it does not cause any problems in the systemd or deployment log.

oleorhagen: Good point, I guess we should first decide

1. if you also think this is better (the collected output).

-> If so, then we can make the writer buffered without too much trouble I guess :) ?

kacf: >     1. if you also think this is better (the collected output).

Copy pasting is more annoying with the current method, but I still think I lean towards that one. Strictly speaking it is what I would expect from a logger. It's ugly, but it's consistent, and every log line will look the same! :-)

Oh, and I just discovered that the Update Module output is already printed this way.

> If so, then we can make the writer buffered without too much trouble I guess :) ?

It's already buffered before `Write()` is called. We definitely should not add additional buffering, because then we lose timestamps and ordering as I mentioned earlier, even if the buffer is small. For example it might print two things several seconds apart, and only on the second print is the buffer full and we print it. This will be misleading in the log.

oleorhagen: Makes sense. I yield :)

oleorhagen (approved):

oleorhagen:
Rebased here: https://github.com/mendersoftware/mender/pull/814

PR #803, opened 2021-06-17
Switch IRC link from Freenode to Libera.

kacf:
Changelog: None


kacf:
I will do the other repos too, but am starting with this one.

michaelatmender (changes requested):
I would suggest using [https://web.libera.chat/?#mender](https://web.libera.chat/?#mender) as the URL

kacf:
> I would suggest using https://web.libera.chat/?#mender as the URL

Good point! I tried to find that link syntax myself, but didn't even think to try the obvious URL. :-)

kacf:
@michaelatmender Updated!

michaelatmender (approved):

PR #807, opened 2021-06-18
MEN-4785: Add support for v2/deployments/next endpoint.

kacf:
Changelog: The client now supports the HTTP Device API v2.

### DIFF
--- a/client/client_update.go
+++ b/client/client_update.go
@@ -56,10 +56,9 @@ func NewUpdate() *UpdateClient {
 // CurrentUpdate describes currently installed update. Non empty fields will be
 // used when querying for the next update.
 type CurrentUpdate struct {
-	Artifact         string
-	DeviceType       string
-	Provides         map[string]string
-	UpdateControlMap bool `json:"update_control_map"`
+	Artifact   string
+	DeviceType string
+	Provides   map[string]string
 }
 
 func (u *CurrentUpdate) MarshalJSON() ([]byte, error) {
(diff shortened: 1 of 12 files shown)


kacf:

client/client_update.go
```diff
@@ -71,6 +70,15 @@ func (u *CurrentUpdate) MarshalJSON() ([]byte, error) {
 	return json.Marshal(u.Provides)
 }
 
+type updateV1Body struct {
```
oleorhagen, line 73: Suggestion:
I'm guessing you're really just looking for a type alias?

client/client_update.go
```diff
@@ -84,39 +92,54 @@ func (u *UpdateClient) GetScheduledUpdate(api ApiRequester, server string,
...
-
-		// Fall back to the GET (Open-Source) functionality on all error codes
-		if r.StatusCode >= 400 && r.StatusCode < 600 {
-
-			log.Debugf("device provides not accepted by the server. Response code: %d", r.StatusCode)
-
-			r, err = api.Do(getReq)
+	ok := false
+	var r *http.Response
+	for _, req := range reqs {
```
oleorhagen, line 102: Musing:
It's OK, I personally would have extended the tuple with a third request return parameter but.

It's also a little symptomatic that we start relying on go-to's :/

oleorhagen (approved):
Does the trick, but does not warm my heart

client/client_update.go
kacf: Essentially yes. Strictly speaking I don't even need this type at all, it's intended to make it clearer when reading, as a counterpart to `updateV2Body`. You're the one looking from the outside, so you can tell me which one you prefer. :-)

oleorhagen: I figured :)
And I like it, but you can just write it:
`type updateV1Body = currentUpdate` (or similar, don't remember exactly the syntax)
But this is all splitting hairs ofc

client/client_update.go
kacf: How about: I'll put it inside a separate function instead. Then it should be easier to read, and I can replace the gotos with `return` (which is to exit `switch` and `for` simultaneously).

kacf: Better now? I think I disagree with the tuples, since then you have to repeat the code three times, and can't do it list-style like I've done. But it's nice to get rid of the gotos, and it's also smaller and easier on the eyes now I think.

oleorhagen: much better :)

tranchitella (approved):

kacf:
I tested this locally, the failure happens with the master client as well, so it's not related to this PR. Merging.

PR #808, opened 2021-06-22
Several Update Control Map parser fixes + misc.

kacf:


kacf:
@oleorhagen: I took the liberty of also fixing bad signature handling while I was in this code. It's been bugging me for a while!

client/client_update.go
```diff
@@ -191,7 +191,35 @@ func (u *UpdateClient) FetchUpdate(api ApiRequester, url string, maxWait time.Du
...
+
+type strictUpdateControlMapParser struct {
+	*updatecontrolmap.UpdateControlMap
+
+	// Used to ignore the map after an error has been returned. We need to
+	// reparse without errors then, to get the UpdateInfo.
+	ignoreMap bool
+}
+
+func (s *strictUpdateControlMapParser) UnmarshalJSON(data []byte) error {
```
oleorhagen, line 209: How about we add the marshaller to the `UpdateInfo` type, and then the `disallowUnknownFields` to the udpate control map, and we can do this in one fell swoop?

client/client_update.go
```diff
@@ -235,10 +263,16 @@ func processUpdateResponse(response *http.Response) (interface{}, error) {
 		log.Debug("Have update available")
 		var ur UpdateResponse
 		if err = json.Unmarshal(respBody, &ur); err != nil {
-			return nil, errors.Wrap(err, "failed to parse the HTTP update response")
+			// In case the update control map was invalid, reparse
+			// without it to get the UpdateInfo. Doesn't matter if
+			// it fails, it will just be nil then, and we are
+			// already returning an error.
+			ur.UpdateControlMap.ignoreMap = true
+			_ = json.Unmarshal(respBody, &ur)
```
oleorhagen, line 271: Suggestion:
Simply pass in the `UpdateInfo` structure, and only that part will be unmarshalled

oleorhagen (commented):
Nice finds!

client/client_update.go
```diff
@@ -191,7 +191,27 @@ func (u *UpdateClient) FetchUpdate(api ApiRequester, url string, maxWait time.Du
...
+	// unknown fields. This is in contrast to the rest of the response,
+	// where we allow unknown fields.
+	UpdateControlMap strictUpdateControlMapParser `json:"update_control_map"`
+}
+
+type strictUpdateControlMapParser struct {
+	*updatecontrolmap.UpdateControlMap
+}
+
+func (s *strictUpdateControlMapParser) UnmarshalJSON(data []byte) error {
```
lluiscampos, line 209: I wonder if we could now reuse this `strictUpdateControlMapParser` struct in D-Bus Update Control Map parsing [here](https://github.com/mendersoftware/mender/blob/master/app/updatemanager.go#L114-L121).

I think that reusing the struct we have less chances for them to diverge in the future (as it happened now, right? Because D-Bus parsing was disallowing extra fields, but here we were not)

client/client_update.go
kacf: Good idea!

client/client_update.go
kacf: But what's the benefit? `UpdateInfo` only uses defaults, so it doesn't need an unmarshaller. Maybe I'm misunderstanding...

client/client_update.go
kacf: Good idea, I'll try to unify them.

client/client_update.go
oleorhagen: I'm sorry, I meant the `UpdateResponse` struct of course!

client/client_update.go
oleorhagen: Yes, totally agree! But should we then add it to the `UpdateControl` type?

client/client_update.go
kacf: This doesn't work, because `DisallowUnknownFields` is a property of the parser, so you can't use it on only one sub element. The sub element itself needs to have an unmarshaller which sets it. It would be nice if it was settable with a JSON tag like `omitempty`, but I didn't find anything like that.

kacf: Oh, and just to be clear on this point as well: We can't do two parse passes, one over `UpdateInfo` and one over `UpdateControlMap`, because the latter's key is not available before you have started parsing, it's just a blob of bytes at this point.

client/client_update.go
kacf: Yes! I will try this. Maybe we can even get rid of `Validate` as a separate step! :smiley:

client/client_update.go
oleorhagen: Ur, right. thx

client/client_update.go
kacf: This actually required quite a big change because the struct needs to be initialized differently, so I added it as a separate commit. I recommend viewing that one in particular with whitespace turned off.

oleorhagen (approved):

kacf:
I have verified that the issue is not due to this pull request, it just makes an already spurious failure more likely because the client is quicker. I will follow up independently on that.

PR #810, opened 2021-06-24
MEN-3360: Make state trans. counter ignore certain states.

kacf:
We divide the states into three groups: Update states which are not
allowed to loop, update states which are allowed to loop, and other
states (which are always allowed to loop). For update states
(`updateState`) we use the new `PermitLooping` interface to determine
whether it permits looping, for other states (`baseState`) it's enough
to know that they are not an `updateState`.

The motivation behind all this is to allow going back and forth
between states as long as at least one of them allows looping, which
will allow us to implement limitless state transitions during certain
operations in a deployment. This is important to support Update
Control, and may have other uses in the future as well (the ticket has
some useful links).

However, using the three groups of states, we *also* maintain the
ability to break out of loops between states that do *not* permit
looping, which typically is caused by update failures. Notice in
particular that reboots, due to calling `LoadStateData`, will always
increase the counter by at least one, even if the states involved do
permit looping.

This also reverts eb15ff94bf4ed57c, which increased the maximum number
of state transitions in order to deal with some states that can loop
for a limited time. In principle, the limit for these states could be
raised now, but it's not something I will handle in this commit.

Changelog: None

### DIFF
--- a/app/state.go
+++ b/app/state.go
@@ -130,6 +130,12 @@ type WaitState interface {
 type UpdateState interface {
 	State
 	Update() *datastore.UpdateInfo
+	// Signals whether this state is allowed to loop indefinitely. This is
+	// used to track state transitions (see StateDataStoreCount in
+	// datastore.go). States that can loop indefinitely are assumed to have
+	// other means to break out of loops (such as server control), and are
+	// not counted when updating the state count.
+	PermitLooping() bool
 }
 
 // baseState is a helper state with some convenience methods
@@ -234,6 +240,12 @@ func (us *updateState) HandleError(ctx *StateContext, c Controller, err menderEr
 		setBrokenArtifactFlag(ctx.Store, us.Update().ArtifactName())
 		return NewUpdateErrorState(err, us.Update()), false
 	}
+}
+
+func (us *updateState) PermitLooping() bool {
+	// By default, states should not permit looping so that the client can
+	// break out of Update Module or reboot loops.
+	return false
 }
 
 type idleState struct {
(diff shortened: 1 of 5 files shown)


kacf:

lluiscampos (approved):
I am sorry I can't provide much valuable feedback - it just looks good to me :heavy_check_mark:

app/state.go
```diff
@@ -1253,6 +1271,10 @@ func (ue *updateErrorState) Update() *datastore.UpdateInfo {
 	return &ue.update
 }
 
+func (ue *updateErrorState) PermitLooping() bool {
```
oleorhagen, line 1274: Q:
Why is this needed, if the default implementation is `false` ? :thinking:

datastore/datastore.go
```diff
@@ -26,12 +26,10 @@ import (
...
 
-	// This number 30 should be kept quite a lot higher than the number of
+	// This number should be kept quite a lot higher than the number of
 	// expected state storage operations, which is usually roughly
-	// equivalent to the number of state transitions. 40 is added as an
-	// extra buffer for StatusReportRetry states, which can run up to 10
-	// times each (10 * two states * enter and exit state = 10 * 2 * 2 = 40)
-	MaximumStateDataStoreCount = 30 + 40
+	// equivalent to the number of state transitions.
+	MaximumStateDataStoreCount = 30
```
oleorhagen, line 32: Much nicer :)

oleorhagen (approved):
Nice!

Elegant solution

app/state.go
kacf: Unlike most of the update states, `updateErrorState` doesn't have `updateState` as a base component. I looked into changing this, but this turned out to be quite tricky because it has `errorState` as a base component instead, so if you try to use both you get ambiguous member functions. Therefore we need to provide it manually here to fulfill the interface.

You can see that it already provides `Update()`, which is there for the same reason.

kacf:
Seems to be only spurious failures. Merging.

PR #811, opened 2021-06-25
MEN-4808: Deprecate single hyphen commands

lluiscampos:
On deprecated single hyphen commands, the CLI now prints and error and
    exits.
    
    Note: the previous code was also amending single hyphen flags with
    double hyphen syntax. However, closer inspection to urfave/cli package
    showed that both syntax are accepted for flags (for example --log-level
    and -log-level are both valid). Remove that part of the amending just
    for clarification (they were some flags not caught there too).
    
    Tests modified accordingly.
    
    And uses of flags modified also for double hyphen (both work, but we
    prefer this syntax).
    
    Changelog: CLI commands prefixed with hyphen are now deprecated, use
    instead directly the command name. For example `mender daemon`, `mender
    commit`, `mender show-artifact`, etc.

Additionally:

    MEN-4808: Remove deprecated flag --log-modules
    
    Which had no effect anymore.
    
    Changelog: Title

### DIFF
--- a/app/standalone.go
+++ b/app/standalone.go
@@ -1,4 +1,4 @@
-// Copyright 2020 Northern.tech AS
+// Copyright 2021 Northern.tech AS
 //
 //    Licensed under the Apache License, Version 2.0 (the "License");
 //    you may not use this file except in compliance with the License.
@@ -231,7 +231,7 @@ func doStandaloneInstallStates(art io.ReadCloser, key []byte,
 	}
 
 	if rollbackSupport {
-		fmt.Println("Use -commit to update, or -rollback to roll back the update.")
+		fmt.Println("Use 'commit' to update, or 'rollback' to roll back the update.")
 	} else {
 		fmt.Println("Artifact doesn't support rollback. Committing immediately.")
 		err = DoStandaloneCommit(device, stateExec)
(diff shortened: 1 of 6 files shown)


lluiscampos:
Testing together with:
* https://github.com/mendersoftware/integration/pull/1432
* https://github.com/mendersoftware/meta-mender/pull/1403
* https://github.com/mendersoftware/mender-image-tests/pull/52
at [Pipeline 327175179](https://gitlab.com/Northern.tech/Mender/mender-qa/-/pipelines/327175179)

Note that I am also running builds from the other PRs with current `mender` master to verify backwards compatibility of the test changes.

app/standalone.go
```diff
@@ -231,7 +231,7 @@ func doStandaloneInstallStates(art io.ReadCloser, key []byte,
 	}
 
 	if rollbackSupport {
-		fmt.Println("Use -commit to update, or -rollback to roll back the update.")
+		fmt.Println("Use commit to update, or rollback to roll back the update.")
```
oleorhagen, line 234: Suggestion:
Maybe we should single-quote the command line arguments now, that nothing is making them stand out from the rest of the text?

oleorhagen (approved):
Much better :)

app/standalone.go
lluiscampos: I thought about the same issue (command not standing out) but I wasn't sure about the solution. But now that you commented on it, makes sense to add :+1: 

I will not re-run the full pipeline, though, as this exact log line seems not to be tested for.

PR #825, opened 2021-07-07
Fix bug which caused rollback after commit if status report fails.

kacf:
The problem stems from two separate sources:

1. The reportErrorState has long been wrong in handling errors when
   the original status submitted into it was StatusSuccess. In this
   case it tries to roll back. However, we can't do that, because this
   state is always invoked after committing, so it's too late for
   that. In fact, there is nothing we can do, and it should do the
   same as all the other statuses, namely go back to the Idle
   state. So there is no need for this state at all, and it has been
   removed.

2. If a spontaneous reboot happens, the updateStatusReportState and
   updateStatusReportRetryState states were not being handled
   correctly, and the client would go down the ArtifactRollback or
   ArtifactFailure path, both of which are wrong. The mentioned
   StatusReport states always happen after one or both of those
   failure states have been visited, or after we have committed, so we
   can't visit them again. Instead, go back to try to submit the
   report, which will either succeed or eventually expire if
   unsuccessful.

Existing tests have been updated to cope with the new flow, but I have
not added any test for the second point above. There does not seem to
be much need, since all we will test is that two values from the
database leads to a particular state being returned, which is already
clear as day in the code. Everything afterwards (the actual reporting
itself), is tested elsewhere.

Fixes MEN-4830.

Changelog: Fix a bug which could sometimes lead the client to do a
rollback after it had already committed. This could happen if the
client happened to spontaneously reboot or fail during the status
update to the server. Doing this is not correct according to the state
flow, and can have unexpected consequences depending on the
combination of Update Modules and State Scripts.

(cherry picked from commit 608dc31141a334dd2f7c237b0f5478c7b53abb46)

### Description

Thank you!

### DIFF
--- a/app/state.go
+++ b/app/state.go
@@ -365,6 +365,24 @@ func (i *initState) getNextState(ctx *StateContext, sd *datastore.StateData,
 	case datastore.MenderStateUpdateCleanup:
 		return NewUpdateCleanupState(&sd.UpdateInfo, client.StatusFailure), false
 
+	// Status reports should be retried. It is possible that the original
+	// status report had a different status than Failure, but worst case
+	// this is simply a wrong report, the device will be fine, and the logs
+	// will reveal what happened.
+	case datastore.MenderStateUpdateStatusReport,
+		datastore.MenderStateStatusReportRetry:
+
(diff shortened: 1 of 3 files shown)


kacf:

PR #834, opened 2021-07-16
Add artifact_name to device provides if not found in store

prashanthjbabu:
Signed-off-by: Prashanth Joseph Babu <prashanthjbabu@gmail.com>
Changelog: Commit

- [x] Make sure that all commits have a [`Changelog`](https://github.com/mendersoftware/mender/blob/master/CONTRIBUTING.md#changelog-tags) tag. If nothing should be added to the Changelog, add a `Changelog: None` tag. If there is a change, add `Changelog: Commit`, or `Changelog: Title`, depending on what should be included in the changelog.

### Description

This PR fixes a corner case for delta updates. 
Consider a fresh device which has no device provides . 
Now when we ask for an update(`deployment/next` API) , we provide `artifact_name` as device provides by getting it from `GetCurrentArtifactName`
Now once the client gets the artifact from the server and when it comes to installing the artifact , the device calls `maybeVerifyArtifactDependsAndProvides` to ensure that the `depends` of the artifact matches with the `provides` of the device .
In such a case for a fresh device which has `no provides` in the datastore , the `LoadProvides` function does not have `artifact_name` in it causing the artifact install to fail since it wouldnt match with the `artifact_name` depends of the artifact.
To take care of this , we need to ensure that if the datastore does not return `artifact_name` we take it from `GetCurrentArtifactName`( which ensures it takes it from the artifact_info file if not present in the store.)

This issue is not seen if the first deployment is not a delta deployment since after the first deployment , the mender client stores the `artifact_name` in provides section of data store.

### DIFF
--- a/app/mender.go
+++ b/app/mender.go
@@ -637,3 +637,18 @@ func (m *Mender) InventoryRefresh() error {
 func (m *Mender) CheckScriptsCompatibility() error {
 	return m.stateScriptExecutor.CheckRootfsScriptsVersion()
 }
+
+func verifyAndSetArtifactNameInProvides(provides map[string]string, getArtifactName func() (string, error)) error {
+	if _, ok := provides["artifact_name"]; !ok {
+		artifactName, err := getArtifactName()
+		if err != nil || artifactName == "" {
+			log.Error("could not get the current Artifact name")
+			if err == nil {
+				err = errors.New("artifact name is empty")
+			}
+			return err
+		}
+		provides["artifact_name"] = artifactName
+	}
+	return nil
+}
--- a/app/standalone.go
+++ b/app/standalone.go
@@ -149,6 +149,17 @@ func doStandaloneInstallStatesDownload(art io.ReadCloser, key []byte,
 		if err != nil {
 			return nil, err
 		}
+		if _, ok := currentProvides["artifact_name"]; !ok {
+			artifactName, err := device.GetCurrentArtifactName()
+			if err != nil || artifactName == "" {
+				log.Error("could not get the current Artifact name")
+				if err == nil {
+					err = errors.New("artifact name is empty")
+				}
+				return nil, err
+			}
+			currentProvides["artifact_name"] = artifactName
+		}
 		if err = verifyArtifactDependencies(depends, currentProvides); err != nil {
 			log.Error(err.Error())
 			return nil, err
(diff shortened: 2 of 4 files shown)


prashanthjbabu:
@lluiscampos can you please take a look at this

kacf:
While I'm not against this fix, I have to ask: Does it actually solve the problem? Delta artifacts depend on checksums, and this will not be improved by this PR.

For your information, we have a much more substantial and complete fix planned for this [here](https://tracker.mender.io/browse/MEN-2582), which will also solve the checksum problem, as well as the artifact name problem which you fixed. I'm not sure exactly when this will be implemented, but I believe the third or fourth quarter this year.

prashanthjbabu:
@kacf thanks for your response . This is only for consistency reasons , since we are using the `artifact_info` fallback mechanism while asking the server for an update , the same should be maintained while validating the `provides`/`depends`  incase there is an artifact which has `artifact_name` as a key in its `depends`. 

Considering the long timeline , I feel it would be good if we can get this in till  the complete fix is in.

kacf (changes requested):
Ok, that's reasonable. I also talked to our product manager, and it's unlikely that [MEN-2582](https://tracker.mender.io/browse/MEN-2582) will make it into 3.1, our next release. So then it makes sense to include this as a partial solution in the meantime.

Content wise, this looks good, but our code checker is complaining about cyclomatic complexity. We use it to keep functions to a reasonable length. Your two sections do pretty much the same thing, so maybe you can move the logic into its own function (perhaps in `mender.go`?) and then call it from both places. That should be enough to make it happy, I think.

prashanthjbabu:
@kacf Thanks for taking a look at this . I've made the necessary changes .

kacf (approved):
Thanks! I rebased to code to fix some unrelated test errors.

kacf:
There is an error in one of the unit tests. I think I know why, but I won't have time to look at it today, and I'll be gone for a bit. I'll try to pick this up the week after next week. Given that there will not be any releases for some time, it's not that urgent anyway.

prashanthjbabu:
@kacf friendly ping :)

kacf:
Hehe, thanks. I found the issue, just a one liner. Will retest now.

kacf:
Thanks! Cherry-picked to 3.0.x and 2.5.x as well. 2.6.x won't get any more releases.

PR #835, opened 2021-07-16
Extend logs for docker module

TheMeaningfulEngineer:

### Description

Was doing a demo with the docker module so I ended up extending the
logging to understand what's going on and show meaningful messages on a
failure.

Need to retest it again, it's been a while.
Leaving in Draft until I find some time.

### DIFF
--- a/support/modules/docker
+++ b/support/modules/docker
@@ -17,18 +17,30 @@ case "$STATE" in
         echo "Yes"
         ;;
 
+    Download)
+        command -v docker >/dev/null 2>&1 || { echo >&2 "ERROR: Must have docker installed to run this update module."; exit 1; }
+        ;;
+
     ArtifactInstall)
+        >&2 echo "Storing reference on running containers for rollback needs to $prev_containers_file"
         docker ps -q > $prev_containers_file
+
+        >&2 echo "Stopping all running containers"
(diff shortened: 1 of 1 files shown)


support/modules/docker
```diff
@@ -18,17 +18,28 @@ case "$STATE" in
         ;;
 
     ArtifactInstall)
+
+        command -v docker >/dev/null 2>&1 || { echo >&2 "ERROR: Must have docker installed to run this update module."; return 1; }
```
kacf, line 22: I would recommend doing this check in `Download` instead. The reason is that if `ArtifactInstall` fails, it will try to run `ArtifactRollback`, which will probably also fail, resulting in the `INCONSISTENT` label. `Download` however, does not change the system, and hence the update immediately stops after a failure there.

kacf (approved):
One small improvement suggestion, otherwise this looks good to me!

lluiscampos (approved):

support/modules/docker
TheMeaningfulEngineer: Yes, great comment! 
Will do the check in the downloads state

TheMeaningfulEngineer: Done

lluiscampos:
The CI failure is a known issue that I am fixing in #837. I'll take care of this PR.

lluiscampos:
Rebased on top of latest master

PR #842, opened 2021-08-19
MEN-3420: Invalidate the authorization token on configuration changes

oleorhagen:


app/auth.go
```diff
@@ -208,6 +214,40 @@ func NewAuthManager(conf AuthManagerConfig) *MenderAuthManager {
 	return mgr
 }
 
+// handleCachedTenantToken Handle the cached tenant token if it is not the same
+// as the one in the configuration
+func handleCachedAuthorizationToken(db store.Store, serverURL, tenantToken []byte) error {
+	dbValue := append(serverURL, tenantToken...)
```
oleorhagen, line 220: Concatenating and using both as a value, instead of storing both. Tell me if you disagree with this approach

app/auth.go
```diff
@@ -180,6 +182,16 @@ func NewAuthManager(conf AuthManagerConfig) *MenderAuthManager {
...
 	}
 
+	tenantToken := client.AuthToken(conf.TenantToken)
+
+	if err := handleCachedAuthorizationToken(
+		conf.AuthDataStore,
+		[]byte(serverURL),
+		[]byte(tenantToken),
+	); err != nil {
+		log.Errorf("Issues handling the caching of the tenant token: %s", err.Error())
```
kacf, line 192: "Issues" sounds a bit off. Maybe "Error" instead?

app/auth.go
```diff
@@ -208,6 +220,40 @@ func NewAuthManager(conf AuthManagerConfig) *MenderAuthManager {
 	return mgr
 }
 
+// handleCachedTenantToken Handle the cached tenant token if it is not the same
+// as the one in the configuration
+func handleCachedAuthorizationToken(db store.Store, serverURL, tenantToken []byte) error {
+	dbValue := append(serverURL, tenantToken...)
+	cachedToken, err := db.ReadAll(datastore.AuthTenantTokenName)
+	if err != nil {
+		if os.IsNotExist(err) {
```
kacf, line 229: This feels a bit opposite, the only thing which happens if this is not true is to return. Isn't it better to invert the condition, and return here, then you can put the entire next block outside instead.

app/auth.go
```diff
@@ -208,6 +214,40 @@ func NewAuthManager(conf AuthManagerConfig) *MenderAuthManager {
...
+			err = db.WriteAll(datastore.AuthTenantTokenName, dbValue)
+			if err != nil {
+				return fmt.Errorf("Failed to cache the currently used tenant token to the DB. Error %s", err.Error())
```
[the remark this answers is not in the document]
kacf: I think it's ok. It makes it easy to locate where in the source code errors are happening when users are submitting reports.

app/auth.go
```diff
@@ -208,6 +220,40 @@ func NewAuthManager(conf AuthManagerConfig) *MenderAuthManager {
 	return mgr
 }
 
+// handleCachedTenantToken Handle the cached tenant token if it is not the same
+// as the one in the configuration
+func handleCachedAuthorizationToken(db store.Store, serverURL, tenantToken []byte) error {
```
kacf, line 225: Even though it probably doesn't matter for this specific case, we should always do database updates in transactions, so that there are no race conditions. So we should use `WriteTransaction` at the top here and put the all the code inside, replacing `db` with `txn`.

app/auth.go
kacf: I think that's a good approach, but I suggest something to separate them, like `___` or something.

Also, I would change the name of the database key then. Maybe this?

```
AuthTokenCacheInvalidatorName = "auth-token-cache-invalidator"
```

kacf (changes requested):

app/auth.go
```diff
@@ -180,6 +182,10 @@ func NewAuthManager(conf AuthManagerConfig) *MenderAuthManager {
 		}
 	}
 
+	if err := handleCachedAuthorizationToken(conf.AuthDataStore, []byte(serverURL), conf.TenantToken); err != nil {
```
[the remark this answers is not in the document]
kacf: Yes, I like that name!

app/auth.go
oleorhagen: https://github.com/mendersoftware/mender/pull/842/commits/300f183f58745aedd94e2ac2d6fc9cbd9b3934b9
This more like what you had in mind?
I don't really see how else to guard it in a transaction

kacf: Yes, exactly! This is the same pattern which is used in several other places in the client.

kacf (approved):

oleorhagen:

app/auth.go
```diff
@@ -208,6 +220,46 @@ func NewAuthManager(conf AuthManagerConfig) *MenderAuthManager {
 	return mgr
 }
 
+// handleCachedTenantToken Handle the cached tenant token if it is not the same
```
lluiscampos, line 223:
```suggestion
// maybeInvalidateCachedAuthorizationToken Handle the cached tenant token if it is not the same
```

oleorhagen: thx

PR #846, opened 2021-09-08
MEN-5012: Remove ability to compile without glib and D-Bus support.

kacf:
Clients from now on will be relying on a separate authentication
manager on the D-Bus.

Changelog: Commit

### DIFF
--- a/cli/commands.go
+++ b/cli/commands.go
@@ -151,15 +151,11 @@ func commonInit(config *conf.MenderConfig, opts *runOptionsType) (*app.Mender, *
 		return nil, nil, errors.New("error initializing authentication manager")
 	}
 
-	if config.DBus.Enabled {
-		api, err := dbus.GetDBusAPI()
-		if err != nil {
-			// close DB store explicitly
-			dbstore.Close()
-			return nil, nil, errors.Wrap(err, "DBus API support not available, but DBus is enabled")
-		}
-		authmgr.EnableDBus(api)
+	if !config.DBus.Enabled {
+		log.Warn(`Support for turning off DBus has been removed. "DBus.Enabled: false" setting ignored.`)
 	}
+
+	authmgr.EnableDBus(dbus.GetDBusAPI())
 
 	mp := app.MenderPieces{
 		Store:       dbstore,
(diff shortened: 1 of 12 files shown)


dbus/dbus_test.go
```diff
@@ -80,9 +77,3 @@ func TestMain(m *testing.M) {
 
 	m.Run()
 }
-
-func TestGetDBusAPI(t *testing.T) {
-	api, err := GetDBusAPI()
-	assert.NoError(t, err)
-	assert.NotNil(t, api)
```
oleorhagen, line 87: I guess this test still applies, no?
Except for the `NoError` part?

oleorhagen (approved):
Looks good.

Just an observation.

I was expecting more code change here, considering we made the whole interface in `dbus.go` in order to be able to conditionally link this out of the build, so I was sort of expecting this to go away now? And hence we could unify more code, but this works just fine too. Depends on what we want I guess.

lluiscampos (approved):
> I was expecting more code change here, considering we made the whole interface in `dbus.go` in order to be able to conditionally link this out of the build, (...)

I first had the same gut feeling: for all I know, we don't need the `interface` at all anymore and we can just convert it into a simple struct. However having the interface facilitates testing, and at the end of the day is not making the code much harder to understand neither. So I am okey with leaving it as-is.

cli/commands.go
```diff
@@ -150,16 +150,12 @@ func commonInit(config *conf.MenderConfig, opts *runOptionsType) (*app.Mender, *
...
-			// close DB store explicitly
-			dbstore.Close()
-			return nil, nil, errors.Wrap(err, "DBus API support not available, but DBus is enabled")
-		}
-		authmgr.EnableDBus(api)
+	if !config.DBus.Enabled {
+		log.Warn(`Support for turning off DBus has been removed. "DBus.Enabled: false" setting ignored.`)
 	}
 
+	authmgr.EnableDBus(dbus.GetDBusAPI())
```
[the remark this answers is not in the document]
kacf: Possibly. I'll leave it alone for now. I have to rewrite this part in the next task anyway, so this will probably change completely.

dbus/dbus_test.go
kacf: This cannot return anything but one value now, so the value of testing it is questionable. In any case, as mentioned, this part is going to change in the next task, so I'll probably write a new test as part of that instead.

kacf:
> I was expecting more code change here, considering we made the whole interface in `dbus.go` in order to be able to conditionally link this out of the build, so I was sort of expecting this to go away now? And hence we could unify more code, but this works just fine too. Depends on what we want I guess.

I was surprised too, I found this easier than I expected. But this is just the compile part. The refactoring of DBus itself comes in the next task. I have started it, and especially the parts around initialization I will have to change significantly. Plus I have to add functions for *calling* DBus now, not just serving.

PR #853, opened 2021-09-20
3.0.x: Fix crash when device has no Provides and artifact has Depends info.

kacf:
The circumstances triggering this bug are a bit specific. There needs
to be an available artifact_name, either through the database or
through `/etc/mender/artifact_info`, the artifact must come with
"Depends" fields, and the `artifact-provides` key in the database must
exist, but contain a null value. I have not figured out the exact
steps which lead to such an `artifact_provides` value, I only know
that it must be at least somewhat rare, since we have not seen in our
tests at all. However, one of our users came across it as described in
the linked ticket. In the tests I have recreated the circumstances
artificially, since it is in any case a valid database entry.

In the described circumstances, the `provide` map is nil, but since
114df81b8bcd5, we always add artifact_name to the map, so we need to
create the map if it's nil.

Fixes MEN-5073.

Changelog: None
No changelog, since this was never released in any version.

(cherry picked from commit 532374ba5345c9e4c901eaacfb119dacb641c3f0)

### DIFF
--- a/app/mender.go
+++ b/app/mender.go
@@ -638,7 +638,7 @@ func (m *Mender) CheckScriptsCompatibility() error {
 	return m.stateScriptExecutor.CheckRootfsScriptsVersion()
 }
 
-func verifyAndSetArtifactNameInProvides(provides map[string]string, getArtifactName func() (string, error)) error {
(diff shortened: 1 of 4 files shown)


kacf:

PR #855, opened 2021-09-21
Fix: entrypoint.sh, do not break if server URL contains slashes

tranchitella:
Changelog: none


oleorhagen (approved):
haha, classic

kacf (approved):

lluiscampos (approved):

PR #864, opened 2021-10-04
MEN-5098: Capture and pretty print output from scripts executed

oleorhagen:
Previously captured output would show up in the logs without any context,
seemingly in between lines, and out of place.

This buffers the output, and pretty prints it on a line-by-line basis, preceeded
by a string prefix like so:

```
Output (stdout|stderr) from command \"/usr/bin/foo\": bar
Output (stdout|stderr) from command \"/usr/bin/foo\": baz
```

Changelog: The client now shows output from scripts executed in the log as regular
log messages, prefixed by the filedescriptor they are written to, and the script
executed, like this: 'Output (stdout|stderr) from command \"/usr/bin/foo\": bar.'


system/system.go
```diff
@@ -82,28 +82,90 @@ func (c *Cmd) StdoutPipe() (io.ReadCloser, error) {
...
+			c.commandName,
+			line[:len(line)-1], // Do not write the newline delimeter
+		)
 	}
+}
 
-	return len(buf), nil
+func (c *cmdLogger) Flush() {
+	c.writeLines()
+	// Empty the remaining bytes
```
oleorhagen, line 138: A hack could be to add a newline, and then call writelines, but it felt a tad hacky

inventory/inventory_data.go
```diff
@@ -75,7 +76,8 @@ func (id *InventoryDataRunner) Get() (client.InventoryData, error) {
 
 	idec := NewInventoryDataDecoder()
 	for _, t := range tools {
-		cmd := id.cmd.Command(t)
+		cmd, logger := id.cmd.Command(t).LogStderr()
```
oleorhagen, line 79: For now only added to the inventory capturing. 
You reckon this should be added every single place we call a script?

kacf: I think "assume everywhere unless it doesn't make sense" is a good strategy.

system/system.go
kacf: Agree!

inventory/inventory_data.go
oleorhagen: With this approach, we will have to be explicit about where it is added though, so I will just go through our calls manually, and add as I see fit.

kacf: Indeed. This is why I initially wanted to have it as a property of the `system.Cmd` struct. But since this turned out pretty tricky, we'll have to do the explicit approach.

oleorhagen: Added to `identity` and `inventory` fetching.

Other places this has been added on a per command basis it looks to me. Like the update modules have something similar in place, where they write to a log struct, which wraps the output.

Do you want to unify this, or should we keep it as it is?

I know the natural answer is unify it, but this will change quite a lot of our output, so the logging will be quite different. I favour this approach, but not sure if you think it too invasive?

kacf: I also favor it. This will be minor-release material, so I think it's ok to change it. One could argue that it's a behavior change, but logs should be allowed to change, IMHO.

oleorhagen:
@mender-test-bot  start pipeline

oleorhagen:

cli/setup.go
```diff
@@ -973,11 +974,10 @@ func (opts *setupOptionsType) installDemoCertificateLocalTrust() error {
 	}
 
 	cmd := system.Command("update-ca-certificates")
-	out, err := cmd.CombinedOutput()
+	err = cmd.Run()
 
 	if err != nil {
-		return errors.Wrapf(err,
-			"update-ca-certificates returned %q", out)
```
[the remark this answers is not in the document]
kacf: The behavior is subtly different. Now it always prints the output, I think the idea was to hide the output if there is no error. I think I would vote for keeping it as is (see also comment about `CombinedOutput`).

system/system.go
```diff
@@ -59,9 +61,35 @@ type Cmd struct {
...
 	c.Stdout = nil
-	return c.Cmd.Output()
+	b, err := c.Cmd.Output()
+	if logger, ok := c.Stderr.(Flusher); ok {
+		logger.Flush()
+	}
+	return b, err
 }
 
 func (c *Cmd) CombinedOutput() ([]byte, error) {
```
kacf, line 95: `Flush()` should be called in here as well, or?

kacf (changes requested):
> @kacf I did sort of a 180 here, this is the attempt, whilst implicitly logging the output from the commands run, then closing them through the interfaces which do execute, like _Run_, _Wait_, _Output_, etc.

Yes, I like it! Overriding `Run()` & friends is a very good idea.

system/system.go
oleorhagen: No, it should not :)
If you think about it you know why ;)

kacf: Haha, you're right, I see it now!

oleorhagen:
@mender-test-bot 
cherry-pick to:
* 3.1.x
* 3.0.x
* 2.6.x

oleorhagen:
@mender-test-bot 
cherry-pick to:
* 3.0.x

oleorhagen:
@mender-test-bot 
cherry-pick to:
* 3.0.x

PR #866, opened 2021-10-04
CI: Create a Junit test report for Gitlab

oleorhagen:
Changelog: None

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -18,8 +18,10 @@ test:
   before_script:
     - apt-get update && apt-get install -yyq liblzma-dev libssl-dev libglib2.0-dev dbus clang-format-9
     - make get-tools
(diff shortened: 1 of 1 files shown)


lluiscampos:
> I'm thinking it won't hurt?

I'm thinking: it is a nice addition!

> 
> Thinking:
> 
>     * I added it to the `gitlab-ci` file, since this is really only CI specific (can be part of the makefile if you think it valid)

I agree, best here.

> 
>     * I run it in series, after the other make targets, instead of adding it to the current coverage make target (it's double the work, but it's not hugely expensive - should probably be added prior to the other test targets, so that we do get the output in case of test error -- me genius)

I agree too. Not only due to the effort, but also because it is pointless to run it locally while developing so no need to alter the current flow.

> 
> 
> What you think?

lluiscampos (approved):
Looks very nice on GitLab!

PR #867, opened 2021-10-04
style: sort the imports by global, external and local

oleorhagen:
Ran goimports -local=github.com/mendersoftware, to sort the imports properly.

Changelog: None

### DIFF
--- a/app/client_closures.go
+++ b/app/client_closures.go
@@ -15,9 +15,10 @@
 package app
 
 import (
+	log "github.com/sirupsen/logrus"
+
 	"github.com/mendersoftware/mender/client"
 	"github.com/mendersoftware/mender/conf"
-	log "github.com/sirupsen/logrus"
 )
 
 // see client.go: ApiRequest.Do()
--- a/app/deployment_logger.go
+++ b/app/deployment_logger.go
@@ -26,8 +26,9 @@ import (
 	"strings"
 	"syscall"
 
-	"github.com/mendersoftware/mender/conf"
 	log "github.com/sirupsen/logrus"
+
+	"github.com/mendersoftware/mender/conf"
 )
 
 // error messages
--- a/app/standalone.go
+++ b/app/standalone.go
@@ -22,6 +22,9 @@ import (
 	"os"
 	"strings"
 
+	"github.com/pkg/errors"
+	log "github.com/sirupsen/logrus"
+
 	"github.com/mendersoftware/mender/client"
 	"github.com/mendersoftware/mender/conf"
 	"github.com/mendersoftware/mender/datastore"
@@ -30,8 +33,6 @@ import (
 	"github.com/mendersoftware/mender/statescript"
 	"github.com/mendersoftware/mender/store"
 	"github.com/mendersoftware/mender/utils"
-	"github.com/pkg/errors"
-	log "github.com/sirupsen/logrus"
 )
 
 var (
--- a/app/standalone_test.go
+++ b/app/standalone_test.go
@@ -23,6 +23,9 @@ import (
 	"strings"
 	"testing"
 
+	"github.com/stretchr/testify/assert"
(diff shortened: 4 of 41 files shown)


oleorhagen:
Necessary you say?

Probably not, but it sure does not hurt.

conf/paths.go
```diff
@@ -12,6 +12,7 @@
 //    See the License for the specific language governing permissions and
 //    limitations under the License.
 
+//go:build !local
```
lluiscampos, line 15: Where are these comments coming from?

dbus/mocks/DBusAPI.go
```diff
@@ -17,8 +17,9 @@
 package mocks
 
 import (
-	dbus "github.com/mendersoftware/mender/dbus"
 	mock "github.com/stretchr/testify/mock"
+
+	dbus "github.com/mendersoftware/mender/dbus"
 )
 
 // DBusAPI is an autogenerated mock type for the DBusAPI type
```
lluiscampos, line 25: This one being an auto-generated file, I think we should not change.

lluiscampos (commented):
I think we should revert the mock change and these comments on the build flags.

conf/paths.go
oleorhagen: So, appareantly this is coming with `Go 1.17`, and is the new way of adding tags, so this is a `gofmt` thing.

So no need to fear, but still, removing them from this

dbus/mocks/DBusAPI.go
oleorhagen: Good catch! Did not strike me!

oleorhagen: Should be safe though, considering we will just regenerate it next time, but still, let's keep it pristine - thanks!

dbus/dbus_libgio.go
```diff
@@ -11,6 +11,7 @@
 //    WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 //    See the License for the specific language governing permissions and
 //    limitations under the License.
+
```
lluiscampos, line 14: Please don't add these empty spaces

lluiscampos (changes requested):

kacf:
Could this wait until after [MEN-5011](https://tracker.mender.io/browse/MEN-5011) is done, and then done on the `feature-unified_comms` branch? This is going to make it a nightmare to merge with master otherwise, since almost every file has changed there. I assume since this was done in an automatic fashion, it shouldn't be too hard, right?

oleorhagen:
> Could this wait until after [MEN-5011](https://tracker.mender.io/browse/MEN-5011) is done, and then done on the `feature-unified_comms` branch? This is going to make it a nightmare to merge with master otherwise, since almost every file has changed there. I assume since this was done in an automatic fashion, it shouldn't be too hard, right?

Sure, just ping me here once ready

conf/paths.go
lluiscampos, line 15: We have now added these back :shrug:

oleorhagen: Moth*&*%(*&%,

lluiscampos: Man, do `git add -p` and review patch by patch what you are adding to your commit :)

oleorhagen: Yeh, lesson learnt here. Will stop trusting the tooling :see_no_evil: 
sry

lluiscampos: Trust, but verify :heavy_check_mark:

lluiscampos (approved):

conf/paths.go
oleorhagen: haha, alright.
Manually staged every chunk, scrolled through it in GitHub.
Will commence to buy coffee in case of discovered errors.

PR #872, opened 2021-10-25
MEN-5011: Split Mender client and its Auth Manager into two separate components

kacf:
Some considerations when reviewing:

* I encourage reading the commit messages

* For the big one, I recommend viewing locally with `git show -M30 --color-moved=dimmed_zebra ':!vendor'`. Even with this the diff is huge, but it's at least a little bit easier to see where things are moving. Interestingly the `--shortstat` of the above command shows that less than 300 lines of actual code has been added!

* Keep in mind that a lot of stuff here is pre-existing and just moved / slightly tweaked.

I have several TODOs that I still am going to fix in here, but none are too major, so it can be done in parallel with reviewing. I will not change the existing commits in here unless absolutely necessary, only add new ones.

One thing I want to ask about though: I'm unsure about the `common/tls` package name. It started out as TLS (HTTPS) only, but then took on HTTP as well. I considered renaming it to `net`, but this is even more common in Golang world than the names I've already used, so maybe not great. Suggestions?


kacf:
@lluiscampos: Two new small fixes.

The warning message has been there for a long time, and was picked up by the syslogger test. Why it hasn't failed before, I don't know. It's possible that slight changes in config parsing hid it before. I didn't look too deep, because the fix is a good one regardless.

kacf:
Despite a very high failure rate for mtls tests (I have not been able to get a clean pass in three consecutive runs), I have verified manually on Gitlab that they *do* in fact pass. I can't let it block this anymore. If the failure rate has indeed increased, I propose that we take a look at solving [QA-243](https://tracker.mender.io/browse/QA-243) instead.

Merging this.

PR #877, opened 2021-11-05
Resolv symlink /dev/root while getting active partition

wjax:
Changelog: Fix error not finding active partition for systems where /dev/root is a symlink

Signed-off-by: Jesus <wjaxxx@gmail.com>

### Description

Fix an error where mender is not able to match the returned mount device with RootfsA or B. This happens in systems where /dev/root is a symlink to the actual partition

### DIFF
--- a/installer/partitions.go
+++ b/installer/partitions.go
@@ -1,4 +1,4 @@
-// Copyright 2020 Northern.tech AS
+// Copyright 2021 Northern.tech AS
 //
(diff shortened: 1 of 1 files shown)


kacf (approved):
Thanks, I think this should work, but the CI will be the ultimate judge.

kacf:
@mender-test-bot
cherry-pick to:
* 3.0.x
* 2.5.x
* 3.1.x

PR #879, opened 2021-11-08
Disable HTTP Keep-Alive.

merlin-northern:
ChangeLog:Allow to disable HTTP Keep-Alive from config.

goes with:
* https://github.com/mendersoftware/mender-docs/pull/1668
* https://github.com/mendersoftware/integration/pull/1921

### DIFF
--- a/client/client.go
+++ b/client/client.go
@@ -546,8 +546,16 @@ func newHttpsClient(conf Config) (*http.Client, error) {
 		return nil, err
 	}
 
+	disableKeepAlive := false
+	idleConnTimeoutSeconds := 0
+	if conf.Connectivity != nil {
+		disableKeepAlive = conf.Connectivity.DisableKeepAlive
+		idleConnTimeoutSeconds = conf.Connectivity.IdleConnTimeoutSeconds
+	}
 	transport := http.Transport{
-		Proxy: http.ProxyFromEnvironment,
+		DisableKeepAlives: disableKeepAlive,
+		IdleConnTimeout:   time.Duration(idleConnTimeoutSeconds) * time.Second,
+		Proxy:             http.ProxyFromEnvironment,
 		DialTLS: func(network string, addr string) (net.Conn, error) {
 			return dialOpenSSL(ctx, &conf, network, addr)
 		},
@@ -579,6 +587,19 @@ type Security struct {
 	SSLEngine      string `json:",omitempty"`
 }
 
+// Connectivity instructs the client how we want to treat the keep alive connections
+// and when a connection is considered idle and therefore closed
+// NOTE: Careful when changing this, the struct is exposed directly in the
+// 'mender.conf' file.
+type Connectivity struct {
+	// If set to true, there will be no persistent connections, and every
+	// HTTP transaction will try to establish a new connection
+	DisableKeepAlive bool `json:",omitempty"`
+	// A number of seconds after which a connection is considered idle and closed.
+	// The longer this is the longer connections are up after the first call over HTTP
+	IdleConnTimeoutSeconds int `json:",omitempty"`
+}
+
 func (h *HttpsClient) Validate() {
 	if h == nil {
 		return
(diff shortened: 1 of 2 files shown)


merlin-northern:
ah one thing I need to fix here. wait.

conf/config.go
```diff
@@ -288,6 +288,8 @@ func SaveConfigFile(config *MenderConfigFromFile, filename string) error {
 func maybeHTTPSClient(c *MenderConfig) *client.HttpsClient {
 	if c.HttpsClient.Certificate != "" && c.HttpsClient.Key != "" {
 		return &c.HttpsClient
+	} else if c.HttpsClient.DisableKeepAlive {
+		return &c.HttpsClient
```
kacf, line 292: AFAICS, `HttpsClient` is only checked [here](https://github.com/mendersoftware/mender/blob/1b3b946337573e1f20ab205d8a1868c9c7bc13ee/client/client.go#L512), which is right next to the part where you add `DisableKeepAlive` to the transport. So is this really needed?

kacf (commented):
Remember to also add an entry [here](https://docs.mender.io/client-installation/configuration-file/configuration-options).

conf/config.go
merlin-northern: I will check that -- but this was the first version that worked.

merlin-northern:
> Remember to also add an entry [here](https://docs.mender.io/client-installation/configuration-file/configuration-options).

thanks for the reminder. I want to add a whole section on the "Network Usage" by the Client and explain everything with this regard, including the keep alive connections and the statistics of typical usage. makes sense?

merlin-northern:
what I am doing now is: gathering more data from the live device connected to hosted Mender. let's see what is the out come before we merge this.

conf/config.go
merlin-northern: yes, it is, since `newHttpsClient` gets called from `New(conf Config)` and it is not what is in the config file, but what `GetHttpConfig() client.Config` returns.

kacf: Ok, now I see what you mean.

But this made me realize that `DisableKeepAlive` does not belong in `HttpsClient`, which is about client side HTTPS settings. `DisableKeepAlive` is equally valid for HTTP. So how about a `Connections` sub section or something like that?

```json
{
  "Connections": {
    "DisableKeepAlive": true
  }
}
```

I'm thinking that this is a place we could also put `MaxDownloadSpeed` which is another connection related requested feature.

merlin-northern: good idea! in the mean time I added a new option: `IdleConnTimeoutSeconds` -- I am now running tests with that (I think it is something you suggested) it may make more sense to have it as well.

merlin-northern:
I am keeping it here, as I am still waiting for the feed back.

merlin-northern:
this is not stale. we are actively working and testing it on a real infra.

conf/config.go
merlin-northern: I kind of added `Connectivity` looks good?

client/client.go
```diff
@@ -579,6 +587,17 @@ type Security struct {
...
 }
 
+// Connectivity instructs the client how we want to treat the keep alive connections
+// and when a connection is considered idle and therefore closed
+type Connectivity struct {
+	// if set to true, there will be no persistent connections, and every
+	// HTTP transaction will try to establish a new connection
+	DisableKeepAlive bool `json:",omitempty"`
+	// a number of seconds after which a connection is considered idle and closed
+	// the longer this is the longer connections are up after the first call over HTTP
```
kacf, line 597: Nitpick: It took me a while to understand what this meant because it looks like one sentence stretched over two lines, but it's really two sentences. Please add a period after "closed", and even capital letters would be nice! :-)

merlin-northern: done.

merlin-northern:
> 

hehe. thanks for the reminder: https://github.com/mendersoftware/mender-docs/pull/1668

merlin-northern:
@kacf this was needed: https://github.com/mendersoftware/mender/compare/c33cd96469b0644bdfaf41ad3260c1c0999bfcde..9129c81f5d47a9f23a13b2c5bce03f6ecc94c540 (during one of the rebases I dropped that, I do not think it was in during the last review)

kacf (approved):
:+1:

PR #885, opened 2021-11-12
installer/bootenv: support systemd-boot tools

zach-welch-timesys:
This provides support for using `systemd-boot` tools to managed the boot environment.  This patch is required as part of mendersoftware/meta-mender#1521


kacf (approved):

zach-welch-timesys:
It would be nice to have this feature backported to dunfell (release 3.1.x).

kacf:
3.2 is right around the corner and will be released at the same time as the next 3.1, both on master and dunfell.

PR #888, opened 2021-11-17
[tentative] Feature Unified Communications

lluiscampos:


lluiscampos:
This PR is only for future reference. We have decided to abandon this design. Closing.

kacf:
> This PR is only for future reference. We have decided to abandon this design. Closing.

👍 I was just thinking we should have something like this!

PR #890, opened 2021-11-24
A couple of fixes for update control map wait logic.

kacf:
See individual commits.

### DIFF
--- a/app/state.go
+++ b/app/state.go
@@ -2104,6 +2104,12 @@ func (c *controlMapPauseState) Handle(ctx *StateContext, controller Controller) 
 
 	updateMapFromServerIn := time.Until(nextServerMapRefresh)
 
+	// For rapid feedback, use the smallest number of the map refresh time
+	// and the update poll interval. The latter is often much shorter.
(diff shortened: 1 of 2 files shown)


kacf:

lluiscampos (approved):
They look good to me

app/state.go
```diff
@@ -2075,6 +2075,12 @@ func (c *controlMapPauseState) Handle(ctx *StateContext, controller Controller)
 
 	updateMapFromServerIn := time.Until(nextServerMapRefresh)
 
+	// For rapid feedback, use the smallest number of the map refresh time
+	// and the update poll interval. The latter is often much shorter.
+	if updateMapFromServerIn > controller.GetUpdatePollInterval() {
```
oleorhagen, line 2080: I think we have a Min function somewhere? :thinking:

oleorhagen (approved):

app/state.go
kacf: Yes, but for ints. Golang is really picky about types, so here it made more sense to just a do a check.

PR #903, opened 2021-12-03
MEN-5217, MEN-5218: Alternative path in MEN-4953 epic.

kacf:


client/client.go
```diff
@@ -129,154 +140,130 @@ func (a *APIError) Unwrap() error {
...
-			log.Warnf("Reauthorization failed with error: %s", e.Error())
+	// First prefill newURL with existing URL.
+	newURL, _ := url.Parse(req.URL.String())
+
+	// Then selectively fill in the prefix from ServerURL.
+	newURL.Scheme = serverURL.Scheme
+	newURL.Host = serverURL.Host
+	newURL.Path = strings.TrimRight(serverURL.Path, "/") +
+		"/" +
+		strings.TrimLeft(req.URL.Path, "/")
```
lluiscampos, line 179: Why we need the first prefill here?

lluiscampos (commented):
The change-set looks good to me. Maybe identical, but everything read very familiar from what I remember from the original epic review.

Before final approval, I want to have a second look at `app/auth.go` once you do the final rebase from master (my PR was merged yesterday).

client/client.go
kacf: The Golang docs state "relative paths may omit leading slash". I have no idea under what circumstances this can happen, and if it is in fact applicable here. But I thought it's better to be safe than sorry, and just remove and readd it every time.

For some context: This took me ages to debug, I was getting successful connections, but it simply refused to use POST, always using GET. It turns out that what happened was that it was connecting to an endpoint with an extra slash in the beginning, and this was being automatically redirected by the Golang server using 301. But 301 redirects do not preserve the method, hence it resorted to GET. This all happened behind my back, outside of my defined handlers, so it took quite some time to find it.

kacf:
> The change-set looks good to me. Maybe identical, but everything read very familiar from what I remember from the original epic review.

Hehe, I expected you to feel familiar. Some of it is identical, but some parts, especially the authorization code, is largely rewritten, even though it is based on the same principle.

> Before final approval, I want to have a second look at `app/auth.go` once you do the final rebase from master (my PR was merged yesterday).

Ack. Will do it after fixing the Gitlab errors.

kacf:
@lluiscampos: Updated!

lluiscampos (approved):

app/auth.go
```diff
@@ -534,7 +471,7 @@ func (m *menderAuthManagerService) broadcast(message AuthManagerResponse) {
 	if m.dbus != nil {
 		tokenAndServerURL := dbus.TokenAndServerURL{
 			Token:     string(message.AuthToken),
-			ServerURL: m.localProxy.GetServerUrl(),
+			ServerURL: string(message.ServerURL),
```
lluiscampos, line 474: :+1:

lluiscampos (approved):

kacf:
Only spurious failures. Merging.

PR #904, opened 2021-12-03
Fix several bugs in unit test reporting.

kacf:
* Make sure we post the log from the tests.

* The eval section which is supposed to capture failing tests doesn't
  work, because the eval happens at parse time, so the file is empty.
  Fix by doing it in shell.

Changelog: None

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -253,9 +253,10 @@ htmlcover: coverage
 
 coverage:
 	rm -f coverage.txt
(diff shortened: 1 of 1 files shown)


oleorhagen (approved):

kacf:
There is still something weird in here, the log is always truncated. Does it throttle output that arrives very rapidly or something..? I think this was already happening before though, so I'm not blaming this PR. It's just bringing back that the output is there at all.

PR #905, opened 2021-12-07
MEN-5273: proxy: Fix websocket connection for advanced auth settings

lluiscampos:
There is an implementation detail of websocket library, where the method
websocket.Dialer.Dial (set by in our case by the client to use OpenSSL
advance auth features) is ignored for wss/https requests. Force here to
use ws/http for our function to be called. See
https://github.com/gorilla/websocket/blob/e8629af678b7fe13f35dff5e197de93b4148a909/client.go#L313

Changelog: None
No changelog, commit 84204a3 claims to support websockets, this commit
just fixes a bug there which has not been released.

### DIFF
--- a/app/proxy/proxy_ws.go
+++ b/app/proxy/proxy_ws.go
@@ -77,14 +77,20 @@ func (pc *proxyControllerInner) DoWsUpgrade(w http.ResponseWriter, r *http.Reque
 	requestHeader.Del("Connection")
 
 	wsUrl := url.URL{
+		// nolint:lll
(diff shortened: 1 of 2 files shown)


lluiscampos:
@mender-test-bot start pipeline :snake:

lluiscampos:
> @mender-test-bot start pipeline snake

:repeat_one:

kacf (approved):

lluiscampos:
The workaround was enough for certain cases (CustomCert, mTLS) but not for the case when the Server certificate is installed in the system CA root. I'll submit an alternative fix in a new PR.

PR #908, opened 2021-12-14
Fix minor typo in log entry

lluiscampos:


merlin-northern (approved):

tranchitella (approved):

app/proxy/proxy_ws.go
```diff
@@ -221,7 +221,7 @@ func (pc *ProxyController) CloseWsConnections() {
 		c.connClientWriteMutex.Lock()
 		errWrite = c.connClient.WriteMessage(websocket.CloseMessage, m)
 		if errWrite != nil {
-			log.Errorf("error while sending close message to backend: %v", errWrite.Error())
+			log.Errorf("error while sending close message to client: %v", errWrite.Error())
```
merlin-northern, line 224: wait. it is correct.

merlin-northern: the log refers to the `connClient.` so it should stay, right?

PR #922, opened 2022-01-11
Set Mender client's umask to `0077` in the systemd service

tranchitella:
The Mender client does not explicitly change permissions of files and
directories created when processing artifacts. With this change, we
ensure that all the temporary files and directories created by the
Mender client will be readable by the root user only.

Changelog: title

### DIFF
--- a/support/mender-client.service
+++ b/support/mender-client.service
@@ -10,6 +10,7 @@ User=root
 Group=root
 ExecStart=/usr/bin/mender daemon
 Restart=on-abort
+UMask=0077
 
 [Install]
 WantedBy=multi-user.target


tranchitella:

kacf:
Looks good. However, I think this deserves a runtime test. Perhaps you can insert it in one of the two tests [here](https://github.com/mendersoftware/mender-image-tests/blob/master/tests/test_update_modules.py), between the install and commit stages.

tranchitella:
No need to change the umask, we are already okay as the path where we unpack the artifact has permissions set to `0700`:
https://github.com/mendersoftware/mender/blob/master/installer/modules.go#L137

PR #931, opened 2022-01-28
Add reboot type AutomaticRebootSkipStatusReport for update modules

agreen17:
Changelog: Allow update modules to optionally skip reporting status to the
server before rebooting

We have encountered a situation where this functionality would be useful - would the mender team be open to accepting a patch for this?

## Description

Before this change, after installing an update module, the mender client would unconditionally try to notify the server of the client's upcoming state transition from ArtifactInstall -> ArtifactReboot. This poses a problem for update modules which (for some reason or another) are expected to break upstream connectivity for the target device. In this case, the mender client will "hang" for a while trying to report status until eventually the client errors out and the update fails.

This change allows an update module to respond with "AutomaticSkipStatusReport" in order to reboot the host device after installing the update module while bypassing mender's normal routine of notifying the server before the state transition into ArtifactReboot. This option should be used by update modules which have an install stage that is expected to break upstream connectivity to the mender server.

### DIFF
--- a/Documentation/update-modules-v3-file-api.md
+++ b/Documentation/update-modules-v3-file-api.md
@@ -84,6 +84,12 @@ The module should print one of the valid responses:
   reboot. **This is usually the best choice** for all modules that just require
   a normal reboot, but modules that reboot a peripheral device may need to use
   `Yes` instead, and implement their own method.
+* `AutomaticSkipStatusReport` - Almost identical semantics to `Automatic` - the
+  only difference being that **the mender client will not try to send a status
+  report to the server saying that the device is rebooting before carrying out
+  the reboot.** This response should be used specifically for update modules
+  which are expected to break upstream connectivity such that the client is
+  expected to be unable to report status to the server until after rebooting.
 * `Yes` - Mender will run the update module with the `ArtifactReboot`
   argument. Use this when you want to reboot a peripheral device that's
   connected to the host. Don't use this if you want to reboot the host that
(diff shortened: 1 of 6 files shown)


kacf:
> In this case, the mender client will "hang" for a while trying to report status until eventually the client errors out and the update fails.

It's true that the client will hang for a bit while trying to report status, but it is not supposed to fail. See [this block](https://github.com/mendersoftware/mender/blob/3.2.1/app/mender.go#L392-L398): A fatal error is only produced if the deployment has been explicitly aborted, otherwise the update continues. So this change should not be necessary.

Are you able to find out exactly what error is being returned by the `ReportUpdateStatus` call?

agreen17:
@kacf Thanks for the reply and apologies - that initial explanation was not entirely accurate!

What we saw was that after the update module was successfully installed (and upstream connectivity from the device was broken, as expected) the mender client was hanging attempting to report an error to the server for a very long time. We did not see the call to report status fail even though upstream connectivity was lost. Manually disabling the network interface in question via `ip link set dev $dev down` also did not seem to fix the client hanging trying to report update status. Through another mechanism outside of mender, we rebooted the target device after some time since it could not reach upstream. So in our case the update eventually "succeeds" in that the target successfully installed the update and the device rebooted, but our target was left in a non-operational state for some time before it was usable again.

We thought it reasonable to skip phoning home altogether since the update module in question will break phoning home by design, Instead of trying to figure out how to successfully get the mender client to quickly fail phoning home.

Given all this, do you still think the change should not be necessary? If so, I would greatly appreciate some guidance on how to avoid this situation.

kacf:
It is a bit surprising that the client hangs, especially after bringing the network down. As far as I know, there is no retry logic inside any of the status report calls, except for the last one before committing, but this is not relevant here. In other words the calls should fail immediately.

I've got a couple of things to try which could maybe help identify the issue:

* If the network is down, it may be DNS which is hanging instead of the call itself. Sometimes these are local services which have their own retry mechanisms. Try shutting down DNS services as well, such as `systemd-resolved`.
* Create an `ArtifactReboot_Enter` state script and attach to the artifact. Inside, try to do a call to the server manually, for example:
  ```bash
  date
  curl https://hosted.mender.io | head
  date
  ```
  Does that also hang? Check the timestamps between the two `date` calls. If it hangs, it's an indication that the network is not fully down.
* Try running the client with `--log-level debug` and see if you get more info.

I'm a bit reluctant to accept this change because it is such a special case, and should not be necessary if things are working as they should.

PR #932, opened 2022-01-29
MEN-5421 - Only fall back to older endpoints on 404's

oleorhagen:


oleorhagen:

oleorhagen:
Only tested with hosted mender manually, works.

Bug does not reproduce.

Integration tests coming

oleorhagen:
It could also be an idea to add a minimum time in between requests to `/deployments/next`.

What do you think @kacf ?

oleorhagen:
I have not added an integration test for this yet, but it is in the works.

client/client_update.go
```diff
@@ -143,13 +143,20 @@ func findFirstWorkingEndpoint(api ApiRequester, reqs []*http.Request) (*http.Res
 		default:
 			r.Body.Close()
 
-			// Fall back to alternative methods/endpoints on other error codes.
-			if r.StatusCode >= 400 && r.StatusCode < 600 {
+			// Fall back to alternative methods/endpoints on 404's
+			if r.StatusCode == 404 {
```
kacf, line 147: In tests I suppose numbers are ok, but in production code, I think we should use `http.StatusNotFound` instead.

kacf (approved):

kjaskiewiczz (approved):
looks good

oleorhagen:
@mender-test-bot 
cherry-pick to:
* 3.2.x

oleorhagen:
@lluiscampos I only cherried to 3.2.x.

Technically the fix applies for older branches, too. But the bug is never triggered.

I'll leave it to you to decide what to do for the cherry-picks.

lluiscampos:
@oleorhagen Thanks. I think is best not to cherry-pick. With no re-fetching of the update control maps the bug won't be triggered, while in the other hand there is a risk of adding some sort of regression in an area that is hard to test (wrt rate limiting the client)

PR #941, opened 2022-02-21
feat: Inventory push retries and backoff.

merlin-northern:
We use RetryPollIntervalSeconds in inventory with the exponential
backoff via GetExponentialBackoffTime together with a new setting:
* RetryPollCount -- the max number of tries
* inventory by default tries 3 times with one minute intervals
  (GetExponentialBackoffTime defaults)

ChangeLog:commit
ChangeLog:RetryPollCount applies to all places where backoff is present

### DIFF
--- a/app/mender.go
+++ b/app/mender.go
@@ -1,4 +1,4 @@
-// Copyright 2021 Northern.tech AS
+// Copyright 2022 Northern.tech AS
 //
 //    Licensed under the Apache License, Version 2.0 (the "License");
 //    you may not use this file except in compliance with the License.
@@ -45,6 +45,7 @@ type Controller interface {
 	GetUpdatePollInterval() time.Duration
 	GetInventoryPollInterval() time.Duration
 	GetRetryPollInterval() time.Duration
+	GetRetryPollCount() int
 
 	CheckUpdate() (*datastore.UpdateInfo, menderError)
 	FetchUpdate(url string) (io.ReadCloser, int64, error)
@@ -444,6 +445,10 @@ func (m *Mender) GetRetryPollInterval() time.Duration {
 	return t
 }
 
+func (m *Mender) GetRetryPollCount() int {
+	return m.Config.RetryPollCount
+}
+
 func (m *Mender) SetNextState(s State) {
 	m.state = s
 }
@@ -564,7 +569,7 @@ func transitionState(to State, ctx *StateContext, c Controller) (State, bool) {
 }
 
 func (m *Mender) InventoryRefresh() error {
-	ic := client.NewInventory()
+	ic := client.NewInventory(m.GetRetryPollInterval(), m.GetRetryPollCount())
 	idg := inv.NewInventoryDataRunner(path.Join(conf.GetDataDirPath(), "inventory"))
 
 	artifactName, err := m.GetCurrentArtifactName()
--- a/client/client.go
+++ b/client/client.go
@@ -651,17 +651,23 @@ var MaxRetriesExceededError = errors.New("Tried maximum amount of times")
 // Simple algorithm: Start with one minute, and try three times, then double
 // interval (maxInterval is maximum) and try again. Repeat until we tried
 // three times with maxInterval.
-func GetExponentialBackoffTime(tried int, maxInterval time.Duration) (time.Duration, error) {
+func GetExponentialBackoffTime(tried int,
+	maxInterval time.Duration,
+	maxAttempts int) (time.Duration, error) {
 	const perIntervalAttempts = 3
 
 	interval := 1 * ExponentialBackoffSmallestUnit
 	nextInterval := interval
 
+	if maxAttempts > 0 && tried >= maxAttempts {
+		return 0, MaxRetriesExceededError
+	}
+
 	for c := 0; c <= tried; c += perIntervalAttempts {
 		interval = nextInterval
 		nextInterval *= 2
 		if interval >= maxInterval {
-			if tried-c >= perIntervalAttempts {
+			if maxAttempts <= 0 && tried-c >= perIntervalAttempts {
 				// At max interval and already tried three
 				// times. Give up.
 				return 0, MaxRetriesExceededError
--- a/client/client_inventory.go
+++ b/client/client_inventory.go
@@ -1,4 +1,4 @@
-// Copyright 2021 Northern.tech AS
+// Copyright 2022 Northern.tech AS
 //
 //    Licensed under the Apache License, Version 2.0 (the "License");
 //    you may not use this file except in compliance with the License.
@@ -17,6 +17,7 @@ import (
 	"bytes"
 	"encoding/json"
 	"net/http"
+	"time"
 
 	"github.com/pkg/errors"
 	log "github.com/sirupsen/logrus"
@@ -27,21 +28,37 @@ type InventorySubmitter interface {
 }
 
 type InventoryClient struct {
+	retryInterval time.Duration
+	maxTries      int
 }
 
-func NewInventory() InventorySubmitter {
-	return &InventoryClient{}
+func NewInventory(retryInterval time.Duration,
+	maxTries int) InventorySubmitter {
+	return &InventoryClient{
+		retryInterval: retryInterval,
+		maxTries:      maxTries,
+	}
 }
 
 // Submit reports status information to the backend
 func (i *InventoryClient) Submit(api ApiRequester, url string, data interface{}) error {
 	// PATCH used to be the only method available in Mender Product 2.5, so
 	// fall back to that if PUT fails.
-	r, err := doSubmitInventory(api, http.MethodPut, url, data)
+	r, err := doSubmitInventory(api,
+		http.MethodPut,
+		url,
+		data,
+		i.retryInterval,
+		i.maxTries)
 	if err == nil {
 		defer r.Body.Close()
 	} else if r != nil && r.StatusCode == http.StatusMethodNotAllowed {
-		r, err = doSubmitInventory(api, http.MethodPatch, url, data)
+		r, err = doSubmitInventory(api,
+			http.MethodPatch,
+			url,
+			data,
+			i.retryInterval,
+			i.maxTries)
 		if err == nil {
 			defer r.Body.Close()
 		}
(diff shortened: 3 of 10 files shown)


kacf:
It's not really clear to me what is being solved here. What's the problem? Inventory is constantly reported, so what's the point of having retrying?

Regardless of that, I don't think we should introduce retry flags specifically for inventory. If so we should rather fix our existing flags. The generic idea of number of retries as well as the retry policy would be applicable to our existing mechanism, I think. It would be great to have a ticket for this so we can discuss it before diving into implementation though. The mechanism is fragile enough as it is, and I'd like input from everyone before we build yet another solution that people will complain about.

merlin-northern:
> 

discussed and agreed upon. I am reworking. thanks Kristian.

conf/config.go
```diff
@@ -69,6 +69,8 @@ type MenderConfigFromFile struct {
 
 	// Global retry polling max interval for fetching update, authorize wait and update status
 	RetryPollIntervalSeconds int `json:",omitempty"`
+	// Global max retry poll count
+	RetryPollCount int `json:",omitempty"`
```
merlin-northern, line 73: @kacf do you think we should use that outside inventory too?

merlin-northern:
and I do remember about the docs update :)

client/client_inventory.go
```diff
@@ -68,8 +87,26 @@ func doSubmitInventory(
 
 	r, err := api.Do(req)
 	if err != nil {
-		log.Error("Failed to submit inventory data: ", err)
-		return r, errors.Wrapf(err, "inventory submit failed")
+		if tries != 0 {
+			for try := 0; try < tries; try++ {
+				sleepInterval, _ := GetExponentialBackoffTime(try, time.Duration(tries)*time.Minute)
```
kacf, line 92: Converting `tries` to `time.Duration`..?

Anyway, this will not work correctly, I believe. `GetExponentialBackoffTime` internally keeps track of the number of retries, and will return error once it is exceeded. That's why I think it should have an extra argument. If it's zero, it uses the existing self-calculated number of retries. If it's non-zero, it uses that exact number of retries. Then you also need to base the `for` loop on the return from `GetExponentialBackoffTime`, not directly on the number of retries.

Giving it an extra argument also allows you to easily extend the flag to work with the existing retries as well, which is the point of introducing a common flag instead of one just for inventory.

If you do change it, a couple of new test cases for this inside `TestExponentialBackoffTimeCalculation` would also be great!

And lastly, I think this change actually warrants *two* changelog entries! The second one can just say that the new `RetryPollCount` flag works for existing retries as well. Rest assured then that you have implemented a lot of value! 😉

client/client_inventory.go
```diff
@@ -68,8 +87,26 @@ func doSubmitInventory(
 
 	r, err := api.Do(req)
 	if err != nil {
-		log.Error("Failed to submit inventory data: ", err)
-		return r, errors.Wrapf(err, "inventory submit failed")
+		if tries != 0 {
```
kacf, line 90: I think we might as well drop this check and let `GetExponentialBackoffTime` handle it always. That would mean that we would introduce the default number of retries for everybody, but I don't think that's a problem. Agree?

conf/config.go
kacf: Yes. Should be relatively easy if you change `GetExponentialBackoffTime` as mentioned.

app/mender.go
```diff
@@ -444,6 +444,10 @@ func (m *Mender) GetRetryPollInterval() time.Duration {
 	return t
 }
 
+func (m *Mender) GetRetryPollCount() int {
```
kacf, line 447: You should also add this [here](https://github.com/mendersoftware/mender/blob/4e9236e404f0acc20b71ffe1b82ba556c544ab2b/app/mender.go#L38-L66) (which will probably trigger an avalanche of other places you need to add it!).

kacf (changes requested):

kacf:
> and I do remember about the docs update :)

Hehe, good!

merlin-northern:
@kacf I would really love to argue :) but I am reworking as indicated instead.

merlin-northern:

merlin-northern:
@mender-test-bot start pipeline please sugar

client/client_inventory.go
```diff
@@ -68,8 +87,26 @@ func doSubmitInventory(
...
+			for try := 0; try < tries; try++ {
+				sleepInterval, _ := GetExponentialBackoffTime(try, time.Duration(tries)*time.Minute)
+				log.Infof("Send inventory waiting for re-try: %d/%d in %.0f",
+					try+1, tries, sleepInterval.Seconds())
```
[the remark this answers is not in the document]
merlin-northern: I left it in a similar form.

client/client_inventory.go
merlin-northern: see below, I really wanted to argue a lot, but fixed all (unless I missed something) instead.

client/client_inventory.go
merlin-northern: dropped, and now `GetExponentialBackoffTime` controls all.

app/mender.go
merlin-northern: done-done.

merlin-northern:
believe it or not, but for some mysterious reason, the initial check in the keep alive tests: https://github.com/mendersoftware/integration/pull/1943 which only checks if we have connections to backend started failing. the integration/pull/1943 changes that condition.

merlin-northern:
pipeline: https://gitlab.com/Northern.tech/Mender/mender-qa/-/pipelines/477990734

client/client_inventory.go
```diff
@@ -60,16 +77,43 @@ func doSubmitInventory(
 	api ApiRequester,
 	method, url string,
 	data interface{},
+	retryInterval int,
+	tries int,
```
merlin-northern, line 81: wait I missed the retryInterval -- to use it.

merlin-northern: fixed

merlin-northern:
@kacf some monitoring tests started failing all of a sudden -- I completely do not understand why. I am trying to find out. please review, I will not merge until I have an answer to the tests failing.

client/client_inventory.go
```diff
@@ -60,16 +77,43 @@ func doSubmitInventory(
...
 	req, err := makeInventorySubmitRequest(method, url, data)
 	if err != nil {
 		return nil, errors.Wrapf(err, "failed to prepare inventory submit request")
 	}
 
 	r, err := api.Do(req)
+	outputTries := tries
+	if outputTries == 0 {
+		outputTries = defaultTries //this is the default from the GetExponentialBackoffTime
+	}
```
kacf, line 93: That's not how it works. It tries three times at each interval, including three times at the max interval. If you really want to print this information, make a function next to `GetExponentialBackoffTime`:

```golang
func GetExponentialBackoffCount(maxInterval, maxRetries int) int {
    if maxRetries > 0 {
        return maxRetries
    }
    if maxInterval < 1 * time.Minute {
        maxInterval = 1 * time.Minute
    }
    return ceil(log2(maxInterval / time.Minute) + 1) * 3
}
```

I'm not exactly sure what those math functions are called in Go, but you get my point.

OTOH, you could also just omit this information.

client/client_test.go
```diff
@@ -383,63 +383,75 @@ func TestLoadingTrust(t *testing.T) {
...
+	_, err = GetExponentialBackoffTime(3, 1*time.Second, 0)
 	assert.Error(t, err)
+
+	maxAttempts := 8
+	expectedIntervalMinutes := 2
+	for try := 1; try < maxAttempts; try++ {
+		intvl, err := GetExponentialBackoffTime(try, 1*time.Minute, maxAttempts)
+		assert.NoError(t, err)
+		assert.Equal(t, time.Duration(expectedIntervalMinutes)*time.Minute, intvl)
+		expectedIntervalMinutes *= 2
```
kacf, line 450: This will need to be adjusted as well. I would not repeat the formula here, but rather test exactly the values we expect at various extremes.

kacf (changes requested):

client/client.go
```diff
@@ -651,12 +651,25 @@ var MaxRetriesExceededError = errors.New("Tried maximum amount of times")
...
+	if maxAttempts > 0 {
+		if tried > maxAttempts {
+			return 0, MaxRetriesExceededError
+		}
+
+		for c := 0; c < tried; c++ {
+			interval *= 2
+		}
+		return interval, nil
+	}
```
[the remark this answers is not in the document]
merlin-northern: that was on purpose, thats what I understood from your comment: that we just try given number of times, not honouring the original scenario in case of maxAttempts given. since it is not maxAttempts, because we try 3 times each. ok. I will rework.

client/client_inventory.go
merlin-northern: I will omit then.

client/client_test.go
merlin-northern: ok.

client/client.go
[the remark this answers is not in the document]
kacf: It is using maxAttempts. I don't mean three *additional* times. With my proposed change, it follows maxAttempts exactly, *but* it does it by using the exponential fallback which is increased after each third attempt, like we have always done. It's possible it will never reach maxInterval in that case, but that was possible with your code as well.

merlin-northern:

client/client.go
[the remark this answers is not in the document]
merlin-northern: no problema, I immediately started fixing. I already pushed, tested, and updated the tests. I dropped `maxAttempts <= 0` I think we should return `MaxRetriesExceededError` at the last interval, even if the `maxAttempts` are set. agree?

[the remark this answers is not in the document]
kacf: I think we should keep it because customers have complained about exactly this. That's also why I've pushed for having this change applied to existing retry operations as well. By adding the `RetryPollCount` flag, and allowing the number of retries to be set higher than what the default gives you, we give customers with long blackout windows, but short retry times, a way to avoid failing updates.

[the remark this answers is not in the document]
merlin-northern: ok re-instanted.

kacf (approved):
Thanks for sticking with it!

merlin-northern:
> Thanks for sticking with it!

thanks for not giving up! now the only thing: tests failures. (status: at the moment they are passing locally after fixes but not in the CI)

PR #952, opened 2022-03-07
fix: log the fallback to the artifact_info file at log level Warn

oleorhagen:
I feel like this fallback deserves a proper log mention, as the behaviour is pretty magic otherwise.

### DIFF
--- a/device/device.go
+++ b/device/device.go
@@ -1,4 +1,4 @@
-// Copyright 2021 Northern.tech AS
+// Copyright 2022 Northern.tech AS
 //
 //    Licensed under the Apache License, Version 2.0 (the "License");
 //    you may not use this file except in compliance with the License.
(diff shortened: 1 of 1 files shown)


device/device.go
```diff
@@ -136,7 +136,7 @@ func (d *DeviceManager) GetCurrentArtifactName() (string, error) {
 			log.Errorf("Could not read artifact name from database: %s", err.Error())
 		}
 	}
-	log.Debugf("Returning artifact name from %s file.", d.ArtifactInfoFile)
+	log.Infof("Returning artifact name from %s file.", d.ArtifactInfoFile)
```
lluiscampos, line 139: I would say this deserves a warning, actually.

lluiscampos (approved):
I think we could upgrade it to warning. Approving and letting you take the decision on it.

device/device.go
oleorhagen: Yes, unless it is having issues with the database of course.

oleorhagen:

oleorhagen:

oleorhagen:
@mender-test-bot cherry-pick to: `3.0.x`, `3.1.x`, `3.2.x`

svenschwermer:
I started seeing this warning in logs since upgrading to 3.3.0. Can somebody explain what this is about? What database is this warning talking about?

oleorhagen:
> I started seeing this warning in logs since upgrading to 3.3.0. Can somebody explain what this is about? What database is this warning talking about?

Hi @svenschwermer 

This refers to the fact that in the case the client has not populated it's database yet (happens upon the first update installed), the client falls back to the `artifact_info` file for knowledge of the Artifact it is running (the initial flashed image).

PR #983, opened 2022-04-19
Fix: Don't accumulate zombies when identity/inventory parsing fails.

kacf:
Changelog: Title
Ticket: MEN-5587

### DIFF
--- a/device/identity_data.go
+++ b/device/identity_data.go
@@ -1,4 +1,4 @@
-// Copyright 2021 Northern.tech AS
+// Copyright 2022 Northern.tech AS
 //
 //    Licensed under the Apache License, Version 2.0 (the "License");
 //    you may not use this file except in compliance with the License.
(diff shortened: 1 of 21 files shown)


kacf:

oleorhagen (approved):
Nice!

Have you checked for the same errors in `bootenv` and `inventory_data` also?

kacf:
> Have you checked for the same errors in `bootenv` and `inventory_data` also?

`inventory_data` is already there, but good thinking about `bootenv`, it appears to also be affected. I'll look at it.

device/identity_data.go
```diff
@@ -69,14 +69,16 @@ func (id IdentityDataRunner) Get() (string, error) {
 	}
 
 	p := utils.KeyValParser{}
-	if err := p.Parse(out); err != nil {
-		return "", errors.Wrapf(err, "failed to parse identity data")
-	}
+	parseErr := p.Parse(out)
 
 	if err := cmd.Wait(); err != nil {
 		return "", errors.Wrapf(err, "wait for helper failed")
```
merlin-northern, line 75: super nit: technically here the wait error might be caused by the not-nil `parseErr`, so we could add parseErr.Error() if parseErr not nil to the error message.

merlin-northern (approved):
lgtm. one nit-picky comment.

device/identity_data.go
merlin-northern: btw, not waiting for children is a common mistake, and we should have an explicit test for this scenario, as it can fill up the process table easily -- we have to be sure it never comes back.

kacf: > super nit: technically here the wait error might be caused by the not-nil `parseErr`, so we could add parseErr.Error() if parseErr not nil to the error message.

Yeah, I tried `Wrap` with a nil error, but this panics, so we need extra logic then. But you're right, the error from `parseErr` can be relevant.

> btw, not waiting for children is a common mistake, and we should have an explicit test for this scenario, as it can fill up the process table easily -- we have to be sure it never comes back.

Alright, I'll try to add it.

Originally I thought that the garbage collection would eventually run `Wait` for us, which would mean that any zombie accumulation would be temporary. But it doesn't seem to have a finalizer for this, so this is indeed important.

kacf:
All issues addressed.

oleorhagen:

PR #986, opened 2022-04-21
QA-365: Update the contributor pull request template

oleorhagen:
Changelog: None

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -31,7 +31,7 @@ to stay on top to review and hopefully merge your work.
 If your work is significant, it can make sense to discuss the idea with the
 maintainers and relevant project members upfront. Start a discussion on our [Mender Hub forum](https://hub.mender.io/c/general-discussions).
 
-Using commit signoffs and changelog tags is mandatory for all commits; we also
+Using commit signoffs tags is mandatory for all commits; we also
 encourage that each commit is small and cohesive. See the next sections for details.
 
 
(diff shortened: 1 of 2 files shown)


CONTRIBUTING.md
```diff
@@ -17,7 +17,7 @@ will publish the fix and acknowledge your finding on our site if you so wish.
 
 ## Proposed tasks to get started
 
-There is a `helpwanted` tag on some tasks in the Mender issue tracker
+There is a `helpwanted`, and a `Part` tag on some tasks in the Mender issue tracker
```
kacf, line 20: I don't think we should document the `Part` tag, which is specifically for part timers. This is not the same thing as it may include tasks that are easy, but which require special access, for example. That's why we have two. 🙂

CONTRIBUTING.md
```diff
@@ -185,78 +110,32 @@ This is to ensure that:
...
-```
-
-```
-Add tests for Y.
+<More detailed explanation of the commit>
 
-Changelog: None
+Changelog: <None|Title|Commit|All>
+Ticket: <None|Ticket Nr>
 ```
```
kacf, line 139: Should we top it off by linking to https://www.github.com/mendersoftware/mendertesting/commitlint/grammar.md somewhere in this file? So far it's only in the pull request template.

kacf (changes requested):

CONTRIBUTING.md
oleorhagen: Ohhh... TIL

CONTRIBUTING.md
oleorhagen: Good point!

kacf (approved):

PR #998, opened 2022-05-31
A few changes from playing with rootfs v2.

drewmoseley:

the form below:

```

<OPTIONAL LONGER DESCRIPTION>

```

### Description

Thank you!

### DIFF
--- a/tests/rootfs-image-v2
+++ b/tests/rootfs-image-v2
@@ -29,6 +29,7 @@ else
     passive=$MENDER_ROOTFS_PART_A
     passive_num=$MENDER_ROOTFS_PART_A_NUMBER
 fi
+passive_num_hex=$(printf '%x' $passive_num)
 
 ensure_correct_root_mounted() {
     if [ "$(stat -L -c %02t%02T $active)" != "$(stat -L -c %04D /)" ]; then
@@ -52,6 +53,7 @@ case "$1" in
         ensure_correct_root_mounted
         ${SETENV} -s - <<EOF
 mender_boot_part=$passive_num
+mender_boot_part_hex=$passive_num_hex
 upgrade_available=1
 bootcount=0
 EOF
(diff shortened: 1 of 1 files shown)


drewmoseley:
@kacf @lluiscampos I'm not sure this is the cleanest solution since it may break compatibility with the grub utilities, but there is definitely something here that needs to be addressed.

tests/rootfs-image-v2
```diff
@@ -29,6 +29,7 @@ else
     passive=$MENDER_ROOTFS_PART_A
     passive_num=$MENDER_ROOTFS_PART_A_NUMBER
 fi
+passive_num_hex=echo $((16#$passive_num))
```
kacf, line 32: Isn't this missing `$(...)` around the command?

Regardless, I think it's better to keep this script compatible with `sh`, so that it can be used on minimal setups. I suggest you use this instead:

```
passive_num_hex=$(printf '%x' $passive_num)
```

tests/rootfs-image-v2
```diff
@@ -1,4 +1,4 @@
-#!/bin/sh
+#!/bin/bash
```
kacf, line 1: Following the other comment, this change should be removed.

kacf (changes requested):
> It is unclear if this will break compatibility with grub though.

I think it should be fine. In the [first fix we made for this](https://github.com/mendersoftware/mender/commit/b2507398889da5de4d859a97cf9fd8d0daf1449c), we were running with `=` only, and the tests seemed to be ok with that.

You'll need to fix the commit formatting though. We have a new template now, take a look at the checklist at the top of this pull request.

tests/rootfs-image-v2
drewmoseley: Fixed in v2

drewmoseley: Fixed in v2

tests/rootfs-image-v2
drewmoseley: Fixed in v2

drewmoseley:
v2 submitted with fixes.

kacf (approved):
Thanks, @drewmoseley!

PR #1001, opened 2022-06-01
Bind to 127

merlin-northern:
fix: Both bind to and return over D-Bus 127.0.0.1 

At the moment we bind to localhost and we also return that as a ProxyHost. We decided it to be safer to use `127.0.0.1` in both cases.

Goes with:
* https://github.com/mendersoftware/meta-mender/pull/1704/head

Changelog: Use 127.0.0.1 both as a default bind host and ProxyHost
Ticket: None

### DIFF
--- a/app/proxy/proxy.go
+++ b/app/proxy/proxy.go
@@ -33,7 +33,7 @@ import (
 )
 
 const (
-	ProxyHost        = "localhost"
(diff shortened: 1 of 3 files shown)


merlin-northern:

merlin-northern:
ok, I can see the acceptance tests also need fixing.

PR #1005, opened 2022-06-06
fix: Update module stdout should display in real-time

DomenicP:
### Description

Hi all, I recently upgraded a project from an older version of the Mender client to 3.2.1 (was held back waiting on vendor support for a newer Yocto version for a while). After making the upgrade to the latest Mender client I noticed that our custom update module script stdout was no longer showing up. After some digging I found two bugs in the stdout handling for update modules for which I have proposed fixes. I am not a Go expert by any means (this was my first time doing any sort of significant work with Go) so I am definitely open to feedback if there are better ways to handle these issues.

The first issue was that `ModuleInstaller.callModule()` was always capturing the contents of `stdout` in a buffer regardless of the value of the `capture` parameter. I believe this bug was introduced accidentally in #864. With `stdout` always captured the data was never printed to the console.

After fixing this issue the stdout console output started appearing again but it would not print until the very end of the script. Additionally the output would print with the first line on its own and all of the rest of the content in one long `\n` filled string.

After tracing through I believe this is due to the `cmdLogger` struct embedding `*bytes.Buffer`. Using print statements I confirmed that `cmdLogger.Write()` was never being called when the logger was assigned to `cmd.Stdout`. Instead, it appears that `bytes.Buffer.Write()` was being used instead since data still ended up in the buffer to be flushed at the end. By changing `cmdLogger` to use composition with a `buf *bytes.Buffer` field instead of embedding I verified that `cmdLogger.Write()` was now being called and that the stdout data would be printed to the console in real-time and on the correct lines as expected.

I put together a minimal example in a Gist for the behavior that I observed for the second issue: https://gist.github.com/DomenicP/10b024397c46a03854276e93edcbcf0e. Again, since I'm not that familiar with Go I'd love to hear if someone more familiar with the language has more insight here or ideas on if there's a better way to approach this. Thanks!


system/system.go
```diff
@@ -131,34 +131,34 @@ func Command(name string, arg ...string) *Cmd {
 type cmdLogger struct {
 	commandName string
 	stream      string
-	*bytes.Buffer
+	buf         *bytes.Buffer
```
kacf, line 134: For any struct which is passed to some function by interface value, it is always the "outermost" struct (the one which embeds others) which receives the call. This is the same as one would typically see in traditional OOP languages like C++, Java and Python, as long as it's called through an interface.

So I don't understand how this can help. To me it looks like this gives exactly the same behavior as before this change. Are you saying that `cmdLogger.Write()` was not called prior to this change, but with this change, it is?

kacf (commented):
Thanks. The first commit looks good. The second I'm a bit puzzled by.

Perhaps @oleorhagen would like to take a look at this as well.

oleorhagen (approved):
Makes sense. Good detective work @DomenicP - thanks :smiley_cat:

kacf (approved):

system/system.go
DomenicP: That makes a lot more sense, thanks for looking into it. I knew that `cmdLogger.Write()` wasn't being called but I wasn't sure exactly what was going on behind the scenes other than the data was still being written to the buffer through some other method.

kacf:
Only unrelated errors. Merging!

DomenicP:
@mender-test-bot cherry-pick to: `3.2.x`

DomenicP:
(I'm not sure if the test bot will respond to my comments since I'm an external contributor, could we cherry pick this fix back to `3.2.x` as well? From what I checked in the code history it looks like this fix only applies to `3.2.x` and newer, `3.1.x` is still using the old logging code from before #864)

DomenicP:
Is `3.3.x` available on the Yocto Dunfell branch of `meta-mender`? Or should it at least be compatible for me to backport if it's not available? Because I am currently vendor bound to Dunfell and when I looked previously `3.2.1` was the latest client version available for that Yocto version.

oleorhagen:
> Is `3.3.x` available on the Yocto Dunfell branch of `meta-mender`? Or should it at least be compatible for me to backport if it's not available by default? Our SBC board vendor currently only supports up to Yocto Dunfell and when I looked previously `3.2.1` was the latest client version available for that Yocto version.

@DomenicP we are currently in the process of releasing `3.3.0`, so in a few days you should hopefully have it available :)

PR #1014, opened 2022-07-06
Fix race condition in TestControlMapPauseState

alfrunes:


app/state_test.go
```diff
@@ -5511,7 +5511,7 @@ func TestControlMapPauseState(t *testing.T) {
 	}
 	c := &stateTestController{
 		controlMap:      pool,
-		updatePollIntvl: 30,
+		updatePollIntvl: 30 * time.Second,
```
kacf, line 5514: Lol, that's one impatient test!

app/state_test.go
```diff
@@ -1148,7 +1148,7 @@ type stateTransitionsWithUpdateModulesTestCase struct {
 
 var stateTransitionsWithUpdateModulesTestCases []stateTransitionsWithUpdateModulesTestCase = []stateTransitionsWithUpdateModulesTestCase{
 
-	stateTransitionsWithUpdateModulesTestCase{
+	{
```
kacf, line 1151: Yes!

But just FYI, IIRC I originally coded it like this, but had to add the explicit types due to old Golang versions in Yocto. That was a while back though, if dunfell is ok with this, then we are good to go.

app/state.go
```diff
@@ -34,6 +35,49 @@ const (
 		"by currently installed artifact (%v != %v)."
 )
 
+type timerPool sync.Pool
```
kacf, line 38: Nothing against this change, but the performance impact was probably more important before the nanosecond fix! 😀

kacf (approved):

app/state.go
alfrunes: Yeah, it was just a random thing I came across as I traced the code for the unit test. I was contemplating leaving it out of the PR, but decided to leave it in anyway.

app/state_test.go
alfrunes: I usually run `go fmt` with the `-s` option. The only time you have to use explicit struct declarations are for slices of interfaces as far as I know.

alfrunes:
@kacf The client acceptance tests passed and the integration test failures are known spurious and not related specifically to the client alone. Do you think we can merge this one?

kacf:
> @kacf The client acceptance tests passed and the integration test failures are known spurious and not related specifically to the client alone. Do you think we can merge this one?

Yes! 🙂

PR #1015, opened 2022-07-19
fix: Resume update after manually rebooting during pre-reboot pause.

kacf:
In order for the spontaneous reboot recovery code to be able to
resume, we have to have recorded that we were in the
`MenderStateReboot` state when the device rebooted. This is tricky for
several reasons:

* The states that deal with Update Control are shared, so they occur
  at multiple points throughout the flow, and only the ArtifactReboot
  case should have this behavior.

* We have to check that the Update Control Maps do not resolve to a
  "fail" before we store `MenderStateReboot`. Otherwise it could be
  that the a local process requested failure and rollback, but Mender
  continues anyway.

Fortunately, the Update Control states (like `controlMapState` &
friends) do not implement the `UpdateState` interface, and hence no
database state is recorded for them. We take advantage of this by
creating one new state which does implement `UpdateState`, and which
records the `MenderStateReboot` state ID. We arrange for this state to
be invoked when there is a pause before a reboot. Transitioning to
this state then records the proper ID in the database, and we loop
once more through the `controlMapState` to do the real pausing.

For the non-reboot states we pass nil when we enter the Update Control
states, and hence they use the old behavior of pausing without
entering any explicit state.

Changelog: If paused before ArtifactReboot, and then manually
rebooting the device outside of Mender, the client will properly
resume the update now, instead of failing and rolling back.

Ticket: MEN-5709

### DIFF
--- a/app/daemon_test.go
+++ b/app/daemon_test.go
@@ -216,16 +216,14 @@ func (d *daemonTestController) RefreshServerUpdateControlMap(deploymentID string
 
 func TestDaemonRun(t *testing.T) {
 	t.Run("Testrun daemon", func(t *testing.T) {
-		if testing.Short() {
-			t.Skip("skipping periodic update check in short tests")
+		t.Parallel()
 
-		}
-
-		pollInterval := time.Duration(30) * time.Second
+		pollInterval := time.Duration(1) * time.Second
 
 		dtc := &daemonTestController{
 			stateTestController{
 				updatePollIntvl: pollInterval,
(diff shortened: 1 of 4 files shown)


kacf:
This turned out to be tricky to test in unit tests due to needing both a real update, Update Control, and a spontaneous restart all at the same time. So I made [an acceptance test](https://github.com/mendersoftware/meta-mender/pull/1753) instead.

tranchitella (approved):

PR #1024, opened 2022-08-17
MEN-2583: bootstrap Artifact

lluiscampos:
```
* chore: Upgrade mender-artifact library to latest

And update client code accordingly.


* feat(bootstrap-artifact): Install bootstrap Artifact on first start-up.

On start-up, Mender checks for the existence of an special bootstrap
Artifact in path `/var/lib/mender/bootstrap.mender` and installs it in
order to initialize the device database.

This applies both for `daemon` start and cli commands `bootstrap` and
`install`.

The Artifact is not installed if the device already has a database. When
the Artifact is not found (and the database is empty) the database is
initialized with `artifact-name=unknown`.

In addition, Mender can also understand other kinds of "empty" Artifacts
and install them either in managed or standalone modes.

Changelog: Commit
Ticket: MEN-2583


* feat(bootstrap-artifact): Remove support for artifact_info file

Changelog: Remove support for `artifact_info` file. The initial artifact
name will be populated in the database using a bootstrap Artifact.

Ticket: MEN-2583
```


lluiscampos:
Build @ [Pipeline 615738811](https://gitlab.com/Northern.tech/Mender/mender-qa/-/pipelines/615738811)

installer/modules_test.go
```diff
@@ -512,15 +516,17 @@ func TestStreamsTreeClearsProvidesAttribute(t *testing.T) {
 		CompatibleDevices: []string{"test-device"},
 	}
 
+	updateType := "test-type"
+
 	headerInfo := artifact.NewHeaderInfoV3([]artifact.UpdateType{
 		artifact.UpdateType{
-			Type: "test-type",
```
oleorhagen, line 517: Wow, that was annoying! And I guess if this was a proper lib version, it'd actually be breaking

app/mender.go
```diff
@@ -219,6 +223,18 @@ func (m *Mender) FetchUpdate(url string) (io.ReadCloser, int64, error) {
 	return m.updater.FetchUpdate(m.download, url, m.GetRetryPollInterval())
 }
 
+func filterArtifactDepends(depends map[string]interface{}) map[string]interface{} {
```
oleorhagen, line 226: Filtered for what?

app/mender.go
```diff
@@ -670,3 +686,195 @@ func verifyAndSetArtifactNameInProvides(
...
+func (m *Mender) HandleBootstrapArtifact(s store.Store) error {
+
+	var err error
+	updatePath := m.BootstrapArtifactFile
+
+	var databaseInitialized bool = false
+	var bootstrapArtifactFileFound bool = false
+
+	defer func() {
+		// Remove file even on faliures. Otherwise we will try to install it on every boot
```
oleorhagen, line 702: Nit: typo

app/mender.go
```diff
@@ -670,3 +686,195 @@ func verifyAndSetArtifactNameInProvides(
...
+	ar := areader.NewReader(f)
+
+	ar.CompatibleDevicesCallback = func(devices []string) error {
+		for _, dev := range devices {
+			if dev == dt {
+				return nil
+			}
+		}
+		return fmt.Errorf(
+			"image (device types %v) not compatible with device %v",
```
oleorhagen, line 809: Q: Aren't they both strings?

app/mender.go
```diff
@@ -670,3 +686,195 @@ func verifyAndSetArtifactNameInProvides(
...
+		signed = true
+		return nil
+	}
+	err = ar.ReadArtifact()
+	if err != nil {
+		return "", "", nil, err
+	}
+
+	if ar.GetArtifactName() == "" {
+		return "", "", nil, errors.New("artifact_name cannot be emtpy")
```
oleorhagen, line 832: Nit: typo

app/mender.go
```diff
@@ -670,3 +686,195 @@ func verifyAndSetArtifactNameInProvides(
...
+
+	depends, err := update.GetUpdateDepends()
+	if err != nil {
+		return "", "", nil, errors.Wrap(err, "error reading depends")
+	}
+	if depends != nil {
+		return "", "", nil, errors.New("update artifact_depends must be empty")
+	}
+
+	if len(update.GetUpdateAllFiles()) != 0 {
```
oleorhagen, line 878: Q: Not against being careful, but isn't this something the library should already guard against?

app/state.go
```diff
@@ -589,12 +589,15 @@ func (uc *updateAfterFirstCommitState) Handle(ctx *StateContext, c Controller) (
...
 	var firstErr error
 
-	for _, i := range c.GetInstallers()[1:] {
-		err := i.CommitUpdate()
-		if err != nil {
-			log.Errorf("Error committing %s payload: %s", i.GetType(), err.Error())
-			if firstErr == nil {
-				firstErr = err
+	installers := c.GetInstallers()
+	if len(installers) > 1 {
```
oleorhagen, line 593: Q: More of a fix, than new functionality, am I right?

oleorhagen (approved):
Only superficial stuff from me, I'll let Kristian do the digging

oleorhagen:
Oh, also typo in the Changelog, last commit

Makefile
```diff
@@ -99,7 +99,6 @@ install-bin: mender
```
kacf, line 99: Typo in commit message: `s/de populated/be populated/`

Makefile
```diff
@@ -139,7 +138,6 @@ install-examples:
 	install -m 0644 support/demo.crt $(prefix)$(docexamplesdir)/
 
 uninstall: uninstall-bin \
-	uninstall-conf \
```
kacf, line 142: This should still be here, no? The target is still intact.

kacf (changes requested):

app/mender.go
lluiscampos: You are right, I'll modify. I took it from [the installer code](https://github.com/mendersoftware/mender/blob/1fd97067f6a50e7dc6446c63cc540b210a7ad3a2/installer/installer.go#L132), though.

Makefile
lluiscampos: Yes, sorry. An leak from the commit amending I did.

app/mender.go
```diff
@@ -670,3 +686,195 @@ func verifyAndSetArtifactNameInProvides(
...
+	updatePath := m.BootstrapArtifactFile
+
+	var databaseInitialized bool = false
+	var bootstrapArtifactFileFound bool = false
+
+	defer func() {
+		// Remove file even on faliures. Otherwise we will try to install it on every boot
+		if bootstrapArtifactFileFound {
+			log.Debugf("Removing bootstrap Artifact %s", updatePath)
+			_ = os.Remove(updatePath)
```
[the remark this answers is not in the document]
lluiscampos: Sure, that is a good idea.

app/mender.go
```diff
@@ -670,3 +686,195 @@ func verifyAndSetArtifactNameInProvides(
...
+	if sdErr == nil {
+		databaseInitialized = true
+	} else {
+		log.Debug("No state data stored")
+		p, pErr := datastore.LoadProvides(s)
+		if pErr == nil && len(p) > 0 {
+			databaseInitialized = true
+		} else {
+			log.Debug("No provides stored")
+			log.Debug("Database not initialized")
```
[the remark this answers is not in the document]
lluiscampos: My idea was than "No provides stored" goes with "No state data stored", and only if both are printed you are concluding that database is not initialized. I'll rework the messages to have only two. (One might go way anyway in relation to the open discussion above).

app/mender.go
```diff
@@ -670,3 +686,195 @@ func verifyAndSetArtifactNameInProvides(
...
+	if err != nil {
+		return "", "", nil, err
+	}
+
+	if ar.GetArtifactName() == "" {
+		return "", "", nil, errors.New("artifact_name cannot be emtpy")
+	}
+
+	if signed {
+		return "", "", nil, errors.New("Artifact cannot be signed")
```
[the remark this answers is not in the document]
lluiscampos: Good reasoning. I think you are right - I'll remove this check (and adapt unit tests).

app/mender.go
```diff
@@ -670,3 +686,195 @@ func verifyAndSetArtifactNameInProvides(
...
+	}
+	if len(scripts) > 0 {
+		return "", "", nil, errors.New("Artifact cannot contain state scripts")
```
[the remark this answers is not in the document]
lluiscampos: Okey, I am amending it. That is how I originally had them but later I thought it was redundant.

app/mender.go
```diff
@@ -670,3 +686,195 @@ func verifyAndSetArtifactNameInProvides(
...
+	}
+
+	info := ar.GetInfo()
+	if info.Format != "mender" {
+		return "", "", nil, errors.New("wrong Artifact format")
+	}
+	if info.Version != 3 {
```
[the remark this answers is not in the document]
lluiscampos: `< 3`, but yes.

app/mender.go
lluiscampos: How exactly? If you create a bootstrap Artifact yes, the library guard against it, but for any other Artifact it wont. I could craft an Artifact with type `nil` but still have actual payload (the current code will anyway ignore the payload, but I think is good to be proactive here).

app/standalone.go
```diff
@@ -142,20 +142,23 @@ func doStandaloneInstallStatesDownload(art io.ReadCloser, key []byte,
...
-			device.GetCurrentArtifactName,
-		); err != nil {
-			log.Error(err.Error())
-			return nil, err
-		}
-		if err = verifyArtifactDependencies(depends, currentProvides); err != nil {
-			log.Error(err.Error())
-			return nil, err
+		actualDepends := filterArtifactDepends(depends)
+		if len(actualDepends) > 0 {
```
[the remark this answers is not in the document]
lluiscampos: Thank you for noticing this!

Long story short: this is not needed and I am reverting it (also removing the `filterArtifactDepends` method).

Long story long: at the early days of my prototype code I wrote this to be able to install a bootstrap Artifact (or any artifact, really) that only depended on `device_type` (which is not really a dependency checked in `verifyArtifactDependencies`, it is checked somewhere else). The problem raised later when calling `GetCurrentArtifactName` before we initialized the database (as we removed the fallback to file). However now we have the "hook" in install and bootstrap commands that will either install the bootstrap Artifact or initialize the database with `unknown` before this code is going to run, so I can clean it up.

Sorry for not caching this up before final submit.

app/mender.go
lluiscampos: Filter `device_type`. This was a poor name, thanks for noticing.

In any case I am [removing the method](https://github.com/mendersoftware/mender/pull/1024/files#r948825773).

app/state.go
```diff
@@ -491,18 +491,18 @@ func (uc *updateCommitState) Handle(ctx *StateContext, c Controller) (State, boo
...
-	err = installers[0].CommitUpdate()
-	if err != nil {
-		// we need to perform roll-back here; one scenario is when
-		// u-boot fw utils won't work after update; at this point
-		// without rolling-back it won't be possible to perform new
-		// update
-		merr := NewTransientError(errors.Errorf("update commit failed: %s", err.Error()))
-		return uc.HandleError(ctx, c, merr)
+	if len(installers) == 0 {
+		log.Warn("GetInstallers() returned empty list: Artifact with no payload")
```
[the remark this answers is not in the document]
lluiscampos: This was a key discovery: a bootstrap Artifact has 1 _installer_ but 0 _updateStorers_. From `mender` client point of view, we name variables as installers when we really mean storers :man_shrugging:.

So yes, this check is to detect a bootstrap Artifact and do not commit it (it has no _updateStorer_ that knows how to commit it).

app/state.go
lluiscampos: Kind of.

Before `GetInstallers()` it was guaranteed to have at least one element, so `GetInstallers()[1:]` would just be an empty slice. However now it can have zero elements, so `[1:]` just crashes badly.

app/mender.go
```diff
@@ -670,3 +686,195 @@ func verifyAndSetArtifactNameInProvides(
...
+					err = errors.Wrapf(err, dbErr.Error())
+				} else {
+					err = dbErr
+				}
+			}
+		}
+	}()
+
+	// Check if database already exists
+	_, sdErr := datastore.LoadStateData(s)
```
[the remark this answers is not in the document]
lluiscampos: Discussed on a call,

The code is "correct", and it needs to check for in-progress updates both upgrading from 1.x and any other client before the bootstrap Artifact support. I'll create a better comment here to make sure we understand it in the future.

app/mender.go
oleorhagen: I understand :)

lluiscampos:
@kacf I've seen now few times the following test failing:
```
=== CONT  TestDaemonRun/Testrun_daemon
    daemon_test.go:246: poke count: 6
    daemon_test.go:248: 
        	Error Trace:	/builds/Northern.tech/Mender/mender/app/daemon_test.go:248
        	Error:      	"6" is not less than "6"
        	Test:       	TestDaemonRun/Testrun_daemon
```

If you retry the CI job, passes. Locally also fails some times.

I don't really understand the relation with my work. If any, my code should have made the daemon start-up _slower_ and then fail in the previous (`GreaterOrEqual` the minimum poll times).

Any idea? Any advise on how can I make this test more robust?

lluiscampos:
@mender-test-bot start pipeline --pr integration/2123 :raised_hands:

lluiscampos:
Build with https://github.com/mendersoftware/integration/pull/2123 @ [Pipeline 618125353](https://gitlab.com/Northern.tech/Mender/mender-qa/-/pipelines/618125353)

oleorhagen:
> @kacf I've seen now few times the following test failing:
> 
> ```
> === CONT  TestDaemonRun/Testrun_daemon
>     daemon_test.go:246: poke count: 6
>     daemon_test.go:248: 
>         	Error Trace:	/builds/Northern.tech/Mender/mender/app/daemon_test.go:248
>         	Error:      	"6" is not less than "6"
>         	Test:       	TestDaemonRun/Testrun_daemon
> ```
> 
> If you retry the CI job, passes. Locally also fails some times.
> 
> I don't really understand the relation with my work. If any, my code should have made the daemon start-up _slower_ and then fail in the previous line (`GreaterOrEqual` the minimum poll times).
> 
> Any idea? Any advise on how can I make this test more robust?

Hmm, sounds suspicious, I changed parts of the state logic in https://github.com/mendersoftware/mender/pull/1025

I can take a closer look at this now actually. I bet this is on me

app/mender.go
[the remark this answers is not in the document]
kacf: 🤦, right

app/state.go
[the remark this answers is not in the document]
kacf: Ok, another example of the excellent (not!) API we have in mender-artifact. 🙂

Still, I am a little bit confused. You say "do not commit it", but yet the code continues after the log statement. And don't we need to commit, even for a bootstrap artifact? Maybe this is better discussed on a call, because I'm probably missing something here...

app/mender.go
```diff
@@ -670,3 +670,200 @@ func verifyAndSetArtifactNameInProvides(
...
+			}
+		}
+	}()
+
+	// Check if the database is already initialized. Or, in other words, if the database has
+	// either state data (for an in-progress update) or provides data. Only when none of these
+	// is true we consider the database not initialized and proceed with either installing the
+	// bootstrap Artifact or falling back to initializing artifact-name to "unknown".
+	// Any earlier version of the client not supporting bootstrap Artifact will have empty
+	// provides until the first update is _committed_, so we must check state data in adition.
```
kacf, line 725: Nitpick: "adition"

app/mender.go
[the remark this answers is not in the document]
kacf: Great comment!

app/state.go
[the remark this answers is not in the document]
lluiscampos: Discussed on a call, the "commit" of the meta-data is done later on calling `CommitArtifactData`. Here in this `if` condition what we avoid is to "commit" the payload.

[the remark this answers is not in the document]
kacf: Discussed f2f, will change log level, but otherwise this is good.

kacf (approved):

lluiscampos:
The [last force-push](https://github.com/mendersoftware/mender/compare/bebff6eb2487bd219dab8bfdc498c3b97a75fb18..4358157803e365b3e4b3d2571eeba2b94e930fba) had only cosmetic changes. Merging.

PR #1025, opened 2022-08-18
fix: the periodic update check bug in checkWaitState

oleorhagen:
Changelog: The client update and inventory checks are now unaffected by the use
of the `check-update` and `send-inventory` commands. While previously, this
would both move the intervals at which checks we're done, and also extend them
beyond the expected polling intervals configured.

Ticket: MEN-5547

In short:

Simplified the `checkWait` logic a lot by moving the last attempt out of the `state.go` file, and into `daemon.go`.

While maybe not completely kosher on the surface, but if you look through the code, you will realize that this does indeed simplify our somewhat fragile logic here a lot, and should now be fairly fool-proof, as opposed to previously, where it has broken for us multiple times.

I also, added an exception for the manual forced update and inventory checks from the cli, so that they do no longer interfere with the regularity of the intervals at which the update checks are done by the state machine. I'm sure we could do something clever here, but I figured a simple bool if, INF-1000 style would suffice here.

Have a lookover, and see what you think :)

### DIFF
--- a/app/daemon.go
+++ b/app/daemon.go
@@ -14,6 +14,8 @@
 package app
 
 import (
+	"time"
+
 	"github.com/pkg/errors"
 	log "github.com/sirupsen/logrus"
 
@@ -105,6 +107,7 @@ func (d *MenderDaemon) Run() error {
 	cancelled := false
 	for {
 		// If signal SIGUSR1 or SIGUSR2 is received, force the state-machine to the correct state.
+		updateLastCheckAttempt := true
 		select {
 		case nState := <-d.ForceToState:
 			switch toState.(type) {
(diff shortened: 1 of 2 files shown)


app/daemon.go
```diff
@@ -114,13 +117,25 @@ func (d *MenderDaemon) Run() error {
...
 			}
 
 		default:
 			// Identity op - do nothing.
 		}
+		// Set the time for the last attempts
+		switch toState.(type) {
+		case *updateCheckState:
+			if updateLastCheckAttempt {
+				d.Sctx.lastUpdateCheckAttempt = time.Now()
```
kacf, line 132: I'm not really arguing that it's better, but rather a suggestion: Could the forced state, and the timer-expired state, be two different states? One would update the timestamp, and one would not. It's not radically different from what you are already doing, but perhaps more state-idiomatic. I dunno..

kacf (approved):
👍

app/daemon.go
oleorhagen: Hmm, I kinda like the idea, but is it okay to skip it for this simple if?

oleorhagen:

app/daemon.go
kacf: Yes, absolutely, it was just an opportunistic suggestion 🙂

oleorhagen: :smiley_cat:

oleorhagen:
@mender-test-bot cherry-pick to: `3.0.x`

oleorhagen:
@mender-test-bot cherry-pick to `3.3.x`

oleorhagen:
@mender-test-bot cherry-pick to: `3.3.x`

PR #1037, opened 2022-09-10
fix: Resolve symlinks for /dev/disk/by-partlabel

Michaelhobo:
We started labelling our partitions to solve disk naming issues when swapping disks between disk slots. However, when we pass /dev/disk/by-partlabel/root2 and /dev/disk/by-partlabel/root3 to Mender, we can't update because Mender throws an error: Active root partition matches neither RootfsPartA nor RootfsPartB.

Resolve by-partlabel as well to solve this problem.

Changelog: Resolve symlinks for /dev/disk/by-partlabel
Ticket: None

### DIFF
--- a/installer/partitions.go
+++ b/installer/partitions.go
@@ -1,4 +1,4 @@
-// Copyright 2021 Northern.tech AS
+// Copyright 2022 Northern.tech AS
 //
 //    Licensed under the Apache License, Version 2.0 (the "License");
 //    you may not use this file except in compliance with the License.
@@ -40,6 +40,11 @@ var (
 	ErrorPartitionNoMatchActive = errors.New(
 		"Active root partition matches neither RootfsPartA nor RootfsPartB.",
 	)
+	// We resolve paths for symlinks directly in these directories.
+	resolvePathsInDirs = map[string]bool{
+		"/dev/disk/by-partuuid":  true,
+		"/dev/disk/by-partlabel": true,
+	}
 )
 
 type partitions struct {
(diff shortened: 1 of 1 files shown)


kacf (approved):
Thanks, looks good to me!

kacf (changes requested):
Actually the CI is complaining about a couple of things: Can you make sure that:

* The copyright year at the top is updated.
* The code is formatted using `go fmt installer/partitions.go`.

installer/partitions.go
```diff
@@ -263,7 +268,7 @@ func maybeResolveLink(unresolvedPath string) string {
 	}
 	// MEN-2302
 	// Only resolve /dev/disk/by-partuuid/ and /dev/root
```
lluiscampos, line 270: Would be nice to update this comment too.

Michaelhobo:
> Thanks, looks good to me!

Michaelhobo:
> Actually the CI is complaining about a couple of things: Can you make sure that:
> 
> * The copyright year at the top is updated.
> * The code is formatted using `go fmt installer/partitions.go`.

Done

kacf (approved):
Thanks!

Michaelhobo:
It seems like something is still failing, but I don't have access to see what it is?

kacf:
Yes, it's part of our internal CI. I'm pretty sure it's unrelated to your change, but I'll make a final check tomorrow.

kacf:
Yep, unrelated. Our internal CI is helpful but not perfectly stable.

lluiscampos:
* 3.3.x

PR #1038, opened 2022-09-14
chore: Upgrade mender-artifact library to latest

lluiscampos:
And update client code accordingly.

Ported from commit 0d6035 but vendoring today's latest.


kacf (approved):
Nitpick: Commit message is slightly wrong, you're actually updating several things.

lluiscampos:
> Nitpick: Commit message is slightly wrong, you're actually updating several things.

They are indirect dependencies, so I don't think it is worth listing them. Think about it this way: if this PR would have been made by dependabot, we will only be adding a changelog related to mender-artifact, not the indirect dependencies.

lluiscampos:

PR #1045, opened 2022-09-25
feat: Support multiple verification keys

Michaelhobo:
https://hub.mender.io/t/multiple-artifactverifykeys/4309/6

Creates a new client config parameter called ArtifactVerifyKeys that is a list of paths to keys. Any matching key can be used to verify an artifact - e.g. if 5 verification keys are provided, only 1 needs to match to verify the artifact.

This also adds the ability to filter keys by update type - so some keys can be specified for use only for certain update types.

This enables a few key features:
- key rotation with a longer grace period by specifying multiple keys while rotating keys.
- separate verification keys for different update types.
- verifying artifacts signed for multiple environments/devices.

Changelog: Commit
Ticket: None

Signed-off-by: Michael Ho <callmemikeh@gmail.com>

### DIFF
--- a/app/standalone.go
+++ b/app/standalone.go
@@ -50,7 +50,7 @@ type standaloneData struct {
 
 // This will be run manually from command line ONLY
 func DoStandaloneInstall(device *dev.DeviceManager, updateURI string,
-	clientConfig client.Config, vKey []byte,
+	clientConfig client.Config,
 	stateExec statescript.Executor, rebootExitCode bool) error {
 
 	var image io.ReadCloser
@@ -93,10 +93,10 @@ func DoStandaloneInstall(device *dev.DeviceManager, updateURI string,
 	p := utils.NewProgressWriter(imageSize)
 	tr := io.TeeReader(image, p)
 
-	return doStandaloneInstallStates(ioutil.NopCloser(tr), vKey, device, stateExec, rebootExitCode)
+	return doStandaloneInstallStates(ioutil.NopCloser(tr), device, stateExec, rebootExitCode)
 }
 
-func doStandaloneInstallStatesDownload(art io.ReadCloser, key []byte,
+func doStandaloneInstallStatesDownload(art io.ReadCloser,
 	device *dev.DeviceManager, stateExec statescript.Executor) (*standaloneData, error) {
 
 	dt, err := device.GetDeviceType()
@@ -113,7 +113,7 @@ func doStandaloneInstallStatesDownload(art io.ReadCloser, key []byte,
 		// No doStandaloneFailureStates here, since we have not done anything yet.
 		return nil, err
 	}
-	installer, installers, err := installer.ReadHeaders(art, dt, key,
+	installer, installers, err := installer.ReadHeaders(art, dt, &device.Config,
 		device.StateScriptPath, &device.InstallerFactories)
 	standaloneData := &standaloneData{
 		installers: installers,
@@ -179,11 +179,11 @@ func doStandaloneInstallStatesDownload(art io.ReadCloser, key []byte,
 	return standaloneData, nil
 }
 
-func doStandaloneInstallStates(art io.ReadCloser, key []byte,
+func doStandaloneInstallStates(art io.ReadCloser,
 	device *dev.DeviceManager, stateExec statescript.Executor,
 	rebootExitCode bool) error {
 
-	standaloneData, err := doStandaloneInstallStatesDownload(art, key, device, stateExec)
+	standaloneData, err := doStandaloneInstallStatesDownload(art, device, stateExec)
 	if err != nil {
 		return err
 	}
--- a/app/standalone_test.go
+++ b/app/standalone_test.go
@@ -98,9 +98,9 @@ func Test_doManualUpdate_noParams_fail(t *testing.T) {
 	deviceType := zeroLengthDeviceTypeFile(t)
 	defer os.Remove(deviceType)
 
-	dualRootfsDevice := installer.NewDualRootfsDevice(nil, nil, installer.DualRootfsDeviceConfig{})
+	dualRootfsDevice := installer.NewDualRootfsDevice(nil, nil, conf.DualRootfsDeviceConfig{})
 	if err := DoStandaloneInstall(getTestDeviceManager(dualRootfsDevice, &config, deviceType, dbdir),
-		"", client.Config{}, nil, dev.NewStateScriptExecutor(&config), false); err == nil {
+		"", client.Config{}, dev.NewStateScriptExecutor(&config), false); err == nil {
 
 		t.FailNow()
 	}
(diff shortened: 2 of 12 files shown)


Michaelhobo:
Hi, I'm still fixing bugs and writing tests, but I wanted to get initial feedback on this feature, especially the changes to the Mender client config file.

estenberg:
@Michaelhobo had a call about the use cases. Here's the summary (please correct if I misstating anything):

* Use case 1: Key rotation of verification keys. Although it's possible to rotate keys today, it may be a bit risky without support for multiple verification keys.
* Use case 2: Ensure attacker cannot install an Artifact on an *incompatible* hardware, by ensuring two different hardware types use different signing keys. This was some of the motivation for splitting signing key by update type. However, during the discussion, we found that this is already secure because device_types_compatible meta-data in the Artifact *is signed as well*. So an attacker cannot change this and install it on a different hardware anyway today.

The conclusion:
* Multiple verification keys is desirable.
* Would likely need to introduce a "key ID", so that the artifact meta-data not only contains the signature, but which key it was signed with (so the client can first look up the key and use it to verify the signature); simply trying all keys it has is llikely not a good idea; may have low performance, and ambiguity and maybe leads to some risk/vulnerability.
* Afraid that Artifact format need to be bumped with the above.
* This PR will be dropped, make use with one key for now.

kacf:
Thanks for the clarifications.

> * Would likely need to introduce a "key ID", so that the artifact meta-data not only contains the signature, but which key it was signed with (so the client can first look up the key and use it to verify the signature); simply trying all keys it has is llikely not a good idea; may have low performance, and ambiguity and maybe leads to some risk/vulnerability.
> * Afraid that Artifact format need to be bumped with the above.

Yes, for the key ID it would be needed. We could introduce it as an optional thing, but we cannot mandate the key ID without bumping the format version. Assuming we don't want to do that, we would have to support both: Looking up directly when possible, and trying iteratively if not.

Is it really that expensive though? Two TLS connections, and you already have the same cost. Also, I suppose that multiple keys is a feature often used only in a transitional period. The way I see it, it's not worth adding the key ID, and even if it is, it can be treated as a separate optimization.

estenberg:
Makes sense, performance is probably not the best argument (even though you might have one or more "wasted" attempts to verify the signature every time you do an update, this may not have a huge total compute impact).

However, I think security is a good argument for making this explicit. I think "randomly" trying several keys for signature verification until at least one passes can lead to attack vectors. Although I don't have a very good example now, it is generally a good idea to be very explicit / narrow when it comes to security...

kacf:
I agree it's good to have as narrow a scope as possible when it comes to security, but I think this is pretty well defined and understood. I think I am on fairly solid ground if I state that trying different keys is solid from a security perspective. After all, protecting against random keys is exactly what encryption is supposed to do. Of course it does increase the code size, where there could be mistakes, but this is quite trivial to implement.

I'm not against storing the key id, but given that we have

* Possibility of multiple keys in configuration - Small and easy to implement
* Key id stored along with signature - Either format extension with fallback code (which gains none of the mentioned security), or full format bump with two incompatible versions. Big job with a long tail of compatibility issues.

Because of this difference in scope, I'm quite reluctant to hold back this PR, since I see no fundamental flaw with having multiple keys, even without key ids.

Michaelhobo:
> I agree it's good to have as narrow a scope as possible when it comes to security, but I think this is pretty well defined and understood. I think I am on fairly solid ground if I state that trying different keys is solid from a security perspective. After all, protecting against random keys is exactly what encryption is supposed to do. Of course it does increase the code size, where there could be mistakes, but this is quite trivial to implement.

We do increase the number of keys that can be compromised, though - so an attacker can now compromise ANY key, instead of just the one key. This is mitigated by having verification keys scoped to specific types.

> 
> I'm not against storing the key id, but given that we have
> 
> * Possibility of multiple keys in configuration - Small and easy to implement
> * Key id stored along with signature - Either format extension with fallback code (which gains none of the mentioned security), or full format bump with two incompatible versions. Big job with a long tail of compatibility issues.
> 
> Because of this difference in scope, I'm quite reluctant to hold back this PR, since I see no fundamental flaw with having multiple keys, even without key ids.

Having an interim solution that doesn't require bumping the artifact version IS a pretty big advantage of this approach.

Michaelhobo:
> We do increase the number of keys that can be compromised, though - so an attacker can now compromise ANY key, instead of just the one key. This is mitigated by having verification keys scoped to specific types.

Now that I think about this again, this is probably even better than the current security posture, since a scoped key decreases the blast radius of having a compromised signing key.

kacf:
> Now that I think about this again, this is probably even better than the current security posture, since a scoped key decreases the blast radius of having a compromised signing key.

The current security model is not prepared for this kind of scoping though. It's trivial to break out of the containment by using a state script, which executes with root privileges, and which will work with any key.

I think that having multiple keys is well defined and understood, and we could accept that into Mender. Yes, you can potentially be exposed to compromises of any of the keys you specify, but opening up for this is voluntary; the default, and the recommendation, is still to use one key, and the security of that does not change.

OTOH scoped keys are unlikely to provide extra security, so I think this should be dropped from the pull request.

Michaelhobo:
@estenberg @kacf is the consensus to have a list of ArtifactVerifyKeys, that's basically a list of the existing ArtifactVerifyKey config parameter? I can make this refactor, but I've been short on time recently, so I want to make sure the next one is the right approach.

```json
{
	ArtifactVerifyKey: "/path/to/key0",
	ArtifactVerifyKeys: ["/path/to/key1", "/path/to/key2"],
}
```

A config like this would result in 3 keys being allowed: key0, key1, and key2, all of the same priority. Another option is to disallow both ArtifactVerifyKey and ArtifactVerifyKeys from being set, or overriding the single key with the list of keys. I don't know if that could have some compatibility issues with the fallback Mender config though.

kacf:
Yes, I think `ArtifactVerifyKeys` with a list value is good.

I think we should disallow having both `ArtifactVerifyKey` and `ArtifactVerifyKeys`, mostly for usability/confusion reasons. Better to be 100% explicit when it comes to security, and only allow one of the two.

It should not affect the fallback config. If you wanted to use it you would have to override the whole value, which is the same as today. And I think having the `ArtifactVerifyKey` in `/data/mender/mender.conf` is anyway uncommon and not recommended.

da77a:
I came here looking for the opposite ie artifact with multiple signature (files) as that is (correct me if I'm wrong) is needed to allow a "late" change of key. By late I mean when you find RSA is broken you issue a signed artifact with the now weak) RSA signature AND the non-broken new signature type not even known to existing deployed client versions, signature. The artifact contains an image that includes the new client that understands the new signature type and a mender.conf with ArtifactVerifyKey: asecurekey - obviously there is no way to make that change without a new artifact format version. The fact the format lacks any extension point for issues like this (e.g. just the ability to skip/ignore file elements in the tar that it doesn't recognise - like - say - a manifest.signatures present after or before the manifest.sig) seems problematic - maybe the first req is for format version 4 to address that?

kacf:
It's possible that would require require a new artifact format, yes.

There are workarounds for this issue though, at least on Mender Enterprise: Give each of your artifacts an extra "Provides" field which contains the checksum of the key inside that artifact (the one it contains, not the one it's signed with). Make sure that each artifact from that point onwards has a "Depends" field on the key checksum. This checksum should match the key the artifact is signed with. When you need an artifact signed with two keys, just sign two copies with one key each, and upload both. The server will then take care of giving the correct artifact to each device.

I know, not perfectly ideal, and it is some manual work, but once it's set up, it should be pretty seamless.

Michaelhobo:
Closing in favor of https://github.com/mendersoftware/mender/pull/1108

PR #1051, opened 2022-10-13
fix: Expand the check for new openssl version

TheMeaningfulEngineer:
[Docs preauth flow](https://docs.mender.io/server-integration/preauthorizing-devices#generate-a-client-keypair) fails if openssl 3 is used.

Changelog: None

Signed-off-by: Alan <alan.martinovic@northern.tech>

### DIFF
--- a/support/keygen-client
+++ b/support/keygen-client
@@ -20,7 +20,7 @@ OPENSSL_VERSION_REGEX_MAJOR_BACKREF="OpenSSL ([0-9]+).*"
 OPENSSL_VERSION_STRING=$(openssl version)
 OPENSSL_VERSION_MAJOR=$(echo "$OPENSSL_VERSION_STRING" | sed -En "s/$OPENSSL_VERSION_REGEX_MAJOR_BACKREF/\1/p")
 
-if [ "$OPENSSL_VERSION_MAJOR" != "1" ]; then
+if [ "$OPENSSL_VERSION_MAJOR" -lt "1" ]; then
   echo "ERROR: openssl is too old, need version 1.0.0 or newer"
   echo "ERROR: OPENSSL_VERSION_STRING=$OPENSSL_VERSION_STRING"
   exit 1


support/keygen-client
```diff
@@ -20,7 +20,7 @@ OPENSSL_VERSION_REGEX_MAJOR_BACKREF="OpenSSL ([0-9]+).*"
 OPENSSL_VERSION_STRING=$(openssl version)
 OPENSSL_VERSION_MAJOR=$(echo "$OPENSSL_VERSION_STRING" | sed -En "s/$OPENSSL_VERSION_REGEX_MAJOR_BACKREF/\1/p")
 
-if [ "$OPENSSL_VERSION_MAJOR" != "1" ]; then
+if [ "$OPENSSL_VERSION_MAJOR" != "1" ] && [ "$OPENSSL_VERSION_MAJOR" != "3" ]; then
```
kacf, line 23: This would be more future proof:

```
if [ "$OPENSSL_VERSION_MAJOR" -lt "1" ]; then
```

That would also include "2", but I think no such version exists.

kacf (changes requested):
Remember to use the new [semantic commit](https://github.com/mendersoftware/mender/blob/master/CONTRIBUTING.md#structuring-your-commits) prefixes. You should prefix your commit title with `fix: Expand...`.

I would consider using `Changelog: Title` also. It is a bug fix, even though it's a small one.

support/keygen-client
TheMeaningfulEngineer: Thanks.
Tested `./keygen-client` with the openssl 3 and this change. Works.

kacf (approved):
From my side it looks good. To make the CI happy you still need to include a `Ticket: <TICKET>` reference in the commit message, or just set it to `Ticket: None` if there is none.

kacf:
Sorry, I didn't notice it before, you also need to rebase on the latest master, your version is very old (check if the copyright year at the top of the file is 2022, btw, you may need to change it).

TheMeaningfulEngineer:
@kacf 
Rebased to master.
Didn't get the `check if the copyright year at the top of the file is 2022` part.

kacf:
Ok, now it passed, thanks!

kacf:
Cherry-picked to 3.3.x and 3.4.x as well.

PR #1060, opened 2022-10-26
chore: remove `py` dependency from tests

lluiscampos:


support/modules-artifact-gen/tests/requirements.txt
```diff
@@ -1,8 +1,8 @@
...
-pluggy==0.13.1
-py==1.10.0
-pyparsing==2.4.7
-pytest==6.2.3
-toml==0.10.2
+packaging==21.3
+pluggy==1.0.0
+pyparsing==3.0.9
+pytest==7.2.0
+tomli==2.0.1
```
merlin-northern, line 8: wow some of those versions were really old. with this PR we will also catch them, right?

merlin-northern (approved):

support/modules-artifact-gen/tests/requirements.txt
lluiscampos: They are the result of installing `pytest` and freezing the dependencies. Dependabot will be sending new PRs for the dependencies that have updates :+1:

PR #1061, opened 2022-10-26
Bootstrap artifact fixes

kacf:


lluiscampos (approved):
Looks good.

I would remove both `Changelog` entries because the feature is not released yet (iow such entries give no meaning coming from `3.4.x` versions).

kacf:
Ah good point.

kacf:
@oleorhagen, @lluiscampos: The nasty upgrade bug I was talking about in the last commit!

oleorhagen:
Hah! Good one!

kacf:
Finally https://gitlab.com/Northern.tech/Mender/mender-qa/-/pipelines/683420593 passed. The one failure which is there is some infrastructure error, and happened after all the tests had already passed.

PR #1062, opened 2022-10-28
feat(logger): Add a thread: <proxy> field to the proxy logger

oleorhagen:
Now the proxy logger stands out in the logs more, and can thus be filtered
easier with tools parsing structured logs. This means that a log line from the
`proxy` thread goes from looking like:

```
Oct 05 08:50:06 qemux86-64 mender[259]: time="2022-10-05T08:50:06Z" level=error msg="error forwarding from client to backend: websocket: close 1006 (abnormal closure): unexpected EOF"
```

To:

```
Oct 05 08:50:06 qemux86-64 mender[259]: time="2022-10-05T08:50:06Z" level=error thread="proxy" msg="error forwarding from client to backend: websocket: close 1006 (abnormal closure): unexpected EOF"
```

Changelog: Commit
Ticket: None

### DIFF
--- a/app/proxy/proxy.go
+++ b/app/proxy/proxy.go
@@ -27,10 +27,12 @@ import (
 
 	"github.com/gorilla/websocket"
 	"github.com/pkg/errors"
-	log "github.com/sirupsen/logrus"
+	"github.com/sirupsen/logrus"
 
 	"github.com/mendersoftware/mender/client"
 )
+
+var log = logrus.WithFields(logrus.Fields{"thread": "proxy"})
 
 const (
 	ProxyHost        = "127.0.0.1"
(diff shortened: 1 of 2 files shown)


snusifer (approved):
Handy 🙌

lluiscampos (approved):
We are not using `thread` anywhere else but it is a good practice that we could extend :+1:

oleorhagen:
> We are not using `thread` anywhere else but it is a good practice that we could extend +1

Yes, I was pondering if we should add it to the regular client. In the end I decided against it. I might have been wrong though. Perhaps we should make it symmetric :thinking:

lluiscampos:
> Perhaps we should make it symmetric thinking

I think it is fine to keep the main client without this "thread" field. But maybe in the future is we are debugging other areas we might add more (authmanager code comes to my head as a good candidate). Anyway, it is good to go from my side!

oleorhagen:

PR #1063, opened 2022-10-29
ME-5 and MEN-6009: http-proxy fixes

snusifer:
Two commits:

fix: websocket connectivity through http-proxy if configured

Enables websocket connections to be established through an http-proxy configurable by setting the `HTTPS_PROXY` environment variable. This renders services that relies on websocket connections, such as `mender-connect`, compatible with http-proxying.

Changelog: All
Ticket: ME-5
Signed-off-by: Mikael Torp-Holte <mikael.torp-holte@northern.tech>

`---`

fix: client not to skip custom TLS if an http-proxy is configured

Previously, Mender client supported http-proxying but ignored custom TLS client configuration if present. This change renders any custom TLS configurations, such as Mutual TLS, compatible with http-proxying.

Changelog: All
Ticket: MEN-6009
Signed-off-by: Mikael Torp-Holte <mikael.torp-holte@northern.tech>

### DIFF
--- a/app/proxy/proxy_ws_test.go
+++ b/app/proxy/proxy_ws_test.go
@@ -434,3 +434,27 @@ func TestProxyWsConnectMutualTLS(t *testing.T) {
 
 	runTestSendReceiveWs(t, srv, proxyController)
 }
+
+func TestProxyWsConnectCustomCertWithReverseProxy(t *testing.T) {
+	httpProxy, err := cltest.NewTestHttpProxy(1, false)
+	require.NoError(t, err)
+	defer func() {
+		err := httpProxy.Stop()
+		require.NoError(t, err)
+	}()
+	t.Setenv("HTTPS_PROXY", httpProxy.GetUrl())
+
+	TestProxyWsConnectCustomCert(t)
+}
+
+func TestProxyWsConnectMutualTLSWithReverseProxy(t *testing.T) {
+	httpProxy, err := cltest.NewTestHttpProxy(1, true)
+	require.NoError(t, err)
+	defer func() {
+		err := httpProxy.Stop()
+		require.NoError(t, err)
+	}()
+	t.Setenv("HTTPS_PROXY", httpProxy.GetUrl())
+
+	TestProxyWsConnectMutualTLS(t)
+}
--- a/client/client.go
+++ b/client/client.go
@@ -506,8 +506,12 @@ func dialOpenSSL(
 	ctx *openssl.Ctx,
 	conf *Config,
 	_, addr string,
-	proxyURL *url.URL,
 ) (net.Conn, error) {
+	proxyURL, err := ProxyURLFromHostPortGetter(addr)
+	if err != nil {
+		return nil, errors.Wrapf(err,
+			"Failed to get http-proxy configurations during dial")
+	}
 	flags := openssl.DialFlags(0)
 
 	if conf.NoVerify {
@@ -586,9 +590,8 @@ func newHttpsClient(conf Config) (*http.Client, error) {
 	transport := http.Transport{
 		DisableKeepAlives: disableKeepAlive,
 		IdleConnTimeout:   time.Duration(idleConnTimeoutSeconds) * time.Second,
-		Proxy:             http.ProxyFromEnvironment,
 		DialTLS: func(network string, addr string) (net.Conn, error) {
-			return dialOpenSSL(ctx, &conf, network, addr, nil)
+			return dialOpenSSL(ctx, &conf, network, addr)
 		},
 	}
 
@@ -740,12 +743,7 @@ func newWebsocketDialerTLS(conf Config) (*websocket.Dialer, error) {
 
 	dialer := websocket.Dialer{
 		NetDialTLSContext: func(_ context.Context, network string, addr string) (net.Conn, error) {
-			proxyURL, err := ProxyURLFromHostPortGetter(addr)
-			if err != nil {
-				return nil, errors.Wrapf(err,
-					"Failed to get http-proxy configurations during websocket dial")
-			}
-			return dialOpenSSL(ctx, &conf, network, addr, proxyURL)
+			return dialOpenSSL(ctx, &conf, network, addr)
 		},
 	}
 
--- a/client/client.go
+++ b/client/client.go
@@ -14,7 +14,9 @@
 package client
 
 import (
+	"bufio"
 	"context"
+	"encoding/base64"
 	"encoding/json"
 	"fmt"
 	"io"
@@ -484,14 +486,35 @@ func loadClientTrust(ctx *openssl.Ctx, conf *Config) (*openssl.Ctx, error) {
 	return ctx, nil
 }
 
-func dialOpenSSL(ctx *openssl.Ctx, conf *Config, _ string, addr string) (net.Conn, error) {
+func establishSSLConnection(
+	addr string,
+	proxyURL *url.URL,
+	ctx *openssl.Ctx,
+	flags openssl.DialFlags,
+) (*openssl.Conn, error) {
+	if proxyURL != nil {
+		proxyConn, err := dialProxy("tcp", addr, proxyURL)
+		if err != nil {
+			return nil, errors.Wrap(err, "Failed to connect to proxy")
+		}
+		return openssl.DialUpgrade(addr, proxyConn, ctx, flags)
+	}
+	return openssl.Dial("tcp", addr, ctx, flags)
+}
+
+func dialOpenSSL(
+	ctx *openssl.Ctx,
+	conf *Config,
+	_, addr string,
+	proxyURL *url.URL,
+) (net.Conn, error) {
 	flags := openssl.DialFlags(0)
 
 	if conf.NoVerify {
 		flags = openssl.InsecureSkipHostVerification
 	}
 
-	conn, err := openssl.Dial("tcp", addr, ctx, flags)
+	conn, err := establishSSLConnection(addr, proxyURL, ctx, flags)
 	if err != nil {
 		return nil, err
 	}
@@ -565,7 +588,7 @@ func newHttpsClient(conf Config) (*http.Client, error) {
 		IdleConnTimeout:   time.Duration(idleConnTimeoutSeconds) * time.Second,
 		Proxy:             http.ProxyFromEnvironment,
 		DialTLS: func(network string, addr string) (net.Conn, error) {
-			return dialOpenSSL(ctx, &conf, network, addr)
+			return dialOpenSSL(ctx, &conf, network, addr, nil)
 		},
 	}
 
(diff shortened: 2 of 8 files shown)


snusifer:
Verified this works with wireshark as well as the tests

oleorhagen:
Since @lluiscampos is in the loop on this one, I'm skipping this one.

Skimmed through it, and on the surface looks fine :)

client/client.go
```diff
@@ -484,7 +486,69 @@ func loadClientTrust(ctx *openssl.Ctx, conf *Config) (*openssl.Ctx, error) {
...
+func upgradeTCPConnectionToSSL(
+	ctx *openssl.Ctx,
+	conn net.Conn,
+	addr string,
+	flags openssl.DialFlags,
+) (*openssl.Conn, error) {
+	host, _, err := net.SplitHostPort(addr)
+	if err != nil {
+		return nil, err
+	}
```
lluiscampos, line 498: In the original `openssl.DialSession` there is support for `ctx` being `nil`. Should we add it here also?
```
	if ctx == nil {
		var err error
		ctx, err = NewCtx()
		if err != nil {
			return nil, err
		}
		// TODO: use operating system default certificate chain?
	}
```

client/client.go
```diff
@@ -484,7 +486,69 @@ func loadClientTrust(ctx *openssl.Ctx, conf *Config) (*openssl.Ctx, error) {
...
+	if flags&openssl.InsecureSkipHostVerification == 0 {
+		err = tlsConn.VerifyHostname(host)
+		if err != nil {
+			conn.Close()
+			return nil, err
+		}
+	}
+	return tlsConn, nil
+}
```
lluiscampos, line 523: The whole `upgradeTCPConnectionToSSL` function should be moved to `openssl` package (we maintain and use our [own fork](https://github.com/mendersoftware/openssl/)).

client/client.go
```diff
@@ -484,7 +486,69 @@ func loadClientTrust(ctx *openssl.Ctx, conf *Config) (*openssl.Ctx, error) {
...
+
+	conn, err := upgradeTCPConnectionToSSL(ctx, proxyConn, addr, flags)
+	if err != nil {
+		return nil, err
+	}
+
+	return verifyTLSResult(conf, conn)
+}
+
+func dialOpenSSL(ctx *openssl.Ctx, conf *Config, _, addr string) (net.Conn, error) {
```
lluiscampos, line 551: I see you moved out `verifyTLSResult(conf, conn)` for reuse. I think we should do the same for the preamble part (getting the flags) so that we repeat no code between `dialOpenSSL` and `proxyDialOpenSSL`.

lluiscampos: An alternative is to refactor and merge both functions, just handling when `proxyUrl` is not `nil`. This way the calls from lines 791-794 would look nicer (but there the code will read uglier). Your call.

client/client.go
```diff
@@ -717,13 +785,98 @@ func newWebsocketDialerTLS(conf Config) (*websocket.Dialer, error) {
 
 	dialer := websocket.Dialer{
 		NetDialTLSContext: func(_ context.Context, network string, addr string) (net.Conn, error) {
+			proxyUrl, err := getProxyUrlFromHostPort(addr)
+			if err != nil {
+				log.Errorf("Failed to get http-proxy configurations during websocket dial: %v", err)
```
lluiscampos, line 790: I would argue that if this fails we want the whole operation to fail (not to continue dialing with no proxy). This is what I think a user expects, so that he/she can see what is going on and possibly amend the env variables.

client/client.go
```diff
@@ -717,13 +785,98 @@ func newWebsocketDialerTLS(conf Config) (*websocket.Dialer, error) {
...
 }
 
+func getProxyUrlFromHostPort(addr string) (*url.URL, error) {
+	u, err := url.Parse("https://" + addr)
+	if err != nil {
+		return u, err
+	}
+	// presumably for security reasons, http.ProxyFromEnvironment returns
+	// nil if req.URL is localhost. To aid testing, allow localhost proxying
+	// only if both HTTPS_PROXY and req.URL are locahosts
```
lluiscampos, line 808: We should not do this. We need to find a way for the tests to override this method without hard-coding "test aide" code in the routine. We can have a public method like `client.ProxxyUrlFromHostPortGetter` in the package set to the safe function and for the tests to be able to override it with this special tweaking.

client/client.go
```diff
@@ -717,13 +785,98 @@ func newWebsocketDialerTLS(conf Config) (*websocket.Dialer, error) {
...
+			}
 			return dialOpenSSL(ctx, &conf, network, addr)
 		},
 	}
 
 	return &dialer, nil
 }
 
+func getProxyUrlFromHostPort(addr string) (*url.URL, error) {
+	u, err := url.Parse("https://" + addr)
```
lluiscampos, line 802: `http` proxy is not supported?

client/client.go
```diff
@@ -717,13 +785,98 @@ func newWebsocketDialerTLS(conf Config) (*websocket.Dialer, error) {
...
+		}
+	}
+	return hostPort
+}
+
+type httpProxyDialer struct {
+	proxyURL *url.URL
+}
+
+func (hpd *httpProxyDialer) Dial(network string, addr string) (net.Conn, error) {
```
lluiscampos, line 838: Where is the base of this code coming from? Please document it.

client/client.go
```diff
@@ -717,13 +785,98 @@ func newWebsocketDialerTLS(conf Config) (*websocket.Dialer, error) {
...
+	}
+	br := bufio.NewReader(conn)
+	resp, err := http.ReadResponse(br, connectReq)
+	if err != nil {
+		conn.Close()
+		return nil, err
+	}
+	defer resp.Body.Close()
+
+	if resp.StatusCode != 200 {
```
lluiscampos, line 872:
```suggestion
	if resp.StatusCode != http.StatusOK {
```

client/client.go
```diff
@@ -717,13 +785,98 @@ func newWebsocketDialerTLS(conf Config) (*websocket.Dialer, error) {
...
+		conn.Close()
+		f := strings.SplitN(resp.Status, " ", 2)
+		return nil, errors.New(f[1])
```
lluiscampos, line 875: What is this?

client/client.go
```diff
@@ -717,13 +785,98 @@ func newWebsocketDialerTLS(conf Config) (*websocket.Dialer, error) {
...
+			connectHeader.Set("Proxy-Authorization", "Basic "+credential)
+		}
+	}
+
+	connectReq := &http.Request{
+		Method: http.MethodConnect,
+		URL:    &url.URL{Opaque: addr},
+		Host:   addr,
+		Header: connectHeader,
+	}
```
lluiscampos, line 859: The details here go a bit over my head.

lluiscampos (changes requested):
Man, great work figuring out this stuff! Really solid investigation and proposal.

I have not yet reviewed the tests, but I collected a good bunch of comments so here they go :+1:

lluiscampos (commented):
Good commit message, but remember that you can use markdown syntax to make even nicer changelog entries. So `HTTP_PROXY`, `HTTPS_PROXY` and `mender-connect` can be formatted as code keywords.

client/client.go
snusifer: It's from `gorillla/websocket`'s `proxy.go` which in turn took it from `net/http`'s `transport.go` I believe (don't remember exactly where but the place where they handle `CONNECT` request/response).

Will document👍

client/client.go
snusifer: Could you elaborate? which details?

client/client.go
snusifer: Wow, this slipped my mind, but I didn't write this. Code taken from `gorilla/websocket`'s `proxy.go`.

client/client.go
snusifer: No, I agree with this. Yes would be a bit uglier, but its only a quick branching

client/client.go
snusifer: Why now, The functions are supposed to mirror each other anyhow

client/client.go
lluiscampos: Yes, if we do the branching we can undo the refactoring of `verifyTLSResult`. I trust your :eye: on this

client/client.go
lluiscampos: I meant that I myself are no expert in HTTP protocol and can't really say this is 100% correct. Nothing specific that looks bad, just a warning note.

client/client.go
```diff
@@ -484,7 +486,69 @@ func loadClientTrust(ctx *openssl.Ctx, conf *Config) (*openssl.Ctx, error) {
 	return ctx, nil
 }
 
-func dialOpenSSL(ctx *openssl.Ctx, conf *Config, _ string, addr string) (net.Conn, error) {
+func upgradeTCPConnectionToSSL(
```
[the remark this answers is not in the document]
snusifer: No need when moved to `openssl` package

client/client.go
snusifer: Ah perhaps I misread this comment. I thought u might have referred to possibility that an index-out-of-bounce error could occur. This line returns the an error with a value equal to the response text. But if the response doesn't have a propper readline like `HTTP/1.1 202 Accepted`, but `HTTP/1.1 202` (without the reason part) then this will cause a runtime error. 
I should also wrap this error giving info that this is in fact coming from the response itself.

client/client.go
snusifer: Ah, but nevermind. When refactoring I saw why I did it like this in the first place. `dialOpenSSL` is also used here:
https://github.com/mendersoftware/mender/blob/46a4419ffa64902bd6570b408df9da8734da4309/client/client.go#L563-L570
This would be OK, but doing so will actually solve MEN-6009, which is not the scope of this PR, even thought they are closely related. 
So either do it, then extend the scope of this PR to cover both of the issues, or have something similar to what its like now, then refactor later when solving MEN-6009.

What do you think?

client/client.go
snusifer: Problem: 

The `Proxy` field in the `websocket.Dialer` and `http.Transport` is a function that takes an `*http.Request` and returns an `*url,URL`. Because it is a function it can be customized to dynamically decide if proxying should be used and what the proxy target is given a request. The request itself has an `*url.URL` which is what is used to decide the aforementioned points.
 
However, in `NetDialTLSContext` we don`t have the full url. Instead we have `addr` whos value is `URL.Host` (scheme is unknown). So what this does is to assume that the target is using `HTTPS`, which I think is OK since the mender-client won't talk to servers using unencrypted `HTTP` anyway.

client/client.go
snusifer: Resolving convo because function is now in `mendersoftware/openssl`

client/client.go
```diff
@@ -717,13 +785,98 @@ func newWebsocketDialerTLS(conf Config) (*websocket.Dialer, error) {
...
+		if targetHost == "127.0.0.1" && proxyHost == "127.0.0.1" {
+			return proxyUrl, err
+		}
+	}
+	return http.ProxyFromEnvironment(&http.Request{URL: u})
+}
+
+func getHostPort(u *url.URL) (hostPort string) {
+	hostPort = u.Host
+	if i := strings.LastIndex(u.Host, ":"); i > strings.LastIndex(u.Host, "]") {
```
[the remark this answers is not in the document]
snusifer: This function adds port number if not present. `net.SplitHostPort` returns `error` if no port exists. I could check the error return value, then add the port number, but there is no guarantee that the missing port number is the reason for the error.

client/client.go
snusifer: I downgraded the function to private. Does it still need documentation?

client/client.go
snusifer: I did this, but somehow I liked `verifyTLSResult` also. Im 50/50 split on that

client/client.go
snusifer: @lluiscampos 
A side note here: we don't really support the `HTTP_PROXY` variable. It implies that the `ServerURL`'s scheme is HTTP and we don't allow application layer communication without any form of encryption.

snusifer:
@mender-test-bot 
start pipeline

client/client.go
```diff
@@ -724,6 +751,70 @@ func newWebsocketDialerTLS(conf Config) (*websocket.Dialer, error) {
...
+		URL:    &url.URL{Opaque: addr},
+		Host:   addr,
+		Header: connectHeader,
+	}
+	if err := connectReq.Write(conn); err != nil {
+		conn.Close()
+		return nil, err
+	}
+	br := bufio.NewReader(conn)
+	resp, err := http.ReadResponse(br, connectReq)
```
snusifer, line 804: i think that this needs a timeout

client/client.go
```diff
@@ -724,6 +751,89 @@ func newWebsocketDialerTLS(conf Config) (*websocket.Dialer, error) {
...
+		br := bufio.NewReader(conn)
+		resp, err = http.ReadResponse(br, connectReq)
+	}()
+	select {
+	case <-connectCtx.Done():
+		conn.Close()
+		<-didReadResponse
+		return nil, connectCtx.Err()
+	case <-didReadResponse:
+	}
```
snusifer, line 822: I changed this so that it will timout if no response was received after a minute. could u verify that his is correct?

snusifer, line 822: I changed this so that it will timeout if no response was received after a minute. @lluiscampos could you take a second look at this?

snusifer: I probably should send down the `ctx` from `NetDialTLSContext` and `DialTLSContext` so that we can spawn `connectCtx` from `ctx`.

client/client.go
lluiscampos: Okey.

client/client.go
lluiscampos: Thank you.

client/test/http_proxy.go
```diff
@@ -0,0 +1,278 @@
...
+	"strings"
+	"sync"
+
+	"github.com/mendersoftware/mender/client"
+
+	"github.com/pkg/errors"
+)
+
+func init() {
+	client.ProxyURLFromHostPortGetter = func(addr string) (*url.URL, error) {
```
lluiscampos, line 35: Nice :+1:

app/proxy/proxy_ws_test.go
```diff
@@ -434,3 +434,51 @@ func TestProxyWsConnectMutualTLS(t *testing.T) {
 
 	runTestSendReceiveWs(t, srv, proxyController)
 }
+
+func TestProxyWsConnectWithReverseProxy(t *testing.T) {
```
lluiscampos, line 438: Could we add another test with MutualTLS and ReverseProxy?

client/client.go
```diff
@@ -717,13 +740,101 @@ func newWebsocketDialerTLS(conf Config) (*websocket.Dialer, error) {
...
+					"Failed to get http-proxy configurations during websocket dial")
+			}
+			return dialOpenSSL(ctx, &conf, network, addr, proxyURL)
 		},
 	}
 
 	return &dialer, nil
 }
 
+var ProxyURLFromHostPortGetter = func(addr string) (*url.URL, error) {
```
lluiscampos, line 753: As this is such an usual method to be public, please add a minimal documentation for it.

app/proxy/proxy_ws_test.go
lluiscampos: Also, both tests using `NewTestHttpProxy` require authentication. So for the sake of don't leaving test code unused, we could do one of the tests with no auth (the new one I suggested or one of the existing ones).

lluiscampos (approved):
The second changelog entry could be more verbose. It is an important fix and honors a better explanation. Particularly I would add something like like "Previously Mender client supported http-proxying but ignoring custom TLS client configuration, if present, which was rending Mutual TLS and http-proxy settings incompatible.

I added a suggestion for more testing.

Other than that, solid work. Thank you :rocket:

app/proxy/proxy_ws_test.go
snusifer: Two tests in `proxy_ws_tests.go` now: `TestProxyWsConnectCustomCertWithReverseProxy` and `TestProxyWsConnectMutualTLSWithReverseProxy`. 

Both identical to their non-proxy counterpart. The former does not require basic auth, the latter does.

client/client.go
snusifer: Added! :)

snusifer:
@mender-test-bot 
start pipeline

PR #1080, opened 2022-11-22
mender-device-identity: Skip private MAC addresses

nandra:
Signed-off-by: Marek Belisko <marek.belisko@open-nandra.com>

### DIFF
--- a/support/mender-device-identity
+++ b/support/mender-device-identity
@@ -44,6 +44,14 @@ for dev in $SCN/*; do
 	continue
     fi
 
+    # Skip private MAC addresses (see https://superuser.com/questions/907827/)
+    case $(cat $dev/address | cut -c2) in
+        2*) continue;;
+        6*) continue;;
+        a*) continue;;
+        e*) continue;;
+    esac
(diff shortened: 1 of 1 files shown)


kacf:
I think we need a bit more background here. I get that the addresses are private, but why shouldn't they be used?

And even if we should avoid them, I'm not sure if we can. Due to the changing identity, this could orphan existing devices on a massive scale.

nandra:
We have specific situation where we use multiple network cards and for some of them MAC is not available so we set same for all of them. In this case without patch devices will be shows with same MAC address.  So with patch we try to avoid that and in real scenario usually adapters have specified MAC address which is not private.

kacf:
Ok. I think this is still fairly specific to your use case. For some people the private address might be all they have, so it's better to rely on the `ifindex` to select the preferred one if there are multiple cards. IMHO, it is too risky to apply this change to everyone, so you should probably keep this in a fork.

nandra:
OK then can be closed. Thanks for comments.

PR #1086, opened 2022-11-23
fix: duplicate journald log entries in some systemd configurations

snusifer:
Since systemd v213, `journald` symlinks `/dev/log` to point to itself. The syslogger in mender-client is configured by default to forward log messages to `/dev/log`. `journald` will receive duplicate log messages; one from `/dev/log` and one from `stdout`.

Changelog: Add `--no-syslog` to the service file to ensure no duplicate log messages in the journal.

Ticket: MEN-6070
Signed-off-by: Mikael Torp-Holte <mikael.torp-holte@northern.tech>

### DIFF
--- a/support/mender-client.service
+++ b/support/mender-client.service
@@ -8,7 +8,7 @@ Conflicts=mender.service
 Type=idle
 User=root
 Group=root
-ExecStart=/usr/bin/mender daemon
+ExecStart=/usr/bin/mender --no-syslog daemon
 Restart=always
 
 [Install]


snusifer:
@mender-test-bot 
start pipeline

support/mender-client.service
```diff
@@ -8,7 +8,7 @@ Conflicts=mender.service
 Type=idle
 User=root
 Group=root
-ExecStart=/usr/bin/mender daemon
+ExecStart=/usr/bin/mender --no-syslog daemon
```
lluiscampos, line 11: We should not do this here, only in `meta-mender` for the branches that this behavior changed (`master` and `kirkstone`).

@kacf and @olepor have the last word :+1:

kacf: @lluiscampos: We agreed yesterday to do it here because since this is the systemd service file, it is guaranteed that the logs are always available using journald. It's possible that someone who were previously using a separate syslog daemon won't see the logs there anymore, which is a behavior change, but we concluded that it is justified since nothing will be lost, and logs are just an informational tool. We do want to keep it in master only though, so at least you won't get this surprise in a patch release.

kacf (approved):
Good commit message!

support/mender-client.service
lluiscampos: Thanks for the explanation

oleorhagen (approved):

PR #1088, opened 2022-12-01
feat: First version of Mender googletest framework in C++.

kacf:
```
commit 4c7e84dcfc1886a3e1ab2be6f7117511a222fdae
Date:   Thu Dec 1 09:35:44 2022

    feat: First version of Mender googletest framework in C++.
    
    This comes with several cleanups in the gitignore file, as well as
    provisionary Makefile changes to get this off the ground. The Makefile
    is expected to change substantially once we get the new build system
    in place though.
    
    Changelog: None
    Ticket: MEN-5794
    
commit 3dfc9a02478d6e0fd8b8cb53b9f8255eb9225021
Date:   Wed Nov 30 10:43:35 2022

    feat: Remove Golang version of client in preparation for C++ version.
    
    Note that I have removed some files that we may still use in the
    future, in tests particularly, but I figured it's better to fish them
    out of the history if/when we need them instead of leaving a lot of
    litter around.
    
    Changelog: None
    Ticket: MEN-5798
    
```


.gitlab-ci.yml
```diff
@@ -4,39 +4,30 @@ stages:
...
   - project: 'Northern.tech/Mender/mendertesting'
     file: '.gitlab-ci-check-license.yml'
 
+variables:
+  GIT_SUBMODULE_STRATEGY: recursive
+
 test:
   stage: test
-  image: golang:1.17
+  image: debian:11
```
vpodzime, line 20: My experience is that Fedora provides the newest versions of compilers, linters, static checkers,... which often mean more issue found. But not a super-strong preference here. For inspiration, this is what we have in CFEngine/Core: https://github.com/cfengine/core/blob/master/Makefile.am#L170

.gitlab-ci.yml
```diff
@@ -62,31 +53,26 @@ test:docker:
...
-    #  and pass few others as command line arguments.
-    #  See also https://docs.coveralls.io/api-reference
-    - export CI_BRANCH=${CI_COMMIT_BRANCH}
+    - apt update && apt install -yyq lcov
+    - pip install cpp-coveralls pyyaml
+
+    # eddyxu/cpp-coveralls appears dead, but there doesn't seem to be an
+    # alternative. Use this patch from someone who attempted to fix it. An
+    # alternative to this is to use pyyaml<6, but it's better to use just one
+    # old component than two.
```
vpodzime, line 67: :+1:

.gitlab-ci.yml
```diff
@@ -62,31 +53,26 @@ test:docker:
...
-      -jobid $(git rev-parse HEAD)
-      -covermode set
-      -flagname unittests
-      -parallel
-      -coverprofile $(find tests/unit-coverage -name 'coverage.txt' | tr '\n' ',' | sed 's/,$//')
+    - 'echo "service_name: gitlab-ci" > .coveralls.yml'
+    - cpp-coveralls
+      --repo-token ${COVERALLS_TOKEN}
+      --no-gcov
+      --lcov-file coverage.lcov
```
vpodzime, line 75: YAML bash is horrible...

Makefile
```diff
@@ -202,86 +163,44 @@ uninstall-examples:
...
+		$(LDFLAGS)
 
 .PHONY: build
 .PHONY: clean
-.PHONY: get-tools
 .PHONY: test
 .PHONY: check
-.PHONY: cover
-.PHONY: htmlcover
-.PHONY: coverage
```
vpodzime, line 284: `coverage` is still a PHONY target.

vpodzime (commented):
> it's better to fish them
out of the history if/when we need them instead of leaving a lot of
litter around

Cannot agree more! Thanks!

Looks good to me otherwise. A good starting point. :+1:

.gitlab-ci.yml
lluiscampos: Shouldn't we apply the same to https://github.com/mendersoftware/mender-binary-delta/blob/master/.gitlab-ci.yml#L173 ?

Even better if we move this to a template. Our "unit tests" template assumes golang, but now we'll have two projects C/C++ so it makes sense to fork into a different template for them.

lluiscampos (approved):

.gitlab-ci.yml
kacf: It's more about convention on the Mender team. We generally stick to alpine if we can, mainly due to speed and size benefits, and then move to either Ubuntu or Debian if we need something a bit more esoteric. So far it hasn't let us down, so I think introducing a third package manager is unnecessary, at least for now.

Makefile
kacf: Oops, good catch!

.gitlab-ci.yml
kacf: Hehe, the hyphen behavior is a bit surprising, but at least it's readable!

.gitlab-ci.yml
kacf: It does make sense, but it makes testing it much harder, and I'm still not sure if this works completely the way it should (it's calling the branch "HEAD", and the mender-binary-delta version has some references to Travis variables... is it hardcoded in cpp-coveralls?). I think we still need to test this a bit in practice, so I think I'll create a task for it instead.

kacf: https://tracker.mender.io/browse/QA-511

.gitlab-ci.yml
vpodzime: Yeah, that's what I was afraid of. :wink: Just to support my point:
```
$ gcc --version
gcc (Debian 10.2.1-6) 10.2.1 20210110
$ clang --version
Debian clang version 11.0.1-2
```
on Debian 11,  vs. this on Fedora 36:
```
$ gcc --version
gcc (GCC) 12.2.1 20221121 (Red Hat 12.2.1-4)
$ clang --version
clang version 14.0.5 (Fedora 14.0.5-1.fc36)
```
and https://github.com/cfengine/core/pull/4959#issue-1249629787

PR #1091, opened 2022-12-05
Prototype build system with Autotools

lluiscampos:
This prototype is _functional_. One of you needs to take over and complete it :toolbox: 

### Issues

- [x] Discuss the proposal for source files hierarchy. What do you think about this way for organizing libraries and their implementations?
- [x] Fix the boost libraries situation. We cannot vendor 130MiB of code... :sweat_smile: I was not able to find a suitable solution and I run out of time.
- [x] use subtargets
- [x] Implement in `Makefile.am` all functionality that I have nuked (install targets for inventory, update modules, etc)
- [x] Restore test targets in the Makefile.
- [x] support (ARM) cross-compilation

###  Commits

* feat: Skeleton/proposal on how to organize the source code

    Taking json as an example, this proposal is based on:
    * `common/<lib>/<lib>.hpp` to describe the interface (abstract class)
      for the library
    * `common/<lib>/impl/<impl>` to encapsulate concrete implementations, in
      this example json interface using Boost JSON library
    * `mender-auth`, `mender-update` to hold the main of the application,
      able to instanciate concrete objects for the interface(s).

* feat: Vendor Boost libraries v1.80

    We might want to revisit this later.
    
    Ideally we should get them from the distributed packages in the OS, but
    on latest Debian and Ubuntu Boost is still on 1.74 while JSON library,
    for example, requires 1.75...
    
    Alternatively we could have it as a submodule but I have not been able
    to find a suitable way do it cleanly.

* feat: Prototype build system with Autotools


common/json/impl/boost/boost_json.cpp
```diff
@@ -0,0 +1,17 @@
+#include <iostream>
+#include <common/json/impl/boost/boost_json.hpp>
+#include <boost/json/src.hpp>
+
+
+void json::BoostJson::hello_world()
```
kacf, line 6: It is perhaps easier to put most of the file inside a `namespace json { }`, or? This also avoids leaking into the unnamed namespace if you forget it somewhere.

common/json/impl/boost/boost_json.cpp
```diff
@@ -0,0 +1,17 @@
...
+{
+    boost::json::value hello_world = {{"hello", "Boost"}};
+
+    boost::json::serializer sr;
+    sr.reset(&hello_world);
+    do
+    {
+        char buf[16];
```
kacf, line 14: I'm tempted to say that we should use this instead:

```
vector<char> buf(16);
```

Because it has better range checking. `char buf[16]` behaves more or less like a pointer, which we are trying to avoid.

However, `serializer` doesn't seem to have an overload for that (nor for `string`), so I'm not sure if it will work well. I guess we are already entering into the world of compromise which is C++.

common/json/json.hpp
```diff
@@ -0,0 +1,12 @@
+#pragma once
+
+namespace json
+{
+
+class Json
+{
+   public:
+    virtual void hello_world() = 0;
```
kacf, line 9: How to print "Hello world" has always been a great mystery! :laughing:

mender-auth/main.cpp
```diff
@@ -0,0 +1,26 @@
+#include <iostream>
+
+using namespace std;
+
+
+#include <common/json/json.hpp>
+
+// This method only knows about the interface Json
+void hello_world(json::Json *j)
```
kacf, line 9: I think this is unused, or? (same for the other one)

In any case, I was going to comment on the raw pointer, which should be replaced, most likely by `shared_ptr`.

Makefile.am
```diff
@@ -0,0 +1,32 @@
...
+
+# Auth daemon
+AUTH_SOURCES = mender-auth/main.cpp
+men_authd_SOURCES = $(COMMON_SOURCES) $(AUTH_SOURCES)
+men_authd_LDADD = $(LIBRARIES)
+
+# Update daemon
+UPDATE_SOURCES = mender-update/main.cpp
+men_updated_SOURCES = $(COMMON_SOURCES) $(UPDATE_SOURCES)
+men_updated_LDADD = $(LIBRARIES)
```
kacf, line 32: Ideally, we should try to preserve the non-binary related Makefile targets from the original, to install update modules, inventory scripts, etc.

Makefile.am
```diff
@@ -0,0 +1,32 @@
...
+# Headers for libraries and utils under common to be listed here
+COMMON_SOURCES = common/json/json.hpp \
+	common/json/impl/boost/boost_json.hpp
+
+# Libraries objects to be listed here
+noinst_LIBRARIES = \
+	libjson.a
+
+# For each library, set the source(s)
+libjson_a_SOURCES = common/json/impl/boost/boost_json.cpp
```
kacf, line 22: Should these perhaps be sub-targets instead?

kacf (commented):
The autotools part is smaller than I feared! :rocket: 

I think we should invest some time in getting Boost in as a submodule instead. The whole package is massive, and we are only using 10% of it. If we used a submodule then we have the option of trimming the size later, if we find that we either don't have to vendor it, or we can vendor a partial copy.

Makefile.am
kacf: Never mind, this is on your list already.

kacf:
Good work. I like the split you have done, I have no further comments on that really!

Makefile.am
```diff
@@ -0,0 +1,32 @@
...
+
+ACLOCAL_AMFLAGS = -I m4
+
+AM_CXXFLAGS = -std=c++11 -Wall -Werror
+AM_CPPFLAGS = \
+	-I$(srcdir)/common/json \
+	-I$(srcdir)/vendor/boost_1_80_0
+
+# Two daemons
+bin_PROGRAMS = men-authd men-updated
```
oleorhagen, line 11: Finally could have gotten `menderd` :p

Makefile.am
oleorhagen: +1

lluiscampos:
> Yeh, not keen. How about a submodule?

I tried but failed with the submodule. I am trying again today :) If I don't get it before sprint change then whoever takes over will implement it this way

Makefile.am
oleorhagen: @kacf

Makefile.am
```diff
@@ -0,0 +1,32 @@
+AUTOMAKE_OPTIONS = subdir-objects
```
oleorhagen, line 1: Personally I prefer recursive subdirs, but it's only a matter of preference

lluiscampos:
I removed the monster commit :see_no_evil: and instead added Boost libraries as a Git submodule :heavy_check_mark:

common/json/impl/boost/boost_json.cpp
lluiscampos: I didn't expect a review of the actual code :sweat_smile: 

Yes, `read` takes only `char *`. This code we will remove but as you said there will need to be compromises in some areas.

mender-auth/main.cpp
lluiscampos: Thank you, it was meant to be used by the main below! Otherwise the whole point of this demo is lost :see_no_evil: 

The idea is to showcase that this function only knows about the interface (pure abstract class) while the caller instantiates a concrete class.

I'll fix and also add the `shared_ptr` :+1:

Makefile
```diff
@@ -42,6 +42,27 @@ DBUS_POLICY_FILES = \
...
 build: mender
 
+## Hardcoded prototype
+CXXFLAGS = \
+	-I . \
+	-I vendor/
+
+LDFLAGS = \
+	-L common/json/ \
+	-l json
```
vpodzime: > LDFLAGS
    Extra flags to give to compilers when they are supposed to invoke the linker, ‘ld’, such as -L. Libraries (-lfoo) should be added to the LDLIBS variable instead.

Might as well stick to the above.

common/json/impl/boost/boost_json.cpp
```diff
@@ -0,0 +1,21 @@
+#include <iostream>
+#include <common/json/impl/boost/boost_json.hpp>
+#include <boost/json/src.hpp>
```
vpodzime, line 3: I would prefer us not using header-only versions of the libraries and just statically-link the libraries with our code. Would that be a major obstacle/complication?

common/json/impl/boost/boost_json.hpp
```diff
@@ -0,0 +1,15 @@
+#pragma once
```
vpodzime, line 1: Not sure if we want to care about it now, but `#pragma`s are generally not very portable.

mender-auth/main.cpp
```diff
@@ -0,0 +1,27 @@
+#include <iostream>
+#include <memory>
+
+using namespace std;
+
+
+#include <common/json/json.hpp>
+
+// This method only knows about the interface Json
```
vpodzime, line 9: It's a function :stuck_out_tongue:

mender-update/main.cpp
```diff
@@ -0,0 +1,27 @@
...
+    j->hello_world();
+}
+
+
+#include <common/json/impl/boost/boost_json.hpp>
+
+int main()
+{
+    // It is here that we make an object from a concrete type, BoostJson.
+    shared_ptr<json::BoostJson> j = make_shared<json::BoostJson>();
```
vpodzime, line 21: How will this work with multiple implementations? My assumption was that we would have multiple different implementations for a class and **compilation+linking** will make sure to use the proper implementation on the given platform. So this code would not have to care about the actual implementation at all.

Makefile.am
vpodzime: recursive subdirs? Could you please expand on that?

Regarding `subdir-objects`, I'm not sure it's necessary and beneficial. Doing builds outside of the source tree with the products all ending up in the build tree is nice and `make distcheck` nicely checks that it works properly (i.e. that `$(srcdir)` and `$(builddir)` are used properly.

Makefile.am
vpodzime: `men-update` -- wow, updating men!

Makefile.am
vpodzime: > recursive subdirs? Could you please expand on that?

Oh, you mean `SUBDIRS` is missing here?

Makefile.am
vpodzime: +1

configure.ac
```diff
@@ -0,0 +1,42 @@
...
+AC_INIT([mender], [version], [mender@discoursemail.com])
+AC_CANONICAL_TARGET
+AM_INIT_AUTOMAKE([-Wall -Werror foreign])
+AM_SILENT_RULES([yes])
+
+dnl Use the m4 directory for m4 macros (convention):
+AC_CONFIG_MACRO_DIR([m4])
+
+AM_PROG_AR
+LT_INIT([shared])
```
vpodzime, line 11: I'd drop `([shared])` here. We don't build any shared libraries (yet).

mender-update/main.cpp
lluiscampos: This is a bit down the road and I wanted to keep that "road" open. Specially because I will not be deep involved in these design decisions.

FWIW, I was rather envisioning an Abstract Factory pattern, where the client code only this this to get the different objects, and then have a concrete factory, customized at build time (compilation + linking) for the target platform.

Again, this is a bit down the road. We should study how this is done in other projects :nerd_face:

common/json/impl/boost/boost_json.cpp
lluiscampos: Boost JSON seems to be a an exception, in that provides both static library and single header file approach. For the other Boost libraries that we have looked into they either provide one or the other.

I _think_ that the end result is the same, and with static library approach the compile time reduces greatly during local development (like you build it once, while with the header you will need to compile the objects every time you make a change in the client code, `bost_json.cpp` in this case).

No major obstacle/complication, I just wanted to keep my prototype simple :+1: 

So for json please go ahead and rework this to use the static library. Just keep in mind that for other libraries we'll go for the header file approach because they don't provide the static library approach.

common/json/impl/boost/boost_json.hpp
lluiscampos: I don't remember if we discussed this in the first meeting... If you have experience of not being portable we should switch to header guards as soon as possible.

I'll do this :memo:

mender-auth/main.cpp
lluiscampos: I'll fix :memo:

Makefile.am
lluiscampos: I leave this for you to rework @vpodzime :+1:

configure.ac
lluiscampos: Thanks :memo:

Makefile.am
lluiscampos: I leave this for you to rework @vpodzime :+1:

lluiscampos:
@vpodzime  I am reworking the small things that I marked with :memo: and leave the rest for your takeover. Then you can decide on keep pushing to this PR or launch a new one from work.

> Sorry for my questions being a bit over the scope of this initial work, but this is not a terrible place for discussions about them.

This is _the_ right place for such discussions.

Makefile
lluiscampos: Ah, wait. This is from the hard-coded code in `Makefile` that I removed when actually introducing autotools. So no-op here :see_no_evil:

common/json/impl/boost/boost_json.hpp
kacf: +1. The Google style guide also explicitly mentions that pragmas should be avoided, and header guards should be used instead.

Makefile.am
kacf: I have added this to the next client team meeting agenda on Monday. @lluiscampos: Could you invite @vpodzime to this meeting?

mender-update/main.cpp
vpodzime: > OTOH, maybe I'm just moving the problem around. slightly_smiling_face What do you think? Other ideas?

It's definitely an improvement because it moves the `#ifdef` as close as possible to the place where the actual decision has to be made. My idea, based on Lluis' point and my "C approach" was to have factory functions on top of costructors. So in the above case it would be

```
JsonClass &get_json();
```
where `get_json()` would have multiple implementations using various classes sub-classed from `JsonClass` and the compilation/build process would decide which one to use. Unlike the classes themselves, the function prototype would be just one for all the possible implementations. And the individual (sub-)classes would then have arbitrary members and only the factory function implementations would have to care about the specifics.

kacf: I think it's not possible to put instances on the stack with that method, is it? It would require using a smart pointer. So `myFunc` from above:

```
int MyFunc() {
    JsonClass json;
    json.DoStuff();
}
```

would not be possible. Instead it would have to be:

```
int MyFunc() {
    unique_ptr<JsonClass> json(get_json());
    json->DoStuff();
}
```

vpodzime: Yes, exactly. I figured that out while thinking about this over the weekend. I guess `get_json()` could return `unique_ptr<JsonClass>` directly? Anyway, it's stack allocation replaced by heap allocation which could (would) be a major performance penalty, I guess.

oleorhagen: >  could (would) be a major performance penalty, I guess.

Only thing ever really hampering us is I/O and write times in terms of performance.

vpodzime:
@lluiscampos cannot request a review from you because you are the author of this PR, but please have a look

mender-update/main.cpp
kacf: Yeah, it's mainly practicality I'm worried about here, not performance. I think the only CPU bound performance areas we are worried about are decompression, and to a much lesser degree, encryption. Everything else is miniscule in comparison.

common/json/impl/boost/boost_json.hpp
```diff
@@ -0,0 +1,18 @@
+#ifndef BOOST_JSON_HPP
```
oleorhagen, line 1: This actually conflicts with a definition in `Boost`. Should we prefix all our defs with `MENDER`?

configure.ac
```diff
@@ -0,0 +1,66 @@
+m4_define([mender_version], m4_normalize(m4_esyscmd([git describe --exact HEAD 2>/dev/null || git rev-parse --short HEAD])))
+AC_INIT([mender], [mender_version], [mender@discoursemail.com])
```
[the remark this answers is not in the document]
oleorhagen: Hmm, good question. I would probably favour `contact@northern.tech`.

Also, the package name. Should we keep just `mender` here?

oleorhagen:
It does not seem to build out of the `source` directory for me.

Does it work for you @vpodzime ?

Makefile.am
```diff
@@ -1,52 +1,57 @@
...
+
+all-local: mender
+
+clean-local:
+	rm -f mender main_test
+	( cd vendor/googletest && git clean -df )
 
 mender: main.cpp lib.cpp
-	g++ -o mender main.cpp lib.cpp $(CXXFLAGS) $(LDFLAGS)
+	$(CXX) -o mender main.cpp lib.cpp $(CXXFLAGS) $(LDFLAGS)
```
oleorhagen, line 50: Q: What about `CPPFLAGS`?

mender-auth/Makefile.am
```diff
@@ -0,0 +1,12 @@
+AUTOMAKE_OPTIONS = subdir-objects
+
+ACLOCAL_AMFLAGS = -I m4
+
+AM_CPPFLAGS = -I$(srcdir)/../common/json
```
oleorhagen, line 5: Q: Looking at the includes in the `common/json` files, should this not be:
```
AM_CPPFLAGS = -I$(srcdir)/../
```
? :thinking:

oleorhagen: Then the same for `update`?

Makefile.am
oleorhagen: I'm guessing this is going away anyhow - same as the files it references?

configure.ac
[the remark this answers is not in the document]
kacf: I think `boards@` is not the right address since we don't deal with board integrations here. Some OS-integrations maybe, but not single boards.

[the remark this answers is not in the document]
TheYoctoJester: I agree that its different, but its the closest we have to anything integration. Then its `contact@mender.io`probably.

common/json/impl/boost/boost_json.hpp
vpodzime: Generally the naming convention is not great. If you just compile this into a library, it will be called `libboost_json` which is again a conflict with Boost. And the `common/MODULE/impl/BACKEND/BACKEND_MODULE.[ch]pp` scheme is very redundant, IMHO. `src.[ch]pp` or `impl.[ch]pp` would make sense to me just fine. And the guard macro could be `JSON_BOOST_WHATEVER` or `MEN_JSON_BOOST_WHATEVER`.

oleorhagen: Yes, agree fully now that I have played around with this a bit!

configure.ac
```diff
@@ -0,0 +1,66 @@
...
+AC_CONFIG_FILES([
+    Makefile
+    common/Makefile
+    vendor/Makefile
+    mender-auth/Makefile
+    mender-update/Makefile
+])
+
+dnl Set C++ standard and a few nice compiler options
+CXXFLAGS="$CXXFLAGS -std=c++11 -Wall -Werror"
```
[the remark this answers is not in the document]
oleorhagen: What about: https://www.gnu.org/software/autoconf-archive/ax_cxx_compile_stdcxx_11.html#ax_cxx_compile_stdcxx_11

mender-update/main.cpp
oleorhagen: > I took a look at Zephyr, but it's not a good model because it's all C. So I took a look at Qt instead. They seem to use the "one class, separate implementations" approach (see [header](https://github.com/qt/qtbase/blob/1bc7e9e77b2d6b03b995f376b622d4c8c97dd51c/src/corelib/io/qfilesystemengine_p.h) and [implementation](https://github.com/qt/qtbase/blob/1bc7e9e77b2d6b03b995f376b622d4c8c97dd51c/src/corelib/io/qfilesystemengine_unix.cpp)). This means that their header file is riddled with ifdefs though.
> 
> I have a suggestion for an alternative approach:
> 
>     * A shared abstract class (interface).
> 
>     * Separate platform classes, with different names, which inherit from the interface.
> 
>     * Some common platform file somewhere which typedefs a common class name to the specific implementation class name.
> 
> 
> Here's an example:
> 
> ```
> common/json.hpp:
> class AbstractJsonClass {...};
> 
> common/platform.hpp:
> #ifdef POSIX
> typedef PosixJsonClass JsonClass;
> #endif
> 
> common/impl/posix_json.hpp:
> class PosixJsonClass : public AbstractJsonClass {...}
> 
> Other source files:
> int MyFunc() {
>     JsonClass json;
>     json.DoStuff();
> }
> ```
> 
> That would keep the naming complexity in one (or at most a few) files, and would also leave platform specific members fully contained in their own files, instead of spread around everywhere.
> 
> OTOH, maybe I'm just moving the problem around. slightly_smiling_face What do you think? Other ideas?

@kacf now that I have played around with this a bit, I would have to say that I favour the `QT` approach over what we are thinking about doing.

Could we have a small discussion around this during stdup tomorrow?

Makefile.am
```diff
@@ -1,52 +1,57 @@
...
+		lib.cpp \
+		main_test.cpp \
+		-pthread \
+		vendor/googletest/lib/libgtest.a \
+		vendor/googletest/lib/libgtest_main.a \
+		-I vendor/googletest/googletest/include \
+		$(CXXFLAGS) \
+		$(LDFLAGS)
+
+# ============= the rest of this file is to be removed =============
```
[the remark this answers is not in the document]
vpodzime: Well, if we continue with Autotools, they should be removed from the top-level _Makefile.am_ and instead the subfolders should have their own _Makefile.am_ files defining the installation directories and files that should go into them the Autotools-way, i.e. `systemddir = $(prefix)/lib/systemd/system` followed by `inst_systemd_DATA = mender-updated.service` and similar instead of having the `install`/`uninstall` targets here.

If we switch to CMake, then I guess we will use the CMake-way of specifying installation directories and their files (no clue how that is done).

Makefile.am
```diff
@@ -1,52 +1,57 @@
...
+
+all-local: mender
+
+clean-local:
+	rm -f mender main_test
+	( cd vendor/googletest && git clean -df )
```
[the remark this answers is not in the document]
vpodzime: Here it might be OK.

vendor/Makefile.am
```diff
@@ -0,0 +1,9 @@
+AUTOMAKE_OPTIONS = subdir-objects
+
+libboost_%.built:
+	$(SHELL) -c "cd $(srcdir)/boost; ./b2 --build-dir=build --prefix=install --with-$* link=static $(BOOST_TOOLSET) install"
+	touch libboost_$*.built
+
+clean-local:
+	$(SHELL) -c "cd $(srcdir)/boost; git clean -df"
```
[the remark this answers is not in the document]
vpodzime: The `-df` is actually on purpose here because it doesn't remove the `b2` utility which is created as part of `./autogen.sh`. `make clean` should only remove the the artifacts from `make [SOMETHING]`. So the `-xdf` variant should go into `distclean-local`.

configure.ac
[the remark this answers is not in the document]
vpodzime: > This somewhat duplicates what it is in the AM_INIT_AUTOMAKE line above though. Can they be combined?

Not really, it's a trap. The options passed to `AM_INIT_AUTOMAKE` are actually for automake. :sweat_smile: It's just GNU guys reusing things like `-Wall` and `-Werror` for other things than compilers.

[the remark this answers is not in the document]
vpodzime: > What about: https://www.gnu.org/software/autoconf-archive/ax_cxx_compile_stdcxx_11.html#ax_cxx_compile_stdcxx_11

I don't think it's worth vendoring the macros or asking people to install/download them before running `./autogen.sh`.

vpodzime:
> Does it work for you @vpodzime ?

It does now that I fixed it. :grin:  Good catch!

Makefile.am
vpodzime: Yes, but we might as well have it right in the git history. :+1:

.gitignore
```diff
@@ -50,3 +50,6 @@
 
 # Test reports
 /reports
+
+# Misc
+/.clang-format
```
oleorhagen, line 55: Hmm, should we not add it to the project instead? :thinking:

oleorhagen (approved):
Nice, seems to work for me :tada:

common/json/impl/boost/boost_json.hpp
oleorhagen: @vpodzime should we not fix this?

configure.ac
```diff
@@ -0,0 +1,66 @@
+m4_define([mender_version], m4_normalize(m4_esyscmd([git describe --exact HEAD 2>/dev/null || git rev-parse --short HEAD])))
+AC_INIT([mender], [mender_version], [contact@mender.io])
```
oleorhagen, line 2: Oh, forgot. Should we make it:
```
AC_INIT([mender-client],...)
```
Or something other than just mender?

@kacf @vpodzime

configure.ac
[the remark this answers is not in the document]
vpodzime: We don't. In the end I didn't use it because it didn't bring anything worth the vendoring/requiring struggle.

configure.ac
vpodzime: `mender` for now, we will see how this is handled in CMake and where/how far it actually propagates.

mender-update/main.cpp
vpodzime: Decision was to use the [Qt approach](https://github.com/qt/qtbase/blob/1bc7e9e77b2d6b03b995f376b622d4c8c97dd51c/src/corelib/io/qfilesystemengine_p.h), but we should avoid having platform-specific **public** class members.

PR #1092, opened 2022-12-06
chore: Update contribution guide with C++ programming style.

kacf:
Ticket: MEN-6032

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -47,6 +47,130 @@ Using commit signoffs tags is mandatory for all commits; we also
 encourage that each commit is small and cohesive. See the next sections for details.
 
 
+### Programming style
+
+#### Organization
+
+The code is organized into shared and platform specific folders. The shared code
+contains application logic which does not depend on platform specific code, as
+well as C++ interface classes for code that does. All code that interacts with
+the operating system, such as file access, network operations, timers, process
+creation, etc, must happen through such as C++ interface.
+
+The interfaces must be organized in a `common/<MODULE>/<MODULE>.hpp`
+hierarchy. An example of a `<MODULE>` name is `json`, for doing JSON
+parsing. The platform implementation must reside in `common/<MODULE>/impl/*.cpp`
+files. It is allowed to use more files in addition to the `<MODULE>` named
+files, if necessary.
+
+All the API must be namespaced inside the `<MODULE>` name. Avoid C macros if
+possible, since they can't be namespaced.
+
+#### Code style
+
+Our code style follows [the Google C++ Style
+Guide](https://google.github.io/styleguide/cppguide.html) with a few
+exceptions. To run automatic code style formatting on files you modify, use our
+[clang-format
+template](https://github.com/mendersoftware/mendertesting/blob/master/.clang-format):
+
+```bash
+curl -O https://github.com/mendersoftware/mendertesting/raw/master/.clang-format
+clang-format -i --style=file:.clang-format <MODIFIED_FILES>
+```
+
+Note that the template requires clang-format version 15 or higher.
+
+##### C++ standard
+
+All shared code must use C++ features from no later than the C++11 standard,
+with a few exceptions:
+
+* Platform code for POSIX platforms is allowed to use C++17 if necessary, but
+  C++11 is preferred for consistency with the rest of the code.
+
+* It is allowed (and encouraged) to use `std::make_unique` from C++14. We will
+  have special arrangements in the code to import this from Boost on platforms
+  where it's not available.
+
+#### File names
+
+Files must be named with `.cpp` and `.hpp` extensions. Only use `.h` if the
+header holds C declarations.
+
+##### Indentation
+
+Indentation in Mender code uses 1 tab per level, no spaces. Function arguments
+or initializers that are broken over multiple lines are indented one level. For
+example:
+
+```
+MyClass::MyFunc(
+	std::string long_argument1,
+	std::string long_argument2,
+	std::string long_argument3,
+	std::string long_argument4) :
+	member1(0),
+	member2(0) {
+	CallAnotherLongFunction(
+		long_argument1,
+		long_argument2,
+		long_argument3,
+		long_argument4);
+}
+```
+
+##### Line length
+
+Our line lengths are capped at 100 characters instead of 80, which is Google's
+cap. When considering line lengths, each tab is considered 4 characters
+wide. [The other guidelines for line
+length](https://google.github.io/styleguide/cppguide.html#Line_Length) still
+apply.
+
+##### Namespaces
+
+All the [rules about
+namespaces](https://google.github.io/styleguide/cppguide.html#Namespaces) apply,
+with two exceptions. All namespaces must be spelled out where they are used
+except for:
+
+* It is allowed to use `using namespace std` to import the `std` namespace.
+* It is allowed to shorten long names using `namespace short_name = long_name`.
+
+
+##### C headers
+
+In general, C++ libraries and wrappers should be preferred over C
+equivalents. However, it is permitted to include C headers if no good
+alternative is available. This must only be done inside platform code.
+
+If including headers from the Standard C Library, use this form:
+
+```
+#include <cstring>  // Good
+#include <string.h> // Bad, should not be used in C++ code.
+```
+
+##### Code blocks
+
+All code blocks occurring as part of `if` statements, `for` loops or `while`
+loops, must be enclosed in curly brackets, even single line blocks.
+
+##### Class access specifiers
+
+Class access specifiers, `public`, `protected` and `private` must be indented at
+the same level as their parent class, in other words at the same level as where
+the `class` keyword appears. For example:
+
+```
+class MyClass {
+public:
+	MyClass();
+}
+```
+
+
 ### Sign your work
 
 Mender is licensed under the Apache License, Version 2.0. To ensure open source


CONTRIBUTING.md
```diff
@@ -47,6 +47,92 @@ Using commit signoffs tags is mandatory for all commits; we also
...
+or initializers that are broken over multiple lines are indented one level.
+
+##### Namespaces
+
+All the [rules about
+namespaces](https://google.github.io/styleguide/cppguide.html#Namespaces) apply,
+with one exception: It is allowed to use `using namespace std` to import the
+`std` namespace. All other namespaces must be spelled out where they are used
+(though it is allowed to shorten long names using `namespace short_name =
+long_name`).
```
kacf, line 109: We didn't discuss the second point explicitly (about short names). But I figured this mirrors what is common to do in the Golang world, and makes sense to allow, or?

kacf (commented):
This is essentially a verbal version of the [the style template](https://github.com/mendersoftware/mendertesting/blob/master/.clang-format).

Some open questions:
* Should it be in the contribution guide, or is it getting too long and should be in its own file?
* Combining CFEngine and Mender guides is probably not a good idea, because the CFEngine guide has a lot of C specific stuff which is not relevant in C++. But that doesn't mean we couldn't move it to a more central place though. Traditionally we have kept in the mender repo because it is the most visible.
* More below

CONTRIBUTING.md
```diff
@@ -47,6 +47,92 @@ Using commit signoffs tags is mandatory for all commits; we also
...
+```bash
+curl -O https://github.com/mendersoftware/mendertesting/raw/master/.clang-format
+clang-format -i --style=file:.clang-format <MODIFIED_FILES>
+```
+
+Note that the template requires clang-format version 15 or higher.
+
+##### C++ standard
+
+All shared code must use C++ features from no later than the C++11
```
TheYoctoJester, line 81: So it is explicitly 11, not 14? Because 14 essentially only was a clean-up release, whereas 17 really brought new stuff.

CONTRIBUTING.md
```diff
@@ -47,6 +47,92 @@ Using commit signoffs tags is mandatory for all commits; we also
...
+operations, timers, process creation, etc, must happen through a C++
+interface. These interfaces must be organized in a
+`common/<MODULE>/<MODULE>.hpp` hierarchy. The platform implementation must
+reside in `common/<MODULE>/impl/<MODULE>.cpp` files. It is allowed to use more
+files in addition to the `<MODULE>` named files, if necessary.
+
+All the API must be namespaced inside the `<MODULE>` name. Avoid C macros if
+possible, since they can't be namespaced.
+
+#### Code style
```
TheYoctoJester, line 64: I personally would prefer to have this as the first paragraph.

CONTRIBUTING.md
```diff
@@ -47,6 +47,92 @@ Using commit signoffs tags is mandatory for all commits; we also
...
+##### Line length
+
+Our line lengths are capped at 100 characters instead of 80, which is Google's
+cap. [The other guidelines for line
+length](https://google.github.io/styleguide/cppguide.html#Line_Length) still
+apply.
+
+##### Indentation
+
+Indentation in Mender code uses 1 tab per level, no spaces. Function arguments
```
TheYoctoJester, line 99: Intentionally not mentioning a preferred tab width?

TheYoctoJester (approved):
Minor thoughts, but fine with me.

CONTRIBUTING.md
kacf: Ok, I moved it.

CONTRIBUTING.md
```diff
@@ -47,6 +47,93 @@ Using commit signoffs tags is mandatory for all commits; we also
...
+
+Our code style follows [the Google C++ Style
+Guide](https://google.github.io/styleguide/cppguide.html) with a few
+exceptions. To run automatic code style formatting on files you modify, use our
+[clang-format
+template](https://github.com/mendersoftware/mendertesting/blob/master/.clang-format):
+
+```bash
+curl -O https://github.com/mendersoftware/mendertesting/raw/master/.clang-format
+clang-format -i --style=file:.clang-format <MODIFIED_FILES>
```
vpodzime, line 62: Looks like a good candidate for a git hook or even multiple possible hooks (pre-commit, pre-push).

CONTRIBUTING.md
```diff
@@ -47,6 +47,93 @@ Using commit signoffs tags is mandatory for all commits; we also
...
+
+#### File names
+
+Files must be named with `.cpp` and `.hpp` extensions. Only use `.h` if the
+header holds C declarations.
+
+##### Indentation
+
+Indentation in Mender code uses 1 tab per level, no spaces. Function arguments
+or initializers that are broken over multiple lines are indented one level.
```
vpodzime, line 81: I think a clarification of how nested arguments should look is needed here. It's still not 100% clear to me so I guess other people may have troubles here too.

CONTRIBUTING.md
vpodzime: Now that I've seen some of the Boost namespaces, I'm all for this.

CONTRIBUTING.md
```diff
@@ -47,6 +47,93 @@ Using commit signoffs tags is mandatory for all commits; we also
...
+with one exception: It is allowed to use `using namespace std` to import the
+`std` namespace. All other namespaces must be spelled out where they are used
+(though it is allowed to shorten long names using `namespace short_name =
+long_name`).
+
+##### C headers
+
+In general, C++ libraries and wrappers should be preferred over C
+equivalents. However, it is permitted to include C headers if no good
+alternative is available. This must only be done inside platform code.
```
vpodzime, line 104: This and also the text above mentions _platform code_. Do we have an explanation of what it means somewhere?

CONTRIBUTING.md
```diff
@@ -47,6 +47,93 @@ Using commit signoffs tags is mandatory for all commits; we also
...
+the `class` keyword appears.
+
+#### Organization
+
+All code that interacts with the operating system, such as file access, network
+operations, timers, process creation, etc, must happen through a C++
+interface. These interfaces must be organized in a
+`common/<MODULE>/<MODULE>.hpp` hierarchy. The platform implementation must
+reside in `common/<MODULE>/impl/<MODULE>.cpp` files. It is allowed to use more
+files in addition to the `<MODULE>` named files, if necessary.
```
vpodzime, line 131: What is an example of `<MODULE>` here? And shouldn't the implementations be in `common/<MODULE>/impl/<PLATFORM>.cpp` or something like that?

CONTRIBUTING.md
lluiscampos: @kacf One thing that I discovered recently is that `make_unique` is C++14 only (ref: https://en.cppreference.com/w/cpp/memory/unique_ptr/make_unique ). This might be important because without the helper we'll need to implement constructors in our classes specifically for `unique_ptr<T>`.

CONTRIBUTING.md
```diff
@@ -47,6 +47,93 @@ Using commit signoffs tags is mandatory for all commits; we also
...
+apply.
+
+##### Namespaces
+
+All the [rules about
+namespaces](https://google.github.io/styleguide/cppguide.html#Namespaces) apply,
+with one exception: It is allowed to use `using namespace std` to import the
+`std` namespace. All other namespaces must be spelled out where they are used
+(though it is allowed to shorten long names using `namespace short_name =
+long_name`).
```
lluiscampos, line 98: It reads to me like two exceptions instead of one: `namespace std` and `short_name = long_name`. I suggest we rewrite it to be more explicit.

For example:
All the rules about namespaces apply, with two exceptions. All namespaces must be spelled out where they are used except for:
* It is allowed to use `using namespace std` to import the `std` namespace
* it is allowed to shorten long names using `namespace short_name = long_name`

lluiscampos (approved):

CONTRIBUTING.md
kacf: Hmm, do you have an example of where this is needed? As far as I can figure, there are two main reasons it is needed:

1. When exceptions are enabled (which they won't be), and more than one function is being evaluated in the statement (see [this thread](https://stackoverflow.com/questions/53870522/why-use-stdmake-unique-in-c17) for a good explanation). In C++17 and later, they clarified the evaluation order, and this reason is moot from then onwards.
2. When passing arrays to the `unique_ptr` constructor, which it doesn't support, and erroneously accepts. I agree it would be nice to have the better type safety that `make_unique` provides in this context. But I think raw arrays will probably be rare, because anywhere I see them I would encourage people to use another `std` structure instead, like `vector`. Main reason is that you can enable range checking for them, which you can't for arrays.

CONTRIBUTING.md
kacf: Good idea. I'll follow this up separately though.

CONTRIBUTING.md
kacf: You're right, examples are probably warranted here, I'll add some.

CONTRIBUTING.md
kacf: Lol, yes, I thought of the second only later, of course.

CONTRIBUTING.md
kacf: It's mentioned under "Organization", but I agree it's not super clear. I'll try to open that paragraph by declaring what shared and platform code sections are.

@TheYoctoJester wanted "Organization" to be below the code style, but this is actually a good reason to put it back on top, so that you read that part first.

CONTRIBUTING.md
kacf: For this part I just followed what was done in https://github.com/mendersoftware/mender/pull/1091, so this is open to change. Let's settle it there first, and return here after.

CONTRIBUTING.md
kacf: Added some here and under "access specifiers".

CONTRIBUTING.md
```diff
@@ -47,6 +47,130 @@ Using commit signoffs tags is mandatory for all commits; we also
...
+example:
+
+```
+MyClass::MyFunc(
+	std::string long_argument1,
+	std::string long_argument2,
+	std::string long_argument3,
+	std::string long_argument4)
+	: member1(0),
+	  member2(0) {
```
kacf, line 107: This is not really my suggestion, but is how clang-format does it with the current settings. May I suggest enabling `BreakConstructorInitializers: BCIS_AfterColon`, which would result in this:

```
MyClass::MyFunc(
	std::string long_argument1,
	std::string long_argument2,
	std::string long_argument3,
	std::string long_argument4) :
	member1(0),
	member2(0) {
```

Another option would be `BreakConstructorInitializers: BCIS_BeforeComma`, which would result in 

```
MyClass::MyFunc(
	std::string long_argument1,
	std::string long_argument2,
	std::string long_argument3,
	std::string long_argument4)
	: member1(0)
	, member2(0) {
```

but I think I vote for the first one (and I will choose it if no one says anything).

vpodzime: Same here, I also vote for the first one.

kacf: https://github.com/mendersoftware/mendertesting/pull/233

Also updated example here.

CONTRIBUTING.md
vpodzime: Are the Boost alternatives nicer?

CONTRIBUTING.md
```diff
@@ -47,6 +47,130 @@ Using commit signoffs tags is mandatory for all commits; we also
...
+apply.
+
+##### Namespaces
+
+All the [rules about
+namespaces](https://google.github.io/styleguide/cppguide.html#Namespaces) apply,
+with one exception: It is allowed to use `using namespace std` to import the
+`std` namespace. All other namespaces must be spelled out where they are used
+(though it is allowed to shorten long names using `namespace short_name =
+long_name`).
```
vpodzime, line 131: This paragraph is a duplicate of the (better) one below.

vpodzime (commented):
Looks good to me otherwise.

CONTRIBUTING.md
kacf: It looks like the Boost alternative is there exactly to fill the gap between C++11 and C++14, but is otherwise the same. So this we could indeed use.

The only (slight) disadvantage is that for historical reasons, their `make_unique` returns a `std::unique_ptr`, whereas their `make_shared` returns `boost::shared_ptr`. IOW we should not mix those two.

If we do use it, I would propose making a special exception for it:

```
using boost::make_unique;
```

Which gives several advantages:

* It means we can use it with its plain name, instead of prefixing it with `boost::`. This will make it much easier to transition to `std::make_unique` later once we move beyond C++11.
* We won't accidentally use `boost::make_shared`.

If that is acceptable I will add this to the style guide.

CONTRIBUTING.md
kacf: Oops, good eyes! I meant to remove that, obviously. Updated locally, will push here soon.

CONTRIBUTING.md
kacf: Updated and made an example of what `<MODULE>` can be.

While I agree we might need `<PLATFORM>.cpp` later, I'm leaving this one for later. It won't really come up until Q2 or Q3 next year, at which point we probably need to change many other things we hadn't thought of as well.

PR #1100, opened 2022-12-15
Remove vendored Boost and add Nlohmann JSON

oleorhagen:
* Removes boost vendor dir
* Add nlohmann/json as a submodule (header only json lib)
* Integrate our common directory with the nlohmann/json library

Note: I have not removed the m4 boost files. We will need them sooner or later. If you think this is dirty, speak up.


.gitmodules
```diff
@@ -2,6 +2,6 @@
 	path = vendor/googletest
 	url = https://github.com/google/googletest.git
 	branch = v1.12.x
-[submodule "vendor/boost"]
-	path = vendor/boost
-	url = https://github.com/boostorg/boost.git
+[submodule "vendor/json"]
```
oleorhagen, line 5: I named it vendor/json... Not sure if this is completely kosher or not.

configure.ac
```diff
@@ -18,38 +18,16 @@ dnl (Make these variables available for use in Makefile.am)
...
+dnl TODO: Check which is the actual minimum we support!
+AX_BOOST_BASE([1.74.0],
+    [AC_SUBST([boost_CPPFLAGS], [${BOOST_CPPFLAGS}])
+        AC_MSG_NOTICE([boost_CPPFLAGS : ${boost_CPPFLAGS}])
+        AC_MSG_NOTICE([boost_LDFLAGS : ${boost_LDFLAGS}])],
+    [AC_MSG_ERROR([Boost 1.74.0 or later is required but was not found.])])
+
+AX_BOOST_JSON
+AC_SUBST([boost_json_LIBS], [${BOOST_JSON_LIB}])
+AC_MSG_NOTICE([boost_json_LIBS : ${boost_json_LIBS}])
```
vpodzime, line 30: Revert (`reset -p`?) artifact here.

oleorhagen: Hmm, I'm not following. What do you mean?

vpodzime: Oh, I thought you were removing it completely. But you are, **properly**, first just making it not vendored. :+1:

common/json/impl/nlohmann/nlohmann_json.cpp
```diff
@@ -0,0 +1,33 @@
...
+#include <nlohmann/json.hpp>
+using njson = nlohmann::json;
+
+#include <common/json/impl/nlohmann/nlohmann_json.hpp>
+
+namespace json {
+
+void NlohmannJson::hello_world() {
+	std::ifstream f("example.json");
+	njson data = njson::parse(f);
```
vpodzime, line 26: Could we use `istream` instead? Otherwise we need the file with proper content.

vpodzime (commented):
Looks good to me otherwise.

configure.ac
oleorhagen: Yes, I wanted to have every commit be functional still :smiley_cat:

common/json/impl/nlohmann/nlohmann_json.cpp
oleorhagen: Done!

vpodzime (approved):
ACK

PR #1103, opened 2022-12-16
Rename the device binaries and services to the agreed upon names

oleorhagen:
Simply renames the client binaries to:

* mender-auth
* mender-update

And creates the systemd service files:

* mender-authd
* mender-updated

As well as removing the old `mender-client.service` file.


vpodzime (approved):
The two commits here look good.

PR #1104, opened 2022-12-19
build: automakify the rest of the root Makefile.am

oleorhagen:
This just solves all the DESTDIR, prefix, stuff automatically, and also
symmetrically mirrors it all for install and uninstall.

Just makes life easier.

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -3,8 +3,6 @@ ACLOCAL_AMFLAGS = -I m4
 SUBDIRS = common mender-auth mender-update
 
 systemd_unitdir ?= /lib/systemd
-datadir ?= /usr/share
-docexamplesdir ?= /usr/share/doc/mender-client/examples
 
 check-local: test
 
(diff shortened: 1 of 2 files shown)


lluiscampos (commented):
If any, this should be part of the `configure.ac`, right? Also see [this note](https://github.com/mendersoftware/mender/pull/1104/files#diff-0462e381b2fb3286568215681c8983490a37ac9ae0f0c5ee304df7fa6426d4afR44) hinting that we will remove all it anyway.

oleorhagen:
> If any, this should be part of the `configure.ac`, right? Also see [this note](https://github.com/mendersoftware/mender/pull/1104/files#diff-0462e381b2fb3286568215681c8983490a37ac9ae0f0c5ee304df7fa6426d4afR44) hinting that we will remove all it anyway.

Yes, you are right. And it will all be removed anyway. But this was the easy approach, and I need it for the Yocto recipes atm, and it is playing a lot nicer with the autotools recipe than our hand-written stuff.

Basically this is a no-op.

lluiscampos:
> > If any, this should be part of the `configure.ac`, right? Also see [this note](https://github.com/mendersoftware/mender/pull/1104/files#diff-0462e381b2fb3286568215681c8983490a37ac9ae0f0c5ee304df7fa6426d4afR44) hinting that we will remove all it anyway.
> 
> Yes, you are right. And it will all be removed anyway. But this was the easy approach, and I need it for the Yocto recipes atm, and it is playing a lot nicer with the autotools recipe than our hand-written stuff.
> 
> Basically this is a no-op.

As long as we keep that note in that we will eventually clean up the file (including the changes from this PR), then I am okey with it :+1:

lluiscampos (approved):

oleorhagen:
> > > If any, this should be part of the `configure.ac`, right? Also see [this note](https://github.com/mendersoftware/mender/pull/1104/files#diff-0462e381b2fb3286568215681c8983490a37ac9ae0f0c5ee304df7fa6426d4afR44) hinting that we will remove all it anyway.
> > 
> > 
> > Yes, you are right. And it will all be removed anyway. But this was the easy approach, and I need it for the Yocto recipes atm, and it is playing a lot nicer with the autotools recipe than our hand-written stuff.
> > Basically this is a no-op.
> 
> As long as we keep that note in that we will eventually clean up the file (including the changes from this PR), then I am okey with it +1

Yeh, it will happen no matter as soon as we move to CMake

oleorhagen:
No need, that that CMake is just around the corner

PR #1106, opened 2022-12-20
feat: do not check if the artifact is already installed

tranchitella:
The Mender Client checked if the artifact is already installed, comparing the Artifact's name of the update with the one of the latest installed on the device. However, since the introduction of Update Modules and application updates, this check is weak because it relies on the latest installed artifact only. We want to move all the `already-installed` logic to the backend. Therefore, we remove this redundant check from the client and rely on the Mender Server only.

Changelog: Title
Ticket: MEN-6129

BREAKING-CHANGE: the Mender Client won't reject updates in case the Artifact's name is the same as the latest one installed

the form below:

```

<OPTIONAL LONGER DESCRIPTION>

```

### Description

Thank you!


tranchitella:

oleorhagen (commented):
I guess this needs a minor discussion on if this means the c++ client is now `5.0`.

Otherwise looks good.

I would separate out the header fixes into a separate commit though :pray:

tranchitella:
> I guess this needs a minor discussion on if this means the c++ client is now `5.0`.
> Otherwise looks good.
> I would separate out the header fixes into a separate commit though 🙏

Done, I separated the header fixes into a dedicated commit.
I am not 100% sure we should bump the major version number (4.0) for this change.
We didn't do it for the update control maps or other client features, why should we do it in this case?
Please note that in managed mode, the Server will enforce the `already-installed` status anyway and no change will apply to the users' flow.

oleorhagen (approved):

tranchitella:

lluiscampos (approved):

kacf:
FWIW, I agree that with the combined behavior of the server and the client, there is no need for a breaking change here. In fact, this check may have never triggered except in our tests, since the server always has done the same check.

PR #1107, opened 2022-12-21
fix: Rework handling of platform-dependent implementations

vpodzime:
The goal is to have the platform-dependent code isolated with the internal interface, the 'common_' libs and headers, encapsulating the implementation details.

That can be achieved by:

  - header files in the `common` directory declare the classes and functions for the common interfaces,

  - separate implementations provide definitions for the classes and functions declared in the common headers,

  - build system chooses the right implementation for each common interface and compiles the common interface library accordingly and

  - #ifdefs are used in the common interface header files if the individual implementations need specific **private** methods or attributes.

With the above, the code that uses a common interface only links to the particular `common_IFACE` library and includes the `common/IFACE.hpp` header.

Ticket: MEN-6112
Changelog: None


oleorhagen (approved):
Makes sense to me.

But can we not remove the `impl` dir also?

vpodzime:
> But can we not remove the `impl` dir also?

We surely can and I would do it. But we didn't discuss that part, IIRC.

common/CMakeLists.txt
```diff
@@ -1,12 +1,19 @@
-add_library(common_json STATIC json/impl/nlohmann/nlohmann_json.cpp)
-target_include_directories(common_json PRIVATE ../vendor/json/include/)
+set(json_sources "$<$<STREQUAL:${PLATFORM},linux_x86>:nlohmann/nlohmann_json.cpp>")
+set(kv_db_sources "$<$<STREQUAL:${PLATFORM},linux_x86>:lmdb/lmdb_kv_db.cpp>")
```
kacf, line 2: CMake syntax is so beautiful! 😆

common/CMakeLists.txt
```diff
@@ -1,12 +1,19 @@
...
+
+add_library(common_json STATIC json/impl/${json_sources})
 target_include_directories(common_json PRIVATE ../)
+if(${json_sources} MATCHES ".*nlohman.*")
```
kacf, line 6: Nitpick: nlohmann has two n's.

kacf (approved):
Nice one!

common/CMakeLists.txt
vpodzime: Oh, right! This has bitten my ass multiple times already.

kacf (approved):

oleorhagen (approved):

PR #1109, opened 2022-12-23
build: Require C++11

oleorhagen:

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -12,6 +12,9 @@ endif (POLICY CMP0135)
 
 project(mender)
 
+set(CMAKE_CXX_STANDARD 11)
+set(CMAKE_CXX_STANDARD_REQUIRED True)
+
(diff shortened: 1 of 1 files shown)


CMakeLists.txt
```diff
@@ -12,6 +12,9 @@ endif (POLICY CMP0135)
 
 project(mender)
 
+set(CMAKE_CXX_STANDARD 11)
+set(CMAKE_CXX_STANDARD_REQUIRED True)
```
vpodzime, line 16: AFAICT, I think we don't to do this globally.

oleorhagen: what do you mean? That we shouldnt do it globally.
that this does not change it globally?

vpodzime: ooops, dropped some words. I was trying to say that we don't **want** to do this globally :sweat_smile:

PR #1112, opened 2022-12-28
Make the CMake build work with our Yocto recipes

oleorhagen:
Goes along with: https://github.com/mendersoftware/meta-mender/pull/1871

* Conditionally download googletest
* Fall back to find_library in case pkg-config fails
* fix the install of the systemd service files (should ignore <prefix>)
* Move to respect the GNUInstallDirs variables


kacf (approved):
LGTM, but I'm a noob in CMake. 🙂

oleorhagen:
> LGTM, but I'm a noob in CMake. slightly_smiling_face

As am I :p

PR #1113, opened 2023-01-03
chore: Respect `prefix` in Makefile

lluiscampos:
Previously, the `Makefile` was miss-using `DESTDIR` as an override to `prefix`. Instead, the latter should only be used for setting `bindir` and `datadir` while the former should be used elsewhere in the install targets.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,10 +1,10 @@
 DESTDIR ?= /
-prefix ?= $(DESTDIR)
-bindir=/usr/bin
-datadir ?= /usr/share
+prefix ?= /usr/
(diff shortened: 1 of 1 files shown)


oleorhagen (approved):

oleorhagen:
Thanks! looks good.

lluiscampos:
> Thanks! looks good.

Thank you.

The short term decision is to revert:
* https://github.com/mendersoftware/integration/pull/2297

So I'll re-take this PR when working on [QA-521](https://tracker.mender.io/browse/QA-521).

PR #1119, opened 2023-01-09
MEN-6127: Event loop API and coverage fixes

kacf:


kacf:
Once again, coveralls is useless. Those lines *are* covered.

common/events.hpp
```diff
@@ -0,0 +1,61 @@
...
+#include <functional>
+#include <system_error>
+
+typedef std::function<void(std::error_code)> EventHandler;
+
+#if EVENTS_USE_BOOST
+#include <boost/asio.hpp>
+#endif // EVENTS_USE_BOOST
+
+namespace mender {
```
oleorhagen, line 24: S: `mender::events`

common/events.hpp
```diff
@@ -0,0 +1,61 @@
...
+namespace events {
+
+#if EVENTS_USE_BOOST
```
oleorhagen, line 27: S: `MENDER_EVENTS_USE_BOOST`

CMakeLists.txt
```diff
@@ -15,6 +15,10 @@ project(mender)
 # set(CMAKE_VERBOSE_MAKEFILE ON)
 enable_testing()
 
+option(COVERAGE "Turn coverage instrumentation on (Default: OFF)" OFF)
+if($CACHE{COVERAGE})
+  set(CMAKE_CXX_FLAGS "--coverage $CACHE{CMAKE_CXX_FLAGS}")
```
vpodzime, line 20: Isn't `--coverage` a linker flag as well?

vpodzime (commented):
Looks good to me otherwise.

CMakeLists.txt
```diff
@@ -18,7 +18,7 @@ enable_testing()
 
 set(GTEST_VERSION 1.12.1)
 
-option(MENDER_DOWNLOAD_GTEST "Download google test if it is not found (Default: Yes)" ON)
+option(MENDER_DOWNLOAD_GTEST "Download google test if it is not found (Default: ON)" ON)
```
oleorhagen, line 25: Did I trip you up? :p

CMakeLists.txt
```diff
@@ -52,7 +52,8 @@ add_library(mender_compiler_flags INTERFACE)
 target_compile_features(mender_compiler_flags INTERFACE cxx_std_11)
 
 add_executable(mender main.cpp lib.cpp)
-target_link_libraries(mender PUBLIC mender_compiler_flags)
+target_link_libraries(mender PUBLIC common_events mender_compiler_flags)
```
oleorhagen, line 55: I was actually thinking we should nuke the toplevel main file.

Why keep it?

common/events/platform/boost/events.cpp
```diff
@@ -0,0 +1,59 @@
...
+Timer::Timer(EventLoop &loop) :
+	timer_(loop.ctx_) {
+}
+
+template <typename Duration>
+void Timer::Wait(Duration duration) {
+	timer_.expires_after(duration);
+	timer_.wait();
+}
+template void Timer::Wait<std::chrono::milliseconds>(std::chrono::milliseconds duration);
```
oleorhagen, line 41: Q:
Why template the function definition and explicitly initiate it here, as opposed to having two overloaded functions?

common/events.hpp
```diff
@@ -0,0 +1,61 @@
...
+#include <system_error>
+
+typedef std::function<void(std::error_code)> EventHandler;
+
+#if EVENTS_USE_BOOST
+#include <boost/asio.hpp>
+#endif // EVENTS_USE_BOOST
+
+namespace mender {
+namespace events {
```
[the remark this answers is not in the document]
kacf: For a very important reason: I didn't know you could! 😃

Fixing.

CMakeLists.txt
kacf: I don't think so. It's not mentioned on the `ld` man page.

CMakeLists.txt
kacf: This is because I remember an earlier open source project I built where this was driving me crazy. I tried `true`, `TRUE`, `1`, `on`, `yes`, `YES`, and probably several others before I finally realized what it should be. Let's give the next guy the soft way, instead of the hard way. 🙂

CMakeLists.txt
kacf: I agree, but there's no rush. We can do that when we have the first pieces of an actual daemon. For now the `main.cpp` binary is still handy as a playground.

common/events.hpp
```diff
@@ -0,0 +1,61 @@
...
+//    WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+//    See the License for the specific language governing permissions and
+//    limitations under the License.
+
+#include <functional>
+#include <system_error>
+
+typedef std::function<void(std::error_code)> EventHandler;
+
+#if EVENTS_USE_BOOST
```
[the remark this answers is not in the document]
kacf: Yeah, I've seen autotools use 0 and 1 as values, and hence they use `#if`, but I also think `#ifdef` is more common in general. I'll switch, since I can't find a good reason for the `#if` approach for us.

common/events/platform/boost/events.cpp
kacf: As mentioned f2f, I was conflicted about this decision. My main motivation was to try to keep it out of the header file, and only in the source file. Which is tidy, but comes with the rather significant downside that you can only use the exact types you defined. I thought maybe this was ok for seconds and milliseconds.

But now that you have to move some of your definitions into header files as well, I will just do the same, there is no reason to keep it like this.

CMakeLists.txt
kacf: Hmm, indeed. Looking at the output from the Makefile, it looks like CMake passes the `--coverage` argument to the linker even if it is only in `CMAKE_CXX_FLAGS`. And it's accepted. That's two unexpected behaviors for the price of one, but the net result is that we don't seem to need any more flags.

kacf:
@oleorhagen, @vpodzime: Updated!

CMakeLists.txt
vpodzime: Great! And "Great..."!

vpodzime (approved):
*Practical example* :laughing: 

Looks good to me now otherwise. Let's hope Ubuntu 22.04 repositories in GitLab work now. :crossed_fingers:

PR #1122, opened 2023-01-12
Config parser

vpodzime:
Based on the internal JSON API work, only the last commit is relevant here.

PR #1124, opened 2023-01-17
Mender configuration parser

oleorhagen:
This is a follow-up PR from https://github.com/mendersoftware/mender/pull/1122


common/config_parser.hpp
```diff
@@ -0,0 +1,161 @@
...
+	/** Command to get the partition which will boot next */
+	string boot_utilities_get_next_active_part;
+
+	/** Path to the device type file */
+	string device_type_file;
+
+	/** DBus configuration */
+	// TODO: struct DbusConfig { bool enabled; }; ?
+	bool dbus_enabled = false;
+	;
```
kacf, line 92: Remove `;`?

common/config_parser/config_parser.cpp
```diff
@@ -0,0 +1,412 @@
...
+		const json::Json value_json = e_cfg_value.value();
+		const json::ExpectedInt e_cfg_int = value_json.GetInt();
+		if (e_cfg_int) {
+			this->module_timeout_seconds = e_cfg_int.value();
+			applied = true;
+		}
+	}
+
+	/* Vectors/arrays now */
+	e_cfg_value = cfg_json.Get("ArtifactVerifyKeys");
```
kacf, line 276: Remember to implement [this conditional](https://github.com/mendersoftware/mender/blob/28bc08ac0ed6816d96372d1c7ff50dc8d03f7a1d/conf/config.go#L265-L268).

common/config_parser.hpp
```diff
@@ -0,0 +1,161 @@
...
+public:
+	/** Path to the public key used to verify signed updates.
+		Only one of artifact_verify_key/artifact_verify_keys can be specified. */
+	string artifact_verify_key;
+
+	/** List of verification keys for verifying signed updates.
+		Starting in order from the first key in the list,
+		each key will try to verify the artifact until one succeeds.
+		Only one of artifact_verify_key/artifact_verify_keys can be specified. */
+	vector<string> artifact_verify_keys;
```
kacf, line 66: In Golang we had to keep both members because the struct is directly used in the parsing. However, in C++ I'm wondering if we should omit the singleton and use only the list. We can just populate it with one element if the singleton exists in the conf file, and it should make the rest of the logic simpler.

common/config_parser.hpp
```diff
@@ -0,0 +1,161 @@
...
+
+	/** Path to deployment log file */
+	string update_log_path;
+
+	/** Server JWT TenantToken */
+	string tenant_token;
+
+	/** List of available servers, to which client can fall over */
+	// TODO: struct MenderServer {string url;}; ?
+	vector<string> servers;
```
kacf, line 140: Same here with regards to keeping only `servers` and dropping `server_url`.

common/config_parser/config_parser.cpp
```diff
@@ -0,0 +1,412 @@
...
+							this->artifact_verify_keys.push_back(item_value);
+							applied = true;
+						}
+					}
+				}
+			}
+		}
+	}
+
+	e_cfg_value = cfg_json.Get("Servers");
```
kacf, line 301: Remember to implement [this conditional](https://github.com/mendersoftware/mender/blob/28bc08ac0ed6816d96372d1c7ff50dc8d03f7a1d/conf/config.go#L185-L197).

common/CMakeLists.txt
```diff
@@ -60,11 +60,11 @@ add_dependencies(check log_test)
 add_library(common_config_parser STATIC config_parser/config_parser.cpp)
 target_include_directories(common_config_parser PRIVATE ../ ${CMAKE_BINARY_DIR})
 if(${json_sources} MATCHES ".*nlohmann.*")
-  target_include_directories(common_config_parser PUBLIC ../vendor/json/include/)
+  target_include_directories(common_config_parser PUBLIC ${CMAKE_SOURCE_DIR}/vendor/json/include/)
```
kacf, line 63: 👍

kacf (approved):
Nice!

A couple of comments below, but nothing that requires re-reviewing.

common/config_parser.hpp
vpodzime: To me it makes more sense to have them separate in this class because then it reflects what the file contained. `MenderConfig` could have them merged, but then it couldn't be based on this (`MenderConfigFromFile`) class which would be quite unfortunate.

PR #1127, opened 2023-01-19
Add the in-memory key-value db interface

oleorhagen:
Still a bit rough around the edges, but I think the direction should be pretty clear.

It still does not lock the DB read-only - Initially I thought marking the functions const would be enough, but it seems I am wrong (please prove me wrong here). If not, then I will add  an explicit implementation of this.

It stores the data as a vector of bytes, but all the interfaces are still string, it was just easier to with the implementation initially this way, I'm guessing overloads is the way to go here, or what do you think? Also casting back to strings is a pain!


CMakeLists.txt
```diff
@@ -51,6 +51,10 @@ set(PLATFORM linux_x86)
 
 option(MENDER_LOG_BOOST "Use Boost as the underlying logging library provider (Default: ON)" ON)
 
+# TODO - set LMDB as the default
```
oleorhagen, line 54: Just easier now whilst I'm developing :)

common/CMakeLists.txt
```diff
@@ -14,24 +14,45 @@ target_include_directories(json_test PRIVATE ../)
...
 
+# Test MenderLog
+add_executable(kv_db_test EXCLUDE_FROM_ALL kv_db_test.cpp)
+target_link_libraries(kv_db_test PRIVATE common_kv_db GTest::gtest_main gmock)
+target_include_directories(kv_db_test PRIVATE ../)
+gtest_discover_tests(kv_db_test)
+add_dependencies(check kv_db_test)
+
 add_library(common_events STATIC events/platform/${events_sources})
+target_include_directories(common_events PUBLIC /home/olepor/scratchpad/boost/install/include/)
```
oleorhagen, line 55: Whops. Removing locally

common/kv_db/platform/inmem/inmem.cpp
```diff
@@ -0,0 +1,73 @@
...
+	return ExpectedBool(true);
+}
+
+const ExpectedEntry KVDB::OpenRead(const string &key) const {
+	if (auto search = map_.find(key); search != map_.end()) {
+		std::string string1(
+			reinterpret_cast<const char *>(&search->second[0]), search->second.size());
+		return ExpectedEntry(DBEntry {search->second});
+	}
+	// TODO - This should mirror the I/O errors in mender-go
```
oleorhagen, line 53: I'm not fully sure what to do about the errors, should we create some global ones, or should we have some sort of error  creator?

common/kv_db/platform/inmem/inmem.cpp
```diff
@@ -0,0 +1,73 @@
...
+	KVDBError err = {KVDBErrorCode::KeyError, "Key Not found!"};
+	return ExpectedEntry(err);
+}
+
+DBEntry KVDB::OpenWrite(const string &key) {
+	// TODO - Make sure we don't do this on a read-only db
```
oleorhagen, line 59: Mentioned in the PR comment. :point_up:

common/kv_db/platform/inmem/inmem.cpp
```diff
@@ -0,0 +1,73 @@
...
+	return DBEntry {key};
+}
+
+ExpectedBool KVDB::Close() const {
+	return ExpectedBool(true);
```
oleorhagen, line 64: Do you want this mocked?

common/kv_db_test.cpp
```diff
@@ -0,0 +1,128 @@
...
+	{
+		// Write
+		auto ret = db_.WriteAll("key", "val");
+		EXPECT_TRUE(ret);
+	}
+
+	{
+		// Read
+		auto entry = db_.ReadAll("key").value().Read();
+		std::string string1(reinterpret_cast<const char *>(&entry[0]), entry.size());
```
oleorhagen, line 38: I really need some util interface for working with this...

kacf (changes requested):
I started looking at the main API, but I think we should make some bigger changes there, so I did not comment on the specifics. I meant it quite literally when I said we should copy the Golang API, which means using interface types in the function arguments. This also means we need to start using shared pointers in the API, since you can't pass interface types by value.

But instead of trying to explain it here, should I take it over, try to shape the API in the described way and we can rather go through it afterwards as a peer-to-peer review and learning exercise?

CMakeLists.txt
```diff
@@ -51,6 +51,10 @@ set(PLATFORM linux_x86)
...
+option(MENDER_DB_LMDB "Use LMDB as the key-value store (Default: ON)" OFF)
+option(MENDER_DB_IN_MEM "Use an in-memory key-value store (Default: OFF)" ON)
```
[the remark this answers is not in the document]
oleorhagen: Makes sense

common/kv_db.hpp
```diff
@@ -12,16 +12,97 @@
...
+
+	std::vector<std::byte> Read() {
+		return buf;
+	}
+};
+
+using ExpectedEntry = expected::Expected<DBEntry, KVDBError>;
+using ExpectedBool = expected::Expected<bool, KVDBError>;
+
+class KVDB {
```
kacf, line 73: @oleorhagen: Another RFC: I think following [this principle](https://google.github.io/styleguide/cppguide.html#General_Naming_Rules) we should spell out this name rather than abbreviate it. Agree?

I like the name better than the original `Store` which was used in Golang though.

oleorhagen: Yes, agree completely!

PR #1130, opened 2023-01-20
KeyValueDataBase API

kacf:
Very early draft of how the API might look.


common/io.hpp
```diff
@@ -0,0 +1,58 @@
...
+public:
+	virtual ~Closer() {};
+
+	virtual error_condition Close() = 0;
+};
+
+class ReadWriter : virtual public Reader, virtual public Writer {};
+class ReadCloser : virtual public Reader, virtual public Closer {};
+class WriteCloser : virtual public Reader, virtual public Closer {};
+class ReadWriteCloser : virtual public ReadWriter, virtual public ReadCloser, virtual public WriteCloser {};
```
kacf, line 54: So far I've only used `Closer` from here, but I think it makes sense to have all, for later use.

common/io.hpp
```diff
@@ -0,0 +1,58 @@
...
+
+using namespace std;
+
+using mender::common::expected::Expected;
+
+class Reader {
+public:
+	virtual ~Reader() {};
+
+	virtual Expected<int64_t, error_condition> Read(vector<byte> &dst) = 0;
```
kacf, line 34: RFC about the error being returned: In `std` the only "official" error types are `error_condition` (platform independent) and `error_code` (platform dependent). The former comes with cross platform error codes so one can do things like `if (error.value() == no_such_file_or_directory)` and such, so it seems better for portability.

Then we have our own error type [here](https://github.com/mendersoftware/mender/blob/feature-c%2B%2B-client/common/error.hpp), which is template based, and where we can add a string as well, for context, which `error_condition` cannot do on its own (you can only get an error string for the code itself).

Now, there's an important property we have to satisfy, which is that `Reader` can not be a template, because `template` and `virtual` do not mix. So we have to decide on one, and only one, error type to be returned.

Here are some pros and cons I can think:

* Advantages when using ` error_condition`:
    * Does not need to allocate anything.
    * Uses only `std`. If we even assume that `std::expected` would be used in the future, then this would be an interface which needs no external types. Since this is an interface we have used *a lot* in Golang, maybe this makes it more reusable (not sure about that though)?
* Advantages when using `mender::common::error::Error<error_condition>`:
    * Can potentially describe the problem better by providing more context in the string.

I'm almost exactly 50/50 undecided on this one. What do you think?

common/kv_db.hpp
```diff
@@ -12,16 +12,53 @@
...
+	virtual ExpectedBool Remove(const string &key) = 0;
+};
+
+// Works as a transaction interface as well, which auto-creates a transaction
+// for each operation.
+class KeyValueDataBase : virtual public Transaction, virtual public mender::common::io::Closer {
 public:
-	void hello_world();
+	virtual ExpectedBool WriteTransaction(function<ExpectedBool(Transaction &)> txnFunc) = 0;
+	virtual ExpectedBool ReadTransaction(function<ExpectedBool(Transaction &)> txnFunc) = 0;
```
kacf, line 59: Note that I omitted `OpenRead` and `OpenWrite` from [the original API](https://github.com/mendersoftware/mender/blob/master/store/store.go#L50-L54). While working on this I realized that these are just duplicating what the other functions already do, and the other functions do it better (most notably they auto-commit). I think the only reason they existed is because `ReadTransaction` and `WriteTransaction` did not exist when they were created.

common/kv_db/in_memory/in_memory.cpp
```diff
@@ -0,0 +1,108 @@
...
+}
+
+ExpectedBytes KeyValueDataBaseInMemory::Read(const string &key)
+{
+	ExpectedBytes ret;
+	auto result = ReadTransaction([&key, &ret](Transaction &txn) -> ExpectedBool {
+		ret = txn.Read(key);
+		return ExpectedBool(true);
+	});
+	return ret;
```
kacf, line 73: Obviously using a transaction for inserting into an in memory hash map is completely overkill, but this exists more for demonstration purposes.

kacf (commented):
@oleorhagen, @vpodzime: Here is my early draft of how I was envisioning our API would look, assuming that we base it on our good experience from Golang. I wanted to submit this early, as I was thinking this would be a common pattern that we would follow, but that's assuming we agree on it, of course. 🙂

It contains two main parts:

* Reader/Writer/Closer types of interfaces, for doing generic I/O, passing arbitrary writers around. This mirrors common Golang interfaces such as [this one](https://pkg.go.dev/io#Reader).
* An API for the KeyValueDataBase. This is quite different from your proposal, Ole, but follows much more closely [the original API from Golang](https://github.com/mendersoftware/mender/blob/master/store/store.go).

The advantage in both cases is that we can freely pass instances to these around without worrying about what their types are, which gives a lot of freedom in implementing mocking, layered I/O (for compression, for example), and in general agnostic APIs.

What do you think? Does it make sense? Or do you have other ideas for how achieve this?

common/io.hpp
oleorhagen: Examples of using all would be nice. Plus why not start adding the surrounding infrastructure also? So something to shuffle bytes from a reader to a writer? Util no-op wrappers?

common/kv_db.hpp
oleorhagen: Makes perfect sense to me!

oleorhagen: Worst case scenario we'll add them in later again

oleorhagen (commented):
Taken to school!

This is absoultely beautiful :rose: :sun_with_face: :rainbow: :unicorn: 

I'll start building on this interface for the Artifact parser right now actually!

common/io.hpp
oleorhagen: Gut feeling is leaning towards the 2nd option, but no deep insight or strong opinions here.

common/io.hpp
kacf: Sure, I can add the trusted and loved `Copy`!

common/io.hpp
kacf: I think I have a 51% leaning towards it as well, so if I see no further comments, I'll go for that.

common/io.hpp
vpodzime: TBH, I don't understand any of this. :sweat_smile:

common/kv_db.hpp
vpodzime: yup!

common/kv_db/in_memory/in_memory.cpp
```diff
@@ -0,0 +1,108 @@
...
+
+private:
+	KeyValueDataBaseInMemory &db_;
+	bool read_only_;
+};
+
+InMemoryTransaction::InMemoryTransaction(KeyValueDataBaseInMemory &db, bool read_only) :
+	db_(db),
+	read_only_(read_only) {
+}
```
vpodzime, line 40: Could be part of the class declaration above.

common/kv_db/in_memory/in_memory.cpp
```diff
@@ -0,0 +1,108 @@
...
+	} else {
+		return ExpectedBytes(Error(KeyError, "Key " + key + " not found in memory database"));
+	}
+}
+
+ExpectedBool InMemoryTransaction::Write(const string &key, const vector<byte> &value)
+{
+	assert(!read_only_);
+	db_.map_[key] = value;
+	return ExpectedBool(true);
```
vpodzime, line 56: Do you think the `assert()` is enough here or should there also be a specific error (code) for this?

common/kv_db/in_memory/in_memory.cpp
```diff
@@ -0,0 +1,108 @@
...
+{
+	InMemoryTransaction txn(*this, false);
+	return txnFunc(txn);
+}
+
+ExpectedBool KeyValueDataBaseInMemory::ReadTransaction(function<ExpectedBool(Transaction &)> txnFunc)
+{
+	InMemoryTransaction txn(*this, true);
+	return txnFunc(txn);
+}
```
vpodzime, line 106: I would move the `ReadTransaction` and `WriteTransaction` definitions above the methods/functions that use them.

vpodzime (commented):
I was going through the individual commits so some comments are out of date (would still be nice to fix things in the commits that introduce them, though).

Generally, I don't quite get a clear picture here. I don't understand why we need interfaces for readers, writers and closers only to have transactions that have both `Write` and `Read` methods/functions where the `Write()` one just does `assert()` on a boolean flag. And why we need transactions that are created internally and passed to functions -- shouldn't it rather be possible to create a transaction, do some stuff and then decide whether to commit or cancel/abort it?

I know this tries to mimic the API from the Go version of the client, but now is a good time to think twice about the required complexity. I am far from being able to tell what the required complexity is, but to me this looks very complicated for potentially simple use cases.

oleorhagen:
>  don't understand why we need interfaces for readers, writers and closers only to have transactions that have both `Write` and `Read` methods/functions where the `Write()` one just does `assert()` on a boolean flag. 
> 

I think the idea is that this will be a global interface we can use. Which is really nice imo.

common/CMakeLists.txt
```diff
@@ -14,24 +14,45 @@ target_include_directories(json_test PRIVATE ../)
...
 
-add_library(common_kv_db STATIC kv_db/platform/${kv_db_sources})
+
+add_library(common_kv_db STATIC)
 target_include_directories(common_kv_db PRIVATE ../)
-if(${kv_db_sources} MATCHES ".*lmdb.*")
-  target_compile_features(common_kv_db PRIVATE cxx_std_17)
+
+if (MENDER_DB_LMDB)
+  set(kv_db_sources "$<$<STREQUAL:${PLATFORM},linux_x86>:lmdb/lmdb_kv_db.cpp>")
```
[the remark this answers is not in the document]
kacf: Indeed, good catch.

common/CMakeLists.txt
```diff
@@ -14,24 +14,45 @@ target_include_directories(json_test PRIVATE ../)
...
 
+# Test MenderLog
+add_executable(kv_db_test EXCLUDE_FROM_ALL kv_db_test.cpp)
+target_link_libraries(kv_db_test PRIVATE common_kv_db GTest::gtest_main gmock)
+target_include_directories(kv_db_test PRIVATE ../)
+gtest_discover_tests(kv_db_test)
+add_dependencies(check kv_db_test)
+
 add_library(common_events STATIC events/platform/${events_sources})
+target_include_directories(common_events PUBLIC /home/olepor/scratchpad/boost/install/include/)
```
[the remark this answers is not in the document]
kacf: All hail the olepor directory!

common/kv_db/in_memory/in_memory.cpp
kacf: Good point, I keep forgetting that this is allowed now.

common/kv_db/in_memory/in_memory.cpp
kacf: Hmm, you're right, I think `assert` is disabled under the `NDEBUG` define isn't it? So we should have a code path here for the production case as well.

I guess I just fell into the trap that people use as an argument against using `assert`. 🙂

kacf:
> > don't understand why we need interfaces for readers, writers and closers only to have transactions that have both `Write` and `Read` methods/functions where the `Write()` one just does `assert()` on a boolean flag.
> 
> I think the idea is that this will be a global interface we can use. Which is really nice imo.

Yes, sorry, things got a bit mixed up, and the draft was meant more as a general direction when it comes to interfaces, and in fact the Reader / Writer / etc interfaces are largely independent from the DB stuff. I'll submit them separately, and hopefully you'll see in time why these are useful (I have enough confidence in that claim from our time with Golang even without having really used them yet 🙂).

One thing we could consider is making them a bit more C++ idiomatic, perhaps by having `<<` as part of the interface, but I don't think I will attempt that without seeing a use case first.

kacf:
Alright, resubmitted the interface/IO part in #1132, I'll get started on the DB part now.

vpodzime:
> Alright, resubmitted the interface/IO part in #1132, I'll get started on the DB part now.

Thanks!

PR #1132, opened 2023-01-24
chore: Add IO interfaces and convenience functions.

kacf:
Since the interfaces need a non-template error type, we need to settle on one and make some changes to our own error type. First of all, we want to base it on the standard C++11 `error_condition` error type, so that we can use error codes from there. This means we need to remove the enum requirement from our error type.

Also make a few other convenience fixes such as the ability to get a no-error object by default construction, a `StdError` typedef, and a bool operator for easy testing in conditionals.

With that, implement the `Reader`, `Writer` and `ReadWriter` interfaces, which are inspired from the Golang interfaces of the same names, as well as the `Copy` functions, and add some some tests for them.


common/CMakeLists.txt
```diff
@@ -1,8 +1,21 @@
+# Cross platform common library.
+add_library(common STATIC io.cpp)
```
kacf, line 2: What do you think about the library name? I was thinking that this is so central that we should have it in a common library, and not split it up into `io` library (and more libs later); we will just end up with so many static libraries in the end. It's still split by namespace, so this is mainly about compilation, not code structure.

common/error.hpp
```diff
@@ -13,31 +13,42 @@
...
 namespace mender::common::error {
 
-template <typename ErrorCodeType>
+template <typename ErrorType>
 class Error {
 public:
-	static_assert(
-		std::is_enum<ErrorCodeType>::value, "Error requires an enum type for error codes");
-	ErrorCodeType error_code;
+	ErrorType error;
```
kacf, line 27: I changed the name to `error` from `error_code`, since now technically any error can be used. I guess `error_condition` is sort-of an error code, but it's a class, neither an enum nor int, so I thought this was more generic. Open to criticism though.

common/error.hpp
```diff
@@ -13,31 +13,42 @@
...
+		error(e.error),
+		message(e.message) {
+	}
+
+	bool operator==(const Error &other) const {
+		return this->message == other.message && this->error == other.error;
+	}
+
+	operator bool() const {
+		return static_cast<bool>(this->error);
```
vpodzime, line 46: How will this works with an `enum` error code?

common/error.hpp
```diff
@@ -13,31 +13,42 @@
...
 
-	Error(const ErrorCodeType &ec, const std::string &msg) {
-		this->error_code = ec;
-		this->message = msg;
-	};
-	Error(const Error &e) {
-		this->error_code = e.error_code;
-		this->message = e.message;
-	};
+	Error() {
```
vpodzime, line 30: What's the default for an `enum` error code?

common/error.hpp
```diff
@@ -13,31 +13,42 @@
...
+	bool operator==(const Error &other) const {
+		return this->message == other.message && this->error == other.error;
+	}
+
+	operator bool() const {
+		return static_cast<bool>(this->error);
+	}
 };
 
+typedef Error<std::error_condition> StdError;
```
vpodzime, line 50: What is the difference between this and `using StdError = Error<std:error_condition>`? I take it the above creates a new type while the `using` clause only gives a shorter name to the template instantiation?

common/expected.hpp
```diff
@@ -46,6 +47,12 @@ class Expected {
 	};
 };
 
+template <typename ExpectedType, typename ErrorType>
+Expected<ExpectedType, ErrorType>::Expected() {
+	this->has_val_ = true;
+	new (&this->ex_) ExpectedType();
+}
```
vpodzime, line 54: I'm not a big fan of this, it might be confusing for things like `ExpectedBool` or similar. I prefer explicit things over implicit. But then again, it's just up to us how (often) we use it.

common/io.cpp
```diff
@@ -0,0 +1,62 @@
...
+	while (true) {
+		if (buffer.size() != orig_size) {
+			buffer.resize(orig_size);
+		}
+
+		auto result = src.Read(buffer);
+		if (!result) {
+			return result.error();
+		} else if (result.value() == 0) {
+			return StdError();
```
vpodzime, line 36: Which `error_condition` is this and why no message?

common/io.cpp
```diff
@@ -0,0 +1,62 @@
...
+		} else if (result.value() > buffer.size()) {
+			return StdError(
+				std::error_condition(std::errc::invalid_argument),
+				"Read returned more bytes than requested. This is a bug in the Read function.");
+		}
+
+		if (result.value() != buffer.size()) {
+			// Because we only ever resize down, this should be very cheap.
+			buffer.resize(result.value());
```
vpodzime, line 45: But then we resize back up, right?

common/io.cpp
```diff
@@ -0,0 +1,62 @@
...
+		}
+
+		auto result = src.Read(buffer);
+		if (!result) {
+			return result.error();
+		} else if (result.value() == 0) {
+			return StdError();
+		} else if (result.value() > buffer.size()) {
+			return StdError(
+				std::error_condition(std::errc::invalid_argument),
```
vpodzime, line 39: I don't think this is `invalid_argument`.

common/io.hpp
```diff
@@ -0,0 +1,61 @@
...
+using namespace std;
+
+using mender::common::error::StdError;
+using mender::common::expected::Expected;
+
+class Reader {
+public:
+	virtual ~Reader() {};
+
+	virtual Expected<size_t, StdError> Read(vector<uint8_t> &dst) = 0;
```
vpodzime, line 36: I prefer naming `using ...` these types in the libraries/namespaces that work with them.

vpodzime (commented):
Looks good to me otherwise. And already simpler than the original stuff in the in-memory-db PR -- the `Closer()` is gone at least!

common/CMakeLists.txt
oleorhagen: Makes no big difference to me, my initial feeling is to keep them seperate, but have at it.

common/error.hpp
oleorhagen: Why not just `code`, I mean, we are already typing:
`.error()` as the prefix a fair few places, so less stuttering?

common/CMakeLists.txt
kacf: Ok, then let's do separate.

common/error.hpp
kacf: Ok, I'll change it to `code` then. I also like that it's shorter, since we will be typing it a lot, I think.

kacf:
> Looks good to me otherwise. And already simpler than the original stuff in the in-memory-db PR -- the `Closer()` is gone at least!

Yep! Once I started looking closer I realized that a one to one mapping to Golang's interfaces doesn't make sense, since C++ uses destructors, not Closers. The others still hold up though.

CMakeLists.txt
```diff
@@ -57,11 +57,27 @@ else()
...
+# If it's a cross platform component, add this:
+#   target_link_libraries(<TARGET> PUBLIC crossplatform_compiler_flags)
+#
+# If it's a platform specific component, with a cross platform interface, add this:
+#   target_link_libraries(<TARGET> PRIVATE platform_compiler_flags)
+#   target_link_libraries(<TARGET> INTERFACE crossplatform_compiler_flags)
+
+add_library(crossplatform_compiler_flags INTERFACE)
+target_compile_features(crossplatform_compiler_flags INTERFACE cxx_std_11)
+target_compile_options(crossplatform_compiler_flags INTERFACE -std=c++11)
```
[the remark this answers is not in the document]
kacf: The important word in that sentence is "if". It doesn't add it for me because my compiler defaults to providing it. But then it also provides more. And we need to match c++11 exactly.

common/error.hpp
kacf: This assumes that an error code of zero means success, but admittedly, that is not very clear from this code. And although I do think it's the case for most error types, it is an assumption that could be wrong.

Question is: Is there a better way to solve it? We need some way to test whether the instance is an error or not...

common/error.hpp
kacf: There's actually two issues here, so let's take one at a time:

* I'm thinking the default is zero, which then becomes the same question as the bool operator above.

* Independently from that, we probably need to explicitly set it to zero, since C++ just uninitialized memory, I believe (I've become too used to Golang here).

common/error.hpp
kacf: Good point. Using `using` (fun to say) is probably better here.

common/expected.hpp
kacf: Good observation. I added this because I needed it in the test, but now that you pointed it out, I realized that it was passing because the mock object constructed a default return value behind my back, which is definitely *not* what we want. It should have been caught instead, which it was, as soon as I removed this. :+1:

common/io.cpp
kacf: This again relates to the discussion about success value: It makes a zero error code object which means no error. `std` does not provide a macro for this AFAICS, but they do document that 0 means success (indirectly: `error_condition` comes from `std::errc` which has the semantics of `errno` and can be compared to it).

Suggestion: Maybe define zero as the success code ourselves, and then use that in construction?

common/io.cpp
kacf: Yes, added comment about that.

common/io.cpp
kacf: I wanted to separate it from `io_error` because ending up here is guaranteed to be a bug, whereas the other `io_error` returns can happen as a result of real I/O conditions. Do you have a better suggestion from [this list](https://en.cppreference.com/w/cpp/error/errc)?

Another alternative could be to use our own error category (which we just talked about) and define it there, but it seems a bit overkill for something that isn't supposed to happen in the first place.

common/io.hpp
kacf: 👍

This got rid of an additional `using` I had added later, so definitely worth it.

CMakeLists.txt
```diff
@@ -57,11 +57,27 @@ else()
...
+# When deciding which one(s) of these to include, the general rule is this:
+#
+# If it's a cross platform component, add this:
+#   target_link_libraries(<TARGET> PUBLIC crossplatform_compiler_flags)
+#
+# If it's a platform specific component, with a cross platform interface, add this:
+#   target_link_libraries(<TARGET> PRIVATE platform_compiler_flags)
+#   target_link_libraries(<TARGET> INTERFACE crossplatform_compiler_flags)
+
+add_library(crossplatform_compiler_flags INTERFACE)
```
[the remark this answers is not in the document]
kacf: See [this reply](https://github.com/mendersoftware/mender/pull/1132#discussion_r1088634549).

common/io.cpp
```diff
@@ -0,0 +1,62 @@
...
+//    WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+//    See the License for the specific language governing permissions and
+//    limitations under the License.
+
+#include <common/io.hpp>
+
+namespace mender::common::io {
+
+StdError Copy(Writer &dst, Reader &src) {
+	vector<uint8_t> buffer(4096);
```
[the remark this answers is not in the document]
kacf: I think we don't need to. In places where it matters, such as the disk writers, we will call it with an appropriately sized buffer instead of relying on this default.

CMakeLists.txt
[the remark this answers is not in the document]
vpodzime: So it would add it for `c++20` because it's not used by default these days? This is more confusing than beneficial, I can tell you that.

common/error.hpp
vpodzime: > Question is: Is there a better way to solve it? We need some way to test whether the instance is an error or not...

As discussed on Slack, it probably makes most sense to use `error_condition` with our custom categories and error codes and respect the rule there that error code 0 always means no error.

common/error.hpp
vpodzime: > * Independently from that, we probably need to explicitly set it to zero, since C++ just uses uninitialized memory, I believe (I've become too used to Golang here).

But we don't have to allow this, right? Can we just mark the default constructor as deleted?

common/io.cpp
vpodzime: > something that isn't supposed to happen in the first place.

I would add `assert(false)` then, in the first place. :)

vpodzime: > Do you have a better suggestion from [this list](https://en.cppreference.com/w/cpp/error/errc)?

I don't think we should limit ourselves to any list.

vpodzime:
> The others still hold up though.

I'm still curious to see a class that is a Reader but not a Writer and how there are multiple such classes used outside of artifact parsing/processing. :wink:

CMakeLists.txt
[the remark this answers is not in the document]
oleorhagen: It seems they have this also: `set_target_properties(myTarget PROPERTIES CXX_EXTENSIONS OFF)` ?

common/io.cpp
oleorhagen: I like that suggestion better :)

CMakeLists.txt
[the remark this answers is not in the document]
kacf: I haven't tried, but I assume so, yes. It's just testing for minimal capability, and adding what's necessary for that.

common/error.hpp
kacf: Yes, let's use the `NoError` constant instead.

CMakeLists.txt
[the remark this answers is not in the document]
kacf: That flag is about "upgrading" from std to GNU extensions, which is a different thing. It probably should be off for us actually (the default is on), but since with this code we are adding an explicit standards flag anyway, it probably does not matter.

common/io.cpp
kacf: Switching to `error_condition` luckily solves this problem as well, because now we can use enums from any category and mix them. So I'll add an error code for programming error.

kacf:
@vpodzime, @oleorhagen: I have now switched `Error` to use `error_condition` everywhere. I think this works quite well! I decided to put this into a separate commit to make it clearer what has actually changed. Apart from rebasing, only that commit is new.

Note that this is going to break other pull requests in flight, and will require making an error category for them, but it should be easy if you look at what I did in the mentioned commit.

common/io.cpp
kacf: (I forgot about this one. It's coming!)

common/json.hpp
```diff
@@ -27,23 +27,33 @@ namespace mender::common::json {
...
+       string message(int code) const override;
+};
+extern const JsonErrorCategoryClass JsonErrorCategory;
+
+error::Error MakeError(JsonErrorCode code, const string &msg);
+
+using ExpectedString = mender::common::expected::Expected<string, error::Error>;
+using ExpectedInt = mender::common::expected::Expected<int, error::Error>;
+using ExpectedBool = mender::common::expected::Expected<bool, error::Error>;
+using ExpectedSize = mender::common::expected::Expected<size_t, error::Error>;
```
vpodzime, line 52: I think these four should actually go to `error.hpp` and probably directly into the `mender::common` namespace.

kacf: The common ones, yes, that's a good idea.

common/io.cpp
kacf: Ok, now it's fixed.

common/json.hpp
kacf: Done! Note that the resulting code actually became longer, because we still have to use `using` to shorten their names where we use them. But it's still nicer I think.

vpodzime (approved):
ACK

oleorhagen (approved):

kacf (commented):
Approve! (I can't approve my own PR :smile:)

PR #1133, opened 2023-01-25
chore: egrep is obsolescent; use grep -E

alfrunes:


oleorhagen (approved):

lluiscampos (changes requested):
Thanks for the fix!

It needs a changelog entry, though.

PR #1134, opened 2023-01-25
feat: support zstd compression in Mender client

Michaelhobo:
This bumps the version of github.com/mendersoftware/mender-artifact to support zstd compression.

Changelog: Add zstd compression support
Ticket: None
Signed-off-by: Michael Ho <callmemikeh@gmail.com>


kacf:
Thanks. Looks like there are still some errors. It turned out to be pretty easy for me to do this, since it's the exact same approach as I used in the deployments service, so I can just take this over from here. Moved to #1135.

PR #1136, opened 2023-01-27
MEN-6181: Identity parser

vpodzime:


common/processes/platform/tiny_process_library/tiny_process_library.cpp
```diff
@@ -0,0 +1,74 @@
...
+				line_end_idx = bytes_view.find("\n", line_start_idx);
+			}
+
+			if ((line_end_idx == string_view::npos) && (line_start_idx != (len - 1))) {
+				trailing_line += string(bytes_view, line_start_idx, (len - line_start_idx));
+			}
+		});
+
+	// waits for the process to finish
+	// TODO: log exit status != 0? Or error?
```
kacf, line 59: Yes, I would even consider to log each line we get, with debug verbosity. This is an external information source, configurable by the user, so it's good to log it well.

common/identity_parser/identity_parser.cpp
```diff
@@ -0,0 +1,57 @@
...
+	auto ex_line_data = proc.GenerateLineData();
+	if (!ex_line_data) {
+		switch (ex_line_data.error().error_code) {
+		case procs::ProcessErrorCode::SpawnError:
+			return ExpectedKeyValuesMap(IdentityParserError(
+				IdentityParserErrorCode::SpawnError, ex_line_data.error().message));
+			break;
+		default:
+			// all cases should be covered above
+			assert(false);
```
kacf, line 37: We should probably return *some* error after this assert, for the production mode. Perhaps also a log message.

common/identity_parser/identity_parser.cpp
```diff
@@ -0,0 +1,57 @@
...
+	auto ex_key_values = kvp::ParseKeyValues(ex_line_data.value());
+	if (!ex_key_values) {
+		switch (ex_key_values.error().error_code) {
+		case kvp::KeyValueParserErrorCode::InvalidData:
+			return ExpectedKeyValuesMap(IdentityParserError(
+				IdentityParserErrorCode::InvalidData, ex_key_values.error().message));
+			break;
+		default:
+			// all cases should be covered above
+			assert(false);
```
kacf, line 50: Same here.

kacf (commented):
Nice!

common/processes.hpp
```diff
@@ -0,0 +1,52 @@
...
+//    See the License for the specific language governing permissions and
+//    limitations under the License.
+
+#include <string>
+#include <vector>
+#include <common/error.hpp>
+#include <common/expected.hpp>
+
+#ifndef MENDER_COMMON_PROCESSES_HPP
+#define MENDER_COMMON_PROCESSES_HPP
```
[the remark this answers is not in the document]
kacf: Yeah, functionally it doesn't really matter. Maybe it will be *slightly* more work for the compiler if outside, but it's not like that makes a difference. To me it just felt natural that the guards would surround everything, but either is fine with me.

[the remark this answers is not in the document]
oleorhagen: Yes, I agree with Kristian here @vpodzime

common/identity_parser/identity_parser.cpp
vpodzime: Well, this is going away with our switch to `error_condition`, the error will be simply propagated with no conversion/translation.

kacf (approved):

PR #1137, opened 2023-01-30
chore: egrep is obsolescent; use grep -E

kacf:
Changelog: Replace obsolescent `egrep` with `grep -E` in inventory script
(cherry picked from commit 75b0acb07c972f5ba4f40b4b221f4a0f1b342b09)


kacf:
This has already been tested on master and is very low risk. Merging without mender-qa pipeline.

PR #1139, opened 2023-01-30
In memory key value database

kacf:


common/error.cpp
```diff
@@ -18,4 +18,25 @@ namespace mender::common::error {
 
 const Error NoError = Error(std::error_condition(), "");
 
+const CommonErrorCategoryClass CommonErrorCategory;
+
+Error MakeError(ErrorCode code, const std::string &msg) {
+	return Error(std::error_condition(code, CommonErrorCategory), msg);
+}
+
+const char *CommonErrorCategoryClass::name() const noexcept {
```
oleorhagen, line 27: Q: Why `const char*`

oleorhagen (approved):

common/error.cpp
kacf: This is an override, so it comes from the `std` interface. I don't know why they chose `const char *` instead of `const string &`, but since it's supposed to be unchanging, I guess it doesn't matter which one it is. :thinking:

kacf:
@oleorhagen: Do you want to look again, or is it good? The only addition is the last commit.

oleorhagen:
> @oleorhagen: Do you want to look again, or is it good? The only addition is the last commit.

Having another quick look :)

common/key_value_database/in_memory/in_memory.cpp
```diff
@@ -0,0 +1,108 @@
...
+
+ExpectedBool InMemoryTransaction::Remove(const string &key) {
+	assert(!read_only_);
+	db_.map_.erase(key);
+	return ExpectedBool(true);
+}
+
+ExpectedBytes KeyValueDatabaseInMemory::Read(const string &key) {
+	ExpectedBytes ret {vector<uint8_t>()};
+	auto result = ReadTransaction([&key, &ret](Transaction &txn) -> ExpectedBool {
```
oleorhagen, line 65: Nit: `result` is unused, no?

common/key_value_database_test.cpp
```diff
@@ -0,0 +1,156 @@
...
+		// Write
+		auto ret = db.Write("key", mender::common::ByteVectorFromString("val"));
+		EXPECT_TRUE(ret);
+	}
+
+	{
+		// Read
+		auto entry = db.Read("key");
+		ASSERT_TRUE(entry);
+		std::string string1(common::StringFromByteVector(entry.value()));
```
oleorhagen, line 43: Nice - can't wait to use this one

common/common.hpp
```diff
@@ -0,0 +1,45 @@
...
+#include <vector>
+
+namespace mender::common {
+
+using namespace std;
+
+inline static vector<uint8_t> ByteVectorFromString(const char *str) {
+	return vector<uint8_t>(
+		reinterpret_cast<const uint8_t *>(str),
+		reinterpret_cast<const uint8_t *>(str + strlen(str)));
```
oleorhagen, line 30: Q: How will this fair with null bytes in the `string` ?

common/key_value_database/in_memory/in_memory.cpp
```diff
@@ -0,0 +1,108 @@
...
+	auto value = db_.map_.find(key);
+	if (value != db_.map_.end()) {
+		return ExpectedBytes(value->second);
+	} else {
+		return ExpectedBytes(MakeError(KeyError, "Key " + key + " not found in memory database"));
+	}
+}
+
+ExpectedBool InMemoryTransaction::Write(const string &key, const vector<uint8_t> &value) {
+	assert(!read_only_);
```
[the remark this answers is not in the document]
kacf: `assert` is only active in debug mode, and I think asserting in debug mode is good; we want early warnings for this kind of thing. But you're right that this leaves the release path open to an erroneous return.

This has popped up several times now, so here's my suggestion: Let's make an `AssertOrError` macro which asserts in debug mode, and returns a `error::MakeError(error::ProgrammingError, "<ASSERT-STRING>")` in production mode. That way we get resiliency in production mode, and also makes it less likely that we will forget this.

If you agree, I will leave this alone for now, and follow up in my next PR, since I'm essentially redoing it for LMDB, so I'll just introduce it for both at the same time.

common/key_value_database/in_memory/in_memory.cpp
kacf: Ooh, good catch! This revealed a whole flurry of mistakes actually. I had not converted the interface from `ExpectedBool` to `Error` as I had intended when I introduced that. I will follow up with a new commit for this.

common/key_value_database/in_memory/in_memory.cpp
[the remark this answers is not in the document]
oleorhagen: Yes, great idea! I was stuck in the either/or kinda mindset

common/common.hpp
kacf: I think the C string and the `string` will then be different, one terminating before the other. I could make an error return I suppose, but that makes the function much more annoying to use. I think I would vote for ignoring this, since the string will simply be shorter in C, there is no other "crashing" behavior I can think of.

common/key_value_database/in_memory/in_memory.cpp
```diff
@@ -0,0 +1,108 @@
...
+	ExpectedBool ret {true};
+	auto result = WriteTransaction([&key, &value, &ret](Transaction &txn) -> ExpectedBool {
+		ret = txn.Write(key, value);
+		return ExpectedBool(true);
+	});
+	return ret;
+}
+
+ExpectedBool KeyValueDatabaseInMemory::Remove(const string &key) {
+	ExpectedBool ret {true};
```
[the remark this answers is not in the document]
kacf: This became irrelevant with the latest commit.

common/key_value_database/in_memory/in_memory.cpp
kacf: Check the latest commit, I switched to Error returns for most functions there!

common/key_value_database/in_memory/in_memory.cpp
```diff
@@ -48,59 +48,57 @@ ExpectedBytes InMemoryTransaction::Read(const string &key) {
...
 ExpectedBytes KeyValueDatabaseInMemory::Read(const string &key) {
 	ExpectedBytes ret {vector<uint8_t>()};
-	auto result = ReadTransaction([&key, &ret](Transaction &txn) -> ExpectedBool {
+	auto err = ReadTransaction([&key, &ret](Transaction &txn) -> Error {
 		ret = txn.Read(key);
-		return ExpectedBool(true);
+		return error::NoError;
 	});
-	return ret;
+	if (err) {
```
oleorhagen, line 69: How tempted on using the `?:` operator were you? :p

common/key_value_database/in_memory/in_memory.cpp
```diff
@@ -48,59 +48,57 @@ ExpectedBytes InMemoryTransaction::Read(const string &key) {
...
+	}
 }
 
-ExpectedBool KeyValueDatabaseInMemory::Write(const string &key, const vector<uint8_t> &value) {
-	ExpectedBool ret {true};
-	auto result = WriteTransaction([&key, &value, &ret](Transaction &txn) -> ExpectedBool {
-		ret = txn.Write(key, value);
-		return ExpectedBool(true);
+Error KeyValueDatabaseInMemory::Write(const string &key, const vector<uint8_t> &value) {
+	return WriteTransaction([&key, &value](Transaction &txn) -> Error {
```
oleorhagen, line 77: Nice

common/key_value_database/in_memory/in_memory.cpp
```diff
@@ -48,59 +48,57 @@ ExpectedBytes InMemoryTransaction::Read(const string &key) {
...
-	return ret;
 }
 
-ExpectedBool KeyValueDatabaseInMemory::WriteTransaction(
-	function<ExpectedBool(Transaction &)> txnFunc) {
+Error KeyValueDatabaseInMemory::WriteTransaction(
+	function<Error(Transaction &)> txnFunc) {
 	InMemoryTransaction txn(*this, false);
 	// Simple, but inefficient rollback support.
 	this->backup_map_ = this->map_;
```
oleorhagen, line 92: Q: Is the backup map ever needed outside this function really?

oleorhagen (approved):
Looking good :star:

common/key_value_database/in_memory/in_memory.cpp
kacf: Haha, you know what? It did not even occur to me! What has happened to me!? :scream:

common/key_value_database/in_memory/in_memory.cpp
kacf: Good point! It isn't. I'll make it function-local.

PR #1143, opened 2023-02-03
Make sure the code-base complies with C++11

oleorhagen:


CMakeLists.txt
```diff
@@ -76,7 +76,7 @@ endif()
 
 add_library(crossplatform_compiler_flags INTERFACE)
 target_compile_features(crossplatform_compiler_flags INTERFACE cxx_std_11)
-target_compile_options(crossplatform_compiler_flags INTERFACE -std=c++11)
+target_compile_options(crossplatform_compiler_flags INTERFACE -std=c++11 -Werror)
```
kacf, line 79: But this still doesn't catch it, correct? Annoying that only clang does...

kacf (approved):

CMakeLists.txt
oleorhagen: Hmm, good point. I only tested it with clang. Doing a `gcc` run now

oleorhagen: Catches it with gnu also

PR #1145, opened 2023-02-10
code review comments addressed 

lukfi:


common/config_parser/config_parser.cpp
```diff
@@ -58,20 +57,19 @@ ExpectedBool MenderConfigFromFile::ValidateArtifactKeyCondition() const {
...
 
-	if (conf.servers.size() == 0) {
-		if (conf.server_url.size() == 0) {
+	if (servers.size() == 0) {
+		if (server_url.size() == 0) {
 			// TODO - Log warning
 		}
 	}
-	return ExpectedBool(true);
+	return true;
```
kacf, line 72: We should replace the return type with `error::Error` actually, and then return `error::NoError`. Returning `ExpectedBool` is a leftover from before we adopted that method.

But this requires changes in the caller as well, and is unrelated to this PR, so I think this can be a followup.

In fact, we should do that pretty much everywhere where we use `ExpectedBool` except if there is a place where we actually use the boolean value (but I don't think there is).

common/io.hpp
```diff
@@ -74,8 +74,7 @@ class StreamReader : virtual public Reader {
 		is_ {stream} {
 	}
 	ExpectedSize Read(vector<uint8_t> &dst) override {
-		is_.read(reinterpret_cast<char *>(&dst[0]), dst.size());
-		return is_.gcount();
+		return is_.readsome(reinterpret_cast<char *>(&dst[0]), dst.size());
```
kacf, line 77: The documentation for this function is a bit unclear, it seems to indicate that it might return nothing if the underlying buffer happens not to be filled. And that the function will not itself fill the buffer?

I don't know, [the documentation](https://en.cppreference.com/w/cpp/io/basic_istream/readsome) is difficult to understand. What do you think?

kacf (changes requested):
* Our commit messages follow the Conventional Commit specification, plus a few extra rules (you have to sign off). Check [this guide](https://github.com/mendersoftware/mendertesting/blob/master/commitlint/grammar.md). Note that changelogs are not very important for this project, since most likely we will make one single entry for the whole project, and not multiple entries. So changelog can be set to `None`.

* Some [style issues](https://gitlab.com/Northern.tech/Mender/mender/-/jobs/3748769201).

common/io.hpp
lukfi: I think we can stay with 'read', the documentation for 'readsome' says:
"The behavior of this function is highly implementation-specific."
So it probably safer to use 'read', especially if we don't expect large delays while extracting data from streams.

common/config_parser/config_parser.cpp
lukfi: Agree, probably 'ExpectedBool Json::GetBool() const " is the only place we should use ExpectedBool

kacf (changes requested):
Thanks!

Note the commit format. Your commit should be squashed and the message needs to consist of what and why you changed the code. It should also be prefixed with a category. `chore: ` should be appropriate for this type of commit.

Other than that take a look at the CI status and fix any issues it tells you about.

kacf (approved):

PR #1149, opened 2023-02-13
Add address and thread sanitization

oleorhagen:


cmake/asan.cmake
```diff
@@ -0,0 +1,26 @@
+get_property(isMultiConfig GLOBAL PROPERTY GENERATOR_IS_MULTI_CONFIG)
+
+if(isMultiConfig)
+    if(NOT "ASan" IN_LIST CMAKE_CONFIGURATION_TYPES)
+        list(APPEND CMAKE_CONFIGURATION_TYPES Asan)
+    endif()
+else()
+    set(allowedBuildTypes ASan Debug Release RelWithDebInfo MinSizeRel)
+    set_property(CACHE CMAKE_BUILD_TYPE PROPERTY STRINGS "${allowedBuildTypes}")
+endif()
```
vpodzime, line 10: Still don't see the world where this is nicer, simpler and easier to understand than autotools...

CMakeLists.txt
```diff
@@ -12,6 +12,9 @@ endif (POLICY CMP0135)
 
 project(mender)
 
+include(cmake/asan.cmake)
+include(cmake/memsan.cmake)
+include(cmake/threadsan.cmake)
```
vpodzime, line 17: I guess you took these from somewhere? Would be nice to mention it in the commit message. Also, what's the license?

vpodzime (commented):
Looks OK to me otherwise. CI doesn't like it, though.

cmake/asan.cmake
oleorhagen: How is this helpful?

CMakeLists.txt
oleorhagen: I adapted them from `learning CMake` the book. I'll mention it in the PR :+1: 

It still feels like this should be able to meta-program a bit though, like this is bacially a function rendering this template with two variables...

oleorhagen: Simplified it quite a bit now. Turns out CMake has functions :tada:

vpodzime (commented):
I think the first 3 commits can be squashed. Looks good to me otherwise.

cmake/helper.cmake
```diff
@@ -0,0 +1,23 @@
+
+
+function (sanitizer_add_compiler_flags CONFIG SANITIZER_BUILD_FLAGS SANITIZER_SHARED_LINKER_FLAGS)
+
+  message (STATUS "Adding the flags: ${SANITIZER_BUILD_FLAGS}")
```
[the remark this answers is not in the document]
oleorhagen: Yes, good point. Added :)

kacf (approved):
👍

PR #1151, opened 2023-02-14
ci: Undefined sanitizer

oleorhagen:


cmake/undefsan.cmake
```diff
@@ -0,0 +1,7 @@
+include(cmake/helper.cmake)
+
+if (CMAKE_BUILD_TYPE STREQUAL "UndefSan")
+  sanitizer_add_compiler_flags(UNDEFSAN
+    "-fsanitize=undefined"
+    "-fsanitize=undefined")
```
vpodzime, line 6: Maybe we should rename the function to `sanitizer_add_compiler_linker_flags()`? This looks a bit confusing otherwise.

vpodzime (approved):

cmake/undefsan.cmake
oleorhagen: Yeh, agreed. Good idea!

oleorhagen: https://github.com/mendersoftware/mender/pull/1151/commits/73dc8de251eba249a343daa1c4309a9a2b3f32d0

oleorhagen: Went for: `sanitizer_add_compiler_and_linker_flags()`

PR #1152, opened 2023-02-15
feat: Inventory parser

vpodzime:
Parsing inventory data from inventory scripts.

Ticket: MEN-6182
Changelog: None


common/key_value_parser/key_value_parser.cpp
```diff
@@ -49,6 +49,15 @@ error::Error MakeError(KeyValueParserErrorCode code, const string &msg) {
...
 	KeyValuesMap ret;
+	error::Error err = AddParseKeyValues(ret, items, delimiter);
+	if (err) {
+		return ExpectedKeyValuesMap(err);
+	} else {
+		return ExpectedKeyValuesMap(ret);
+	}
+}
+
+error::Error AddParseKeyValues(KeyValuesMap &base, const vector<string> &items, char delimiter) {
```
oleorhagen, line 62: S: I would consider returning the map explicitly here, instead of just an error.

Then you can eliminate the `if .. else` above also.

common/key_value_parser/key_value_parser.cpp
```diff
@@ -60,22 +69,23 @@ ExpectedKeyValuesMap ParseKeyValues(const vector<string> &items, char delimiter)
...
 	if (invalid_data != "") {
-		return ExpectedKeyValuesMap(MakeError(
+		return MakeError(
 			KeyValueParserErrorCode::InvalidDataError,
-			"Invalid data given: '" + invalid_data + "'"));
+			"Invalid data given: '" + invalid_data + "'");
 	}
 
-	return ExpectedKeyValuesMap(ret);
+	return MakeError(KeyValueParserErrorCode::NoError, "");
```
oleorhagen, line 87: Q: Should we consider having an overload with just the error, and then we won't have to pass in an empty message?

oleorhagen (approved):
Looks good

common/key_value_parser/key_value_parser.cpp
vpodzime: Hmm, that would allow creation of errors with other codes and no message. And that we don't want.

common/key_value_parser/key_value_parser.cpp
vpodzime: That's what I had initially. But it was copying the whole map in and out. How would I avoid that? Taking a reference and then returning it? How would you use a function with such signature then?

oleorhagen: Why not create the map in the function, and `move` it out of the function afterwards? :thinking: 

```
diff --git a/common/key_value_parser/key_value_parser.cpp b/common/key_value_parser/key_value_parser.cpp
index cc45e14e..8eecc117 100644
--- a/common/key_value_parser/key_value_parser.cpp
+++ b/common/key_value_parser/key_value_parser.cpp
@@ -50,16 +50,7 @@ error::Error MakeError(KeyValueParserErrorCode code, const string &msg) {
 }
 
 ExpectedKeyValuesMap ParseKeyValues(const vector<string> &items, char delimiter) {
-	KeyValuesMap ret;
-	error::Error err = AddParseKeyValues(ret, items, delimiter);
-	if (err) {
-		return ExpectedKeyValuesMap(err);
-	} else {
-		return ExpectedKeyValuesMap(ret);
-	}
-}
-
-error::Error AddParseKeyValues(KeyValuesMap &base, const vector<string> &items, char delimiter) {
+	KeyValuesMap base;
 	string invalid_data = "";
 
 	for (auto str : items) {
@@ -84,10 +75,9 @@ error::Error AddParseKeyValues(KeyValuesMap &base, const vector<string> &items,
 			"Invalid data given: '" + invalid_data + "'");
 	}
 
-	return MakeError(KeyValueParserErrorCode::NoError, "");
+	return base;
 }
 
-
 } // namespace key_value_parser
 } // namespace common
 } // namespace mender
```

Something like this?

common/CMakeLists.txt
```diff
@@ -138,3 +138,15 @@ target_include_directories(identity_parser_test PRIVATE ${CMAKE_SOURCE_DIR})
 gtest_discover_tests(identity_parser_test)
 add_dependencies(check identity_parser_test)
 
+
+find_package(Boost 1.54 REQUIRED COMPONENTS filesystem)
```
oleorhagen, line 142: Filesystem is ways old...

I found it all the way back in `1.30`

common/inventory_parser.hpp
```diff
@@ -0,0 +1,33 @@
...
+#include <common/key_value_parser.hpp>
+
+namespace mender {
+namespace common {
+namespace inventory_parser {
+
+using namespace std;
+namespace kvp = mender::common::key_value_parser;
+
+kvp::ExpectedKeyValuesMap GetInventoryData(const string generators_dir);
```
oleorhagen, line 27: S: Maybe take a reference?

common/inventory_parser/inventory_parser.cpp
```diff
@@ -0,0 +1,90 @@
...
+
+using namespace std;
+namespace kvp = mender::common::key_value_parser;
+namespace procs = mender::common::processes;
+namespace log = mender::common::log;
+namespace err = mender::common::error;
+namespace fs = boost::filesystem;
+
+kvp::ExpectedKeyValuesMap GetInventoryData(const string generators_dir) {
+	bool any_success;
```
oleorhagen, line 34: S: Initialize them also

common/inventory_parser/inventory_parser.cpp
```diff
@@ -0,0 +1,90 @@
...
+	bool any_failure;
+	kvp::KeyValuesMap data;
+
+	fs::path dir_path(generators_dir);
+	if (!fs::exists(dir_path)) {
+		return kvp::ExpectedKeyValuesMap(data);
+	}
+
+	fs::directory_iterator end_itr; // default construction yields past-the-end
```
oleorhagen, line 43: Come again.... :persevere:

common/inventory_parser/inventory_parser.cpp
oleorhagen: Isn't `any_success` pretty much the same as checking if the map has any values?
Which also begs the question, is this necessarily an error? :thinking:

common/inventory_parser/inventory_parser.cpp
```diff
@@ -0,0 +1,90 @@
...
+		auto err = kvp::AddParseKeyValues(data, ex_line_data.value());
+		if (err) {
+			log::Error("Failed to parse data from '" + file_path_str + "': " + err.message);
+			any_failure = true;
+		} else {
+			any_success = true;
+		}
+	}
+
+	if (any_success || !any_failure) {
```
oleorhagen, line 78: I'm not quite sure this is right :thinking: 

So if one out of `5` inventory scripts fail, should we not gather the data from the remaining `4`?

common/inventory_parser/inventory_parser.cpp
```diff
@@ -0,0 +1,90 @@
...
+		return kvp::ExpectedKeyValuesMap(data);
+	} else {
+		err::Error error = MakeError(
```
oleorhagen, line 81: Also, you can just return the error straight away. No need to create it on a separate line, then copy it and return :)

common/inventory_parser/inventory_parser.cpp
vpodzime: > Isn't any_success pretty much the same as checking if the map has any values?

It is, but I like the explicit name and explicit assignment.

>  Which also begs the question, is this necessarily an error?

Hard to say, what exactly should be treated as error. But to me it makes sense to treat no scripts as not an error, some successful and some failing scripts as not an error with issues being logged and all scripts failing as an error. Then again, not sure what a general approach to things like this is/should be in mender.

common/inventory_parser/inventory_parser.cpp
vpodzime: ? This how `boost::filesystem` does it, unfortunately.

common/inventory_parser/inventory_parser.cpp
vpodzime: That's what the code does, right? If there is *any success*, it returns the parsed data. If there is **not** *any failure*, it also returns the parsed data.

common/inventory_parser/inventory_parser.cpp
vpodzime: Tell that to the code formatter.

common/CMakeLists.txt
oleorhagen: But you're sure it works with `1.54`?

common/key_value_parser/key_value_parser.cpp
oleorhagen: discussed f2f (summary):

I _personally_ prefer explictly returning the type, and merging, but merging through side-effects are fine. Do whatever you think works best

common/inventory_parser/inventory_parser.cpp
oleorhagen: Buu @boost

common/inventory_parser/inventory_parser.cpp
oleorhagen: True

common/inventory_parser/inventory_parser.cpp
oleorhagen: Hmm, why not:
```
return MakeError(...)
````

What does it have to do with the formatter ? :thinking:

File: common/inventory_parser/inventory_parser.cpp (hunk 5 above)
Comment (oleorhagen) [reply common/inventory_parser/inventory_parser.cpp:34]: Yeh, makes sense :+1: 

I would still initialize the boolean I think though

File: common/CMakeLists.txt [hunk 9]
```diff
@@ -138,3 +138,15 @@ target_include_directories(identity_parser_test PRIVATE ${CMAKE_SOURCE_DIR})
...
+add_library(common_inventory_parser STATIC inventory_parser/inventory_parser.cpp)
+target_link_libraries(common_inventory_parser PUBLIC crossplatform_compiler_flags)
+target_include_directories(common_inventory_parser PRIVATE ../ ${CMAKE_BINARY_DIR} ${Boost_INCLUDEDIR})
```
kacf, line 145: Make sure you coordinate the `../` cleanup with your other effort.

common/inventory_parser/inventory_parser.cpp
```diff
@@ -0,0 +1,90 @@
...
+namespace mender {
+namespace common {
+namespace inventory_parser {
+
+using namespace std;
+namespace kvp = mender::common::key_value_parser;
+namespace procs = mender::common::processes;
+namespace log = mender::common::log;
+namespace err = mender::common::error;
+namespace fs = boost::filesystem;
```
kacf, line 31: If it uses `boost::filesystem` the file should be in a `platform` directory.

common/inventory_parser/inventory_parser.cpp
```diff
@@ -0,0 +1,90 @@
...
+
+	fs::directory_iterator end_itr; // default construction yields past-the-end
+	for (fs::directory_iterator itr(dir_path); itr != end_itr; ++itr) {
+		fs::path file_path = itr->path();
+		if (!fs::is_regular_file(file_path)) {
+			continue;
+		}
+
+		if (file_path.filename().string().find("mender-inventory-") != 0) {
+			continue;
```
kacf, line 51: I would warn about this too. It's a common user mistake.

common/inventory_parser/inventory_parser.cpp
```diff
@@ -0,0 +1,90 @@
...
+		}
+
+		string file_path_str = file_path.string();
+		fs::perms perms = itr->status().permissions();
+		if ((perms & (fs::perms::owner_exe | fs::perms::group_exe | fs::perms::others_exe))
+			== fs::perms::no_perms) {
+			log::Warning("'" + file_path_str + "' is not executable");
```
kacf, line 58: POSIX has the `access` function, which not only tells you the permissions, but take group membership and such into account, to determine whether you have the access you're asking for. Does boost have anything like that?

common/key_value_parser/key_value_parser.cpp
```diff
@@ -49,6 +51,15 @@ error::Error MakeError(KeyValueParserErrorCode code, const string &msg) {
 
 ExpectedKeyValuesMap ParseKeyValues(const vector<string> &items, char delimiter) {
 	KeyValuesMap ret;
+	error::Error err = AddParseKeyValues(ret, items, delimiter);
+	if (err) {
+		return ExpectedKeyValuesMap(err);
+	} else {
+		return ExpectedKeyValuesMap(ret);
```
kacf, line 58: I think you can just write `return err` and `return ret` here.

common/key_value_parser/key_value_parser.cpp
kacf: Could also return a shared_ptr if value-copy is expensive. But modification is also fine, IMHO.

common/key_value_parser/key_value_parser.cpp
kacf: You should return `error::NoError` instead. The `NoError`s for other error categories need to be there for code correctness purposes, but there is no need to use them.

common/CMakeLists.txt
vpodzime: I thought that was what our CI was (used to be?) using.

common/inventory_parser/inventory_parser.cpp
vpodzime: Cannot agree more!

common/inventory_parser/inventory_parser.cpp
vpodzime: > I would still initialize the boolean I think though

common/inventory_parser/inventory_parser.cpp
vpodzime: Because it's `return kvp::ExpectedKeyValuesMap(MakeError(...))` and that's too long so it puts it into multiple lines making me feel better about the separate construction and assignment.

common/inventory_parser/inventory_parser.cpp
vpodzime: I haven't found anything like that. But if there is not enough access, the execution will fail. Too bad the *tiny-process-library* doesn't tell us about that...

common/key_value_parser/key_value_parser.cpp
vpodzime: How does that work? I saw `return true;` in a function returning `ExpectedBool()`, but no idea how/why it works.

common/CMakeLists.txt
oleorhagen: No, we're on like `1.7x` something

common/inventory_parser/inventory_parser.cpp
oleorhagen: Oh, but I don't think you need to wrap it in `ExpectedKeyValuesMap`, it will do the conversion implicitly.

common/CMakeLists.txt
vpodzime: This will be taken care of in #1154, I want to merge this one first.

common/inventory_parser/inventory_parser.cpp
vpodzime: And I still don't understand why `return MakeError()` would probably work even though `MakeError()` doesn't return the same type as this function does.

common/inventory_parser/inventory_parser.cpp
kacf: Ok, it's fine, it would be an improvement, but by no means necessary.

common/key_value_parser/key_value_parser.cpp
kacf: Since the compiler already knows that the return type of the function is `ExpectedKeyValuesMap`, it just puts two and two together and transforms `return <RET>` into `return ExpectedKeyValuesMap(<RET>)` for you. I suppose one could argue that it's less explicit, but I think that's fine as long as we name our variables sensibly.

kacf (approved):

common/inventory_parser/inventory_parser.cpp
oleorhagen: It does an `implicit conversion`. See [here](https://en.cppreference.com/w/cpp/language/implicit_conversion), and especially the `return` type conversions.
`when initializing a new object of type T2, including return statement in a function returning T2; `

vpodzime: I see. I still don't like it, though, so forgive for not using it much. Especially in C++ where this "return type conversion" can actually mean execution of more lines than the function containing the `return` I don't like it to be implicit.

common/key_value_parser/key_value_parser.cpp
vpodzime: Like I said above in https://github.com/mendersoftware/mender/pull/1152#discussion_r1115568345, I don't like this so I'd prefer not to use it.

PR #1153, opened 2023-02-15
fix: Call setup when creating the global logger

oleorhagen:
Ticket: None
Changelog: None


vpodzime (approved):
Looks good to me.

PR #1154, opened 2023-02-16
feat: Add default paths and environment handling

vpodzime:
Default paths in mender::common::conf::paths and
mender::common::conf:GetEnv() for getting values of environment variables with a fallback (default) value.

Ticket: MEN-6313
Changelog: None

### DIFF
--- a/common/CMakeLists.txt
+++ b/common/CMakeLists.txt
@@ -133,7 +133,7 @@ add_dependencies(check processes_test)
 add_library(common_key_value_parser STATIC key_value_parser/key_value_parser.cpp)
 target_link_libraries(common_key_value_parser PUBLIC crossplatform_compiler_flags)
 target_link_libraries(common_key_value_parser PUBLIC common_error)
-target_include_directories(common_key_value_parser PRIVATE ../ ${CMAKE_BINARY_DIR})
+target_include_directories(common_key_value_parser PRIVATE ${CMAKE_SOURCE_DIR} ${CMAKE_BINARY_DIR})
 
 add_executable(key_value_parser_test EXCLUDE_FROM_ALL key_value_parser_test.cpp)
 target_link_libraries(key_value_parser_test PUBLIC common_key_value_parser GTest::gtest_main gmock)
@@ -144,7 +144,7 @@ add_dependencies(check key_value_parser_test)
 
 add_library(common_identity_parser STATIC identity_parser/identity_parser.cpp)
 target_link_libraries(common_identity_parser PUBLIC crossplatform_compiler_flags)
-target_include_directories(common_identity_parser PRIVATE ../ ${CMAKE_BINARY_DIR})
+target_include_directories(common_identity_parser PRIVATE ${CMAKE_SOURCE_DIR} ${CMAKE_BINARY_DIR})
 target_link_libraries(common_identity_parser PUBLIC common_key_value_parser common_processes)
 
 add_executable(identity_parser_test EXCLUDE_FROM_ALL identity_parser_test.cpp)
--- a/common/CMakeLists.txt
+++ b/common/CMakeLists.txt
@@ -169,3 +169,20 @@ target_link_libraries(inventory_parser_test PUBLIC common_inventory_parser commo
 target_include_directories(inventory_parser_test PRIVATE ${CMAKE_SOURCE_DIR})
 gtest_discover_tests(inventory_parser_test)
 add_dependencies(check inventory_parser_test)
+
+
+add_library(common_conf STATIC conf/conf.cpp)
+target_link_libraries(common_conf INTERFACE crossplatform_compiler_flags)
+if(MENDER_USE_BOOST_FS)
+  find_package(Boost 1.54 REQUIRED COMPONENTS filesystem)
+  target_sources(common_conf PRIVATE conf/platform/boost_filesystem/paths.cpp)
+  target_include_directories(common_conf PRIVATE ${CMAKE_SOURCE_DIR} ${CMAKE_BINARY_DIR} ${Boost_INCLUDEDIR})
+  target_link_libraries(common_conf PRIVATE platform_compiler_flags)
+  target_link_libraries(common_conf ${Boost_libraries})
+endif()
+
+add_executable(conf_test EXCLUDE_FROM_ALL conf_test.cpp)
+target_link_libraries(conf_test PUBLIC common_conf GTest::gtest_main gmock)
+target_include_directories(conf_test PRIVATE ${CMAKE_SOURCE_DIR})
+gtest_discover_tests(conf_test)
+add_dependencies(check conf_test)
--- a/common/conf/conf.cpp
+++ b/common/conf/conf.cpp
@@ -1,0 +1,37 @@
+// Copyright 2023 Northern.tech AS
+//
+//    Licensed under the Apache License, Version 2.0 (the "License");
+//    you may not use this file except in compliance with the License.
+//    You may obtain a copy of the License at
+//
+//        http://www.apache.org/licenses/LICENSE-2.0
+//
+//    Unless required by applicable law or agreed to in writing, software
+//    distributed under the License is distributed on an "AS IS" BASIS,
+//    WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+//    See the License for the specific language governing permissions and
+//    limitations under the License.
+
+#include <common/conf.hpp>
+
+#include <string>
+#include <cstdlib>
+
+namespace mender {
+namespace common {
+namespace conf {
+
+using namespace std;
+
+string GetEnv(const string &var_name, const string &default_value) {
+	const char *value = getenv(var_name.c_str());
+	if (value == nullptr) {
+		return string(default_value);
+	} else {
+		return string(value);
+	}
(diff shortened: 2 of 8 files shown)


common/CMakeLists.txt
```diff
@@ -138,3 +138,13 @@ target_include_directories(identity_parser_test PRIVATE ${CMAKE_SOURCE_DIR})
 gtest_discover_tests(identity_parser_test)
 add_dependencies(check identity_parser_test)
 
+
+add_library(common_conf STATIC conf/conf.cpp conf/paths.cpp)
+target_link_libraries(common_conf PUBLIC crossplatform_compiler_flags)
+target_include_directories(common_conf PRIVATE ../ ${CMAKE_BINARY_DIR})
```
kacf, line 144: Use `${CMAKE_SOURCE_DIR}` instead of `../`. The latter has broken for me in some circumstances.

common/conf/paths.cpp
```diff
@@ -0,0 +1,44 @@
...
+
+using namespace std;
+namespace conf = mender::common::conf;
+
+string DefaultPathConfDir = conf::GetEnv("MENDER_CONF_DIR", "/etc/mender");
+string DefaultPathDataDir = conf::GetEnv("MENDER_DATA_DIR", "/usr/share/mender");
+string DefaultDataStore = conf::GetEnv("MENDER_DATASTORE_DIR", "/var/lib/mender");
+string DefaultKeyFile = "mender-agent.pem";
+
+string DefaultConfFile = DefaultPathConfDir + "/mender.conf";
```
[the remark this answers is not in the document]
kacf: Ideally, yes, but we can't expose the boost functions in this file. If so you'll have to put it under `platform`.

Alternatively you could make a directory separator variable which we ifdef per platform, which is probably good enough.

common/conf/paths.cpp
```diff
@@ -0,0 +1,44 @@
...
+string DefaultFallbackConfFile = DefaultDataStore + "/mender.conf";
+
+// device specific paths
+string DefaultArtScriptsPath = DefaultDataStore + "/scripts";
+string DefaultRootfsScriptsPath = DefaultPathConfDir + "/scripts";
+string DefaultModulesPath = DefaultDataStore + "/modules" + "/v3";
```
kacf, line 37: Careful! This should be `DataPathDir` (otherwise it's the same as the one below).

kacf (changes requested):

common/conf/paths.cpp
vpodzime: Good :eye: !

common/conf/paths.cpp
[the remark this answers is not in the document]
vpodzime: I thought we would use `boost::filesystem` on all platforms where we need to work with file system (paths).

common/CMakeLists.txt
vpodzime: Will fix it in other places so that I don't take one of them as a template again next time.

common/conf/paths.cpp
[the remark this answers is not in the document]
vpodzime: It makes more sense to me to add a platform-specific something like `PathJoin(string path, string suffix)` rather than just the dir separator constant. But like I said, I thought we would leave this to `boost::filesystem` everywhere where we need to work with paths. And with that, I guess all these constants should use `boost::filesystem::path` functions and operators to put together the full paths and then maybe do call `string()` on the result to get a string?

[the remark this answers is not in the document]
kacf: As discussed in the standup, we will try to only return the "payload" from filesystems operations to the cross platform logic, and deal with paths and file lists and such under the hood. So should we put this file in `platform` then, perhaps?

vpodzime:
https://github.com/mendersoftware/mender/pull/1157/commits/2f322e50d90a09a0e1373a93d4a6d8c1c2d21b9e slightly modifies things done here.

common/CMakeLists.txt
```diff
@@ -140,6 +140,7 @@ gtest_discover_tests(identity_parser_test)
 add_dependencies(check identity_parser_test)
 
 
+
```
oleorhagen, line 143: Nit: Leftover new newline

common/CMakeLists.txt
```diff
@@ -155,3 +156,20 @@ target_link_libraries(inventory_parser_test PUBLIC common_inventory_parser commo
...
+add_library(common_conf STATIC conf/conf.cpp)
+target_link_libraries(common_conf INTERFACE crossplatform_compiler_flags)
+if(MENDER_USE_BOOST_FS)
+  find_package(Boost 1.54 REQUIRED COMPONENTS filesystem)
+  target_sources(common_conf PRIVATE conf/platform/boost_filesystem/paths.cpp)
+  target_include_directories(common_conf PRIVATE ${CMAKE_SOURCE_DIR} ${CMAKE_BINARY_DIR} ${Boost_INCLUDEDIR})
+  target_link_libraries(common_conf PRIVATE platform_compiler_flags)
+  target_link_libraries(common_conf ${Boost_libraries})
+endif()
```
oleorhagen, line 169: I know we don't really do this anywhere, but should we starting adding `else` statements to this?

common/conf/conf.cpp
```diff
@@ -0,0 +1,37 @@
...
+
+namespace mender {
+namespace common {
+namespace conf {
+
+using namespace std;
+
+string GetEnv(const string &var_name, const string &default_value) {
+	const char *value = getenv(var_name.c_str());
+	if (value == nullptr) {
```
oleorhagen, line 28: How badly did you wanna use the ternary?

common/conf/platform/boost_filesystem/paths.cpp
```diff
@@ -0,0 +1,52 @@
...
+//
+//    Unless required by applicable law or agreed to in writing, software
+//    distributed under the License is distributed on an "AS IS" BASIS,
+//    WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+//    See the License for the specific language governing permissions and
+//    limitations under the License.
+
+#include <common/conf/paths.hpp>
+
+#include <string>
```
oleorhagen, line 17: Q: Add a space in between?

common/CMakeLists.txt
```diff
@@ -155,3 +156,20 @@ target_link_libraries(inventory_parser_test PUBLIC common_inventory_parser commo
...
+
+
+add_library(common_conf STATIC conf/conf.cpp)
+target_link_libraries(common_conf INTERFACE crossplatform_compiler_flags)
+if(MENDER_USE_BOOST_FS)
+  find_package(Boost 1.54 REQUIRED COMPONENTS filesystem)
+  target_sources(common_conf PRIVATE conf/platform/boost_filesystem/paths.cpp)
+  target_include_directories(common_conf PRIVATE ${CMAKE_SOURCE_DIR} ${CMAKE_BINARY_DIR} ${Boost_INCLUDEDIR})
+  target_link_libraries(common_conf PRIVATE platform_compiler_flags)
+  target_link_libraries(common_conf ${Boost_libraries})
```
oleorhagen, line 168: Hmm, all of them? And is it in lowercaps? :thinking:

oleorhagen (approved):

common/conf/conf.cpp
vpodzime: It didn't even cross my mind, actually.

common/CMakeLists.txt
vpodzime: You mean all of the Boost libraries? I believe `find_package(Boost 1.54 REQUIRED COMPONENTS filesystem)` above sets this variable only to libraries identified by it.

common/conf/platform/boost_filesystem/paths.cpp
```diff
@@ -0,0 +1,52 @@
...
+const string DefaultKeyFile = "mender-agent.pem";
+
+const string DefaultConfFile = (fs::path(DefaultPathConfDir) / "mender.conf").string();
+const string DefaultFallbackConfFile = (fs::path(DefaultDataStore) / "mender.conf").string();
+
+// device specific paths
+const string DefaultArtScriptsPath = (fs::path(DefaultDataStore) / "scripts").string();
+const string DefaultRootfsScriptsPath = (fs::path(DefaultPathConfDir) / "scripts").string();
+const string DefaultModulesPath = (fs::path(DefaultPathDataDir) / "/modules" / "/v3").string();
+const string DefaultModulesWorkPath = (fs::path(DefaultDataStore) / "/modules" / "/v3").string();
```
kacf, line 45: Forgot to remove the `/` in the string (x4).

common/CMakeLists.txt
vpodzime: You mean with a an error message?

kacf: It will fail to compile anyway. No danger of this going unnoticed 🙂

kacf (approved):

common/CMakeLists.txt
oleorhagen: Oh, nice!

vpodzime: As long as it's the case. :sweat_smile:

PR #1156, opened 2023-02-20
Fix logger levels

oleorhagen:
Leftover fix from the logger, and one cleanup

* Now loggers no longer interfere with eachothers log-levels, as the filtering is done in each logger.
* Chore: Removed the standard log-level setting in Setup, as this is already done when constructing the class.

### DIFF
--- a/common/log.hpp
+++ b/common/log.hpp
@@ -81,6 +81,8 @@ private:
 
 	void AddField(const LogField &field);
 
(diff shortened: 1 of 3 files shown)


vpodzime (approved):
Cool, thanks! I think a little explanation of what was going on/what the fix is would be nice in the commit message of the first commit.

PR #1157, opened 2023-02-20
MEN-6309: Basic global options

vpodzime:


common/conf/conf.cpp
```diff
@@ -32,6 +57,33 @@ string GetEnv(const string &var_name, const string &default_value) {
...
+		}
+		if (arg.find("--fallback-config=") == 0) {
+			fallback_config_path = arg.substr(18, arg.size());
+		}
+	}
+
+	auto ret = this->LoadFile(config_path);
+	if (!ret) {
+		// XXX: this->Reset() ???
+		ret = this->LoadFile(fallback_config_path);
```
kacf, line 78: One clarification: The fallback is always loaded, but the settings in the main config override it per key.

vpodzime: Yeah, right. It's a fallback, but it's read first. Thanks for the heads-up!

mender-auth/main.cpp
```diff
@@ -12,10 +12,32 @@
...
+int main(int argc, char *argv[]) {
+	mender::common::conf::MenderConfig config;
+	if (argc > 1) {
+		vector<string> args(argv + 1, argv + argc);
+		auto err = config.ProcessCmdlineArgs(args);
+		if (err) {
+			cerr << "Failed to process command line options: " + err.message << endl;
+			return 1;
+		}
+	}
```
vpodzime, line 33: I realized (while driving a car home) we need to call something like `config.LoadDefaults()` here in an `else` branch. Will do it.

common/log/platform/boost/boost_log.cpp
```diff
@@ -184,6 +187,31 @@ void SetLevel(LogLevel level) {
 	global_logger_.SetLevel(level);
 }
 
+error::Error SetupFileLogging(const string &log_file_path, bool exclusive) {
+	typedef sinks::synchronous_sink<sinks::text_ostream_backend> text_sink;
+	boost::shared_ptr<text_sink> sink = boost::make_shared<text_sink>();
+
+	// Add a stream to write log to
+	auto log_stream = boost::make_shared<std::ofstream>(log_file_path);
```
oleorhagen, line 195: Should we have a short discussion on whether to use the rotating log-file backend?

I'm a little undecided. I _think_ we do in the old client, but I would have to double check.

common/log.hpp
```diff
@@ -74,7 +74,7 @@ enum class LogLevel {
 
 using ExpectedLogLevel = expected::Expected<LogLevel, error::Error>;
 
-inline string to_string_level(LogLevel lvl) {
+inline string ToStringLogLevel(LogLevel lvl) {
```
oleorhagen, line 77: Doesn't matter now, but this commit should be either a `chore` or a `style` category. This probably would not be a new patch release ;)

common/conf/conf.cpp
```diff
@@ -32,6 +59,71 @@ string GetEnv(const string &var_name, const string &default_value) {
...
+				ConfigErrorCode::InvalidOptionsError,
+				"Option " + option + " doesn't expect a value");
+			return ExpectedOptionValue(err);
+		}
+	} else {
+		value = args_[pos_];
+		pos_++;
+	}
+
+	return ExpectedOptionValue({std::move(option), std::move(value)});
```
oleorhagen, line 127: Q: I think the `move` is unnecessary here, as the values go out of scope, and will be automatically moved.

oleorhagen: I have to admit I find it a little harder to read when errors and regular expected returns look exactly the same.

common/conf_test.cpp
```diff
@@ -35,3 +36,221 @@ TEST(ConfTests, GetEnvTest) {
...
+		args, {"--opt1", "-o2", "--opt4", "-o6"}, {"--opt3", "--opt5", "-o7"});
+	auto ex_opt_val = opts_iter.Next();
+	ASSERT_FALSE(ex_opt_val);
+	EXPECT_EQ(ex_opt_val.error().message, "Option --opt1 missing value");
+}
+
+TEST(ConfTests, CmdlineOptionsIteratorOptionExtraValueTest) {
+	vector<string> args = {
+		"--opt3=val3",
+		"-o2",
```
oleorhagen, line 247: I don't remember what the conclusion was, but I thought we had to support `-o 2` also?

common/conf/conf.cpp
```diff
@@ -32,6 +59,71 @@ string GetEnv(const string &var_name, const string &default_value) {
 	}
 }
 
+ExpectedOptionValue CmdlineOptionsIterator::Next() {
```
oleorhagen, line 65: Well written! Didn't look too deep, but looks clean and mean, skimmed through it. Structure and organisation is neat. No problem to understand the intention here.

Good job!

oleorhagen: I'm not too sure about just the name `conf` though :thinking:

common/conf/conf.cpp
```diff
@@ -124,6 +127,69 @@ ExpectedOptionValue CmdlineOptionsIterator::Next() {
...
+		return ex_log_level.error();
+	}
+	SetLevel(ex_log_level.value());
+
+	auto ret = this->LoadFile(fallback_config_path);
+	if (!ret || !this->ValidateConfig()) {
+		this->Reset();
+	}
+	ret = this->LoadFile(config_path);
+	if (!ret || !this->ValidateConfig()) {
```
oleorhagen, line 187: :+1:

common/conf/paths.hpp
```diff
@@ -24,6 +24,8 @@ namespace paths {
 
 using namespace std;
 
+string JoinPaths(const string &prefix, const string &suffix);
```
oleorhagen, line 27: S:
`paths::Join` should be plenty I think. In case we will probably use this other places as well.

oleorhagen (approved):
Well done!

common/conf/conf.cpp
kacf: I suppose this will soon be changed to `unexpected` though, so they won't the same much longer.

common/conf_test.cpp
kacf: I think it does, it actually is doing `-o2 val1`. Using single-dash multi-letter options is not something are going to use, but perfectly fine that it supports it.

common/conf/conf.cpp
```diff
@@ -124,6 +127,69 @@ ExpectedOptionValue CmdlineOptionsIterator::Next() {
...
+		this->Reset();
+	}
```
kacf, line 189: Shouldn't we return error here?

common/conf/paths.hpp
kacf: +1

common/conf/conf.cpp
```diff
@@ -160,6 +160,7 @@ error::Error MenderConfig::ProcessCmdlineArgs(const vector<string> &args) {
 		} else if ((opt_val.option == "--log-level") || (opt_val.option == "-l")) {
 			log_level = opt_val.value;
 		}
+		ex_opt_val = opts_iter.Next();
```
kacf, line 163: Nitpick: Unrelated to "feat: Add MenderContext", should have been in the earlier commit which introduced this loop.

common/context.hpp
```diff
@@ -0,0 +1,53 @@
...
+class MenderContext {
+public:
+	error::Error Initialize(const conf::MenderConfig &config);
+	kv_db::KeyValueDatabase &GetMenderStoreDB();
+
+private:
+#if MENDER_USE_LMDB
+	kv_db::KeyValueDatabaseLmdb mender_store_;
+#else
+	kv_db::KeyValueDatabaseInMemory mender_store_;
```
kacf, line 45: Actually, it's better if we don't have a fallback. The in-memory one is definitely not suitable, so we should get a compile error instead.

kacf (commented):
Agree with Ole, nice work on the arguments loop!

common/log/platform/boost/boost_log.cpp
oleorhagen: I wasn't referring to the deployment rotation.

However, I did go and check the regular log-file functionality, and it does not seem we rotate it, just dump everything to file, so this is fine .

common/log/platform/boost/boost_log.cpp
```diff
@@ -184,6 +187,31 @@ void SetLevel(LogLevel level) {
...
+	if (!(log_stream.get())) {
```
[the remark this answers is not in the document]
vpodzime: `get()` only gives me the stream from the `boost::make_shared()` wrapper. It's the `!()` that checks the stream, using the [bool operator](https://en.cppreference.com/w/cpp/io/basic_ios/operator_bool). Or am I missing something?

common/conf/conf.cpp
vpodzime: I don't know which cases are errors, TBH. There are the cases in which I would return an error:
- failing to load the config given as a cmdline option, missing or invalid (because it was explicitly specified)
- failing to load the fallback config given as a cmdline option, missing or invalid (because it was explicitly specified)
- present but invalid config at the default path
- present but invalid fallback config at the default path

Do we have a spec of what the current Go-based implementation does?

common/conf/conf.cpp
vpodzime: Good catch!

common/log.hpp
vpodzime: You don't like it's **fixing** the ugly name you gave it? :stuck_out_tongue_winking_eye:

common/conf/conf.cpp
vpodzime: I don't like too, TBH. But it's only used by the `MenderConfig` class now and wasn't there unit-testing, it would be solely in the `.cpp` file.

common/conf/paths.hpp
vpodzime: :+1:

common/log.hpp
oleorhagen: :laughing:

common/log/platform/boost/boost_log.cpp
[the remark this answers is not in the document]
kacf: Ah ok, yes, that will also work, I think. But there is a sneaky bug then, because `.get()` returns a raw pointer, so what you're testing is the C-style "is pointer non-null?". To do it correctly you need to write: `(!(*log_stream.get()))`.

I would argue that using `->bad()` is perhaps a bit more explicit and harder to get wrong, but I won't object to the `operator bool` style.

common/conf/conf.cpp
kacf: The Golang client returns errors for all of those cases, apparently.

common/log/platform/boost/boost_log.cpp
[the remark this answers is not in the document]
vpodzime: I used `!()` because `bool` is a bit more than just `bad()`:
image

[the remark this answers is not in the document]
vpodzime: > Ah ok, yes, that will also work, I think. But there is a sneaky bug then, because .get() returns a raw pointer

Oh, of course. But I was told there will be no pointers in C++! :wink:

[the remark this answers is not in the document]
kacf: Hehe, yeah, that one snuck up on you from under a rock. You should quickly bury it again!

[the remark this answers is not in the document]
kacf: > I used `!()` because `bool` is a bit more than just `bad()`:

You could use `!good()`, which has the same response as the bool variant.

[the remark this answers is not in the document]
vpodzime: It doesn't, actually, it differs in the way it treats `eof()`.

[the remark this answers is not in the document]
kacf: Bah, you're right. So many variants there... But `fail()` is exactly inverse bool, isn't it?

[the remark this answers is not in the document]
vpodzime: I like the use of the `bool` operator (even though I have to say the code is quite cryptic, but that's just C++ syntax to blame).

common/conf/platform/boost_filesystem/paths.cpp
```diff
@@ -28,24 +28,24 @@ using namespace std;
...
-const string DefaultArtScriptsPath = (fs::path(DefaultDataStore) / "scripts").string();
-const string DefaultRootfsScriptsPath = (fs::path(DefaultPathConfDir) / "scripts").string();
-const string DefaultModulesPath = (fs::path(DefaultPathDataDir) / "modules" / "v3").string();
-const string DefaultModulesWorkPath = (fs::path(DefaultDataStore) / "modules" / "v3").string();
-const string DefaultBootstrapArtifactFile =
-	(fs::path(DefaultDataStore) / "bootstrap.mender").string();
+const string DefaultArtScriptsPath = Join(DefaultDataStore, "scripts");
+const string DefaultRootfsScriptsPath = Join(DefaultPathConfDir, "scripts");
+const string DefaultModulesPath = Join(DefaultPathDataDir, "modules/v3");
+const string DefaultModulesWorkPath = Join(DefaultDataStore, "modules/v3");
```
kacf, line 47: Should `modules/v3` also be joined?

kacf (approved):

common/conf/platform/boost_filesystem/paths.cpp
vpodzime: I don't exactly know, TBH. `boost::fs::path` docs mention the [*portable format*](https://www.boost.org/doc/libs/1_39_0/libs/filesystem/doc/reference.html#Pathname-formats) for paths (basically POSIX paths) which should be accepted and properly handled. So my understanding is that by converting `DefaultDataStore` to `boost::fs::path`, appending to it should work fine with paths in the portable format. But like I said, not 100% sure here, I'd have to check it on a system that has a different path format, but no such is easily available for me.

common/log/platform/boost/boost_log.cpp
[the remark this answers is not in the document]
vpodzime: It's just the semantics of the `bool` operator which is something like "is this file stream OK?" and that makes most sense to me. Rather than always going to the docs checking what exactly `fail()`, `bad()`,... mean. But I will, hopefully, change this code soon anyway.

PR #1158, opened 2023-02-21
I/O utilities

oleorhagen:
This adds to new utilities to the `io` library.

1. A `StringReader` to easily wrap a string in a `Reader` interface.
2. A `Discard` writer. Pretty much a `/dev/null` for io Read/Writers.

### DIFF
--- a/common/io.hpp
+++ b/common/io.hpp
@@ -79,6 +79,13 @@ public:
 	}
 };
 
+/* Discards all data written to it */
+class Discard : virtual public Writer {
+	ExpectedSize Write(const vector<uint8_t> &dst) override {
+		return dst.size();
+	}
+};
+
 } // namespace io
 } // namespace common
 } // namespace mender
--- a/common/io.hpp
+++ b/common/io.hpp
@@ -22,6 +22,8 @@
 #include <system_error>
 #include <vector>
 #include <istream>
+#include <sstream>
+#include <string>
 
 namespace mender {
 namespace common {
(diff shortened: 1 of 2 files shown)


common/io.hpp
```diff
@@ -86,6 +88,26 @@ class Discard : virtual public Writer {
...
 };
 
+class StringReader : virtual public Reader {
+private:
+	StreamReader reader_;
+	std::stringstream s_;
+
+public:
+	StringReader(string &str) :
+		s_ {str},
```
vpodzime, line 98: Any reason why you prefer the `{}` initializers instead of parentheses? You seem to be the only one using them.

vpodzime (approved):
> Whenever you want to empty a reader.

Wouldn't it be better to have a way to empty a reader? Having the `Discard` writer makes sense to me, but this use case just suggests wasting CPU cycles.

common/io.hpp
oleorhagen: It's just cause it's what I've learnt, and it is what is recommended to use these days appareantly: https://github.com/isocpp/CppCoreGuidelines/blob/master/CppCoreGuidelines.md#Res-list.

 No hard preference here though.

oleorhagen:
> > Whenever you want to empty a reader.
> 
> Wouldn't it be better to have a way to empty a reader? Having the `Discard` writer makes sense to me, but this use case just suggests wasting CPU cycles.

Mby. It reads pretty nice though I think, and it's a nice drop-in for instance in tests. So in a lot of patterns you will wrap a `Reader` in another `Reader` for instance, and then writer to whatever. But when testing this, you don't necessarily care where you write to. In this case the `Discard` writer is a really nice stand-in.

vpodzime:
> In this case the `Discard` writer is a really nice stand-in.

Yup! Like I said, it makes sense to me. But not to waste CPU cycles outside of tests.

PR #1159, opened 2023-02-21
expected lib alignments with the standard library

oleorhagen:
This adds a few methods to align our `mender::expected` library to align closer with the standard librarys `expected` implementation.

The commits should contain most the information needed.


kacf (approved):
Beautiful!

PR #1160, opened 2023-02-21
feat: Add the shasum library to be used by the artifact parser

oleorhagen:
This adds a simple shasum library which wraps openssl as the shasum provider.
The default shasum used is 256.

Ticket: MEN-6200


oleorhagen:
Relies on https://github.com/mendersoftware/mender/pull/1158

artifact/sha/sha.cpp
```diff
@@ -0,0 +1,126 @@
...
+error::Error MakeError(ErrorCode code, const string &msg) {
+	return error::Error(error_condition(code, ErrorCategory), msg);
+}
+
+
+Reader::Reader(io::Reader &reader) :
+	Reader::Reader {reader, ""} {
+}
+
+Reader::Reader(io::Reader &reader, std::string expected_sha = "") :
```
vpodzime, line 52: &expected_sha

artifact/sha/sha.hpp
```diff
@@ -0,0 +1,77 @@
...
+
+class Reader : virtual public io::Reader {
+private:
+#ifdef MENDER_SHA_OPENSSL
+	std::unique_ptr<EVP_MD_CTX, void (*)(EVP_MD_CTX *)> sha_handle_;
+#endif
+	io::Reader &wrapped_reader_;
+	vector<uint8_t> buffer_;
+	std::string expected_sha_ {};
+	bool initialized {false};
```
vpodzime, line 56: `initialized_`

artifact/sha/sha.cpp
```diff
@@ -0,0 +1,126 @@
...
+	sha_handle_(EVP_MD_CTX_new(), [](EVP_MD_CTX *ctx) { EVP_MD_CTX_free(ctx); }),
+	wrapped_reader_ {reader},
+	buffer_(1024),
```
vpodzime, line 55: Why such a small buffer?

artifact/sha/sha.cpp
```diff
@@ -0,0 +1,126 @@
...
+		return bytes_read;
+	}
+
+	// bytes_read == 0 == EOF marker in our Reader/Writer interface implementation
+	if (bytes_read.value() == 0) {
+		auto expected_sha = ShaSum();
+		if (!expected_sha) {
+			return expected_sha.error();
+		}
+		auto actual_sha = expected_sha.value();
```
vpodzime, line 81: `expected_sha` is a really confusing name here.

artifact/sha/sha.cpp
```diff
@@ -0,0 +1,126 @@
...
+	}
+
+	if (EVP_DigestUpdate(sha_handle_.get(), buffer_.data(), bytes_read.value()) != 1) {
+		return MakeError(ShasumCreationError, "Failed to create the shasum");
+	}
+
+	// Pass on the underlying stream
+	for (int i = 0; i < std::min(dst.size(), bytes_read.value()); ++i) {
+		dst[i] = buffer_[i];
+	}
```
vpodzime, line 98: There must be a better way to do this than one byte at a time.

artifact/sha/sha.cpp
```diff
@@ -0,0 +1,126 @@
...
+
+	return bytes_read.value();
+}
+
+expected::ExpectedString Reader::ShaSum() {
+	vector<uint8_t> hash(EVP_MAX_MD_SIZE);
+	unsigned int hash_length = 0;
+
+	if (EVP_DigestFinal_ex(sha_handle_.get(), hash.data(), &hash_length) != 1) {
+		return MakeError(ShasumCreationError, "Failed to create the shasum. OpenSSL error: ");
```
vpodzime, line 108: Missing stuff for the OpenSSL error?

artifact/sha/sha.cpp
```diff
@@ -0,0 +1,126 @@
...
+	}
+
+	if (hash_length != 32) {
```
vpodzime, line 111: This should be a named constant with a proper name.

artifact/sha/sha.hpp
```diff
@@ -0,0 +1,77 @@
...
+#include <memory>
+
+#include <config.h>
+
+#include <common/io.hpp>
+#include <common/error.hpp>
+#include <common/log.hpp>
+
+#include <common/expected.hpp>
+#include <string>
```
vpodzime, line 14: Group the STL/std includes, please.

vpodzime (commented):
Looks good to me otherwise.

artifact/sha/sha.cpp
```diff
@@ -0,0 +1,126 @@
...
+	ShasumMismatchError:
+		return "Shasum mismatch error";
+	default:
+		return "Unknown";
+	}
+}
+
+error::Error MakeError(ErrorCode code) {
+	return error::Error(error_condition(code, ErrorCategory), "");
+}
```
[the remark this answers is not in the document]
kacf: I agree. We should feel encouraged to add a message, or else we'll have the ugly empty string stare back at us! :smile:

artifact/sha/sha.cpp
kacf: Yes, but use `const` if you use reference (unless you intend to modify it).

artifact/sha/sha.cpp
kacf: Lol, yes. You are only using `actual_sha` twice, one of which is an error path, so maybe just rename `expected_sha` to `actual_sha`, and use it directly with the `value()` function.

artifact/sha/sha.cpp
kacf: Suggest using this:

```
std::copy(buffer_.begin(), buffer_.end(), std::inserter(dst, dst.begin()));
```

Note the use of `buffer_.end()` which relates to my other comment about vector sizes.

artifact/sha/sha.cpp
```diff
@@ -0,0 +1,126 @@
...
+	}
+	initialized = true;
+}
+
+expected::ExpectedSize Reader::Read(vector<uint8_t> &dst) {
+	if (!initialized) {
+		return MakeError(InitializationError);
+	}
+
+	auto bytes_read = wrapped_reader_.Read(buffer_);
```
kacf, line 70: This is incorrect. The buffer needs to be resized to `min(dst.size(), buffer_.size())`. Otherwise it will start losing bytes because it only puts `dst.size()` bytes into that buffer and anything overshooting that, is lost. Then it needs to resized back to original capacity afterwards.

In general I have wondered if using direct `vector`s was a mistake, and we should have had this interface instead:

```
ExpectedSize Read(vector<uint8_t>::iterator start, vector<uint8_t>::iterator end);
```

This aligns better with how the rest of library works, and is also closer, at least in philosophy, to the Golang slices. Do you think we should consider mass-changing it?

kacf (changes requested):

artifact/sha/sha.cpp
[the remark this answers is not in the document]
oleorhagen: kk, I give

artifact/sha/CMakeLists.txt
```diff
@@ -0,0 +1,28 @@
+
+find_package(OpenSSL REQUIRED)
```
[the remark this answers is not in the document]
oleorhagen: It's the same thing I think:
https://cmake.org/cmake/help/latest/command/find_package.html

:thinking:

artifact/sha/sha.cpp
oleorhagen: Yes, horrible naming. What about something like `stream_sha`?
* `calculated_sha`
* `real_sha`

artifact/sha/sha.cpp
oleorhagen: Beautiful! 

I was searching for something like this :tada:

artifact/sha/sha.cpp
oleorhagen: You mean like a separate error code?

artifact/sha/sha.cpp
oleorhagen: Come to think of it. Do you really think this is needed? :thinking:  Maybe we should just rely on `OpenSSL` defaults here ?

artifact/sha/sha.cpp
oleorhagen: hah, I just noticed it actually :p

oleorhagen: Yes, I was just thinking of writing something similar actually, cause I've been runnning into this very same problem in the `tar-reader` implementation too. It's really annoying handling cases in which `size(dst) < size(buf)`, and you have to do intermediate caching, returning, and then fill your buffer anew.

oleorhagen: I had this beautiful `// TODO - What if 'size(dst) < size(buf)'` in my notes :sweat:

kacf: Just to conclude: Ole will fix the function using the current interface, and I will take a look at changing our interface to use iterators instead. But this will be in a followup.

oleorhagen: For now, I solved it by adding a buf-resizer class, which resizes on destruction.

oleorhagen: See: https://github.com/mendersoftware/mender/pull/1160/commits/365700938f8bc1da15a5d60f402ceb21d951b0ac#diff-2c09a25897ac9f57ed5e29eaec90863a06d6656551268cffa1c8497f92291166R72

kacf (approved):
Looks good apart from the commit messages (but I assume you will squash them).

oleorhagen:
> Looks good apart from the commit messages (but I assume you will squash them).

Was planning on it :laughing:

artifact/sha/CMakeLists.txt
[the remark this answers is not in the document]
vpodzime: Why would they add `FindOpenSSL()` then?

artifact/sha/platform/openssl/sha.cpp
```diff
@@ -0,0 +1,155 @@
...
+		buf.resize(original_size);
+	}
+};
+
+expected::ExpectedSize Reader::Read(vector<uint8_t> &dst) {
+	if (!initialized_) {
+		return MakeError(InitializationError, "");
+	}
+
+	auto buffer_resizer = BufferResizer(buffer_, buffer_.size(), dst.size());
```
vpodzime, line 100: Wouldn't it better to allow access to the buffer of the `wrapped_reader`? Why do we need another buffer here and copy things?

vpodzime (commented):
A couple comments. Sorry for being late to the party.

artifact/sha/platform/openssl/sha.cpp
```diff
@@ -0,0 +1,155 @@
...
+		return MakeError(ShasumCreationError, "Failed to create the shasum");
+	}
+
+	// Pass on the underlying stream
+	std::copy_n(buffer_.begin(), bytes_read.value(), std::inserter(dst, dst.begin()));
+
+	return bytes_read.value();
+}
+
+expected::ExpectedString Reader::ShaSum() {
```
vpodzime, line 132: Missing `if (!initialized_)...`

oleorhagen:
> A couple comments. Sorry for being late to the party.

Not at all, they we're good finds. Fixing now!

artifact/sha/platform/openssl/sha.cpp
oleorhagen: Hmm, it's actually a really good point I think!

I can just copy into `dst` directly, and use it internally! 

Good :eye:

kacf: Note: the BufferResizer is being removed in #1164.

oleorhagen: It is also removed in: https://github.com/mendersoftware/mender/pull/1166

oleorhagen:
late-review fixes here: https://github.com/mendersoftware/mender/pull/1166

PR #1162, opened 2023-02-23
(artifact-parser): Add the Version header parser

oleorhagen:
This adds a simple parser for the Mender-Artifact v3 version header.

As well as a small utility ByteWriter class to io.

I do have to admit that I feel this turned out needlessly complicated, so all open for advice on how to improve this!

### DIFF
--- a/artifact/v3/version/CMakeLists.txt
+++ b/artifact/v3/version/CMakeLists.txt
@@ -1,0 +1,19 @@
+
+# Test the parser
+add_executable(artifact_version_parser_test EXCLUDE_FROM_ALL
+  version.cpp
+  version_test.cpp
+)
+target_link_libraries(artifact_version_parser_test PRIVATE
+  GTest::gtest_main
+  gmock
+  common_io
+  common_error
+  common_json
+)
+target_include_directories(artifact_version_parser_test PRIVATE
+  ${CMAKE_SOURCE_DIR}
+  ${CMAKE_BINARY_DIR}
+)
+gtest_discover_tests(artifact_version_parser_test)
+add_dependencies(check artifact_version_parser_test)
--- a/artifact/v3/version/version.cpp
+++ b/artifact/v3/version/version.cpp
@@ -1,0 +1,117 @@
+// Copyright 2023 Northern.tech AS
+//
+//    Licensed under the Apache License, Version 2.0 (the "License");
+//    you may not use this file except in compliance with the License.
+//    You may obtain a copy of the License at
+//
+//        http://www.apache.org/licenses/LICENSE-2.0
+//
+//    Unless required by applicable law or agreed to in writing, software
+//    distributed under the License is distributed on an "AS IS" BASIS,
+//    WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+//    See the License for the specific language governing permissions and
+//    limitations under the License.
+
+#include <artifact/v3/version/version.hpp>
+
+#include <string>
+#include <vector>
+
+#include <common/common.hpp>
+#include <common/expected.hpp>
+#include <common/error.hpp>
+#include <common/json.hpp>
+
+
+namespace artifact {
+namespace v3 {
+namespace version {
+
+namespace io = mender::common::io;
+namespace expected = mender::common::expected;
+namespace error = mender::common::error;
+namespace json = mender::common::json;
+
+const int supported_parser_version {3};
+const string supported_parser_format {"mender"};
+
+const ErrorCategoryClass ErrorCategory {};
+
+const char *ErrorCategoryClass::name() const noexcept {
+	return "VersionParserErrorCategory";
+}
+
+string ErrorCategoryClass::message(int code) const {
+	switch (code) {
+	case NoError:
+		return "Success";
+	case ParseError:
+		return "Parse error";
+	case VersionError:
+		return "Wrong Artifact version";
+	case FormatError:
+		return "Wrong Artifact format";
+	default:
+		return "Unknown";
+	}
+}
+
+error::Error MakeError(ErrorCode code, const string &msg) {
+	return error::Error(error_condition(code, ErrorCategory), msg);
+}
+
+
+ExpectedVersion Parse(io::Reader &reader) {
+	// Read in all the bytes into a tmp buffer
+	std::vector<uint8_t> buf(1024);
+	io::ByteWriter bw {buf};
+
+	auto err = io::Copy(bw, reader);
+	if (err) {
+		return expected::unexpected(err);
+	}
+
+	auto expected_json = json::LoadFromString(mender::common::StringFromByteVector(buf));
+
+	if (!expected_json) {
+		return expected::unexpected(MakeError(
+			ParseError,
+			"Failed to parse the version header JSON: " + expected_json.error().message));
+	}
+
+	const json::Json version_json = expected_json.value();
+
+	auto version =
+		version_json.Get("version").and_then([](const json::Json &j) { return j.GetInt(); });
+
+	if (!version) {
+		return expected::unexpected(MakeError(VersionError, version.error().message));
+	}
+
+	if (version.value() != supported_parser_version) {
+		return expected::unexpected(MakeError(
+			FormatError,
+			"Only version " + std::to_string(supported_parser_version)
+				+ " is supported, received version " + std::to_string(version.value())));
+	}
+
+	auto format =
+		version_json.Get("format").and_then([](const json::Json &j) { return j.GetString(); });
+
+	if (!format) {
+		return expected::unexpected(MakeError(FormatError, format.error().message));
+	}
+
+	if (format != supported_parser_format) {
+		return expected::unexpected(MakeError(
+			FormatError,
+			"The client only understands the 'mender' Artifact type. Got format: "
+				+ format.value()));
+	}
+
+	return Version {.version = supported_parser_version, .format = supported_parser_format};
+}
+
+} // namespace version
+} // namespace v3
+} // namespace artifact
--- a/artifact/v3/version/version.hpp
+++ b/artifact/v3/version/version.hpp
@@ -1,0 +1,65 @@
+// Copyright 2023 Northern.tech AS
+//
+//    Licensed under the Apache License, Version 2.0 (the "License");
+//    you may not use this file except in compliance with the License.
+//    You may obtain a copy of the License at
+//
+//        http://www.apache.org/licenses/LICENSE-2.0
+//
+//    Unless required by applicable law or agreed to in writing, software
+//    distributed under the License is distributed on an "AS IS" BASIS,
+//    WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+//    See the License for the specific language governing permissions and
+//    limitations under the License.
+
+#ifndef MENDER_ARTIFACT_VERSION_PARSER_HPP
+#define MENDER_ARTIFACT_VERSION_PARSER_HPP
+
+#include <string>
+
+#include <common/expected.hpp>
+#include <common/error.hpp>
+#include <common/io.hpp>
+
+namespace artifact {
+namespace v3 {
+namespace version {
+
+using namespace std;
+
+namespace io = mender::common::io;
+namespace expected = mender::common::expected;
+namespace error = mender::common::error;
+
+
+enum ErrorCode {
+	NoError = 0,
+	ParseError,
+	VersionError,
+	FormatError,
+};
+
+class ErrorCategoryClass : public std::error_category {
+public:
+	const char *name() const noexcept override;
+	string message(int code) const override;
+};
+
+extern const ErrorCategoryClass ErrorCategory;
+
+error::Error MakeError(ErrorCode code, const string &msg);
+
+struct Version {
+	const int version;
+	const string format;
+};
+
+using ExpectedVersion = expected::expected<Version, error::Error>;
+
+ExpectedVersion Parse(io::Reader &reader);
+
+} // namespace version
+} // namespace v3
+} // namespace artifact
+
+#endif // MENDER_ARTIFACT_VERSION_PARSER_HPP
--- a/artifact/v3/version/version_test.cpp
+++ b/artifact/v3/version/version_test.cpp
@@ -1,0 +1,132 @@
+// Copyright 2023 Northern.tech AS
+//
+//    Licensed under the Apache License, Version 2.0 (the "License");
+//    you may not use this file except in compliance with the License.
+//    You may obtain a copy of the License at
+//
+//        http://www.apache.org/licenses/LICENSE-2.0
+//
+//    Unless required by applicable law or agreed to in writing, software
+//    distributed under the License is distributed on an "AS IS" BASIS,
+//    WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+//    See the License for the specific language governing permissions and
+//    limitations under the License.
+
+#include <artifact/v3/version/version.hpp>
+
+#include <gtest/gtest.h>
+#include <gmock/gmock.h>
+
+#include <fstream>
+
+
+using namespace std;
+
+
+TEST(ParserTest, TestParseVersion) {
+	std::string json_data = R"(
+  {
+    "version": 3,
+		"format" : "mender"
+  }
+)";
+
+	std::stringstream ss {json_data};
+
+	mender::common::io::StreamReader sr {ss};
+
+	auto version = artifact::v3::version::Parse(sr);
+
+	ASSERT_TRUE(version) << version.error().message << std::endl;
+
+	auto version_unwrapped = version.value();
+
+	EXPECT_EQ(version_unwrapped.version, 3);
+	EXPECT_EQ(version_unwrapped.format, "mender");
+}
+
+TEST(ParserTest, TestParseWrongVersion) {
+	// We don't support version 2
+	std::string json_data = R"(
+  {
+    "version": 2,
+		"format" : "mender"
+  }
+)";
+
+	std::stringstream ss {json_data};
+
+	mender::common::io::StreamReader sr {ss};
+
+	auto version = artifact::v3::version::Parse(sr);
+
+	ASSERT_FALSE(version) << version.error().message << std::endl;
+
+	auto expected_error_message = "Only version 3 is supported, received version 2";
+
+	EXPECT_EQ(version.error().message, expected_error_message);
+}
+
+
+TEST(ParserTest, TestParseWrongFormat) {
+	// We don't support other formats than mender atm
+	std::string json_data = R"(
+  {
+    "version": 3,
+		"format" : "foobar"
+  }
+)";
+
+	std::stringstream ss {json_data};
+
+	mender::common::io::StreamReader sr {ss};
+
+	auto version = artifact::v3::version::Parse(sr);
+
+	ASSERT_FALSE(version) << version.error().message << std::endl;
+
+	auto expected_error_message =
+		"The client only understands the 'mender' Artifact type. Got format: foobar";
+
+	EXPECT_EQ(version.error().message, expected_error_message);
+}
+
+TEST(ParserTest, TestParseMumboJumbo) {
+	std::string json_data = R"(
+foobarbaz
+)";
+
+	std::stringstream ss {json_data};
+
+	mender::common::io::StreamReader sr {ss};
+
+	auto version = artifact::v3::version::Parse(sr);
+
+	ASSERT_FALSE(version) << version.error().message << std::endl;
+
+	auto expected_error_message = "Failed to parse the version header JSON";
+
+	EXPECT_THAT(version.error().message, testing::StartsWith(expected_error_message));
+}
+
+
+TEST(ParserTest, TestParseMalformedInput) {
+	// Missing ending {
+	std::string json_data = R"(
+  {
+    "version": 3,
+		"format" : "mender"
+)";
+
+	std::stringstream ss {json_data};
+
+	mender::common::io::StreamReader sr {ss};
+
+	auto version = artifact::v3::version::Parse(sr);
+
+	ASSERT_FALSE(version) << version.error().message << std::endl;
+
+	auto expected_error_message = "Failed to parse the version header JSON";
+
+	EXPECT_THAT(version.error().message, testing::StartsWith(expected_error_message));
+}
--- a/artifact/CMakeLists.txt
+++ b/artifact/CMakeLists.txt
@@ -1,1 +1,2 @@
 add_subdirectory(sha)
+add_subdirectory(v3/version)


artifact/version/version.cpp
```diff
@@ -0,0 +1,124 @@
...
+		return MakeError(ParseError, expected_json.error().message);
+	}
+
+	const json::ExpectedJson expected_version = expected_json.value();
+
+	if (!expected_version) {
+		return MakeError(ParseError, expected_json.error().message);
+	}
+
+	auto expected_obj = expected_version.value();
```
kacf, line 70: I'm somewhat confused by this code. Is there three layers of ExpectedJson inside each other? Line 64 looks like a no-op, you are just assigning a payload value to an ExpectedJson, and since it is not an error, it will always construct an ExpectedJson with a value, and the next `if` will always evaluate false.

But feel free to enlighten me if I'm misunderstanding this.

artifact/version/version.cpp
```diff
@@ -0,0 +1,124 @@
...
+
+	if (!expected_obj.IsObject()) {
+		return MakeError(
+			ParseError,
```
kacf, line 74: The JSON is valid though, so should this be `FormatError` instead? It's JSON, but a format we don't understand.

artifact/version/version.cpp
```diff
@@ -0,0 +1,124 @@
...
+			"Failed to unmarshal the version from the version JSON data. The received data does not contain a single top-level object");
+	}
+
+	auto version = expected_obj.Get("version");
+	if (!version) {
+		return MakeError(
+			ParseError,
```
kacf, line 81: Same here, maybe `FormatError`.

artifact/version/version.cpp
```diff
@@ -0,0 +1,124 @@
...
+			"Failed to unmarshal the version from the version JSON data. No 'version' key found");
+	}
+	auto version_unwrapped = version.value().GetInt();
+	int version_;
```
kacf, line 85: Trailing underscores are for member variables.

artifact/version/version.hpp
```diff
@@ -0,0 +1,49 @@
...
+#define MENDER_ARTIFACT_VERSION_PARSER_HPP
+
+#include <string>
+
+#include <common/expected.hpp>
+#include <common/error.hpp>
+#include <common/io.hpp>
+
+namespace artifact {
+namespace v3 {
```
kacf, line 11: 👍

But should the files be inside `v3` as well?

artifact/version/version.cpp
```diff
@@ -0,0 +1,124 @@
...
+		return MakeError(
+			ParseError,
+			"Failed to unmarshal the version from the version JSON data. No 'format' field found");
+	}
+	auto format_unwrapped = format.value().GetString();
+	if (!format_unwrapped) {
+		return MakeError(
+			ParseError,
+			"Failed to unmarshal the version from the version JSON data. The 'format' fields needs to contain a string");
+	}
```
kacf, line 104: Suggestion: You could shorten this by doing:

```c++
auto format = expected_obj.Get("format");
if (!format || !format.value().GetString()) {
	return MakeError(
		ParseError,
		"Failed to unmarshal the version from the version JSON data. 'format' field not found or not a string");
}
auto format_unwrapped = format.value().GetString();
```

And similar with `version` above. Not saying it's better though, verbosity is always good. Just a suggestion.

artifact/version/version.cpp
```diff
@@ -0,0 +1,124 @@
...
+	if (format_unwrapped.value() != "mender") {
+		return MakeError(
+			FormatError,
+			"Failed to unmarshal the version from the version JSON data. The client only understands the 'mender' Artifact type");
+	}
+
+	if (version_ != 3) {
+		return MakeError(
+			VersionError,
+			"Failed to unmarshal the version from the version JSON data. Only 'version' 3 is supported received version "
```
kacf, line 115: > Failed to unmarshal the version from the version JSON data

You've used this message in nearly every error so far. Maybe you should make it part of the error code string instead, and use only the more specific strings in this function?

artifact/version/version.hpp
```diff
@@ -0,0 +1,49 @@
...
+namespace io = mender::common::io;
+namespace expected = mender::common::expected;
+namespace error = mender::common::error;
+
+
+enum ErrorCode {
+	NoError = 0,
+	ParseError,
+	VersionError,
+	FormatError,
```
kacf, line 25: I have a question about this, but it's an abstract question, hard to put into words here. Ping me when you have time.

artifact/version/version.cpp
oleorhagen: I don't even know. I think this is horrible...

However, I have not come up with a cleaner way...

oleorhagen: Alright, cleaning it up a little...

artifact/v3/version/version.cpp
```diff
@@ -0,0 +1,124 @@
...
+
+	auto err = io::Copy(bw, reader);
+	if (err) {
+		return err;
+	}
+
+	auto expected_json = json::LoadFromString(mender::common::StringFromByteVector(buf));
+
+	if (!expected_json) {
+		return MakeError(ParseError, "Failed to parse the version JSON");
```
vpodzime, line 78: Might as well just return `expected_json.error()` here.

artifact/v3/version/version.cpp
```diff
@@ -0,0 +1,124 @@
...
+	int version {-1};
+	string format {"unknown"};
+
+
+	auto err = io::Copy(bw, reader);
+	if (err) {
+		return err;
+	}
+
+	auto expected_json = json::LoadFromString(mender::common::StringFromByteVector(buf));
```
vpodzime, line 75: We copy the data into a buffer only to create a string from it here and throw the buffer away. It would be needed if `StringFromByteVector` took a reference and potentially modified the given buffer, but it is (I believe) a `const` reference and so we can safely pass in a buffer without creating a copy of it.

artifact/v3/version/version.cpp
```diff
@@ -0,0 +1,124 @@
...
+expected::Expected<Version, error::Error> Parse(io::Reader &reader) {
+	// Read in all the bytes into a tmp buffer
+	std::vector<uint8_t> buf(1024);
+	io::ByteWriter bw {buf};
+
+	int version {-1};
+	string format {"unknown"};
+
+
+	auto err = io::Copy(bw, reader);
```
vpodzime, line 70: No a real issue here, I believe, but what if the whole JSON doesn't fit into the buffer?

artifact/v3/version/version.cpp
```diff
@@ -0,0 +1,124 @@
...
+			FormatError,
+			"Failed to unmarshal the version from the version JSON data. The received data does not contain a single top-level object");
+	}
+
+	const json::ExpectedJson version_value = version_json.Get("version");
+	if (!version_value) {
+		return MakeError(FormatError, "No version field found");
+	}
+
+	const json::ExpectedInt value_version = version_value.value().GetInt();
```
vpodzime, line 94: Out of  scope of this PR, but `ExpectedInt` should be in the `expected` namespace/header, I think.

vpodzime: Please read this line (maybe out loud) and I'm pretty sure you will be able to come up with better variable names.

artifact/v3/version/version.cpp
```diff
@@ -0,0 +1,124 @@
...
+	if (value_version) {
+		version = value_version.value();
+		if (version != 3) {
+			return MakeError(
+				VersionError,
+				"Failed to unmarshal the version from the version JSON data. Only 'version' 3 is supported received version "
+					+ std::to_string(version));
```
vpodzime, line 101: Failing to get the version integer should also be an error, right?

artifact/v3/version/version.cpp
```diff
@@ -0,0 +1,124 @@
...
+		}
+	}
+
+	const json::ExpectedJson format_value = version_json.Get("format");
+	if (version_value) {
```
vpodzime, line 106: `if (format_value)`?

artifact/v3/version/version.cpp
```diff
@@ -0,0 +1,124 @@
...
+		const json::ExpectedString value_format = format_value.value().GetString();
+		if (value_format) {
+			format = value_format.value();
+			if (format != "mender") {
+				return MakeError(
+					FormatError,
+					"Failed to unmarshal the 'format' from the version JSON data. The client only understands the 'mender' Artifact type. Got format: "
+						+ format);
```
vpodzime, line 114: Again, `!value_format` is an error too, right?

artifact/v3/version/version.cpp
```diff
@@ -0,0 +1,124 @@
...
+			}
+		}
+	}
+
+	return Version {.version = 3, .format = "mender"};
```
vpodzime, line 119: It feels a bit strange to have a function called `Parse()` that doesn't actually return any parsed data, but not a big deal.

artifact/v3/version/version.hpp
```diff
@@ -0,0 +1,63 @@
...
+};
+
+extern const ErrorCategoryClass ErrorCategory;
+
+error::Error MakeError(ErrorCode code, const string &msg);
+
+struct Version {
+	const int version;
+	const string format;
+};
```
vpodzime, line 55: Similar here, a bit weird to have a type for something that's actually just a hard-coded constant pair `3-"mender"`.

artifact/v3/version/version_test.cpp
```diff
@@ -0,0 +1,133 @@
...
+//    WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+//    See the License for the specific language governing permissions and
+//    limitations under the License.
+
+#include <artifact/v3/version/version.hpp>
+
+#include <gtest/gtest.h>
+#include <gmock/gmock.h>
+
+#include <fstream>
```
vpodzime, line 20: `sstream`?

artifact/v3/version/version.cpp
```diff
@@ -0,0 +1,124 @@
...
+
+	const json::ExpectedJson version_value = version_json.Get("version");
+	if (!version_value) {
+		return MakeError(FormatError, "No version field found");
+	}
+
+	const json::ExpectedInt value_version = version_value.value().GetInt();
+	if (value_version) {
+		version = value_version.value();
+		if (version != 3) {
```
kacf, line 97: Come to think of it, maybe this should be a constant.

artifact/v3/version/version.cpp
```diff
@@ -0,0 +1,124 @@
...
+			"Failed to unmarshal the version from the version JSON data. The received data does not contain a single top-level object");
+	}
+
+	const json::ExpectedJson version_value = version_json.Get("version");
+	if (!version_value) {
+		return MakeError(FormatError, "No version field found");
+	}
+
+	const json::ExpectedInt value_version = version_value.value().GetInt();
+	if (value_version) {
```
kacf, line 95: This logic is wrong now. If the `version` field is not an int, the code continues, but it should fail.

artifact/v3/version/version.cpp
kacf: Yes, plus code continues on error here as well.

artifact/v3/version/version.cpp
```diff
@@ -0,0 +1,124 @@
...
+
+#include <common/common.hpp>
+#include <common/expected.hpp>
+#include <common/error.hpp>
+#include <common/json.hpp>
+
+#include <artifact/v3/version/version.hpp>
+
+namespace artifact {
+namespace v3 {
```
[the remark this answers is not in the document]
oleorhagen: Hmm... Why not?

artifact/v3/version/version.cpp
oleorhagen: The absolute best would be if `json` took a `Reader`, and I wouldn't have to do any of this.

Since ur intimately familiar with this. Would it be a lot of work to add you think?

artifact/v3/version/version.cpp
oleorhagen: Then we've received some mumbo-jumbo which we don't support anyways :smile:

kacf (changes requested):

artifact/v3/version/version.cpp
oleorhagen: hahaha. I think so too

artifact/v3/version/version.cpp
oleorhagen: Well, it does, right. The parsed values match exactly this. I just wanted to highlight it here. But if it's odd I can change it.

artifact/v3/version/version.hpp
oleorhagen: Maybe, but it does match the structure of what is parsed to a code-template of sorts. So at least it's symmetric with how we handle the rest of the parsing.

artifact/v3/version/version.cpp
oleorhagen: Oh

My

God....

artifact/v3/version/version.cpp
vpodzime: Yeah, I think I can do that. The JSON stuff we have predates our IO stuff so it wasn't possible back then. The only complication is that to do it right, we should have a UTF8Reader because bytes shouldn't be passed into the JSON parser that expects characters.

oleorhagen: Great :tada: :tada:

common/io.hpp
```diff
@@ -108,6 +109,26 @@ class StringReader : virtual public Reader {
 	}
 };
 
+class ByteWriter : virtual public Writer {
+private:
+	vector<uint8_t> &receiver;
```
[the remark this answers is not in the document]
oleorhagen: I should really add a static check for this in my editor or smthn...

common/io.hpp
```diff
@@ -108,6 +109,26 @@ class StringReader : virtual public Reader {
...
+
+public:
+	ByteWriter(vector<uint8_t> &receiver) :
+		receiver {receiver} {
+	}
+
+	ExpectedSize Write(const vector<uint8_t> &dst) override {
+		int i = 0;
+		for (; i < std::min(dst.size(), receiver.size() - bytes_written_); ++i) {
+			receiver[i + bytes_written_] = dst[i];
```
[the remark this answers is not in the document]
oleorhagen: We can unroll the loop? :p 

But really, I'll wrap it one of the std::copy derivatives

artifact/v3/version/version.cpp
oleorhagen: Can't directly actually, since they are different types...

But I can add a conversion function?

oleorhagen:
Rebased on top of: https://github.com/mendersoftware/mender/pull/1173

artifact/v3/version/version.cpp
```diff
@@ -0,0 +1,106 @@
...
+
+	if (!expected_json) {
+		return MakeError(ParseError, "Failed to parse the version JSON");
+	}
+
+	const json::Json version_json = expected_json.value();
+
+	const int version = version_json.Get("version")
+							.and_then([](const json::Json &j) { return j.GetInt(); })
+							.value_or(-1);
```
vpodzime, line 83: Is this what our formatter produces? :scream:

vpodzime, line 83: I'd drop the `value_or()` and got the `ExpectedInt()` from the chain because `value_or()` discards the error.

artifact/v3/version/version.cpp
```diff
@@ -0,0 +1,106 @@
...
+
+	if (version != supported_parser_version) {
+		return expected::make_unexpected(MakeError(
+			FormatError,
+			"Failed to unmarshal the version from the version JSON data. Only 'version' 3 is supported in this client"));
```
vpodzime, line 88: Should use the exact error (message) our JSON handling has the ability to provide.

artifact/v3/version/version.cpp
```diff
@@ -0,0 +1,106 @@
...
+	}
+
+	const string format = version_json.Get("format")
+							  .and_then([](const json::Json &j) { return j.GetString(); })
+							  .value_or("unknown");
```
vpodzime, line 93: Same here.

artifact/v3/version/version.cpp
```diff
@@ -0,0 +1,106 @@
...
+
+	if (format != supported_parser_format) {
+		return expected::make_unexpected(MakeError(
+			FormatError,
+			"Failed to unmarshal the format from the version JSON data. Only the 'mender' format is supported in this client"));
+	}
+
+	return Version {.version = 3, .format = "mender"};
```
vpodzime, line 101: Should use the constants.

vpodzime (commented):
+Still not a big fan of using the expected library without even trying to expand our own code, but looks like you guys are decided.

artifact/v3/version/CMakeLists.txt
```diff
@@ -0,0 +1,20 @@
...
+add_executable(artifact_version_parser_test EXCLUDE_FROM_ALL
+  version.cpp
+  version_test.cpp
+)
+target_link_libraries(artifact_version_parser_test PRIVATE
+  GTest::gtest_main
+  gmock
+  common_io
+  common_error
+  common_json
```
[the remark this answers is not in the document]
oleorhagen: Yeh, it's good you pointed this out, I forgot to mention it.

So I _did not_ make a module out of this, since it is so small, in which case, the test needs this to link.

However maybe this was a bad idea. I really don't know. 

Since it caught you by surprise, maybe I should make a separate library out of it, it just felt a bit extensive.

I'll let you decide :pray:

artifact/v3/version/version.cpp
```diff
@@ -0,0 +1,106 @@
...
+
+#include <string>
+#include <vector>
+
+#include <common/common.hpp>
+#include <common/expected.hpp>
+#include <common/error.hpp>
+#include <common/json.hpp>
+
+#include <artifact/v3/version/version.hpp>
```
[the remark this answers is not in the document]
oleorhagen: Yes, true. This is actually a really good rule!

artifact/v3/version/version.cpp
oleorhagen: Yes :see_no_evil: Not the prettiest on the :eyes:

artifact/v3/version/version.cpp
oleorhagen: Yeh, let's. I left it there to highlight it, but it's really just duplicating...

artifact/v3/version/version.cpp
```diff
@@ -0,0 +1,106 @@
...
+	}
+}
+
+error::Error MakeError(ErrorCode code, const string &msg) {
+	return error::Error(error_condition(code, ErrorCategory), msg);
+}
+
+
+expected::Expected<Version, error::Error> Parse(io::Reader &reader) {
+	// Read in all the bytes into a tmp buffer
```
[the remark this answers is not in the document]
oleorhagen: Yes, but like we discussed previously, the very best would be `json::From(io::Reader&)`

artifact/v3/version/version.cpp
```diff
@@ -0,0 +1,106 @@
...
+	std::vector<uint8_t> buf(1024);
+	io::ByteWriter bw {buf};
+
+	auto err = io::Copy(bw, reader);
```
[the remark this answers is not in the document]
oleorhagen: At least from what I'm used to, `Copy` is nice ,because it guarantees to read to the end. A single `Read` call can theoretically do short reads, and may need to be called again.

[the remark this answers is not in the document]
vpodzime: But the data is shuffled around (into an internal buffer in `Copy` and then into the `ByteWriter` above) and you can easily get an error if `buf` above is not big enough. Like I said above, to me it would make much more sense to have a function/class/... to read everything into a buffer and then use that buffer.

artifact/v3/version/version.cpp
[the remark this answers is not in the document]
vpodzime: That will require that function to read a file into a buffer anyway. Or a stream interface to our `Reader` because C++ already has IO utilities -- the streams -- an things are setup to work with them not our Go-like implementation.

artifact/v3/version/version.cpp
vpodzime: `value_or()` is still there.

vpodzime (commented):
DId you forget to push some commits? Many of the conversations are marked as resolved without the changes in the code being done.

oleorhagen:
> DId you forget to push some commits? Many of the conversations are marked as resolved without the changes in the code being done.

Yes, just fixed locally, not pushed yet. I didn't know I had re-requested a review here, I'm sorry.
Fixed and pushed now @vpodzime :smile_cat:

artifact/v3/version/version.cpp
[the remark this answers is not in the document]
oleorhagen: This is true, it is just a question about the size of the buffer. Like buffering the whole payload, or doing it in chunks.

I actually had a conversation with @kacf about it this morning, where I suggested to move to the stdlib interface, and his argument is that the `io-streams` add too much code bloat for us, as they are all templated.

I'm not quite sure I fully agree, but I do understand the argument, and it is not our biggest thorn in our sides imo.

Therefore for now, going with this intermediary buffering is the way it seems.

artifact/v3/version/version.cpp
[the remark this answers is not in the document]
oleorhagen: Ur just up too early in the morning :p :sunrise_over_mountains:  :rooster:

artifact/v3/version/version.cpp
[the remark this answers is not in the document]
oleorhagen: This is true in the general case. But the advantage here is that we control the format, and the `version` header is never going to be bigger than what is here available.

artifact/v3/version/version.cpp
oleorhagen: Same as above. You need to sleep in more :laughing:

oleorhagen:
I will not merge this until the `byte-writer` PR is merged!

kacf (approved):

artifact/v3/version/version.cpp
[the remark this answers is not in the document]
oleorhagen: > But the data is shuffled around (into an internal buffer in Copy and then into the ByteWriter above) and you can easily get an error if buf above is not big enough. Like I said above, to me it would make much more sense to have a function/class/... to read everything into a buffer and then use that buffer.

Isn't this what I am doing with the `byte-buffer` above already?

oleorhagen:
> I will not merge this until the `byte-writer` PR is merged!

Merged!

artifact/v3/version/version.cpp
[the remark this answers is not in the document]
vpodzime: Assuming it's at most 1024 bytes. But let's not ride a dead horse here anymore. I'll try to come up with something for this myself and then we can have a discussion there.

vpodzime (approved):
Just a "nitpick of nitpicks". ACK!

artifact/v3/version/version.cpp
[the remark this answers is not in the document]
vpodzime: > So there are 3 copies of the same thing, all with read-only access.

It's 4 actually. `Copy` has its own internal buffer too.

PR #1164, opened 2023-02-24
Iterators

kacf:
```
commit 11ee57d9727333709539538bf37e116685b159f5
Date:   Fri Feb 24 12:02:38 2023

    fix: Handle errors in `StreamReader::Read`.
    
    Changelog: None
    Ticket: None
    
commit 761937834c98522dfc1f7ef7fd11227764de651c
Date:   Fri Feb 24 16:08:24 2023

    chore: Switch Reader/Writer interfaces to use iterators.
    
    This is more in line with how the std interfaces work, and saves us
    from having to resize containers before and after passing them into
    the Read and Write functions. The resulting code reduction can be
    clearly seen.
    
    The tests are a bit more complicated because they do deep comparisons,
    and comparing ranges of iterators is harder than just two values. We
    cannot just match arguments one by one anymore, but we need to use a
    custom function where we call `std::equal`. This is limited to the
    io_test though, and I don't think it affects production code.
    
```

### DIFF
--- a/artifact/sha/platform/openssl/sha.cpp
+++ b/artifact/sha/platform/openssl/sha.cpp
@@ -64,7 +64,6 @@ Reader::Reader(io::Reader &reader) :
 Reader::Reader(io::Reader &reader, const std::string &expected_sha = "") :
 	sha_handle_(EVP_MD_CTX_new(), [](EVP_MD_CTX *ctx) { EVP_MD_CTX_free(ctx); }),
 	wrapped_reader_ {reader},
-	buffer_(4092),
 	expected_sha_ {expected_sha} {
 	if (EVP_DigestInit_ex(sha_handle_.get(), EVP_sha256(), nullptr) != 1) {
 		log::Error("Failed to initialize the shasummer");
@@ -74,32 +73,13 @@ Reader::Reader(io::Reader &reader, const std::string &expected_sha = "") :
 	initialized_ = true;
 }
 
-// Simply takes a reference to a buffer, and if need be resizes it, and then on
-// destruction, if it has been resized, sets it back to the original size
-class BufferResizer {
-	vector<uint8_t> &buf;
-	int original_size;
-
-public:
-	BufferResizer(vector<uint8_t> &buf, int original_size, int max_size) :
-		buf {buf},
-		original_size {original_size} {
-		buf.resize(std::min(original_size, max_size));
-	}
-
-	~BufferResizer() {
-		buf.resize(original_size);
-	}
-};
-
-expected::ExpectedSize Reader::Read(vector<uint8_t> &dst) {
+expected::ExpectedSize Reader::Read(
+	vector<uint8_t>::iterator start, vector<uint8_t>::iterator end) {
 	if (!initialized_) {
 		return MakeError(InitializationError, "");
 	}
 
-	auto buffer_resizer = BufferResizer(buffer_, buffer_.size(), dst.size());
-
-	auto bytes_read = wrapped_reader_.Read(buffer_);
+	auto bytes_read = wrapped_reader_.Read(start, end);
 	if (!bytes_read) {
 		return bytes_read;
 	}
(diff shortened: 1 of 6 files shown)


artifact/sha/platform/openssl/sha.cpp
```diff
@@ -119,13 +99,10 @@ expected::ExpectedSize Reader::Read(vector<uint8_t> &dst) {
 		return 0;
 	}
 
-	if (EVP_DigestUpdate(sha_handle_.get(), buffer_.data(), bytes_read.value()) != 1) {
+	if (EVP_DigestUpdate(sha_handle_.get(), &*start, bytes_read.value()) != 1) {
```
oleorhagen, line 102: Nit: I'm sure you've got your bindings right, but can I still ask for a parenthesis?

artifact/sha/sha_test.cpp
```diff
@@ -34,9 +34,9 @@ TEST(ShasummerTest, TestShaSum) {
 
 	vector<uint8_t> actual(4096);
 
-	auto bytes_read = r.Read(actual);
+	auto bytes_read = r.Read(actual.begin(), actual.end());
 
-	EXPECT_TRUE(bytes_read);
+	ASSERT_TRUE(bytes_read) << bytes_read.error();
```
oleorhagen, line 39: Nit: doing more than the commit says.

common/io.hpp
```diff
@@ -77,6 +77,11 @@ class StreamReader : virtual public Reader {
 	}
 	ExpectedSize Read(vector<uint8_t> &dst) override {
 		is_.read(reinterpret_cast<char *>(&dst[0]), dst.size());
+		if (is_.bad()) {
+			int int_error = errno;
+			return mender::common::error::Error(
+				std::error_code(int_error, std::system_category()).default_error_condition(), "");
```
vpodzime, line 84: No message? Shouldn't we extract the one from `errno` here?

artifact/sha/platform/openssl/sha.cpp
vpodzime, line 102: ouch

common/io.cpp
```diff
@@ -27,34 +27,24 @@ Error Copy(Writer &dst, Reader &src, vector<uint8_t> &buffer) {
 	size_t orig_size = buffer.size();
```
vpodzime, line 27: `orig_size` is no longer needed, right?

vpodzime (approved):
Looks good to me otherwise, to the extent I understand.

artifact/sha/platform/openssl/sha.cpp
kacf: You mean `&(*start)`? Not sure if that makes it clearer.. I think I vote for keeping it as is.

artifact/sha/sha_test.cpp
kacf: Oops, yes, that wasn't even supposed to be there (I'm surprised it works actually. I thought `error()` would throw in that context).

common/io.cpp
kacf: Indeed. Why didn't it warn about this..?

oleorhagen: Does it ever? :thinking: 

I think it's only unhappy about unused functions

artifact/sha/platform/openssl/sha.cpp
kacf: I could do `&start[0]`, although my personal preference is still the former. Let me know which you prefer.

oleorhagen: haha, suit yourself

oleorhagen: I would hugely prefer the `&start[0]` syntax. With emphasis on `I`

kacf: Haha, ok, that settles it!

artifact/sha/sha_test.cpp
kacf: Hehe, I wish. I believe only `&&` and `||` are eligible for lazy evaluation (short circuiting, as I think it's called in C++).

common/io.cpp
vpodzime: Time for `-Wextra`?

kacf: It should have caught it anyway:

> ```
>  -Wunused-but-set-variable
>            Warn whenever a local variable is assigned to, but otherwise unused
>            (aside from its declaration).  This warning is enabled by -Wall.
> ```

PR #1165, opened 2023-02-26
Add the -Wall flag during builds

oleorhagen:
Also fixed the build errors which popped up as a result.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -95,7 +95,7 @@ target_compile_options(crossplatform_compiler_flags INTERFACE -std=c++11 -Werror
 
 add_library(platform_compiler_flags INTERFACE)
 target_compile_features(platform_compiler_flags INTERFACE cxx_std_17)
-target_compile_options(platform_compiler_flags INTERFACE -std=c++17 -Wall)
+target_compile_options(platform_compiler_flags INTERFACE -std=c++17 -Werror -Wall)
 #---------------------------------------------------------------------------------------------------
 
 add_executable(mender main.cpp lib.cpp)
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -95,7 +95,7 @@ target_compile_options(crossplatform_compiler_flags INTERFACE -std=c++11 -Werror
 
 add_library(platform_compiler_flags INTERFACE)
 target_compile_features(platform_compiler_flags INTERFACE cxx_std_17)
-target_compile_options(platform_compiler_flags INTERFACE -std=c++17)
+target_compile_options(platform_compiler_flags INTERFACE -std=c++17 -Wall)
 #---------------------------------------------------------------------------------------------------
 
 add_executable(mender main.cpp lib.cpp)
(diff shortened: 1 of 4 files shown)


common/io.hpp
```diff
@@ -90,8 +90,8 @@ class Discard : virtual public Writer {
 
 class StringReader : virtual public Reader {
 private:
-	StreamReader reader_;
 	std::stringstream s_;
+	StreamReader reader_;
```
vpodzime, line 94: How is this a fix for a warning?

common/log.hpp
```diff
@@ -75,10 +75,10 @@ class Logger {
 	boost::log::sources::severity_logger<LogLevel> logger;
 #endif
 
-	LogLevel level_ {LogLevel::Info};
-
 	string name_ {};
 
+	LogLevel level_ {LogLevel::Info};
```
vpodzime, line 80: Same here.

vpodzime (commented):
Thanks for adding this!

common/io.hpp
oleorhagen: haha, I know it looks weird. But it will be a lesson to you, same as it was to me :)

Ordering mattes appareantly. So now the declaration matches the initialization order in the constructor, because this is the order in which they are initialized.

common/log.hpp
oleorhagen: Same answer ;)

common/io.hpp
kacf: The reason it matters is that destruction is the reverse of the member order, and it's good practice that construction and destruction are always 100% inverse of each other.

CMakeLists.txt
```diff
@@ -91,7 +91,7 @@ endif()
 
 add_library(crossplatform_compiler_flags INTERFACE)
 target_compile_features(crossplatform_compiler_flags INTERFACE cxx_std_11)
-target_compile_options(crossplatform_compiler_flags INTERFACE -std=c++11 -Werror)
+target_compile_options(crossplatform_compiler_flags INTERFACE -std=c++11 -Werror -Wall)
```
kacf, line 94: What about platform flags? Should we not enable both of these options there?

lluiscampos:
Thank you for adding this :+1:

common/io.hpp
oleorhagen: So kudos to the compiler here :1st_place_medal:

oleorhagen: Also, that makes a lot of sense actually!

CMakeLists.txt
oleorhagen: Yeah, good point. Adding

common/io.hpp
vpodzime: Ehm, I don't agree that it makes much sense because it's supposed to be a declaration and thus, by definition, declarative, i.e. with order not being important. :wink: But thank for the clarification.

vpodzime (approved):
ACK

kacf (approved):

PR #1173, opened 2023-02-28
Replace the internal expected implementation with an existing lib

oleorhagen:
This PR replaces our internal std::expected implementation with an existing project. More closely:

```
https://github.com/TartanLlama/expected
```

This fully replaces our implementation, while giving us access to the modern monadic extensions, which can help reduce the overhead surrounding our expactation handling.

See the link for more information.

### DIFF
--- a/artifact/sha/platform/openssl/sha.cpp
+++ b/artifact/sha/platform/openssl/sha.cpp
@@ -78,9 +78,9 @@ Reader::Reader(io::Reader &reader, const std::string &expected_sha = "") :
 expected::ExpectedSize Reader::Read(
 	vector<uint8_t>::iterator start, vector<uint8_t>::iterator end) {
 	if (!initialized_) {
-		return MakeError(
+		return expected::unexpected(MakeError(
 			InitializationError,
-			"The ShaReader was not properly initialized. Shasumming is not possible");
+			"The ShaReader was not properly initialized. Shasumming is not possible"));
 	}
 
 	auto bytes_read = wrapped_reader_.Read(start, end);
@@ -95,16 +95,16 @@ expected::ExpectedSize Reader::Read(
 			return real_sha.error();
 		}
 		if (real_sha.value() != expected_sha_) {
-			return MakeError(
+			return expected::unexpected(MakeError(
 				ShasumMismatchError,
 				"The checksum of the read byte-stream does not match the expected checksum, (expected): "
-					+ expected_sha_ + " (calculated): " + real_sha.value());
+					+ expected_sha_ + " (calculated): " + real_sha.value()));
 		}
 		return 0;
 	}
 
 	if (EVP_DigestUpdate(sha_handle_.get(), &start[0], bytes_read.value()) != 1) {
-		return MakeError(ShasumCreationError, "Failed to create the shasum");
+		return expected::unexpected(MakeError(ShasumCreationError, "Failed to create the shasum"));
 	}
 
 	return bytes_read.value();
@@ -112,22 +112,23 @@ expected::ExpectedSize Reader::Read(
 
 expected::ExpectedString Reader::ShaSum() {
 	if (!initialized_) {
-		return MakeError(
+		return expected::unexpected(MakeError(
 			InitializationError,
-			"The ShaReader was not properly initialized. Shasumming is not possible");
+			"The ShaReader was not properly initialized. Shasumming is not possible"));
 	}
 
 	vector<uint8_t> hash(EVP_MAX_MD_SIZE);
 	unsigned int hash_length = 0;
 
 	if (EVP_DigestFinal_ex(sha_handle_.get(), hash.data(), &hash_length) != 1) {
-		return MakeError(ShasumCreationError, "Failed to create the shasum. OpenSSL error: ");
+		return expected::unexpected(
+			MakeError(ShasumCreationError, "Failed to create the shasum. OpenSSL error: "));
 	}
 
 	if (hash_length != SHA_256_digest_length) {
-		return MakeError(
+		return expected::unexpected(MakeError(
 			ShasumCreationError,
-			"SHA of unexpected length: " + std::to_string(hash_length) + " expected length: 32");
+			"SHA of unexpected length: " + std::to_string(hash_length) + " expected length: 32"));
 	}
 
 	std::stringstream ss {};
--- a/common/expected.hpp
+++ b/common/expected.hpp
@@ -15,6 +15,12 @@
 #ifndef MENDER_COMMON_EXPECTED_HPP
 #define MENDER_COMMON_EXPECTED_HPP
 
+#ifdef __cpp_lib_expected
+#include <expected>
+#else
+#include <tl/expected.hpp>
+#endif
+
 #include <common/error.hpp>
 
 #include <cassert>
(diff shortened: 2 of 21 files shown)


vpodzime (commented):
I don't like the use of `make_unexpected()` because it's not what `std::expected` does or even plans to do in future (if I'm looking correctly). The rest is :+1: as long as there are enough use cases for the extra stuff the library provides (because it's much bigger and full of things we don't need that are hard to understand,...).

vpodzime:
Also, please double-check that the library doesn't introduce exceptions.

common/key_value_database/platform/lmdb/lmdb.cpp
```diff
@@ -118,7 +119,7 @@ expected::ExpectedBytes KeyValueDatabaseLmdb::Read(const string &key) {
 		}
 	});
 	if (err) {
-		return err;
+		return tl::make_unexpected(err);
```
kacf, line 122: IMHO this defeats some of the convenience we already have. I also don't fully understand why it's needed. Isn't `tl::expected<Type1, Type2>(Type2 &)` a valid constructor?

kacf (commented):
Very much much in favor of using this library, but I also have some concerns about `make_unexpected`.

vpodzime:
How hard would it be add the "monads" to our implementation (if that's the thing we are missing)? It's still ~200 lines vs. ~2800 lines and implementing `and_then()` doesn't look too hard.

oleorhagen:
> Also, please double-check that the library doesn't introduce exceptions.

Compiled and ran fine with `no-exceptions`

nickanderson:
Drive by comment ...
 
 "Replace the internal expected implementation with an existing lib" what internal expected implementation?

oleorhagen:
> > Also, please double-check that the library doesn't introduce exceptions.
> 
> Compiled and ran fine with `no-exceptions`

https://github.com/mendersoftware/mender/pull/1173/files#diff-2037103007b16c81a2006bc8dc7950255a3ce7ccbe6e50ce3ea6cc2f47dcecdc

vpodzime:
> How hard would it be add the "monads" to our implementation (if that's the thing we are missing)? It's still ~200 lines vs. ~2800 lines and implementing and_then() doesn't look too hard.

Now we know it's `unexpected()` plus the chaining operations ("monads"), but the above question still applies, AFAICT.

common/key_value_database/platform/lmdb/lmdb.cpp
kacf: Hmm, yes, looks like it. I understand it from their perspective. Our types are not easily confused, but that may not be the case for everyone. A bit annoying, but I think it is an acceptable price to pay.

oleorhagen: Yeh, we can create a conversion operator for this though...

_if we want to_...

kacf:
> > How hard would it be add the "monads" to our implementation (if that's the thing we are missing)? It's still ~200 lines vs. ~2800 lines and implementing and_then() doesn't look too hard.
> 
> Now we know it's `unexpected()` plus the chaining operations ("monads"), but the above question still applies, AFAICT.

I would be quite strongly against adding our own implementation. We should get code out of our code base, not into it. Every line we move out is something we can get others to maintain. Especially when it is becoming part of the C++ standard.

I think the number of lines in their code base is irrelevant; it is of course expected because they do more than we do, and since they are entering the standard they have the whole C++ community behind them to maintain it. Binary size matters a little though, but I would be quite surprised if this has a significant impact, since it depends on what we actually use.

kacf:
> Drive by comment ...
> 
> "Replace the internal expected implementation with an existing lib" what internal expected implementation?

@nickanderson: The library is called "expected". 🙂

common/expected.hpp
```diff
@@ -22,153 +22,24 @@
...
+using namespace std;
 
-template <typename ExpectedType, typename ErrorType>
-Expected<ExpectedType, ErrorType> &Expected<ExpectedType, ErrorType>::operator=(Expected &&e) {
-	if (this->has_val_) {
-		this->ex_.~ExpectedType();
-	} else {
-		this->err_.~ErrorType();
-	}
+using namespace tl;
```
kacf, line 33: I think this will break in the future, because both `std` and `tl` will contain `expected` then. A better approach is maybe to remove the two `using namespace` declarations and replace with this:

```c++
#include <version>

#ifdef __cpp_lib_expected
using expected = std::expected;
#else
using expected = tl::expected;
#endif
```

See [this page](https://en.cppreference.com/w/cpp/feature_test) for more info.

kacf (approved):

vpodzime:
> I would be quite strongly against adding our own implementation. We should get code out of our code base, not into it.

OTOH, we are trying to avoid some big things (that already have implementations for many things we have) and we have already pulled in quite a few unnecessarily huge things.

It might still be harder for a community to maintain 3000 cryptic lines of code than for us to maintain 400 lines of clean and easy-to-understand code.

artifact/sha/platform/openssl/sha.cpp
```diff
@@ -78,9 +78,9 @@ Reader::Reader(io::Reader &reader, const std::string &expected_sha = "") :
 expected::ExpectedSize Reader::Read(
 	vector<uint8_t>::iterator start, vector<uint8_t>::iterator end) {
 	if (!initialized_) {
-		return MakeError(
+		return expected::make_unexpected(MakeError(
```
vpodzime, line 81: As I asked in the other PR -- could we do `using unexpected = make_unexpected;` or something like that? `std::expected` doesn't have `make_unexpected()`, right?

kacf: I reckon we can follow the same rules as everywhere else: We can make such statements in any local namespace/source, but let's keep it out of the global scope. Whether we should do it in this mass replacement.. I don't have a strong opinion, either is fine with me. It does mean it would need to be added multiple times.

kacf:
> It might still be harder for a community to maintain 3000 cryptic lines of code than for us to maintain 400 lines of clean and easy-to-understand code.

All code is a liability. If we have doubts of their ability to maintain something, that's one thing. It could happen if a repository is abandoned, there is only one contributor, or something like that. But in this case we have one the strongest guarantees we could possibly get, that this is fully supported by a dedicated and professional team. So let's use that, and focus on the domain specific code instead of writing code for basic C++ functionality.

common/expected.hpp
oleorhagen: Nice! Thanks for digging this up!!

artifact/sha/platform/openssl/sha.cpp
oleorhagen: We can do this:

```
#ifdef __cpp_lib_expected
using std::expected;
using std::unexpected;
#else
using tl::expected;
template <typename V>
tl::unexpected<V> unexpected(V &&v) {
	return tl::unexpected<V>(v);
}
#endif
```

Which maps nicely to the std-lib then

kacf: 👍

oleorhagen: Done!

vpodzime: :clap:  thanks!

PR #1174, opened 2023-02-28
chore: Make sure LMDB header doesn't need C++17.

kacf:
This is needed because it is compiled together with other headers which are compiled in C++11 mode.


common/key_value_database/platform/lmdb/lmdb.cpp
```diff
@@ -77,7 +79,7 @@ error::Error LmdbTransaction::Remove(const string &key) {
 }
 
 KeyValueDatabaseLmdb::KeyValueDatabaseLmdb() :
-	env_(lmdb::env::create()),
+	env_(new lmdb::env(lmdb::env::create())),
```
vpodzime, line 82: I'd prefer `make_unique()` if possible, but YMMV.

vpodzime (approved):
Looks good to me otherwise.

common/key_value_database/platform/lmdb/lmdb.cpp
kacf: Ah, I *can* use it here. I was thinking that it's not C++11, but this is a platform file. Fixing!

PR #1175, opened 2023-03-01
feat(io): Add a byte-writer

oleorhagen:
Ticket: None
Changelog: None

### DIFF
--- a/common/io.hpp
+++ b/common/io.hpp
@@ -18,12 +18,15 @@
 #include <common/error.hpp>
 #include <common/expected.hpp>
 
+#include <cstdint>
+#include <iterator>
 #include <memory>
 #include <system_error>
 #include <vector>
 #include <istream>
 #include <sstream>
 #include <string>
+#include <algorithm>
 
 namespace mender {
 namespace common {
@@ -120,6 +123,22 @@ public:
 	ExpectedSize Read(vector<uint8_t>::iterator start, vector<uint8_t>::iterator end) override {
 		return reader_.Read(start, end);
 	}
+};
+
+using Vsize = vector<uint8_t>::size_type;
+
+class ByteWriter : virtual public Writer {
+private:
+	vector<uint8_t> &receiver_;
+	Vsize bytes_written_ {0};
+
+public:
+	ByteWriter(vector<uint8_t> &receiver) :
+		receiver_ {receiver} {
+	}
+
+	ExpectedSize Write(
+		vector<uint8_t>::const_iterator start, vector<uint8_t>::const_iterator end) override;
 };
 
 } // namespace io
--- a/common/io.cpp
+++ b/common/io.cpp
@@ -49,6 +49,21 @@ Error Copy(Writer &dst, Reader &src, vector<uint8_t> &buffer) {
 	}
 }
 
+ExpectedSize ByteWriter::Write(
+	vector<uint8_t>::const_iterator start, vector<uint8_t>::const_iterator end) {
+	assert(end > start);
+	Vsize max_write {receiver_.size() - bytes_written_};
+	if (max_write == 0) {
+		return Error(make_error_condition(errc::no_space_on_device), "");
+	}
+	Vsize iterator_size {static_cast<Vsize>(end - start)};
+	Vsize bytes_to_write {min(iterator_size, max_write)};
+	auto it = next(receiver_.begin(), bytes_written_);
+	std::copy_n(start, bytes_to_write, it);
+	bytes_written_ += bytes_to_write;
+	return bytes_to_write;
+}
+
 } // namespace io
 } // namespace common
 } // namespace mender


common/io.hpp
```diff
@@ -115,6 +118,28 @@ class StringReader : virtual public Reader {
...
+public:
+	ByteWriter(vector<uint8_t> &receiver) :
+		receiver_ {receiver} {
+	}
+
+	ExpectedSize Write(
+		vector<uint8_t>::const_iterator start, vector<uint8_t>::const_iterator end) override {
+		long unsigned int max_write {receiver_.size() - bytes_written_};
+		long unsigned int iterator_size {(long unsigned int) (end - start)};
+		long unsigned int bytes_to_write {std::min(iterator_size, max_write)};
```
vpodzime, line 135: `size_t`, please.

vpodzime (commented):
Looks good to me otherwise. I still hope we shouldn't need this copy-writer, but we shall see.

common/io.hpp
oleorhagen: But they're not `size_t`'s

oleorhagen: This annoyed me to no end. I can do `size_t a = end - start` for instance. But I do get a warning it's a narrowing conversion on my setup :persevere:

vpodzime: Wow, that's quite strange. Please add a comment about this.

oleorhagen: Yeh, so I'm guessing the reason is that the range can be negative (although we know it's not in this case, since we are passing in the iterators). But the `distance_type) (between iterators) are signed as a result, while the size is not (always >= 0)...

Pushed a fix, and added an assert for the negative range as a safe-guard. 

Please see what you think.

kacf: According to [this page](https://stackoverflow.com/questions/22021886/subtraction-of-iterators-and-integers), the result type after subtracting two iterators is [`ptrdiff_t`](https://en.cppreference.com/w/cpp/types/ptrdiff_t). I guess it's because negative values are possible, although in practical `end - start` scenarios it should never happen. This is identical to `difference_type`, not `size_type` as you have used.

I suspect that the warnings will go away if you keep the `ptrdiff_t` type for the direct assignments (or just `auto`), but you add it to a `size_t` total with `+=`. This is allowed AFAIK.

oleorhagen: Hmm, I'm not sure I follow ur argument. All these things I already knew, and is what I tried to convey in the comment above.

Why should I add a signed value to unsigned (aka `ptrdiff_t` to `size_t` ?) - Isn't it better to do, as I have done, base all arithmetic in the vector size type, and then do `one`, and `one` only `explicit` cast to unsigned. Which is actually safe here, since it is guarded by the assert?

vpodzime: There's also `ssize_t`.

oleorhagen: Yeh, but point being, I want to use `vector:size_type`

vpodzime: Yeah, I don't know how to do that. But there's no guarantee it's going to be `long unsigned int` everywhere. That's exactly what `size_t` and `ssize_t` are for. They made it more flexible, but I'm not sure we can easily do the same.

oleorhagen:
I will also implement [this](https://github.com/mendersoftware/mender/pull/1177#discussion_r1124183634) nugget

common/io.hpp
```diff
@@ -122,6 +125,34 @@ class StringReader : virtual public Reader {
...
+private:
+	vector<uint8_t> &receiver_;
+	Vsize bytes_written_ {0};
+
+public:
+	ByteWriter(vector<uint8_t> &receiver) :
+		receiver_ {receiver} {
+	}
+
+	ExpectedSize Write(
```
vpodzime, line 140: This should be in the `.cpp` file.

vpodzime (commented):
Looks good to me otherwise. But I have to say I still think every single use of this class will mean an unnecessary and problematic data duplication or manipulation to me. Either we need the raw data from the reader and then we should access it directly or we need a post-processed data from the reader and then we don't know the size of it (and thus the necessary buffer). Maybe it's because I'm not used to the Reader/Writer schema, but I keep seeing potential problems like the above in practically every place where we use it.

common/io.hpp
kacf: My point was that if you stick to `difference_type` or `ptrdiff_t` until you do the addition operation, you don't need any casts at all, because adding signed to unsigned is allowed, unlike assignment.

oleorhagen:
```
using Vsize = vector<uint8_t>::const_iterator::difference_type;
```
```
error: narrowing conversion of ‘((&((mender::common::io::ByteWriter*)this)->mender::common::io::ByteWriter::receiver_)->std::vector<unsigned char>::size() - ((long unsigned int)((mender::common::io::ByteWriter*)this)->mender::common::io::ByteWriter::bytes_written_))’ from ‘long unsigned int’ to ‘mender::common::io::Vsize’ {aka ‘long int’} [-Werror=narrowing]
   55 |         Vsize max_write {receiver_.size() - bytes_written_};
```

So it is not happy with it the other way around.

kacf: Sorry, my mistake, I did not realize that you also use `bytes_written_` at the start. Then you cannot avoid casting, I think.

kacf:
> Looks good to me otherwise. But I have to say I still think every single use of this class will mean an unnecessary and problematic data duplication or manipulation to me. Either we need the raw data from the reader and then we should access it directly or we need a post-processed data from the reader and then we don't know the size of it (and thus the necessary buffer). Maybe it's because I'm not used to the Reader/Writer schema, but I keep seeing potential problems like the above in practically every place where we use it.

You have a point, but the alternatives I'm aware of are not free of problems either. It's not possible to share the raw data from readers in general, because they have different internals, and the caller should not care. An alternative could be to return iterators from the `Reader` interface instead of filling an existing iterator range, but the problem with this is that iterators are template types, so we have to decide upfront which iterator type we want (templates do not mix with `virtual`). Because we almost certainly would have to choose `vector::iterator` or some other memory based variant, it would still require buffering, we have just moved it from the caller to the callee.

The heart of the problem is that for a generic interface we cannot assume that the data we are trying to read resides in memory. It might, and then we'll get an extra buffered copy, or it might be in a file, and in that case the buffer is necessary and unavoidable. At the interface level it's hard to distinguish between these two. We could consider special-case interfaces, but then we are trading developer efficiency for uncertain benefits.

I would say that we should try to avoid premature optimization and assume that buffering is ok. We can always lower their size, which will improve memory footprint at the cost of somewhat higher CPU usage (more callbacks). Right now, as far as we know, neither of those is a problem.

kacf:
> An alternative could be to return iterators from the `Reader` interface instead of filling an existing iterator range, but the problem with this is that iterators are template types, so we have to decide upfront which iterator type we want (templates do not mix with `virtual`). Because we almost certainly would have to choose `vector::iterator` or some other memory based variant, it would still require buffering, we have just moved it from the caller to the callee.

Actually one thing that might work is to do as @oleorhagen suggested: Implement iterator primitives ourselves, and make all those methods pure virtual. That would allow us to subclass them and use one single iterator type everywhere. However, subclassing and implementing these methods has a lot more complexity and caveats than implementing one single Reader method.

vpodzime:
> The heart of the problem is that for a generic interface we cannot assume that the data we are trying to read resides in memory. It might, and then we'll get an extra buffered copy, or it might be in a file, and in that case the buffer is necessary and unavoidable.

There are things like `mmap()` that allow you to work with data in files without (explicit) buffers, but that's a bit off-topic, of course. OTOH, the standard C++ streams work on top of both files (in the broad/POSIX sense of the word *file*) as well as memory chunks.

> I would say that we should try to avoid premature optimization and assume that buffering is ok. We can always lower their size, which will improve memory footprint at the cost of somewhat higher CPU usage (more callbacks). Right now, as far as we know, neither of those is a problem.

This depends on what the target platforms are. Things that don't require unnecessary resources can easily run in places with abundance of resources, but it doesn't work the other way around.

PR #1177, opened 2023-03-01
feat: Add a tar-reader library

oleorhagen:
This adds the initial implementation of the tar-reader library.

I have to admit, I'm not really pleased with this. I think the interface is nice, but the implementation is so-so...

However, it does work (yay).

Feel free to have a look, and help me get this thing into fighting shape!


oleorhagen:
Feel free to ignore the byte-writer for now. It's only used in the tests

artifact/tar/CMakeLists.txt
```diff
@@ -0,0 +1,36 @@
...
+endif()
+
+target_link_libraries(common_tar PUBLIC
+  LibArchive::LibArchive
+  common_error
+  common_log
+  common_io
+)
+target_link_libraries(common_tar PRIVATE platform_compiler_flags)
+target_link_libraries(common_tar INTERFACE crossplatform_compiler_flags)
```
kacf, line 15: This has been changed in the main branch now. Use [this style](https://github.com/mendersoftware/mender/blob/feature-c%2B%2B-client/common/CMakeLists.txt#L25) instead. Cross platform flags can be removed completely.

artifact/tar/platform/libarchive/tar.cpp
```diff
@@ -0,0 +1,129 @@
...
+	case TarReaderError:
+		return "Error reading the tar stream";
+	case TarShortReadError:
+		return "Short read error";
+	case TarEntryError:
+		return "Error reading the tar entry";
+	case TarEOFError:
+		return "Archive EOF reached";
+	default:
+		return "Unknown";
```
kacf, line 51: I know this isn't consistent in the code base, but I suggest switching to [this style](https://github.com/mendersoftware/mender/blob/feature-c%2B%2B-client/common/http/http.cpp#L42) instead, which should help us catch missing cases.

artifact/tar/platform/libarchive/tar.cpp
```diff
@@ -0,0 +1,129 @@
...
+	}
+}
+
+error::Error MakeError(ErrorCode code) {
+	return error::Error(error_condition(code, ErrorCategory), "");
```
kacf, line 56: What was the outcome of this discussions? Didn't we agree not to provide this?

artifact/tar/platform/libarchive/tar.cpp
```diff
@@ -0,0 +1,129 @@
...
+}
+error::Error MakeError(ErrorCode code, const string &msg) {
+	return error::Error(error_condition(code, ErrorCategory), msg);
+}
+
+
+mender::common::expected::Expected<size_t, mender::common::error::Error> Entry::Read(
```
kacf, line 63: Suggest putting this at the top for easier readability:
```
namespace expected = mender::common::expected;
```

Applies to many instances.

artifact/tar/platform/libarchive/tar.cpp
```diff
@@ -0,0 +1,129 @@
...
+	vector<uint8_t>::iterator start, vector<uint8_t>::iterator end) {
+	expected::ExpectedSize read_bytes = reader_.Read(start, end);
+
+	if (!read_bytes) {
+		return read_bytes;
+	}
+
+	if (read_bytes.value() < 0) {
```
kacf, line 71: `ExpectedSize` uses `size_t`, so it can't be below zero. Surprised it didn't warn about this..

artifact/tar/platform/libarchive/tar.cpp
```diff
@@ -0,0 +1,129 @@
...
+		return MakeError(TarReaderError, "Failed to read from the underlying tar stream");
+	}
+
+	nr_bytes_read_ += read_bytes.value();
+
+	if (read_bytes.value() == 0 || read_bytes.value() == ARCHIVE_EOF) {
```
kacf, line 77: What is `ARCHIVE_EOF` exactly? If it's anything else than zero, then it needs to be returned as an error instead. And if it is zero, then you don't need the check on the left.

artifact/tar/tar.hpp
```diff
@@ -0,0 +1,104 @@
...
+	expected::ExpectedSize Read(
+		vector<uint8_t>::iterator start, vector<uint8_t>::iterator end) override;
+};
+
+using ExpectedEntry = expected::Expected<Entry, error::Error>;
+
+class Reader : io::Reader {
+private:
+#ifdef MENDER_TAR_LIBARCHIVE
+	mender::libarchive::wrapper::Handle archive_handle;
```
kacf, line 89: Needs trailing underscore.

artifact/tar/platform/libarchive/wrapper.cpp
```diff
@@ -0,0 +1,105 @@
...
+// the archive. The read callback reads data into a buffer, sets the const void
+// **buffer argument to point to the available data, and returns a count of the
+// number of bytes available. LibArchive will invoke the read callback again
+// only after it has consumed this data. LibArchive imposes no constraints on
+// the size of the data blocks returned.
+//
+// - On EOF return 0.
+// - On error return -1.
+ssize_t reader_callback(archive *archive, void *in_reader_container, const void **buff) {
+	ReaderContainer *p_reader_container = reinterpret_cast<ReaderContainer *>(in_reader_container);
```
kacf, line 34: Should only need `static_cast` here. Casting from `void` to anything is allowed.

artifact/tar/platform/libarchive/wrapper.cpp
```diff
@@ -0,0 +1,105 @@
...
+
+	int bytes_read = ret.value();
+	*buff = &p_reader_container->buff_[0];
+
+	return bytes_read;
+};
+
+Error Handle::Init() {
+	archive_read_support_filter_gzip(archive_.get());
+	archive_read_support_filter_xz(archive_.get());
```
kacf, line 51: Nice! So both of these are built in?

artifact/tar/platform/libarchive/wrapper.cpp
```diff
@@ -0,0 +1,105 @@
...
+	archive_read_support_format_tar(archive_.get());
+	int r =
+		archive_read_open(archive_.get(), &reader_container_, nullptr, reader_callback, nullptr);
+	if (r != ARCHIVE_OK) {
+		return Error(
+			std::error_condition(std::errc::inappropriate_io_control_operation),
```
kacf, line 57: AFAIK, this error type is for terminal errors. I can't really find an appropriate one, so maybe it's better to make your own for this one, or use `GenericError`.

artifact/tar/tar.hpp
```diff
@@ -0,0 +1,104 @@
...
+public:
+	Entry(std::string name, uint64_t archive_size, Reader &reader) :
+		name_ {name},
+		total_size_ {archive_size},
+		reader_ {reader} {
+	}
+
+	std::string Name() {
+		return name_;
+	}
```
kacf, line 78: Can we also get the size? We will need this for [MEN-6273](https://northerntech.atlassian.net/browse/MEN-6273).

artifact/tar/tar_test.cpp
```diff
@@ -0,0 +1,165 @@
...
+tar cvf test.tar testdata
+
+# Create large tar file
+dd if=/dev/random of=testinput.large bs=1M count=4
+tar cvf test-large.tar testinput.large
+
+exit 0
+)";
+
+		const string script_fname = "./test-script.sh";
```
kacf, line 53: Suggest using [`TemporaryDirectory`](https://github.com/mendersoftware/mender/blob/feature-c%2B%2B-client/common/testing.hpp#L26) and grab its `Path()` to put the things in a temporary directory. That both reduces the chance that you overwrite something in the current directory, and you can also remove most of your cleanup code, because it does it automatically. You will need to construct the object here, and destroy it in the cleanup code though.

kacf (changes requested):
Nice!

Even if the internals are complicated, a good and easy API goes a long way!

common/io.hpp
```diff
@@ -115,6 +118,28 @@ class StringReader : virtual public Reader {
...
+
+	ExpectedSize Write(
+		vector<uint8_t>::const_iterator start, vector<uint8_t>::const_iterator end) override {
+		long unsigned int max_write {receiver_.size() - bytes_written_};
+		long unsigned int iterator_size {(long unsigned int) (end - start)};
+		long unsigned int bytes_to_write {std::min(iterator_size, max_write)};
+		auto it = std::next(receiver_.begin(), bytes_written_);
+		std::copy_n(start, bytes_to_write, it);
+		bytes_written_ += bytes_to_write;
+		return bytes_to_write;
```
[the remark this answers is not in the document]
oleorhagen: Yes, good idea!

artifact/tar/platform/libarchive/wrapper.cpp
oleorhagen: Yeh :tada:

artifact/tar/platform/libarchive/wrapper.cpp
```diff
@@ -0,0 +1,105 @@
...
+
+	auto ret = p_reader_container->reader_.Read(
+		p_reader_container->buff_.begin(), p_reader_container->buff_.end());
+	if (!ret) {
+		log::Error("Failed to read from the archive stream: Error: " + ret.error().message);
+		return -1;
+	}
+
+	int bytes_read = ret.value();
+	*buff = &p_reader_container->buff_[0];
```
vpodzime, line 44: `buff_.data()`

artifact/tar/platform/libarchive/wrapper.hpp
```diff
@@ -0,0 +1,94 @@
...
+using namespace std;
+
+namespace error = mender::common::error;
+namespace expected = mender::common::expected;
+namespace io = mender::common::io;
+using mender::common::error::Error;
+
+namespace log = mender::common::log;
+
+class ReaderContainer {
```
vpodzime, line 44: I  believe this could be just a passive `struct` (with public fields) defined in `wrapper.cpp`, it's not used anywhere else, right?

artifact/tar/platform/libarchive/wrapper.cpp
```diff
@@ -0,0 +1,105 @@
...
+
+	return bytes_read;
+};
+
+Error Handle::Init() {
+	archive_read_support_filter_gzip(archive_.get());
+	archive_read_support_filter_xz(archive_.get());
+	archive_read_support_format_tar(archive_.get());
```
vpodzime, line 52: No return values from these?

artifact/tar/platform/libarchive/wrapper.cpp
vpodzime, line 57: Why not `TarReaderError`?

artifact/tar/platform/libarchive/wrapper.cpp
```diff
@@ -0,0 +1,105 @@
...
+			std::error_condition(std::errc::inappropriate_io_control_operation),
+			"Failed to initalize the 'libarchive' C bindings");
+	}
+	this->initalized_ = true;
+	return error::NoError;
+}
+
+int FreeLibArchiveHandle(archive *a) {
+	int r = 0;
+	if (a) {
```
vpodzime, line 66: `a != nullptr`

artifact/tar/platform/libarchive/wrapper.cpp
```diff
@@ -0,0 +1,105 @@
...
+		r = archive_read_free(a);
+		if (r != ARCHIVE_OK) {
+			log::Error("Failed to free the resources from the Archive");
+		}
+	}
+	return r;
```
vpodzime, line 72: Any point in returning `r` here? Aren't the `free` functions `void ...`?

artifact/tar/platform/libarchive/wrapper.cpp
```diff
@@ -0,0 +1,105 @@
...
+}
+
+Handle::Handle(mender::common::io::Reader &reader) :
+	archive_(archive_read_new(), FreeLibArchiveHandle),
+	reader_container_ {reader, 4096} {
```
vpodzime, line 77: No magic numbers, please.

artifact/tar/platform/libarchive/wrapper.cpp
```diff
@@ -0,0 +1,105 @@
...
+		log::Error("Failed to initialize the Archive handle");
+	}
+}
+
+
+mender::common::expected::Expected<size_t, mender::common::error::Error> Handle::Read(
+	vector<uint8_t>::iterator start, vector<uint8_t>::iterator end) {
+	if (!initalized_) {
+		return common::error::MakeError(
+			common::error::GenericError,
```
vpodzime, line 89: Again, `TarReaderError`?

artifact/tar/platform/libarchive/wrapper.cpp
```diff
@@ -0,0 +1,105 @@
...
+			"Unable to read from a tar reader which is not initialized properly");
+	}
+	size_t iterator_size = end - start;
```
vpodzime, line 92: So is this warning about narrowing the value space?

artifact/tar/platform/libarchive/wrapper.cpp
```diff
@@ -0,0 +1,105 @@
...
+	ssize_t read_bytes = archive_read_data(archive_.get(), &start[0], iterator_size);
+
+	if (read_bytes < 0) {
+		return common::error::MakeError(
+			common::error::GenericError, "Failed to read from the underlying tar stream");
```
vpodzime, line 97: `TarReaderError`

artifact/tar/platform/libarchive/wrapper.cpp
```diff
@@ -0,0 +1,105 @@
...
+
+mender::common::expected::Expected<size_t, mender::common::error::Error> Handle::Read(
+	vector<uint8_t>::iterator start, vector<uint8_t>::iterator end) {
+	if (!initalized_) {
+		return common::error::MakeError(
+			common::error::GenericError,
+			"Unable to read from a tar reader which is not initialized properly");
+	}
+	size_t iterator_size = end - start;
+	ssize_t read_bytes = archive_read_data(archive_.get(), &start[0], iterator_size);
```
vpodzime, line 93: `start.data()`

artifact/tar/platform/libarchive/tar.cpp
```diff
@@ -0,0 +1,129 @@
...
+	case TarEOFError:
+		return "Archive EOF reached";
+	default:
+		return "Unknown";
+	}
+}
+
+error::Error MakeError(ErrorCode code) {
+	return error::Error(error_condition(code, ErrorCategory), "");
+}
```
vpodzime, line 57: We don't want this.

artifact/tar/platform/libarchive/tar.cpp
```diff
@@ -0,0 +1,129 @@
...
+error::Error MakeError(ErrorCode code, const string &msg) {
+	return error::Error(error_condition(code, ErrorCategory), msg);
+}
+
+
+mender::common::expected::Expected<size_t, mender::common::error::Error> Entry::Read(
+	vector<uint8_t>::iterator start, vector<uint8_t>::iterator end) {
```
vpodzime, line 64: Feels like we should have a `ByteVector` defined in some common place.

artifact/tar/tar.hpp
vpodzime, line 89: `archive_handle_`

artifact/tar/tar.hpp
```diff
@@ -0,0 +1,104 @@
...
+	std::string name_;
+	uint64_t total_size_;
+
+	Reader &reader_;
+
+	// Reader data
+	size_t nr_bytes_read_ {0};
+
+public:
+	Entry(std::string name, uint64_t archive_size, Reader &reader) :
```
vpodzime, line 70: `size_t`?

artifact/tar/tar.hpp
```diff
@@ -0,0 +1,104 @@
...
+	mender::libarchive::wrapper::Handle archive_handle;
+#endif
+
+	expected::ExpectedSize Read(
+		vector<uint8_t>::iterator start, vector<uint8_t>::iterator end) override;
+
+public:
+	Reader(mender::common::io::Reader &reader);
+
+	ExpectedEntry Next();
```
vpodzime, line 98: Could we move this `Next()` to the `Handle` class and get rid of this `Reader` class?

artifact/tar/platform/libarchive/tar.cpp
```diff
@@ -0,0 +1,129 @@
...
+// * Archive size
+ExpectedEntry Reader::Next() {
+	int r {-1};
+	struct archive_entry *current_entry {};
+
+	if (archive_handle.Get() == nullptr) {
+		return MakeError(TarEntryError, "No underlying stream to read from");
+	}
+
+	r = archive_read_next_header(archive_handle.Get(), &current_entry);
```
vpodzime, line 110: `int r = ...`, no need to define it above.

artifact/tar/platform/libarchive/tar.cpp
```diff
@@ -0,0 +1,129 @@
...
+Reader::Reader(mender::common::io::Reader &reader) :
+	archive_handle {reader} {
+}
+
+// Read the next Tar header, and populate the meta-data:
+// * name
+// * Archive size
+ExpectedEntry Reader::Next() {
+	int r {-1};
+	struct archive_entry *current_entry {};
```
vpodzime, line 104: It's better to keep this unitialized, I think.

artifact/tar/tar_test.cpp
```diff
@@ -0,0 +1,165 @@
...
+	static void SetUpTestSuite() {
+		string script = R"(#! /bin/sh
+
+# Create small tar file
+echo foobar > testdata
+tar cvf test.tar testdata
+
+# Create large tar file
+dd if=/dev/random of=testinput.large bs=1M count=4
+tar cvf test-large.tar testinput.large
```
vpodzime, line 48: Should we use `z` as well to check that compressed tarballs work too?

artifact/tar/tar_test.cpp
```diff
@@ -0,0 +1,165 @@
...
+		EXPECT_EQ(proc.GetExitStatus(), 0) << "error message: " + ex_line_data.error().message;
+	}
+
+	static void TearDownTestSuite() {
+		std::remove("test.tar");
+		std::remove("test-script.sh");
+		std::remove("test-large.tar");
+		std::remove("testdata");
+		std::remove("testinput.large");
+	}
```
vpodzime, line 75: Better use `TemporaryDirectory` and put everything in it.

vpodzime (commented):
Mostly small things, but I believe things could be simplified a bit if moved around and merged.

artifact/tar/platform/libarchive/wrapper.cpp
oleorhagen: Went with `GenericError`

artifact/tar/platform/libarchive/wrapper.hpp
```diff
@@ -0,0 +1,94 @@
...
+#ifndef MENDER_LIB_ARCHIVE_WRAPPER_HPP
+#define MENDER_LIB_ARCHIVE_WRAPPER_HPP
+
+#include <cstddef>
+#include <memory>
+#include <vector>
+
+#include <archive.h>
+#include <archive_entry.h>
+#include <sys/types.h>
```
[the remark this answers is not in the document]
oleorhagen: Hmm, good point. Tbh, I just assumed this was for the `C` wrapper, I'll give `cctype` a spin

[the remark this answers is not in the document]
oleorhagen: Removed it altogether and it's still happy

artifact/tar/platform/libarchive/wrapper.cpp
```diff
@@ -0,0 +1,105 @@
...
+namespace mender {
+namespace libarchive {
+namespace wrapper {
+
+// The reader_callback is invoked whenever the library requires raw bytes from
+// the archive. The read callback reads data into a buffer, sets the const void
+// **buffer argument to point to the available data, and returns a count of the
+// number of bytes available. LibArchive will invoke the read callback again
+// only after it has consumed this data. LibArchive imposes no constraints on
+// the size of the data blocks returned.
```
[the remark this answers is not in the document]
oleorhagen: haha, it was for future me as well ;)

artifact/tar/platform/libarchive/wrapper.cpp
oleorhagen: It will have to be `.data()[0]`. But if that's fine with you, then all for it

artifact/tar/platform/libarchive/wrapper.hpp
oleorhagen: mby, but does it really make a difference?

artifact/tar/platform/libarchive/wrapper.cpp
oleorhagen: :+1: 

Which begs the question, should we treat them all as errors (?) :thinking: 

`Tar` we will have to. But it can function just fine without the other decompression algorithms.

What do you think @kacf ?

artifact/tar/platform/libarchive/wrapper.cpp
oleorhagen: The `C++` free functions you mean?
They don't seem to have a required type.
So I guess `void` is more appropriate

artifact/tar/platform/libarchive/wrapper.cpp
oleorhagen: Hmm, I was planning on it initially, but then I will either have to:

1. Split the errors out from `tar.hpp`, as I don't want to include `tar` -> `libarchive-wrapper`, and hence make the dependency cyclical.
2. Just use the generic error here. I'm not even sure I am going to have individual errors for all the sub-components of the parser. They will add up, as there are ~10 components in the end.

What you think?

artifact/tar/platform/libarchive/wrapper.cpp
oleorhagen: Yes :persevere:

artifact/tar/platform/libarchive/wrapper.cpp
oleorhagen: Same as above: `data()[0]` ?

artifact/tar/platform/libarchive/tar.cpp
oleorhagen: It's just initialized to `nullptr`. 
I think that's fine?
Maybe explicitly set it to `nullptr`?

artifact/tar/tar.hpp
oleorhagen: Hmm, yeh, mby.
You're more familiar with this than me, but `libarchive` defines this:
```
__LA_DECL la_int64_t	 archive_entry_size(struct archive_entry *);
```
What should we map it to internally?

artifact/tar/tar.hpp
oleorhagen: I don't think we can get completely rid of it, something needs to own the `unique_ptr`, or rather, the `handle` in this instance.

artifact/tar/tar.hpp
oleorhagen: I wind up returning into an `ssize_t`, checking for negative values, if not, `static_cast` to `size_t`, and use that value for our internal `tar` interface

artifact/tar/platform/libarchive/wrapper.cpp
kacf: First of all, don't forget that we also need [zstd support](https://northerntech.atlassian.net/browse/MEN-6207) (maybe not in this PR, but before you close the task).

As discussed f2f, error handling is good, but let's make sure they can be compiled out (the current Golang client supports it).

artifact/tar/platform/libarchive/tar.cpp
kacf: Still not quite what I was looking for. You need to move the assert and return outside the `switch` block and remove the `default`. The intention is to have the compiler warn us when there is a enum value which is isn't covered, but the `default` catch-all prevents that.

artifact/tar/tar_test.cpp
```diff
@@ -0,0 +1,197 @@
...
+		ASSERT_TRUE(ex_line_data);
+		EXPECT_EQ(proc.GetExitStatus(), 0) << "error message: " + ex_line_data.error().message;
+	}
+
+	static void TearDownTestSuite() {
+		delete tmpdir;
+		tmpdir = nullptr;
+	}
+
+	static mendertesting::TemporaryDirectory *tmpdir;
```
kacf, line 92: Even though you can't rely on scope to determine the lifetime of `tmpdir`, I suggest using `unique_ptr<mendertesting::TemporaryDirectory>` and call `.reset()` above instead of `delete` followed by null pointer assignment. I know it makes no difference in this particular case, but we never know what might change later. Or be copied to somewhere else.

artifact/tar/platform/libarchive/tar.cpp
oleorhagen: Ahh, alright. With you! Fixed

artifact/tar/tar_test.cpp
oleorhagen: For regular code yes, but for this? 
I've literally pasted this out of the `googletest` manual.

kacf: If it made it more difficult, I could bend, but this should be the same, and actually removes one line, I think. Let's stick to the no-pointers rule as much as we can.

artifact/tar/platform/libarchive/wrapper.cpp
vpodzime: ? `data()` gives a pointer to the buffer, right? So you should be able to use it as a pointer to the buffer (dropping the `&` of course).

artifact/tar/platform/libarchive/tar.cpp
vpodzime: If you keep it uninitialized, you give the compiler a chance to warn you if you use it without assigning into it first.

artifact/tar/tar.hpp
vpodzime: I don't follow, sorry. This holds an instance of `Handle`. It actually wraps the `Handle` and provides one extra function, `Next()` on top of it. So you can either move this `Next()` function to the `Handle` class and get rid of this `Reader` class or you can base this class on top of the `Handle` class (inherit it) and then only implement the extra `Next()` function here.

artifact/tar/platform/libarchive/wrapper.cpp
```diff
@@ -0,0 +1,142 @@
...
+//
+// - On EOF return 0.
+// - On error return -1.
+ssize_t reader_callback(archive *archive, void *in_reader_container, const void **buff) {
+	ReaderContainer *p_reader_container = static_cast<ReaderContainer *>(in_reader_container);
```
vpodzime, line 40: Wouldn't `reinterpret_cast` be a better fit here?

artifact/tar/platform/libarchive/wrapper.cpp
```diff
@@ -0,0 +1,142 @@
...
+
+
+ExpectedSize Handle::Read(vector<uint8_t>::iterator start, vector<uint8_t>::iterator end) {
+	if (!initalized_) {
+		return common::error::MakeError(
+			common::error::GenericError,
+			"Unable to read from a tar reader which is not initialized properly");
+	}
+	size_t iterator_size {static_cast<size_t>(end - start)};
+	ssize_t read_bytes {archive_read_data(archive_.get(), &start[0], iterator_size)};
```
vpodzime, line 119: I'd use good old assignment for these, but YMMV.

artifact/tar/tar_test.cpp
```diff
@@ -0,0 +1,194 @@
...
+		# Create large tar file
+		dd if=/dev/random of=${DIRNAME}/testinput.large bs=1M count=4
+		tar cvf ${DIRNAME}/test-large.tar ${DIRNAME}/testinput.large
+
+    # Create a corrupt tar file
+    cp ${DIRNAME}/test.tar ${DIRNAME}/test-corrupt.tar
+    dd if=/dev/random of=${DIRNAME}/test-corrupt.tar seek=10 count=5 bs=1 conv=notrunc
+
+		exit 0
+		)";
```
vpodzime, line 65: Weird indentation above.

vpodzime (commented):
Just nitpicks. But I added a couple comments above in the conversations from the previous review.

artifact/tar/platform/libarchive/wrapper.cpp
kacf: `static_cast` should always be preferred over `reinterpret_cast` if it works, because it is more restrictive. IOW it can catch later mistakes more effectively than `reinterpret_cast` can.

kacf (approved):

artifact/tar/platform/libarchive/wrapper.cpp
vpodzime: But when casting a `void` pointer, `static_cast` does nothing, right? (just trying to learn here)

kacf: Do you mean runtime checks? It's `dynamic_cast` which does that. `static_cast` never does any runtime checks, but it does check at compile time that the cast is "potentially valid". It allows casts from `void *` because they can be anything, but it does not allow unrelated types such as `Reader *` to `Writer *` or `int *` to `string *`. That is when you need `reinterpret_cast`, which allows practically anything.

kacf: And in case you wonder: We can not use `dynamic_cast` here, because it cannot cast from `void *`.

There are other reasons not to use it too, mainly that the runtime data it requires grows the binary size quite a bit, and some customers might want to keep that disabled (`-fno-rtti` option IIRC).

artifact/tar/platform/libarchive/wrapper.cpp
oleorhagen: Hmm, I thought so too, but didn't get it to work. 

Does now though... Maybe I left the `&` at the front?

Fixed!

artifact/tar/platform/libarchive/wrapper.hpp
oleorhagen: Hmm, kk. If it's the way it's usually done. 

It hides less data though, but have at it.

artifact/tar/platform/libarchive/wrapper.cpp
oleorhagen: Actually, `int archive_read_free` is the signature of the free function, so I will leave this be, to have it match the signature, even though it is not used.

artifact/tar/platform/libarchive/tar.cpp
oleorhagen: Good point!

artifact/tar/platform/libarchive/wrapper.cpp
oleorhagen: https://google.github.io/styleguide/cppguide.html#Variable_and_Array_Initialization

Seems they allow both

artifact/tar/tar.hpp
oleorhagen: Solved f2f. Leaving it be for now. It is OK as is.

PR #1179, opened 2023-03-02
fix(cmake): Avoid inheriting conflicting flags from static libs.

kacf:
Because flags are inherited from other libraries when you link to them, this does not work well for conflicting flags. Instead, introduce a default for all components, and then override using `target_compile_options`, which always puts the flags last.

Changelog: None
Ticket: None

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -79,28 +79,14 @@ else()
   set(MENDER_USE_BOOST_FS 0)
 endif()
 
+# Default for all components.
+add_compile_options(-std=c++11 -Werror -Wall)
 
-#---------------------------------------------------------------------------------------------------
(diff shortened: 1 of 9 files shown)


kacf:
Hah! This even caught some [things we had missed](https://gitlab.com/Northern.tech/Mender/mender/-/jobs/3864316531) because the flags weren't getting set correctly. Instant value.

vpodzime:
> Hah! This even caught some [things we had missed](https://gitlab.com/Northern.tech/Mender/mender/-/jobs/3864316531) because the flags weren't getting set correctly. Instant value.

Mystery solved for why those things haven't being showing up in CI and our local builds.

vpodzime (approved):
ACK

oleorhagen (approved):
Nice!

PR #1180, opened 2023-03-03
Add the manifest parser implementation

oleorhagen:
This adds the initial manifest parser implementation for the Mender-artifact format parser.

### DIFF
--- a/artifact/v3/manifest/manifest.cpp
+++ b/artifact/v3/manifest/manifest.cpp
@@ -1,0 +1,95 @@
+// Copyright 2023 Northern.tech AS
+//
+//    Licensed under the Apache License, Version 2.0 (the "License");
+//    you may not use this file except in compliance with the License.
+//    You may obtain a copy of the License at
+//
+//        http://www.apache.org/licenses/LICENSE-2.0
+//
+//    Unless required by applicable law or agreed to in writing, software
+//    distributed under the License is distributed on an "AS IS" BASIS,
+//    WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+//    See the License for the specific language governing permissions and
+//    limitations under the License.
+
+#include <artifact/v3/manifest/manifest.hpp>
+
+#include <cctype>
+#include <iterator>
+#include <string>
+#include <sstream>
+#include <regex>
+
+#include <common/error.hpp>
+
+#include <artifact/error.hpp>
+
+namespace mender {
+namespace artifact {
+namespace v3 {
+namespace manifest {
+
+namespace io = mender::common::io;
+namespace error = mender::common::error;
+
+struct ManifestLine {
+	const string shasum;
+	const string entry_name;
+};
+
+using ExpectedManifestLine = expected::expected<ManifestLine, error::Error>;
+
+const size_t expected_shasum_length {64};
+const size_t expected_whitespace {2};
+const string manifest_line_regex_string {
+	"^([0-9a-z]{" + to_string(expected_shasum_length) + "})[[:space:]]{"
+	+ to_string(expected_whitespace) + "}([/.[:alnum:]]+)$"};
+
+namespace {
+ExpectedManifestLine Tokenize(const string &line) {
+	const std::regex manifest_line_regex(
+		manifest_line_regex_string, std::regex_constants::ECMAScript);
+	std::smatch base_match;
+	std::regex_match(line, base_match, manifest_line_regex);
+
+	if (base_match.size() != 3) {
+		return expected::unexpected(parser_error::MakeError(
+			parser_error::ParseError,
+			"Line (" + line
+				+ ") is not in the expected manifest format: " + manifest_line_regex_string));
+	}
+
+	return ManifestLine {.shasum = base_match[1], .entry_name = base_match[2]};
+}
+} // namespace
+
+ExpectedManifest Parse(mender::common::io::Reader &reader) {
+	Manifest m {};
+
+	stringstream ss {};
+
+	io::StreamWriter sw {ss};
+
+	auto err = io::Copy(sw, reader);
+	if (err) {
+		return expected::unexpected(err);
+	}
+
+	string line {};
+
+	while (getline(ss, line, '\n')) {
+		auto manifest_line = Tokenize(line);
+		if (!manifest_line) {
+			return expected::unexpected(manifest_line.error());
+		}
+
+		m.map_[manifest_line->entry_name] = manifest_line->shasum;
+	}
+
+	return m;
+}
+
+} // namespace manifest
+} // namespace v3
+} // namespace artifact
+} // namespace mender
--- a/artifact/v3/manifest/manifest.hpp
+++ b/artifact/v3/manifest/manifest.hpp
@@ -1,0 +1,57 @@
+// Copyright 2023 Northern.tech AS
+//
+//    Licensed under the Apache License, Version 2.0 (the "License");
+//    you may not use this file except in compliance with the License.
+//    You may obtain a copy of the License at
+//
+//        http://www.apache.org/licenses/LICENSE-2.0
+//
+//    Unless required by applicable law or agreed to in writing, software
+//    distributed under the License is distributed on an "AS IS" BASIS,
+//    WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+//    See the License for the specific language governing permissions and
+//    limitations under the License.
+
+#ifndef MENDER_ARTIFACT_V3_MANIFEST_PARSER_HPP
+#define MENDER_ARTIFACT_V3_MANIFEST_PARSER_HPP
+
+#include <string>
+#include <unordered_map>
+
+#include <common/expected.hpp>
+#include <common/error.hpp>
+#include <common/io.hpp>
+
+namespace mender {
+namespace artifact {
+namespace v3 {
+namespace manifest {
+
+using namespace std;
+
+namespace expected = mender::common::expected;
+namespace io = mender::common::io;
+namespace error = mender::common::error;
+
+class Manifest {
+public:
+	string Get(const string &key) {
+		auto value = map_.find(key);
+		if (value != map_.end()) {
+			return value->second;
+		}
+		return "";
+	}
+
+	unordered_map<string, string> map_;
+};
+
+using ExpectedManifest = expected::expected<Manifest, error::Error>;
+
+ExpectedManifest Parse(io::Reader &reader);
+
+} // namespace manifest
+} // namespace v3
+} // namespace artifact
+} // namespace mender
+#endif // MENDER_ARTIFACT_V3_MANIFEST_PARSER_HPP
--- a/artifact/v3/manifest/manifest_test.cpp
+++ b/artifact/v3/manifest/manifest_test.cpp
@@ -1,0 +1,147 @@
+// Copyright 2023 Northern.tech AS
+//
+//    Licensed under the Apache License, Version 2.0 (the "License");
+//    you may not use this file except in compliance with the License.
+//    You may obtain a copy of the License at
+//
+//        http://www.apache.org/licenses/LICENSE-2.0
+//
+//    Unless required by applicable law or agreed to in writing, software
+//    distributed under the License is distributed on an "AS IS" BASIS,
+//    WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+//    See the License for the specific language governing permissions and
+//    limitations under the License.
+
+#include <artifact/v3/manifest/manifest.hpp>
+
+#include <string>
+
+#include <gtest/gtest.h>
+#include <gmock/gmock.h>
+
+using namespace std;
+
+TEST(ParserTest, TestParseManifest) {
+	std::string manifest_data =
+		R"(aec070645fe53ee3b3763059376134f058cc337247c978add178b6ccdfb0019f  data/0000.tar
+9f65db081a46f7832b9767c56afcc7bfe784f0a62cc2950b6375b2b6390e6e50  header.tar
+96bcd965947569404798bcbdb614f103db5a004eb6e364cfc162c146890ea35b  version
+)";
+
+	std::stringstream ss {manifest_data};
+
+	mender::common::io::StreamReader sr {ss};
+
+	auto manifest = mender::artifact::v3::manifest::Parse(sr);
+
+	ASSERT_TRUE(manifest) << "error message: " << manifest.error().message;
+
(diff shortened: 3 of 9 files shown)


artifact/v3/manifest/manifest.hpp
```diff
@@ -0,0 +1,54 @@
...
+namespace v3 {
+namespace manifest {
+
+using namespace std;
+
+namespace expected = mender::common::expected;
+namespace io = mender::common::io;
+
+struct Manifest {
+	unordered_map<string, string> m;
```
kacf, line 35: I wonder if we need a `vector` instead, because order matters when we start reading the payloads later.

Although if you can look up by name, then maybe it is enough to store the index together with the checksum, so that you can check what it is while reading.

artifact/v3/manifest/manifest.cpp
```diff
@@ -0,0 +1,68 @@
...
+			return expected::unexpected(error::MakeError(
+				error::GenericError,
+				"Manifest sha's need to be of size " + to_string(expected_shasum_length)
+					+ ". Received: " + to_string(shasum.size())));
+		}
+		/* Ignore the leading whitespace in between entries */
+		ss >> std::ws >> entry_name;
+		if (entry_name == "") {
+			return expected::unexpected(error::MakeError(
+				error::GenericError, "No name found for the manifest entry: " + shasum));
```
kacf, line 58: Shouldn't we at least have a generic `ParsingError` for the artifact parsing? `GenericError` is extremely generic, it's when you don't even have a good category.

artifact/v3/manifest/manifest.cpp
```diff
@@ -0,0 +1,68 @@
...
+#include <artifact/v3/manifest/manifest.hpp>
+
+#include <string>
+#include <sstream>
+
+#include <common/error.hpp>
+
+namespace artifact {
+namespace v3 {
+namespace manifest {
```
kacf, line 24: I probably missed this earlier, but should the first namespace be `mender`?

artifact/v3/manifest/manifest_test.cpp
```diff
@@ -0,0 +1,87 @@
...
+#include <gtest/gtest.h>
+#include <gmock/gmock.h>
+
+using namespace std;
+
+TEST(ParserTest, TestParseManifest) {
+	std::string manifest_data =
+		R"(aec070645fe53ee3b3763059376134f058cc337247c978add178b6ccdfb0019f  data/0000.tar
+9f65db081a46f7832b9767c56afcc7bfe784f0a62cc2950b6375b2b6390e6e50  header.tar
+96bcd965947569404798bcbdb614f103db5a004eb6e364cfc162c146890ea35b  version
```
kacf, line 28: Testing some of the examples I gave above might be prudent here, and they should all fail:

* Wrong number of spaces.
* All on one line, instead of several lines.
* All whitespace replaced with newlines.

kacf (changes requested):

artifact/v3/manifest/manifest.cpp
```diff
@@ -0,0 +1,68 @@
...
+	while (!ss.eof()) {
+		ss >> shasum;
+		if (shasum.size() != expected_shasum_length) {
+			return expected::unexpected(error::MakeError(
+				error::GenericError,
+				"Manifest sha's need to be of size " + to_string(expected_shasum_length)
+					+ ". Received: " + to_string(shasum.size())));
+		}
+		/* Ignore the leading whitespace in between entries */
+		ss >> std::ws >> entry_name;
```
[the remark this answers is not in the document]
oleorhagen: Roger. Strictness-level 11 it is then!

artifact/v3/manifest/manifest.hpp
oleorhagen: I'd rather keep it like this actually, we do lookups by name.

artifact/v3/manifest/manifest.cpp
oleorhagen: I might make an error class to share in between all the artifact stuff then

artifact/v3/manifest/manifest.cpp
oleorhagen: Probs right.

oleorhagen: so, just to be sure: `mender::artifact::v3::manifest` ?

kacf: Yes. I didn't want to type it out in case you thought I meant to change it to the non C++11 style. 🙂

common/io.cpp
```diff
@@ -64,6 +64,12 @@ ExpectedSize ByteWriter::Write(
 	return bytes_to_write;
 }
 
+ExpectedSize StreamWriter::Write(
+	vector<uint8_t>::const_iterator start, vector<uint8_t>::const_iterator end) {
+	os_.write(reinterpret_cast<const char *>(&*start), end - start);
```
vpodzime, line 69: No short writes possible? (would surprise me if not)

artifact/v3/manifest/manifest.cpp
```diff
@@ -0,0 +1,98 @@
...
+	auto err = io::Copy(sw, reader);
+	if (err) {
+		return expected::unexpected(err);
+	}
+
+	string entry_name {};
+	string shasum {};
+	string line {};
+
+	while (getline(ss, line, '\n')) {
```
vpodzime, line 49: Ole, forgive me, no offense to you, this is nothing personal and I know this is how we decided to do things.

But the above is really just crazy to me. We want to read lines from a file. Our answer is to:
1. create a file stream
2. create a Reader on top of the stream
3. create a Writer on top of a string stream
4. use Copy to shovel data between the Reader and Writer and thus from one stream to the other
5. get lines from the string stream

I don't even want to count the buffers used in the above, but if our target is to limit the amount of our code and (re)use what's available, why are we doing this to ourselves?

artifact/v3/manifest/manifest.cpp
```diff
@@ -0,0 +1,98 @@
...
+	io::StreamWriter sw {ss};
+
+	auto err = io::Copy(sw, reader);
+	if (err) {
+		return expected::unexpected(err);
+	}
+
+	string entry_name {};
+	string shasum {};
+	string line {};
```
vpodzime, line 47: All of these can be declared inside the loop, ideally in the place where they are used.

artifact/v3/manifest/manifest.cpp
```diff
@@ -0,0 +1,98 @@
...
+
+	while (getline(ss, line, '\n')) {
+		/* Since the shasums are 64 character, and two whitespaces are required, and the shortest
+		 * name is 1 character */
+		if (line.size() < expected_shasum_length + expected_whitespace + 1) {
+			return expected::unexpected(error::MakeError(
+				error::GenericError, "Malformed input line. Line too short. Line: " + line));
```
vpodzime, line 54: I agree with Kristian we should at least have some generic `ArtifactParseError` instead of using `GenericError` everywhere.

artifact/v3/manifest/manifest.cpp
```diff
@@ -0,0 +1,98 @@
...
+		}
+
+		auto delimeter_pos = line.find("  ");
+		if (delimeter_pos == string::npos) {
+			return expected::unexpected(error::MakeError(
+				error::GenericError,
+				"Malformed manifest line. No double whitespace separator found, line: " + line));
+		}
+		if (delimeter_pos < expected_shasum_length) {
```
vpodzime, line 63: Shouldn't this be `!= expected_shasum_length`? It should be right after the shasum, right?

artifact/v3/manifest/manifest.cpp
vpodzime: Plus the fact that a large input file will just kill any process running this code if I understand correctly what's actually happening in those 5 steps above.

artifact/v3/manifest/manifest.cpp
```diff
@@ -0,0 +1,98 @@
...
+			return expected::unexpected(error::MakeError(
+				error::GenericError,
+				"Manifest sha's need to be of size 64. Received: " + to_string(delimeter_pos)));
+		}
+
+		if (line.substr(expected_shasum_length, expected_whitespace) != "  ") {
```
vpodzime, line 69: If two spaces are required, it should check the two characters (`line[expected_shasum_length]` and `+1`) for being space and then the one after that for not being a space.

vpodzime: I see the check is done on `entry_name` below. Would make more sense to me to do it here, but YMMV.

artifact/v3/manifest/manifest.cpp
```diff
@@ -0,0 +1,98 @@
...
+
+		if (entry_name.at(0) == ' ') {
+			return expected::unexpected(error::MakeError(
+				error::GenericError,
+				"More than two spaces found in between the shasum and the name. This is not allowed"));
+		}
+		if (entry_name.find(" ") != string::npos) {
+			return expected::unexpected(
+				error::MakeError(error::GenericError, "Entry name contains spaces."));
+		}
```
vpodzime, line 87: No further restrictions for the entry name?

artifact/v3/manifest/manifest.hpp
```diff
@@ -0,0 +1,56 @@
...
+namespace manifest {
+
+using namespace std;
+
+namespace expected = mender::common::expected;
+namespace io = mender::common::io;
+namespace error = mender::common::error;
+
+struct Manifest {
+	unordered_map<string, string> m;
```
vpodzime, line 37: `m` is a not a great name for the variable (noticed when I saw `m.m` above). Also, should it be `private` (and thus with a trailing `_`)?

artifact/v3/manifest/manifest.hpp
```diff
@@ -0,0 +1,56 @@
...
+namespace v3 {
+namespace manifest {
+
+using namespace std;
+
+namespace expected = mender::common::expected;
+namespace io = mender::common::io;
+namespace error = mender::common::error;
+
+struct Manifest {
```
vpodzime, line 36: I don't like `struct`s with functions (methods), but YMMV.

vpodzime (commented):
Feel free to stop requesting review from me.

artifact/v3/manifest/manifest.hpp
oleorhagen: I'll have a look in the style-guide. I'm fine either way

artifact/v3/manifest/manifest.cpp
oleorhagen: `Variables needed for if, while and for statements should normally be declared within those statements, so that such variables are confined to those scopes. E.g.:`

Seems ur right: https://google.github.io/styleguide/cppguide.html#Variable_and_Array_Initialization

artifact/v3/manifest/manifest.hpp
oleorhagen: I guess it's a little bit up for impretation:

`Use a struct only for passive objects that carry data; everything else is a class.`

https://google.github.io/styleguide/cppguide.html#Structs_vs._Classes

It is made just for carrying data, but it has an `interface` through `Get`. I'm guessing I'll favour your judgement and go with class :+1:

artifact/v3/manifest/manifest.cpp
vpodzime: > because we still internally build up a JSON structure 

AFAICT, here we don't build any JSON structure, we just read lines from a file.

vpodzime: > getline requires a stream instead of iterators or a buffer. This is easily solvable by implementing the inverse of [StreamReader](https://github.com/mendersoftware/mender/blob/feature-c%2B%2B-client/common/io.hpp#L108).

So we would have `StreamReader` wrapping a stream into a `Reader` and then `ReaderStream` which would provide a stream for a given `Reader`? That sounds twisted to me.

vpodzime: I believe the general issue is that the Reader/Writer concept is not native to C++. C++ has streams, `istream` for input and `ostream` for output. Libraries and even the standard library are implemented the way that they work with streams. Why should we introduce the Reader/Writer concept into that and break/disrupt things? What is the benefit? AFAICT, everywhere where we have a `Reader` an `istream` would do and the same applies to `Writer` and `ostream`. Then we would be able to stream-process data and use the C++-native APIs without any extra code. Instead of wrapping a stream into our class and then providing the stream API on top of it, we could just work with the streams directly. C++ already has an abstraction for I/O, there's no need to introduce another one, IMHO.

The question is how hard it is to provide a stream on top of a buffer-based API, like the one that Beast or tiny-process-lib provide. That I will try to find out.

Also, this discussion is unrelated to this PR, we might as well take it somewhere else.

kacf: Correct me if I'm wrong, but AFAIK the C++ streams are not virtual, and therefore they are unsuitable for interfaces. I believe they also require a lot more methods to be implemented to subclass them. But feel free to suggest improvements which are easier to use than Reader and Writer.

oleorhagen: I am going to close this for now. This can be handled out of band for this PR.

oleorhagen:
Alright, new try. I folded, and decided to use the std regex lib, instead of making an ad-hoc simple parser for this.

I think it turned out pretty clean this time around.

artifact/v3/manifest/manifest.cpp
```diff
@@ -0,0 +1,93 @@
...
+	const string shasum;
+	const string entry_name;
+};
+
+using ExpectedManifestLine = expected::expected<ManifestLine, error::Error>;
+
+const size_t expected_shasum_length {64};
+const size_t expected_whitespace {2};
+const string manifest_line_regex_string {
+	"^([[:alnum:]]{" + to_string(expected_shasum_length) + "})[[:space:]]{"
```
kacf, line 45: Shouldn't `[[:alnum:]]` instead be `[0-9a-f]`?

artifact/v3/manifest/manifest.cpp
```diff
@@ -0,0 +1,93 @@
...
+	+ to_string(expected_whitespace) + "}([/.[:alnum:]]+)$"};
+
+ExpectedManifestLine tokenize(const string &line) {
```
kacf, line 48: Should be `Tokenize` and `static` (if you want to mark it as local).

kacf (approved):

artifact/v3/manifest/manifest.cpp
oleorhagen: Made it an anonymous namespace

PR #1184, opened 2023-03-06
feat: Introduce the UpdateModule class

vpodzime:
Only as a skeleton for now, just to set up the basic structure and placement of the code.

Ticket: MEN-6312
Changelog: none

### DIFF
--- a/mender-update/CMakeLists.txt
+++ b/mender-update/CMakeLists.txt
@@ -1,7 +1,9 @@
 include_directories(${CMAKE_SOURCE_DIR} ${CMAKE_BINARY_DIR})
 
+add_library(update_module STATIC update_module/v3/update_module.cpp)
+
 add_executable(mender-update main.cpp)
-target_link_libraries(mender-update PRIVATE common_json common_log common_http common_conf)
+target_link_libraries(mender-update PRIVATE common_json common_log common_http common_conf update_module)
 install(TARGETS mender-update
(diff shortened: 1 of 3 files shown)


oleorhagen:
I will defer this one to Kristian :)

mender-update/CMakeLists.txt
```diff
@@ -1,7 +1,10 @@
 include_directories(${CMAKE_SOURCE_DIR} ${CMAKE_BINARY_DIR})
 
+add_library(update_module STATIC v3/update_module.cpp)
```
kacf, line 3: I suggest using a path like `update-module/v3` instead, since there are other components at other, unrelated versions, such as D-Bus API at v1.

mender-update/v3/update_module.cpp
```diff
@@ -0,0 +1,73 @@
...
+//    limitations under the License.
+
+#include <mender-update/v3/update_module.hpp>
+
+#include <common/error.hpp>
+#include <common/expected.hpp>
+
+namespace mender {
+namespace update {
+namespace v3 {
```
kacf, line 22: Same here, I propose getting either "module" or "update_module" in between `update` and `v3`.

kacf (approved):
👍

A couple of comments on path and namespace, but it's easy to fix, so I'm pre-approving!

PR #1186, opened 2023-03-09
Artifact Payload parser implementation

oleorhagen:
This adds the payload parser implementation.

Interface:

```
payload::Verify(io::Reader) -> Reader
```

The payload is itself offering a Reader, and will automatically verify the
payload checksum upon write, and if the checksum does not match, will throw an
error at the end of the write.

Interface:

```
p = Payload.Next()
.Read() -> expectedSize
```

Ticket: MEN-6205
Changelog: None

### DIFF
--- a/artifact/v3/payload/payload.cpp
+++ b/artifact/v3/payload/payload.cpp
@@ -1,0 +1,35 @@
+// Copyright 2023 Northern.tech AS
+//
+//    Licensed under the Apache License, Version 2.0 (the "License");
+//    you may not use this file except in compliance with the License.
+//    You may obtain a copy of the License at
+//
+//        http://www.apache.org/licenses/LICENSE-2.0
+//
+//    Unless required by applicable law or agreed to in writing, software
+//    distributed under the License is distributed on an "AS IS" BASIS,
+//    WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+//    See the License for the specific language governing permissions and
+//    limitations under the License.
+
+#include <artifact/v3/payload/payload.hpp>
+
+#include <vector>
+
+#include <common/io.hpp>
+
+namespace mender {
+namespace artifact {
+namespace v3 {
+namespace payload {
+
+using namespace std;
+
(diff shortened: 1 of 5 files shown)


artifact/v3/payload/payload.cpp
```diff
@@ -0,0 +1,35 @@
...
+
+#include <vector>
+
+#include <common/log.hpp>
+#include <common/io.hpp>
+#include <common/expected.hpp>
+
+namespace artifact {
+namespace v3 {
+namespace payload {
```
kacf, line 25: I think it's covered in a different PR already, but should it be `mender::artifact::v3::payload`?

kacf (approved):
You said to review only one commit, which I've done. Was that all of it?

artifact/token.hpp
```diff
@@ -0,0 +1,101 @@
...
+	{Type::Header, "header"},
+	{Type::HeaderAugment, "header-augment"},
+	{Type::Payload, "data"},
+};
+
+class Token {
+public:
+	Type type;
+	// Poor mans optional
+	shared_ptr<tar::Entry> value;
```
vpodzime, line 62: Also here, what about `unique_ptr`?

vpodzime (commented):
My two cents...

artifact/lexer.hpp
```diff
@@ -0,0 +1,74 @@
...
+namespace log = mender::common::log;
+
+class Lexer {
+private:
+	std::shared_ptr<mender::tar::Reader> &tar_reader;
+
+public:
+	token::Token current;
+
+	Lexer(std::shared_ptr<mender::tar::Reader> &tr) :
```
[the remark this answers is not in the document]
kacf: Semantically it makes sense, but I'm not sure if it's a good idea since we often don't know where a `Reader` comes from. Providers of `Reader`s will tend to provide it as `shared_ptr` to be as flexible as possible, and we can't convert a `shared_ptr` to a `unique_ptr`. For example, this is used [in the HTTP API](https://github.com/mendersoftware/mender/blob/ea30d51e0fee413a181dbe69a21d962f5dafab01/common/http.hpp#L85).

PR #1188, opened 2023-03-13
Respect errno in many places

vpodzime:


vpodzime:
> ** ci/gitlab ** — Pipeline failed on Gitlab

Hmm, there's a failure in `test:static:asan` where the `BadFileLogTest.SetupBadFileLogging` test case fails (I guess in relation to the `shared_ptr` change here). But I cannot reproduce it locally with `-DCMAKE_BUILD_TYPE=Asan`, neither with g++ nor clang++.

vpodzime:
> But I cannot reproduce it locally with `-DCMAKE_BUILD_TYPE=Asan`, neither with g++ nor clang++.

Oh, it's A**S**an.

common/common.cpp
```diff
@@ -28,7 +28,7 @@ mender::common::expected::ExpectedLongLong StringToLongLong(const string &str, i
 	if (errno != 0) {
 		int int_error = errno;
 		return mender::common::error::Error(
-			std::error_code(int_error, std::system_category()).default_error_condition(), "");
+			std::generic_category().default_error_condition(int_error), "");
```
kacf, line 31: Are you able to differentiate between https://en.cppreference.com/w/cpp/error/generic_category and https://en.cppreference.com/w/cpp/error/system_category? They appear to be saying almost the same thing, and both use `errno` as example! Not very helpful documentation...

This looks good though, better than my original approach.

common/error.hpp
```diff
@@ -60,6 +60,12 @@ class Error {
 	std::string String() const {
 		return code.message() + ": " + message;
 	}
+
+	bool IsErrno(int errno_value) {
+		return (
+			(this->code.category() == std::generic_category())
+			&& (this->code.value() == errno_value));
+	}
```
kacf, line 68: Nice!

common/io.hpp
```diff
@@ -89,9 +90,9 @@ class StreamReader : virtual public Reader {
 	ExpectedSize Read(vector<uint8_t>::iterator start, vector<uint8_t>::iterator end) override {
 		is_.read(reinterpret_cast<char *>(&*start), end - start);
 		if (is_.bad()) {
-			int int_error = errno;
-			return Error(
-				std::error_code(int_error, std::system_category()).default_error_condition(), "");
+			int io_error = errno;
+			return expected::unexpected(
+				Error(std::generic_category().default_error_condition(io_error), strerror(errno)));
```
kacf, line 95: No need for `strerror` here, you are just duplicating the message.

kacf (approved):
Nice cleanup!

common/io.hpp
vpodzime: We don't have a good way to extract the message from the `error_condition`, AFAICT. And when processing our errors as part of the `ExpectedSomething` types, we do `ex.error().message`. So either the `Error` constructor should do the something like `message(strerror(the_extracted_errno))` or we should do it explicitly like above.

common/common.cpp
vpodzime: It is very confusing. I actually go to this when searching for how to get errors from stream operation failures and it's mentioned [here](https://stackoverflow.com/a/66229636) with the following:

> `std::system_category()` will map the error codes from system's native error code facility. For *nix, this is `errno`. For Win32, it is `GetLastError()`.

So with `system_category` you should use `GetLastError()` on Windows (or whatever system-specific thing), `generic_category` works with `errno` everywhere. But it's confusing as hell, so...

common/io.hpp
kacf: Ah, but we do! Usually we don't look at `message` directly, but use [`String()`](https://github.com/mendersoftware/mender/blob/feature-c%2B%2B-client/common/error.hpp#L60), which combines both messages.

common/common.cpp
kacf: Aha! I guess that sort of makes sense, but it could definitely have been clearer. I guess we should stick to `generic_category` then, since even on Windows we would use the POSIX API as it's currently implemented.

common/io.hpp
vpodzime: > Ah, but we do! Usually we don't look at `message` directly, but use [`String()`](https://github.com/mendersoftware/mender/blob/feature-c%2B%2B-client/common/error.hpp#L60), which combines both messages.

Oh, right! I'm not sure we do use `String()` consistently, but we definitely should. I'll update this.

PR #1195, opened 2023-03-20
Use streams in I/O

vpodzime:
Instead of reading things in memory before processing them.

### DIFF
--- a/artifact/v3/version/version.cpp
+++ b/artifact/v3/version/version.cpp
@@ -72,7 +72,7 @@ ExpectedVersion Parse(io::Reader &reader) {
 		return expected::unexpected(err);
 	}
 
-	auto expected_json = json::LoadFromString(mender::common::StringFromByteVector(buf));
+	auto expected_json = json::Load(mender::common::StringFromByteVector(buf));
 
 	if (!expected_json) {
 		return expected::unexpected(MakeError(
--- a/artifact/v3/version/version.cpp
+++ b/artifact/v3/version/version.cpp
@@ -63,17 +63,7 @@ error::Error MakeError(ErrorCode code, const string &msg) {
 
 
 ExpectedVersion Parse(io::Reader &reader) {
-	// Read in all the bytes into a tmp buffer
-	std::vector<uint8_t> buf(1024);
-	io::ByteWriter bw {buf};
-
-	auto err = io::Copy(bw, reader);
-	if (err) {
-		return expected::unexpected(err);
-	}
-
-	auto expected_json = json::Load(mender::common::StringFromByteVector(buf));
-
+	auto expected_json = json::Load(reader);
 	if (!expected_json) {
 		return expected::unexpected(MakeError(
 			ParseError,
--- a/common/io.cpp
+++ b/common/io.cpp
@@ -14,6 +14,13 @@
 
 #include <common/io.hpp>
 
+#include <cerrno>
+#include <cstdint>
+#include <istream>
+#include <memory>
+#include <streambuf>
+#include <vector>
+
 namespace mender {
 namespace common {
 namespace io {
@@ -64,6 +71,7 @@ ExpectedSize ByteWriter::Write(
 	return bytes_to_write;
 }
 
+
 ExpectedSize StreamWriter::Write(
 	vector<uint8_t>::const_iterator start, vector<uint8_t>::const_iterator end) {
 	os_.write(reinterpret_cast<const char *>(&*start), end - start);
@@ -72,6 +80,90 @@ ExpectedSize StreamWriter::Write(
 	}
 	return end - start;
 }
+
+class ReaderStreamBuffer : public streambuf {
+public:
+	ReaderStreamBuffer(Reader &reader) :
+		reader_ {reader},
+		buf_(buf_size_) {};
+	streambuf::int_type underflow() override;
+
+private:
+	static const Vsize buf_size_ = 1024;
+	Reader &reader_;
+	vector<uint8_t> buf_;
+};
+
+streambuf::int_type ReaderStreamBuffer::underflow() {
+	// eback -- pointer to the first char (byte)
+	// gptr  -- pointer to the current char (byte)
+	// egptr -- pointer past the last char (byte)
+
+	// This function is only called if gptr() == nullptr or gptr() >= egptr(),
+	// i.e. if there's nothing more to read.
+	if (this->gptr() >= this->egptr()) {
+		errno = 0;
+		auto ex_n_read = reader_.Read(buf_.begin(), buf_.end());
+		streamsize n_read;
+		if (ex_n_read) {
+			n_read = ex_n_read.value();
+		} else {
+			// There is no way to return an error from underflow(), generally
+			// the streams only care about how much data was read. No data or
+			// less data then requested by the caller of istream.read() means
+			// eofbit and failbit are set. If the user code wants to get the
+			// error or check if there was an error, it needs to check errno.
+			//
+			// So as long as we don't clear errno after a failure in the
+			// reader_.Read() above, error handling works as usual and returning
+			// eof below is all that needs to happen here.
+			//
+			// In case errno is not set for some reason, let's try to get it
+			// from the error with a fallback to a generic I/O error.
+			if (errno == 0) {
+				if (ex_n_read.error().code.category() == generic_category()) {
+					errno = ex_n_read.error().code.value();
+				} else {
+					errno = EIO;
+				}
+			}
+			n_read = 0;
+		}
+
+		streambuf::char_type *first = reinterpret_cast<streambuf::char_type *>(buf_.data());
+
+		// set eback, gptr, egptr
+		this->setg(first, first, first + n_read);
+	}
+
+	return this->gptr() == this->egptr() ? std::char_traits<char>::eof()
+										 : std::char_traits<char>::to_int_type(*this->gptr());
+};
+
+/**
+ * A variant of the #istream class that takes ownership of the #streambuf buffer
+ * created for it.
+ *
+ * @note Base #istream is designed to work on shared buffers so it doesn't
+ *       destruct/delete the buffer.
+ */
+class istreamWithUniqueBuffer : public istream {
+public:
+	// The unique_ptr, &&buf and std::move() model this really nicely -- a
+	// unique_ptr rvalue (i.e. temporary) is required and it's moved into the
+	// object. The default destructor then takes care of cleaning up properly.
+	istreamWithUniqueBuffer(unique_ptr<streambuf> &&buf) :
+		istream(buf.get()),
+		buf_ {std::move(buf)} {};
+
+private:
+	unique_ptr<streambuf> buf_;
+};
+
+unique_ptr<istream> Reader::GetStream() {
+	return unique_ptr<istream>(
+		new istreamWithUniqueBuffer(unique_ptr<ReaderStreamBuffer>(new ReaderStreamBuffer(*this))));
+};
 
 } // namespace io
 } // namespace common
--- a/common/io.cpp
+++ b/common/io.cpp
@@ -72,15 +72,6 @@ ExpectedSize ByteWriter::Write(
 }
 
 
-ExpectedSize StreamWriter::Write(
-	vector<uint8_t>::const_iterator start, vector<uint8_t>::const_iterator end) {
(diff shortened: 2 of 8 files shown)


common/json.hpp
```diff
@@ -107,9 +107,9 @@ class Json {
 using ExpectedJson = expected::expected<Json, error::Error>;
 
 ExpectedJson LoadFromFile(string file_path);
-ExpectedJson LoadFromString(string json_str);
-ExpectedJson LoadFromStream(istream &str);
-ExpectedJson LoadFromReader(io::Reader &reader);
+ExpectedJson Load(string json_str);
```
oleorhagen, line 110: If we did something like this:

```
diff --git a/common/json.hpp b/common/json.hpp
index 745ea21a..713c7a81 100644
--- a/common/json.hpp
+++ b/common/json.hpp
@@ -55,6 +55,18 @@ using ExpectedInt = mender::common::expected::ExpectedInt;
 using ExpectedBool = mender::common::expected::ExpectedBool;
 using ExpectedSize = mender::common::expected::ExpectedSize;
 
+struct FilePath {
+	string path;
+
+	FilePath(string path) :
+		path {path} {
+	}
+
+	operator string() {
+		return path;
+	}
+};
+
 class Json {
 public:
 	using ExpectedJson = expected::expected<Json, error::Error>;
@@ -91,7 +103,7 @@ public:
 
 	ExpectedSize GetArraySize() const;
 
-	friend ExpectedJson LoadFromFile(string file_path);
+	friend ExpectedJson Load(FilePath file_path);
 	friend ExpectedJson Load(string json_str);
 	friend ExpectedJson Load(istream &str);
 	friend ExpectedJson Load(io::Reader &reader);
@@ -106,7 +118,7 @@ private:
 
 using ExpectedJson = expected::expected<Json, error::Error>;
 
-ExpectedJson LoadFromFile(string file_path);
+ExpectedJson Load(FilePath file_path);
 ExpectedJson Load(string json_str);
 ExpectedJson Load(istream &str);
 ExpectedJson Load(io::Reader &reader);
diff --git a/common/json/platform/nlohmann/nlohmann_json.cpp b/common/json/platform/nlohmann/nlohmann_json.cpp
index d64d5062..46757afc 100644
--- a/common/json/platform/nlohmann/nlohmann_json.cpp
+++ b/common/json/platform/nlohmann/nlohmann_json.cpp
@@ -28,7 +28,7 @@ namespace io = mender::common::io;
 
 namespace mender::common::json {
 
-ExpectedJson LoadFromFile(string file_path) {
+ExpectedJson Load(FilePath file_path) {
 	ifstream f;
 	errno = 0;
 	f.open(file_path);
@@ -36,7 +36,7 @@ ExpectedJson LoadFromFile(string file_path) {
 		int io_errno = errno;
 		auto err = error::Error(
 			std::generic_category().default_error_condition(io_errno),
-			"Failed to open '" + file_path + "': " + strerror(io_errno));
+			"Failed to open '" + string(file_path) + "': " + strerror(io_errno));
 	}
 
@@ -46,7 +46,7 @@ ExpectedJson LoadFromFile(string file_path) {
 		return ExpectedJson(j);
 	} catch (njson::parse_error &e) {
 		auto err = MakeError(
-			JsonErrorCode::ParseError, "Failed to parse '" + file_path + "': " + e.what());
+			JsonErrorCode::ParseError, "Failed to parse '" + string(file_path) + "': " + e.what());
 	}
 }
diff --git a/common/json_test.cpp b/common/json_test.cpp
index 02f894a7..0b004b93 100644
--- a/common/json_test.cpp
+++ b/common/json_test.cpp
@@ -105,7 +105,7 @@ TEST_F(JsonFileTests, LoadFromValidFile) {
 	os << json_example_str;
 	os.close();
 
-	json::ExpectedJson ej = json::LoadFromFile(test_json_fname);
+	json::ExpectedJson ej = json::Load(json::FilePath {test_json_fname});
 	ASSERT_TRUE(ej);
 	EXPECT_FALSE(ej.value().IsNull());
 }
@@ -115,7 +115,7 @@ TEST_F(JsonFileTests, LoadFromInvalidFile) {
 	os << "{ invalid: json";
 	os.close();
 
-	json::ExpectedJson ej = json::LoadFromFile(test_json_fname);
+	json::ExpectedJson ej = json::Load(json::FilePath(test_json_fname));
 	ASSERT_FALSE(ej);
 	EXPECT_EQ(ej.error().code, json::MakeError(json::JsonErrorCode::ParseError, "").code);
 	EXPECT_THAT(
@@ -123,7 +123,7 @@ TEST_F(JsonFileTests, LoadFromInvalidFile) {
 }
 
 TEST_F(JsonFileTests, LoadFromNonexistingFile) {
-	json::ExpectedJson ej = json::LoadFromFile("non-existing-file");
+	json::ExpectedJson ej = json::Load(json::FilePath {"non-existing-file"});
 	ASSERT_FALSE(ej);
 	EXPECT_TRUE(ej.error().IsErrno(ENOENT));
 	EXPECT_THAT(

```

We could get `FromFile` in the same pattern also

artifact/v3/version/version.cpp
```diff
@@ -63,17 +63,7 @@ error::Error MakeError(ErrorCode code, const string &msg) {
...
-	io::ByteWriter bw {buf};
-
-	auto err = io::Copy(bw, reader);
-	if (err) {
-		return expected::unexpected(err);
-	}
-
-	auto expected_json = json::Load(mender::common::StringFromByteVector(buf));
-
+	auto expected_json = json::Load(reader);
```
oleorhagen, line 66: Nice!

common/io.cpp
```diff
@@ -71,15 +71,6 @@ ExpectedSize ByteWriter::Write(
 }
 
 
-ExpectedSize StreamWriter::Write(
```
oleorhagen, line 67: Q:

Shouldn't we extend this instead, to be the `ostream` part of what you have jsut done with `istream`, so that writers also can map to streams?

common/json.hpp
vpodzime: Sure. But then there is `std::filesystem::path` which has many extra features and benefits on top of the above. I don't think having `LoadFromFile()` is a problem.

common/io.cpp
vpodzime: If it turns out we need it then yes. Right now I'm not aware of any place where we would need an `ostream` for a given `Writer`. In fact, we don't use any real (not just shoveling data between buffers) `Writer` at all, AFAICT.

common/io.cpp
```diff
@@ -73,6 +80,79 @@ ExpectedSize StreamWriter::Write(
...
+/**
+ * A variant of the #istream class that takes ownership of the #streambuf buffer
+ * created for it.
+ *
+ * @note Base #istream is designed to work on shared buffers so it doesn't
+ *       destruct/delete the buffer.
+ */
+class istreamWithUniqueBuffer : public istream {
+public:
+	// The unique_ptr, &&buf and std::move() model this really nicely -- a
```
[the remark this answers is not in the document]
vpodzime: The `istream` constructor requires a pointer a to a `streambuf` and passing a `streambuf` around as value didn't work (I believe I tried). Anyway, as the comment says, I think it really fits nicely and models what is happening.

common/io.cpp
```diff
@@ -73,6 +80,79 @@ ExpectedSize StreamWriter::Write(
...
 
+class ReaderStreamBuffer : public streambuf {
+public:
+	ReaderStreamBuffer(Reader &reader) :
+		reader_ {reader},
+		buf_(buf_size_) {};
+	streambuf::int_type underflow() override;
+
+private:
+	static const Vsize buf_size_ = 1024;
```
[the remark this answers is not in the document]
vpodzime: Good catch, I wanted this to be a named constant and then forgot to do it.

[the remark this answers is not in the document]
vpodzime: No particular reason, just a reasonable starting value that we can tweak later.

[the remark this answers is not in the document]
oleorhagen: Maybe we should even consider making this configurable at build time? :thinking:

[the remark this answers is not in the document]
vpodzime: Yes, globally for all I/O buffers we use. But that's for a different PR.

common/io.cpp
```diff
@@ -73,6 +80,79 @@ ExpectedSize StreamWriter::Write(
...
+		}
+
+		streambuf::char_type *first = reinterpret_cast<streambuf::char_type *>(buf_.data());
+
+		// set eback, gptr, egptr
+		this->setg(first, first, first + n_read);
+	}
+
+	return this->gptr() == this->egptr() ? std::char_traits<char>::eof()
+										 : std::char_traits<char>::to_int_type(*this->gptr());
```
kacf, line 119: `clang-format` agrees with this..? :neutral_face:

common/json/platform/nlohmann/nlohmann_json.cpp
```diff
@@ -59,6 +63,23 @@ ExpectedJson LoadFromString(string json_str) {
...
+	} catch (njson::parse_error &e) {
+		auto err = MakeError(
+			JsonErrorCode::ParseError, string("Failed to parse JSON from stream: ") + e.what());
+		return expected::unexpected(err);
+	}
+}
+
+ExpectedJson LoadFromReader(io::Reader &reader) {
+	auto str_ptr = reader.GetStream();
+	return LoadFromStream(*(str_ptr.get()));
```
kacf: Really nice to use! 👍

common/io.cpp
kacf: I'm using `Writer` quite a bit in the HTTP lib. But not `StreamWriter`, indeed.

kacf (approved):
Pre-approving, but a possible improvement mentioned below.

common/io.cpp
```diff
@@ -73,6 +80,79 @@ ExpectedSize StreamWriter::Write(
...
+			// There is no way to return an error from underflow(), generally
+			// the streams only care about how much data was read. No data or
+			// less data then requested by the caller of istream.read() means
+			// eofbit and failbit are set. If the user code wants to get the
+			// error or check if there was an error, it needs to check errno.
+			//
+			// So as long as we don't clear errno after a failure in the
+			// reader_.Read() above, error handling works as usual and returning
+			// eof below is all that needs to happen here.
+			n_read = 0;
```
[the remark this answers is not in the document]
kacf: Err, I meant `generic_category`. So confusing, those categories...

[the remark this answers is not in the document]
vpodzime: I was actually thinking the same. Let's do it then!

common/io.cpp
vpodzime: You think it was me formatting the code like this?!

common/io.cpp
vpodzime: But also just shoveling data between buffers, right? :wink:

common/json.hpp
oleorhagen: Why not use `std::filesystem::path` then? :

vpodzime: Because it's C++17.

common/io.cpp
kacf: We don't know. In the tests it does, but the `Writer` will likely be a pipe to mender-flash in the future.

common/io.cpp
```diff
@@ -64,14 +71,89 @@ ExpectedSize ByteWriter::Write(
...
+			// eofbit and failbit are set. If the user code wants to get the
+			// error or check if there was an error, it needs to check errno.
+			//
+			// So as long as we don't clear errno after a failure in the
+			// reader_.Read() above, error handling works as usual and returning
+			// eof below is all that needs to happen here.
+			//
+			// In case errno is not set for some reason, let's try to get it
+			// from the error with a fallback to a generic I/O error.
+			if (errno == 0) {
```
kacf, line 113: If you test for zero, you should perhaps set it to zero as well, before the call to `Read`. Otherwise it may be that it was non-zero already, but `Read` doesn't use it, and then you end up returning an `errno` which is for an unrelated error. I know that some libc calls explicitly encourage you to set it to zero before calling them.

vpodzime: Good point. Although I believe we should (also) set it to zero in our code where we use the syscalls that can set it.

vpodzime: OTOH, one would think things like `istream.read()` do `errno = 0` before the call the `read()` syscall.

PR #1196, opened 2023-03-21
fix(io): Missing unexpected wrapper around error call

oleorhagen:
Ticket: None
Changelog: None

-------


vpodzime (approved):
This is scary. We should look at why and how this compiles without a single warning.

PR #1199, opened 2023-03-27
feat(vendor): Add an optional library

oleorhagen:
This adds an optional library to vendor. The library is fully compatible with C++17's optional implementation, when built under C++11, which should suit us well.

Added a simple example to the auth binary, to examplify it's use.

Ticket: None
Changelog: None

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -10,3 +10,6 @@
 [submodule "vendor/expected"]
 	path = vendor/expected
 	url = https://github.com/TartanLlama/expected.git
+[submodule "vendor/optional-lite"]
+	path = vendor/optional-lite
+	url = https://github.com/martinmoene/optional-lite.git
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -93,6 +93,9 @@ endif()
 add_subdirectory(vendor/expected)
 include_directories(${CMAKE_SOURCE_DIR}/vendor/expected/include)
 
+add_subdirectory(vendor/optional-lite)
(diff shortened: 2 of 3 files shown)


mender-auth/main.cpp
```diff
@@ -45,5 +57,14 @@ int main(int argc, char *argv[]) {
...
 	}
 
+	char const *int_text = argc > 1 ? argv[1] : "42";
+
+	optional<int> oi = to_int(int_text);
+	if (oi) {
+		std::cout << "'" << oi.value() << "' is " << *oi;
+	} else {
+		std::cout << "'" << int_text << "' isn't a number";
+	}
```
vpodzime, line 67: Please don't pollute this code. :wink:

vpodzime (commented):
~it's~ its (in the commit message)

Where do you/we need this?

mender-auth/main.cpp
oleorhagen: Sure, just wanted to leave an example of how to use it, and that it works.

I'll remove

oleorhagen:
> ~it's~ its (in the commit message)
> 
> Where do you/we need this?

Need it for the Artifact header parser

vpodzime:
> Need it for the Artifact header parser
:+1: just keep in mind that in the [*File API*](https://github.com/mendersoftware/mender/blob/master/Documentation/update-modules-v3-file-api.md#file-api), all the specified files are guaranteed to exist so those should be just empty strings if not present in the artifact, not `optional<string>`.

kacf (approved):

PR #1200, opened 2023-03-27
feat: Basic DB keys and vacuuming

vpodzime:
Adding basic set of DB keys as constants into the MenderContext class and cleanup of unused keys from the DB as part of MenderContext::Initialize().

Ticket: MEN-6308
Changelog: none


kacf (approved):
👍

PR #1203, opened 2023-03-29
Header parser impl

oleorhagen:
Implementation of the header-parser, minus the meta-data parsing, which will be handled as a follow-up in https://northerntech.atlassian.net/browse/MEN-6424.

For now it's directly importing the `vendor/expected` lib, but I will move this out into our own wrapper in a separate PR in parallel, and then rebase on top of it.

### DIFF
--- a/artifact/parser.cpp
+++ b/artifact/parser.cpp
@@ -26,19 +26,7 @@
 
 #include <artifact/lexer.hpp>
 #include <artifact/tar/tar.hpp>
-
-namespace mender {
-namespace artifact {
-namespace v3 {
-namespace header {
-Header Parse() {
-	return Header {};
-}
-} // namespace header
-} // namespace v3
-} // namespace artifact
-} // namespace mender
-
+#include <artifact/sha/sha.hpp>
 
 namespace mender {
 namespace artifact {
@@ -57,7 +45,7 @@ namespace version = mender::artifact::v3::version;
 namespace manifest = mender::artifact::v3::manifest;
 namespace payload = mender::artifact::v3::payload;
 
-ExpectedArtifact Parse(io::Reader &reader) {
+ExpectedArtifact Parse(io::Reader &reader, config::ParserConfig config) {
 	std::shared_ptr<tar::Reader> tar_reader {make_shared<tar::Reader>(reader)};
 
 	auto lexer = lexer::Lexer<token::Token, token::Type> {tar_reader};
@@ -108,17 +96,24 @@ ExpectedArtifact Parse(io::Reader &reader) {
 			parser_error::Code::ParseError,
 			"Got unexpected token " + tok.TypeToString() + " expected 'Header'"));
 	}
-	// auto header = v3::header::Parse(); // TODO (MEN-6178): Enable
+	sha::Reader shasum_reader {*tok.value, manifest.Get("header.tar")};
+	auto expected_header = v3::header::Parse(shasum_reader, config);
+	if (!expected_header) {
+		return expected::unexpected(parser_error::MakeError(
+			parser_error::Code::ParseError,
+			"Failed to parse the header: " + expected_header.error().message));
+	}
+	auto header = expected_header.value();
 
 	log::Trace("Parsing the payload");
 	tok = lexer.Next();
 	if (tok.type != token::Type::Payload) {
 		return expected::unexpected(parser_error::MakeError(
 			parser_error::Code::ParseError,
-			"Got unexpected token " + tok.TypeToString() + "expected 'data/0000.tar"));
+			"Got unexpected token " + tok.TypeToString() + " expected 'data/0000.tar"));
 	}
 
-	return Artifact {version, manifest, lexer}; // TODO (MEN-6178): Add the header
+	return Artifact {version, manifest, header, lexer};
 };
 
 
--- a/artifact/parser.cpp
+++ b/artifact/parser.cpp
@@ -60,7 +60,7 @@ namespace payload = mender::artifact::v3::payload;
 ExpectedArtifact Parse(io::Reader &reader) {
 	std::shared_ptr<tar::Reader> tar_reader {make_shared<tar::Reader>(reader)};
 
-	lexer::Lexer lexer = lexer::Lexer {tar_reader};
+	auto lexer = lexer::Lexer<token::Token, token::Type> {tar_reader};
 
 	token::Token tok = lexer.Next();
 
--- a/artifact/parser_test.cpp
+++ b/artifact/parser_test.cpp
@@ -18,6 +18,7 @@
 #include <gtest/gtest.h>
 #include <gmock/gmock.h>
 
+#include <common/log.hpp>
 #include <common/processes.hpp>
 
 #include <common/testing.hpp>
@@ -39,6 +40,8 @@ class ParserTestEnv : public testing::Test {
 public:
 protected:
 	static void SetUpTestSuite() {
+		mender::common::log::SetLevel(mender::common::log::LogLevel::Trace);
+
 		string script = R"(#! /bin/sh
 
     DIRNAME=$(dirname $0)
--- a/artifact/parser_test.cpp
+++ b/artifact/parser_test.cpp
@@ -134,5 +134,5 @@ TEST(ParserTest, TestParseMumboJumbo) {
 	auto artifact = mender::artifact::parser::Parse(sr);
 
 	ASSERT_FALSE(artifact) << artifact.error().message << std::endl;
-	ASSERT_EQ(artifact.error().message, "Got unexpected token : 'Unrecognized' expected 'version'");
+	ASSERT_EQ(artifact.error().message, "Got unexpected token : 'EOF' expected 'version'");
 }
--- a/artifact/parser_test.cpp
+++ b/artifact/parser_test.cpp
@@ -15,15 +15,15 @@
 #include <artifact/parser.hpp>
 #include <artifact/lexer.hpp>
 
+#include <string>
+#include <fstream>
+
 #include <gtest/gtest.h>
 #include <gmock/gmock.h>
 
 #include <common/log.hpp>
 #include <common/processes.hpp>
-
 #include <common/testing.hpp>
-
-#include <fstream>
 
 
 using namespace std;
--- a/artifact/token.hpp
+++ b/artifact/token.hpp
@@ -39,6 +39,7 @@ namespace expected = mender::common::expected;
 
 enum class Type {
 	Uninitialized = 0,
+	EOFToken,
 	Unrecognized,
 	Version,
 	Manifest,
@@ -51,6 +52,7 @@ enum class Type {
 
 const unordered_map<const Type, const string> type_map {
 	{Type::Uninitialized, "Uninitialized"},
+	{Type::EOFToken, "EOF"},
 	{Type::Unrecognized, "Unrecognized"},
 	{Type::Version, "version"},
 	{Type::Manifest, "manifest"},
--- a/artifact/token.hpp
+++ b/artifact/token.hpp
@@ -17,6 +17,7 @@
 
 #include <unordered_map>
 #include <string>
+#include <regex>
 
 #include <common/log.hpp>
 #include <common/expected.hpp>
@@ -24,6 +25,9 @@
 namespace mender {
 namespace artifact {
 namespace token {
+
+const std::regex payload_regexp {
+	"data/[0-9]{4}\\.tar(\\.(gz|xz|zst))?", std::regex_constants::ECMAScript};
 
 using namespace std;
 
@@ -90,7 +94,7 @@ private:
 		if (type_name.find("header.tar") == 0) {
 			return Type::Header;
 		}
-		if (type_name.find("data/0000.tar") == 0) {
+		if (regex_match(type_name, payload_regexp)) {
 			return Type::Payload;
 		}
 		log::Error("Unrecognized token: " + type_name);
--- a/artifact/v3/header/header.cpp
+++ b/artifact/v3/header/header.cpp
@@ -1,0 +1,191 @@
+// Copyright 2023 Northern.tech AS
+//
+//    Licensed under the Apache License, Version 2.0 (the "License");
+//    you may not use this file except in compliance with the License.
+//    You may obtain a copy of the License at
+//
+//        http://www.apache.org/licenses/LICENSE-2.0
+//
+//    Unless required by applicable law or agreed to in writing, software
+//    distributed under the License is distributed on an "AS IS" BASIS,
+//    WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+//    See the License for the specific language governing permissions and
+//    limitations under the License.
+
+#include <artifact/v3/header/header.hpp>
+
+#include <iomanip>
+#include <memory>
+#include <string>
+#include <system_error>
+#include <vector>
+#include <iostream>
+#include <fstream>
+
+#include <common/expected.hpp>
+#include <common/error.hpp>
+#include <common/io.hpp>
+#include <common/log.hpp>
+#include <common/json.hpp>
+#include <common/common.hpp>
+#include <common/conf/paths.hpp>
+
+#include <artifact/error.hpp>
+#include <artifact/lexer.hpp>
+#include <artifact/tar/tar.hpp>
+
+#include <artifact/v3/header/token.hpp>
+
+namespace mender {
+namespace artifact {
+namespace v3 {
+namespace header {
+
+using namespace std;
+
+namespace expected = mender::common::expected;
+namespace io = mender::common::io;
+namespace error = mender::common::error;
+namespace log = mender::common::log;
+namespace json = mender::common::json;
+
+
+namespace {
+string IndexString(int index) {
+	stringstream index_string {};
+	index_string << setw(4) << setfill('0') << index;
+	return index_string.str();
+}
+} // namespace
+
+ExpectedHeader Parse(io::Reader &reader, ParserConfig conf) {
+	Header header {};
+
+	shared_ptr<tar::Reader> tar_reader {make_shared<tar::Reader>(reader)};
+
+	auto lexer = lexer::Lexer<header::token::Token, header::token::Type> {tar_reader};
+
+	token::Token tok = lexer.Next();
+
+	if (tok.type != token::Type::HeaderInfo) {
+		return expected::unexpected(parser_error::MakeError(
+			parser_error::Code::ParseError,
+			"Got unexpected token: '" + tok.TypeToString() + "' expected 'header-info'"));
+	}
+
+	auto expected_info = header::info::Parse(*tok.value);
+
+	if (!expected_info) {
+		return expected::unexpected(parser_error::MakeError(
+			parser_error::Code::ParseError,
+			"Failed to parse the header-info: " + expected_info.error().message));
+	}
+	header.info = expected_info.value();
+
+	tok = lexer.Next();
+	vector<ArtifactScript> state_scripts {};
+	while (tok.type == token::Type::ArtifactScripts) {
+		log::Trace("Parsing state script...");
+		const string artifact_script_path =
+			mender::common::conf::paths::Join(conf.artifact_scripts_filesystem_path, tok.name);
+		errno = 0;
+		ofstream myfile(artifact_script_path);
+		log::Trace("state script name: " + tok.name);
+		if (!myfile.good()) {
+			auto io_errno = errno;
+			return expected::unexpected(error::Error(
+				std::generic_category().default_error_condition(io_errno),
+				"Failed to create a file for writing the Artifact script: " + artifact_script_path
+					+ " to the filesystem"));
+		}
+		io::StreamWriter sw {myfile};
+
+		auto err = io::Copy(sw, *tok.value);
+		if (err != error::NoError) {
+			return expected::unexpected(err);
+		}
+
+		state_scripts.push_back(artifact_script_path);
+
+		tok = lexer.Next();
+	}
+
+	header.artifactScripts = move(state_scripts);
+
+	vector<SubHeader> subheaders {};
+
+	int current_index {0};
+	while (tok.type != token::Type::EOFToken) {
+		log::Trace("Parsing the sub-header ...");
+
+		// NOTE: We currently do not support multiple payloads
+		if (current_index != 0) {
+			return expected::unexpected(parser_error::MakeError(
+				parser_error::Code::ParseError,
+				"Multiple header entries found. Currently only one is supported"));
+		}
+
+		SubHeader sub_header {};
+		if (tok.type != token::Type::ArtifactHeaderTypeInfo) {
+			return expected::unexpected(parser_error::MakeError(
+				parser_error::Code::ParseError,
+				"Unexpected entry: " + tok.TypeToString() + " expected: type-info"));
+		}
+
+		if (current_index != tok.Index()) {
+			return expected::unexpected(parser_error::MakeError(
+				parser_error::Code::ParseError,
+				"Unexpected index order for the type-info: " + tok.name + " expected: headers/"
+					+ IndexString(current_index) + "/type-info"));
+		}
+		auto expected_type_info = type_info::Parse(*tok.value);
+		if (!expected_type_info) {
+			return expected::unexpected(expected_type_info.error());
+		}
+		sub_header.type_info = expected_type_info.value();
+
+		// NOTE (workaround): Bug in the Artifact format writer:
+		// If the type is a RootfsImage, then the payload-type will be empty. This
+		// is a bug in the mender-artifact tool, which writes the payload. For now,
+		// just work around it.
+		if (header.info.payloads[current_index].type == Payload::RootfsImage) {
+			log::Debug(
+				"Setting the type-info in payload nr " + to_string(current_index)
+				+ " to rootfs-image");
+			sub_header.type_info.type = "rootfs-image";
+		}
+
+		tok = lexer.Next();
+
+		log::Trace("sub-header: looking for meta-data");
+
+		// meta-data (optional)
+		if (tok.type == token::Type::ArtifactHeaderMetaData) {
+			if (current_index != tok.Index()) {
+				return expected::unexpected(parser_error::MakeError(
+					parser_error::Code::ParseError,
+					"Unexpected index order for the meta-data: " + tok.name + " expected: headers/"
+						+ IndexString(current_index) + "/meta-data"));
+			}
+			// auto expected_meta_data = meta_data::Parse(*tok.value);
+			// if (!expected_meta_data) {
+			// TODO - follow-up (MEN-6424)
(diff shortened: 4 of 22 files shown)


artifact/v3/header/header.hpp
```diff
@@ -0,0 +1,178 @@
...
+
+namespace mender {
+namespace artifact {
+namespace v3 {
+namespace header {
+
+using namespace std;
+
+using nonstd::nullopt;
+using nonstd::optional;
```
kacf, line 37: We should do like we did for `expected`, and relabel it `mender::common::optional` instead.

artifact/v3/header/header.hpp
```diff
@@ -0,0 +1,178 @@
...
+namespace expected = mender::common::expected;
+namespace io = mender::common::io;
+namespace error = mender::common::error;
+
+//
+// +---header-info
+//
+
+enum class Payload {
+	RootfsImage,
```
kacf, line 48: Do we actually need `RootfsImage`?

In the new implementation there shouldn't be any difference between `ModuleImage` and `RootfsImage`, because the details are handled by the Update Module.

artifact/v3/header/header.hpp
```diff
@@ -0,0 +1,178 @@
...
+	EmptyPayload,
+};
+
+using PayloadType = pair<Payload, string>;
+
+// {'artifact_name', 'artifact_group'}
+struct Provides {
+	string artifact_name;
+	optional<string> artifact_group;
+};
```
kacf, line 59: And a map as well, or?

artifact/v3/header/token.hpp
```diff
@@ -0,0 +1,111 @@
...
+
+#include <unordered_map>
+#include <string>
+#include <regex>
+
+#include <common/log.hpp>
+#include <common/expected.hpp>
+
+const std::regex artifact_script_regexp {
+	"scripts/Artifact(Install|Reboot|Rollback|RollbackReboot|Failure)_(Enter|Leave)_[0-9][0-9](_[^[:space:]]+)?",
```
kacf, line 27: Should use `(_\S+)?` at the end instead (this is what [the original](https://github.com/mendersoftware/mender-artifact/blob/master/artifact/scripter.go#L50) uses).

artifact/v3/header/token.hpp
```diff
@@ -0,0 +1,111 @@
...
+			return Type::ArtifactScripts;
+		}
+		if (std::regex_match(type_name, regex("headers/[0-9]{4}/type-info"))) {
+			return Type::ArtifactHeaderTypeInfo;
+		}
+		if (regex_match(type_name, regex("headers/[0-9]{4}/meta-data"))) {
+			return Type::ArtifactHeaderMetaData;
+		}
+		log::Error("Unrecognized token: " + type_name);
+		return Type::Unrecognized;
```
kacf, line 102: This might not be the right place to handle it (and maybe you already have), but I'll just point it out: I remember that I fixed a bug in mender-artifact a long time ago where it couldn't read an artifact which was artificially put together using `tar`. While this is of course not common, it has its debugging uses and occasionally even in production. Notice the difference between them:

File list from mender-artifact:

```
headers/0000/type-info
```

File list from `tar`:

```
headers
headers/0000
headers/0000/type-info
```

In other words it contains the directories leading up to the file as well. This should be filtered out by looking at file type.

common/json.hpp
```diff
@@ -121,6 +121,16 @@ ExpectedJson Load(string json_str);
 ExpectedJson Load(istream &str);
 ExpectedJson Load(io::Reader &reader);
 
+
+using ExpectedStringVector = expected::ExpectedStringVector;
+using ExpectedKeyValuesMap = expected::expected<unordered_map<string, string>, error::Error>;
+
+// TODO - Templatize
```
kacf, line 128: AFAIK you cannot template on return type alone, so this may not be possible.

kacf (commented):
> Obviously not completely clean and ready for final review, but I would appreciate a view through still.

Based on this comment I focused on reviewing the headers. The interface looks good to me, I couldn't spot any major issues.

artifact/v3/header/header.hpp
oleorhagen: Very good point!

artifact/v3/header/token.hpp
oleorhagen: It's the same thing, but sure.

kacf: Oh is it? I thought it was only actual space. Then either is fine.

artifact/v3/header/header.hpp
oleorhagen: I decided to keep it in here, due to the fact that we need to workaround that `type` is not written in `type-info:payload_type`

artifact/v3/header/header.hpp
oleorhagen: I'll create a separate PR for this one, and subsequently rebase on top of it.

artifact/v3/header/artifact_scripts.cpp
```diff
@@ -0,0 +1,44 @@
...
+namespace mender {
+namespace artifact {
+namespace v3 {
+namespace header {
+namespace artifact_scripts {
+
+ExpectedArtifactScript Parse(io::Reader &reader) {
+	log::Info("TODO - Parse the scripts names - and verify");
+	// Read the whole script to a string, and store it in memory
+	vector<uint8_t> buf(4096); // TODO - Use a stream, and variable size
```
kacf: I take it this means that the file storing is not part of this PR?

common/json/json.cpp
```diff
@@ -45,6 +48,56 @@ error::Error MakeError(JsonErrorCode code, const string &msg) {
...
+		return expected::unexpected(
+			MakeError(JsonErrorCode::ParseError, "The JSON object is not an array"));
+	}
+	vector<string> vector_elements {};
+	size_t vector_size {j.GetArraySize().value()};
+	for (size_t i = 0; i < vector_size; ++i) {
+		auto element = j.Get(i).and_then([](const json::Json &j) { return j.GetString(); });
+		if (!element) {
+			return expected::unexpected(MakeError(
+				JsonErrorCode::ParseError, "Unexpected element type in the string array"));
```
kacf, line 63: These aren't really `ParseError`s though, are they? More like `WrongFormatError` or something like that.

Goes for all of the errors below.

kacf (commented):
Posting what I have, I will have to continue on Monday.

Note to self: Continue at `while (tok.type != token::Type::EOFToken)`.

artifact/v3/header/header.cpp
```diff
@@ -0,0 +1,156 @@
...
+		if (!myfile.good()) {
+			// TODO - Return what sort of error (?)
+			return expected::unexpected(parser_error::MakeError(
+				parser_error::Code::ParseError,
+				"Failed to create a file for writing the Artifact script: " + artifact_script_path
+					+ " to the filesystem"));
+		}
+		io::StreamWriter bw {myfile};
+
+		auto err = io::Copy(bw, *tok.value);
```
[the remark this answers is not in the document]
kacf: I guess you already answered this f2f.

artifact/v3/header/artifact_scripts.cpp
oleorhagen: Nope, removed.

artifact/v3/header/header.cpp
```diff
@@ -0,0 +1,155 @@
...
+			"Failed to parse the header-info: " + expected_info.error().message));
+	}
+	header.info = expected_info.value();
+
+	tok = lexer.Next();
+	vector<ArtifactScript> state_scripts {};
+	while (tok.type == token::Type::ArtifactScripts) {
+		log::Trace("Parsing state script...");
+		const string artifact_script_path =
+			conf.artifact_scripts_filesystem_path + "/" + tok.TypeToString();
```
vpodzime, line 78: Should use `paths::Join()`.

artifact/v3/header/header.cpp
```diff
@@ -0,0 +1,156 @@
...
+		ofstream myfile(artifact_script_path);
+		log::Trace("state script name: " + tok.TypeToString());
+		if (!myfile.good()) {
+			// TODO - Return what sort of error (?)
```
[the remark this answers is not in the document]
vpodzime: I did [this](https://github.com/mendersoftware/mender/blob/feature-c%2B%2B-client/common/log/platform/boost/boost_log.cpp#L200) in the logging code. But if you don't want to use a custom error code, just use the `generic_category` and the `errno`.

artifact/v3/header/header.cpp
```diff
@@ -0,0 +1,155 @@
...
+			return expected::unexpected(parser_error::MakeError(
+				parser_error::Code::ParseError,
+				"Failed to create a file for writing the Artifact script: " + artifact_script_path
+					+ " to the filesystem"));
+		}
+		io::StreamWriter bw {myfile};
```
vpodzime, line 88: `bw` is a weird name for this variable. Copy-pasta?

artifact/v3/header/header.cpp
```diff
@@ -0,0 +1,155 @@
...
+			// TODO - follow-up (MEN-6204)
+			// return expected::unexpected(expected_meta_data.error());
+			// sub_header.metadata = expected_meta_data.value();
+			// }
+		}
+		log::Trace("sub-header: parsed the meta-data");
+
+		header.subHeaders.push_back(sub_header);
+
+		tok = lexer.Next();
```
vpodzime, line 145: Shouldn't this `Next()` be inside the above `if`? Otherwise this will skip a token if it's not `ArtifactHeaderMetaData`, right?

artifact/v3/header/header.hpp
```diff
@@ -0,0 +1,175 @@
...
+
+struct Provides {
+	string artifact_name;
+	optional<string> artifact_group;
+};
+
+struct Depends {
+	vector<string> device_type;
+	optional<vector<string>> artifact_name;
+	optional<vector<string>> artifact_group;
```
vpodzime, line 65: Shouldn't this be a vector of string pairs?

artifact/v3/header/header.hpp
```diff
@@ -0,0 +1,175 @@
...
+// +---header-info
+//
+
+enum class Payload {
+	RootfsImage,
+	ModuleImage,
+	EmptyPayload,
+};
+
+using PayloadType = pair<Payload, string>;
```
vpodzime, line 55: Would a `struct PayloadType` make sense here? What is the string for? I think it would be nice to give the two parts names.

artifact/v3/header/header_info.cpp
```diff
@@ -0,0 +1,195 @@
...
+
+using ExpectedPayloadType = expected::expected<vector<PayloadType>, error::Error>;
+
+ExpectedPayloadType ToPayloadTypes(const json::Json &j) {
+	if (!j.IsArray()) {
+		return expected::unexpected(parser_error::MakeError(
+			parser_error::Code::ParseError, "The JSON object is not an array"));
+	}
+	vector<PayloadType> vector_elements {};
+	size_t vector_size {j.GetArraySize().value()};
```
vpodzime, line 47: We could initialize the `vector_elements` to the expected size since we know it.

artifact/v3/header/header_info.cpp
```diff
@@ -0,0 +1,195 @@
...
+	for (size_t i = 0; i < vector_size; ++i) {
+		auto expected_element =
+			j.Get(i).and_then([](const json::Json &j) { return j.Get("type"); });
+		if (!expected_element) {
+			return expected::unexpected(parser_error::MakeError(
+				parser_error::Code::ParseError,
+				"Failed to get the type from the payload" + expected_element.error().message));
```
vpodzime, line 54: Missing colon and a space in the message.

artifact/v3/header/header.hpp
vpodzime: Also, do we even need the string? Isn't it just a string of the type?

artifact/v3/header/header_info.cpp
```diff
@@ -0,0 +1,195 @@
...
+
+	auto expected_json = json::Load(reader);
+
+	if (!expected_json) {
+		return expected::unexpected(parser_error::MakeError(
+			parser_error::Code::ParseError,
+			"Failed to parse the header JSON: " + expected_json.error().message));
+	}
+
+	const json::Json header_info_json = expected_json.value();
```
vpodzime, line 91: Why not `auto`?

artifact/v3/header/header_info.cpp
```diff
@@ -0,0 +1,195 @@
...
+	if (!artifact_name) {
+		return expected::unexpected(
+			parser_error::MakeError(parser_error::Code::ParseError, artifact_name.error().message));
+	}
+	info.provides.artifact_name = artifact_name.value();
+
+	// provides:artifact_group (optional)
+	log::Trace("Parsing the header-info provides:artifact_group (if any)");
+	auto artifact_group = provides.value().Get("artifact_group").and_then(json::ToString);
+	if (!artifact_group && artifact_group.error().code.value() != json::KeyError) {
```
vpodzime, line 131: I'd do `Get("artifact_group")` and then `GetString()` if it returns something, but YMMV.

artifact/v3/header/header.hpp
oleorhagen: @kacf actually the `nonstd::optional` wrapper already defaults to `optional` if it exists in `std`.

Should we still wrap it you think?

artifact/v3/header/token.hpp
```diff
@@ -0,0 +1,112 @@
...
+	Uninitialized = 0,
+	EOFToken,
+	Unrecognized,
+	HeaderInfo,
+	ArtifactScripts,
+	ArtifactHeaderTypeInfo,
+	ArtifactHeaderMetaData,
+};
+
+unordered_map<const Type, string> type_map {
```
vpodzime, line 54: Could be an array, no need for a hash map.

artifact/v3/header/token.hpp
```diff
@@ -0,0 +1,112 @@
...
+		value {make_shared<tar::Entry>(entry)} {
+	}
+
+	const string TypeToString() const {
+		return type_map.at(type);
+	}
+
+private:
+	Type StringToType(const string &type_name) {
+		if (type_name.find("header-info") == 0) {
```
vpodzime, line 87: Shouldn't this be an exact match of the whole string?

vpodzime (commented):
Looks good to me otherwise. Smart approach to extracting real data from the JSON elements!

common/json/json.cpp
oleorhagen: That makes it easy then! :tada:

artifact/v3/header/header_test.cpp
```diff
@@ -0,0 +1,368 @@
...
+	CreateTestArtifact(tmpdir, "module-image", {"--type test-module-image"});
+
+	std::fstream fs {tmpdir.Path() + "/header.tar"};
+
+	mender::common::io::StreamReader sr {fs};
+
+	ExpectedHeader expected_header = header::Parse(sr);
+
+	ASSERT_TRUE(expected_header) << expected_header.error().message;
+}
```
kacf, line 368: Perhaps add test cases for:

* Some files out of order.
* Wrong index of header or payload.

artifact/token.hpp
```diff
@@ -47,6 +52,7 @@ enum class Type {
 
 const unordered_map<const Type, const string> type_map {
 	{Type::Uninitialized, "Uninitialized"},
+	{Type::EOFToken, "EOF"},
```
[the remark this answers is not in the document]
oleorhagen: Ur right- fixing

artifact/parser.cpp
```diff
@@ -108,17 +95,23 @@ ExpectedArtifact Parse(io::Reader &reader) {
...
 			"Got unexpected token " + tok.TypeToString() + " expected 'Header'"));
 	}
-	// auto header = v3::header::Parse(); // TODO (MEN-6178): Enable
+	auto expected_header = v3::header::Parse(*tok.value, config);
+	if (!expected_header) {
+		return expected::unexpected(parser_error::MakeError(
+			parser_error::Code::ParseError,
+			"Failed to parse the header: " + expected_header.error().message));
+	}
+	auto header = expected_header.value();
```
kacf, line 104: Are we missing checksumming? (Or am I just missing where it is..?)

artifact/v3/header/header.hpp
kacf: Hmm, in the name of consistency I think I would vote for that. Just to keep "foreign" names out of our cross platform code.

artifact/v3/header/header.cpp
oleorhagen: Absolutely right!!!

Good :eyes:

artifact/v3/header/header.hpp
oleorhagen: Hmm, maybe, but the name is known, so doing it like this the structure of extraction will be:

`header.info.depends.artifact_group -> vector<string>` 

as opposed to:

`header.info.depends["artifact_group"] -> vector<string>`

Where I chose the former, as it's type-checked, and hence typo safe.

artifact/v3/header/header.hpp
oleorhagen: > Isn't it just a string of the type?

No, in fact this needs a value. This is because you can write your own update-modules, and hence the string is free-form.

So we provide the `Type`, (ie, Update-Module), and you as the end user provides the `Update-Module:Typename`.

Does it make sense?

oleorhagen: We could consider a struct however though.

artifact/v3/header/header_info.cpp
oleorhagen: Yeah, could've. I think they both make sense here, so staying with the current impl

artifact/v3/header/token.hpp
oleorhagen: Yup - good find! fixing

artifact/v3/header/header.cpp
```diff
@@ -0,0 +1,155 @@
...
+			// auto expected_meta_data = meta_data::Parse(*tok.value);
+			// if (!expected_meta_data) {
+			// TODO - follow-up (MEN-6204)
+			// return expected::unexpected(expected_meta_data.error());
+			// sub_header.metadata = expected_meta_data.value();
+			// }
+		}
+		log::Trace("sub-header: parsed the meta-data");
+
+		header.subHeaders.push_back(sub_header);
```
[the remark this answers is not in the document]
oleorhagen: Parts of it, I still don't allow it in the payload `Next` call. But it's kinda disabled by default now. Enabling it is easy, when we want to :)

artifact/v3/header/token.hpp
vpodzime: If you have an enum that starts with `0`, you can have this mapping in an array and use the enum value as an index in the array.

common/json/json.cpp
```diff
@@ -45,6 +48,56 @@ error::Error MakeError(JsonErrorCode code, const string &msg) {
...
+			MakeError(JsonErrorCode::ParseError, "The JSON is not an object"));
+	}
+
+	auto expected_children = j.GetChildren();
+	if (!expected_children) {
+		return expected::unexpected(
+			MakeError(JsonErrorCode::ParseError, expected_children.error().message));
+	}
+
+	unordered_map<string, string> kv_map {};
```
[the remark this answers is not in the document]
vpodzime: No, `using ExpectedKeyValueMap = expected::expected<KeyValueMap, error::Error>;`. It can use the new name of the type.

artifact/v3/header/header.hpp
vpodzime: Actually, I meant `artifact_name - artifact_group`  should be pairs, but that's not how it is in the JSON (if I recall correctly what I've seen in the code).

artifact/v3/header/header.hpp
vpodzime: I see, but a `struct` would still be nicer, I think. We do it everywhere else, AFAICT.

artifact/parser.cpp
oleorhagen: Oh my, we're not. Fixing.

artifact/v3/header/header.hpp
oleorhagen: Ahh, yeah. Ideally they probably should.

artifact/v3/header/header_info.cpp
```diff
@@ -0,0 +1,195 @@
...
+		auto json_element = expected_element.value();
+		if (json_element.IsString()) {
+			string payload_type = json_element.GetString().value();
+			if (payload_type == "") {
+				return expected::unexpected(
+					parser_error::MakeError(parser_error::Code::ParseError, "Empty Payload type"));
+			}
+			Payload type {Payload::RootfsImage};
+			if (payload_type != "rootfs-image") {
+				type = Payload::ModuleImage;
```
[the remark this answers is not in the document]
oleorhagen: https://github.com/oleorhagen/mender/blob/header-parser-pr/artifact/v3/header/header_info.cpp#L67

Done here.

artifact/v3/header/header.hpp
oleorhagen: Yup, agreed. Moved to a struct.

artifact/v3/header/token.hpp
```diff
@@ -0,0 +1,114 @@
...
+	}
+
+private:
+	Type StringToType(const string &type_name) {
+		if (type_name == "header-info") {
+			return Type::HeaderInfo;
+		}
+		if (std::regex_match(type_name, artifact_script_regexp)) {
+			const string script_name = type_name.substr(7 + 1); // Strip the scripts + / prefix
+			type_map[Type::ArtifactScripts] = script_name;
```
vpodzime, line 92: Modifying a global `type_map` from a function in `Token` looks quite fishy.

artifact/v3/manifest/manifest.cpp
```diff
@@ -80,6 +96,14 @@ ExpectedManifest Parse(mender::common::io::Reader &reader) {
 	return m;
 }
 
+string Manifest::Get(const string &key) {
+	auto value = this->map_.find(key);
+	if (value != this->map_.end()) {
```
vpodzime, line 101: I'd do `if (this->map_.count(key) != 0)`

oleorhagen: This is pre-existing. I've only moved it into the `cpp` file.

Also, it's how the `cppreference examples` do it

vpodzime:
@oleorhagen I've rebased this and resolved conflicts (and one `if (err)...` comparison).

artifact/v3/header/token.hpp
kacf: I'd have to agree. `type_map` is a singleton, but `Token` isn't, so this is a dangerous relationship.

artifact/v3/header/token.hpp
oleorhagen: I won't handle this here, and maybe we should discuss this a little more.

So in short this means that we will accept Artifacts which do not align with the format we have described, I know that the regular cli tools create this form of structure, but then we should update our documentation also, no?

kacf: I wouldn't say that it doesn't align. It's just that the `tar` tool creates a slightly different internal structure compared to the Go tar writer, but it's still semantically valid. It doesn't really change anything, it's just the same thing represented a bit differently.

It should be pretty easy to filter out though, shouldn't it? Something like (pseudo code):

```
if (entry.type == directory) {
    if (entry.path is a prefix of any valid token) {
        ignore();
    } else {
        error("Forbidden entry");
    }
}
```

kacf (changes requested):

artifact/v3/header/header.cpp
```diff
@@ -0,0 +1,182 @@
...
+			// sub_header.metadata = expected_meta_data.value();
+			// }
+			tok = lexer.Next();
+		}
+		log::Trace("sub-header: parsed the meta-data");
+
+		header.subHeaders.push_back(sub_header);
+
+		i++;
+		break; // NOTE: We currently do not support multiple payloads
```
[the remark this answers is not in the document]
oleorhagen: Hmm, maybe we should keep spinning it actually

artifact/v3/header/token.hpp
```diff
@@ -0,0 +1,131 @@
...
+	const string TypeToString() const {
+		return type_map.at(type);
+	}
+
+	const int Index() const {
+		switch (type) {
+		case Type::ArtifactHeaderTypeInfo:
+			// fallthrough
+		case Type::ArtifactHeaderMetaData:
+			return stoi(this->name.substr(8, 4));
```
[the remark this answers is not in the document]
oleorhagen: 1. Hmm, ur probably right. I figured it was OK since we've already vetted the string.
2. fixing.

artifact/parser_test.cpp
```diff
@@ -39,6 +40,8 @@ class ParserTestEnv : public testing::Test {
 public:
 protected:
 	static void SetUpTestSuite() {
+		mender::common::log::SetLevel(mender::common::log::LogLevel::Trace);
```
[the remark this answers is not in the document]
oleorhagen: Hmm, I don't agree here actually. I think it should always be enabled. However, it should only be shown on a failing test, which is more up to the test framework I would think.

But if ur adamant...

artifact/v3/header/header.cpp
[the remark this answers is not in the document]
oleorhagen: Added this:

```
		// NOTE: We currently do not support multiple payloads
		if (i != 0) {
			return expected::unexpected(parser_error::MakeError(
				parser_error::Code::ParseError,
				"Multiple header entries found. Currently only one is supported"));
		}
```
To the start of the loop for now

artifact/v3/header/token.hpp
oleorhagen: Hmm, alright. But keeping it out of this PR

kacf: Fair enough.

artifact/parser_test.cpp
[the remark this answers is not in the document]
kacf: Not disagreeing with that! But right now it will be noisy. And even if we want to improve the test framework, it would be better to enable Trace level globally, not here in this single test, or?

kacf:
> [oleorhagen](https://github.com/oleorhagen) requested a review from [kacf](https://github.com/kacf)

Did you push any changes..? I can't see anything new.

artifact/parser_test.cpp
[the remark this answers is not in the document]
oleorhagen: If you give the green light, I'm on it asap!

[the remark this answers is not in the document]
kacf: As long as it's hidden for passing tests, I'm all for it. I'd recommend a new PR for this though, it's not really related to this code.

kacf (approved):
🎆 🎇

PR #1206, opened 2023-03-30
MenderContext::LoadProvides() and CommitArtifactData()

vpodzime:
Ticket: MEN-6390
Changelog: none

### DIFF
--- a/common/context/context.cpp
+++ b/common/context/context.cpp
@@ -14,16 +14,26 @@
 
 #include <common/context.hpp>
 
-#include <common/error.hpp>
+#include <common/common.hpp>
 #include <common/conf/paths.hpp>
+#include <common/error.hpp>
+#include <common/expected.hpp>
+#include <common/io.hpp>
+#include <common/json.hpp>
+#include <common/key_value_database.hpp>
 
 namespace mender {
 namespace common {
 namespace context {
 
 using namespace std;
+namespace common = mender::common;
 namespace conf = mender::common::conf;
 namespace error = mender::common::error;
+namespace expected = mender::common::expected;
+namespace io = mender::common::io;
+namespace json = mender::common::json;
+namespace kv_db = mender::common::key_value_database;
 
 error::Error MenderContext::Initialize(const conf::MenderConfig &config) {
 #if MENDER_USE_LMDB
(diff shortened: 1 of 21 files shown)


common/context/context.cpp
```diff
@@ -14,16 +14,28 @@
...
 using namespace std;
+namespace common = mender::common;
 namespace conf = mender::common::conf;
 namespace error = mender::common::error;
+namespace expected = mender::common::expected;
+namespace io = mender::common::io;
+namespace json = mender::common::json;
+namespace kv_db = mender::common::key_value_database;
+
+const string MenderContext::provides_key_ = "artifact-provides";
```
vpodzime, line 38: @kacf no idea what the right key name here is. I guess it should be consistent with the Go client?

common/context/context.cpp
```diff
@@ -53,6 +65,52 @@ kv_db::KeyValueDatabase &MenderContext::GetMenderStoreDB() {
...
+	auto children = ex_children.value();
+	if (!all_of(children.cbegin(), children.cend(), [](const json::ChildrenMap::value_type &it) {
+			return it.second.IsString();
+		})) {
+		auto err = json::MakeError(json::TypeError, "Unexpected non-string data in provides");
+		return ExpectedProvidesData(expected::unexpected(err));
+	}
+
+	ProvidesData ret {};
+	for (const auto &it : ex_children.value()) {
```
kacf, line 100: There is a misunderstanding here: The incorrect keys are not stored in the JSON object, they are stored as separate database keys. In other words, in LMDB, you have [three keys](https://github.com/mendersoftware/mender/blob/feature-c%2B%2B-client/common/context.hpp#L46-L57), `artifact-name`, `artifact-group` and `artifact-provides`. The two first keys contain plain values, the last one JSON. But in the end-result, we want to return only one JSON, containing all keys, including the ones which were originally outside in separate keys in the DB.

And then the rename, which you've already done.

For this you will need to use a transaction, because you will need to read three separate keys from the database.

common/context/context.cpp
```diff
@@ -53,6 +65,52 @@ kv_db::KeyValueDatabase &MenderContext::GetMenderStoreDB() {
...
+ExpectedProvidesData MenderContext::LoadProvides() {
+	auto ex_bytes = mender_store_.Read(provides_key_);
+	if (!ex_bytes) {
+		if (ex_bytes.error().code == kv_db::MakeError(kv_db::KeyError, "").code) {
+			// no such data is not an error, just return an empty result
+			ProvidesData ret {};
+			return ExpectedProvidesData(ret);
+		} else {
+			// everything else is an error
+			return ExpectedProvidesData(expected::unexpected(ex_bytes.error()));
```
kacf, line 77: I know you are a fan of explicit construction, but I must admit I find these harder to read. It's harder to visually distinguish good paths from error paths, because they look more alike. But it's not wrong of course, just my 2 cents.

kacf (changes requested):

common/context/context.cpp
vpodzime: Yeah, I have mixed feelings about this myself.

common/context/context.cpp
vpodzime: Good that I submitted this alone first :)

kacf (approved):
Now it looks correct to me! 👍

kacf:
@vpodzime: I forgot one thing earlier: This functionality is specific to `mender-update`, so it should not be in `common`.

mender-update/context/context.cpp
```diff
@@ -122,6 +122,57 @@ ExpectedProvidesData MenderContext::LoadProvides() {
...
+				artifact_provides_key, common::ByteVectorFromString(artifact_provides_str));
+			if (err != error::NoError) {
+				return err;
+			}
+		}
+		return error::NoError;
+	};
+
+	return mender_store_.WriteTransaction(
+		[&](kv_db::Transaction &txn) { return commit_data_to_db(txn); });
```
kacf, line 173: This has the same signature. Can't you just pass in `commit_data_to_db` directly?

kacf (approved):

mender-update/context/context.cpp
vpodzime: I surely can! :+1:  This is a remnant of copy-pasta from `LoadProvides(). :blush:

common/context_test.cpp
```diff
@@ -42,18 +44,18 @@ TEST_F(ContextTests, LoadProvidesValid) {
 
 	context::MenderContext ctx;
 	auto err = ctx.Initialize(cfg);
-	ASSERT_FALSE(err);
+	ASSERT_EQ(err, error::NoError);
```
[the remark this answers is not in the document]
vpodzime: Yes, good catch! I changed them to make Lukasz's life easier, but it happened later so... Moving them.

PR #1208, opened 2023-04-03
chore: Enforce use of `override` everywhere where it is appropriate.

kacf:
It helps readability by knowing where overrides are, and can also prevent mistakes in cases where an override is unintentional.


oleorhagen (approved):

vpodzime (approved):
ACK

PR #1214, opened 2023-04-11
Async I/O + process improvements + bug fixes

kacf:
All of these are enablers for what is to come in [MEN-6317](https://northerntech.atlassian.net/browse/MEN-6317).

[MEN-6317]: https://northerntech.atlassian.net/browse/MEN-6317?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


common/processes_test.cpp
```diff
@@ -150,3 +155,82 @@ TEST_F(ProcessesTests, SpawnFailGenerateLineDataTest) {
...
+	remove(testfile.c_str());
+
+	auto exit_status = proc.Wait();
+	EXPECT_EQ(exit_status, 0);
+}
+
+TEST_F(ProcessesTests, Terminate) {
+	mtesting::TemporaryDirectory tmpdir;
+
+	string testfile = tmpdir.Path() + "/testfile";
```
oleorhagen, line 197: Nit: copy-pasta error: `testfile` is unused here and below

common/setup/platform/posix/setup.cpp
```diff
@@ -0,0 +1,32 @@
...
+
+namespace mender {
+namespace common {
+namespace setup {
+
+void GlobalSetup() {
+	struct sigaction action;
+	action.sa_handler = SIG_IGN;
+
+	sigaction(SIGPIPE, &action, nullptr);
```
oleorhagen, line 27: Q: When do we ever get a `SIGPIPE` ? :thinking: 

And what about term, abort, etc?

common/events_io.hpp
```diff
@@ -0,0 +1,88 @@
...
+//    distributed under the License is distributed on an "AS IS" BASIS,
+//    WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+//    See the License for the specific language governing permissions and
+//    limitations under the License.
+
+#ifndef MENDER_COMMON_IO_UTIL_HPP
+#define MENDER_COMMON_IO_UTIL_HPP
+
+#include <boost/asio.hpp>
+
```
oleorhagen, line 19: Nit: Missing std-lib imports

common/events_io_test.cpp
```diff
@@ -0,0 +1,386 @@
...
+
+	int fds[2];
+	ASSERT_EQ(pipe(fds), 0);
+
+	io::AsyncFileDescriptorReader reader(loop, fds[0]);
+	io::AsyncFileDescriptorWriter writer(loop, fds[1]);
+
+	const uint8_t data[] = "abcd";
+
+	vector<uint8_t> to_send {data, data + sizeof(data)};
```
oleorhagen, line 52: Hmm, is this right? Should this not be `()` ? :thinking:

common/events_io_test.cpp
```diff
@@ -0,0 +1,386 @@
...
+
+class EventsIo : public testing::Test {
+protected:
+	void SetUp() override {
+		mender::common::setup::GlobalSetup();
+	}
+};
+
+TEST_F(EventsIo, ReadAndWriteWithPipes) {
+	TestEventLoop loop;
```
oleorhagen, line 42: Q: Isn't this really more like: `timeout` or similar?

common/events_io_test.cpp
```diff
@@ -0,0 +1,386 @@
...
+
+			in_read = true;
+		});
+	ASSERT_EQ(err, error::NoError);
+	err =
+		writer.AsyncWrite(to_send.begin(), to_send.end(), [&in_write](size_t n, error::Error err) {
+			EXPECT_EQ(err, error::NoError);
+			EXPECT_EQ(n, 5);
+
+			in_write = true;
```
oleorhagen, line 74: Is this really necessary? Isn't matching the data enough?

common/events_io_test.cpp
```diff
@@ -0,0 +1,386 @@
...
+
+	loop.Run();
+
+	EXPECT_TRUE(in_write);
+}
+
+TEST_F(EventsIo, FileOpen) {
+	mtesting::TemporaryDirectory tmpdir;
+	TestEventLoop loop;
+	string tmpfile = tmpdir.Path() + "file";
```
oleorhagen, line 332: Nit: Previously you joined with `Path() + /<somepath>`

.gitlab-ci.yml
```diff
@@ -31,7 +31,7 @@ test:
     - chmod +x /usr/local/bin/mender-artifact
   script:
     - cmake -D COVERAGE=ON .
-    - make coverage
+    - make -k coverage
```
oleorhagen, line 34: `--keep-going` :pray:

oleorhagen (approved):
Phew! Quite a chunk to go through!

Left a few comments and questions here and there, but no major issues, so pre-approving.

common/processes/platform/tiny_process_library/tiny_process_library.cpp
```diff
@@ -17,13 +17,29 @@
...
 
-#include <process.hpp>
-
-namespace tpl = TinyProcessLib;
 using namespace std;
 
 namespace mender::common::processes {
 
+error::Error Process::Start() {
+	proc_.reset(new tpl::Process(this->args_, ""));
```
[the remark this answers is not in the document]
kacf: I was thinking I couldn't use `make_unique` because it is C++14, but I forgot this is a platform file. Fixed.

`this->args_` was preexisting (just moved), but I nevertheless simplified it now.

lukfi (approved):
only a few questions asked, approved

common/processes/platform/tiny_process_library/tiny_process_library.cpp
```diff
@@ -53,22 +69,36 @@ ExpectedLineData Process::GenerateLineData() {
 			if ((line_end_idx == string_view::npos) && (line_start_idx != (len - 1))) {
 				trailing_line += string(bytes_view, line_start_idx, (len - line_start_idx));
 			}
-		});
+		}));
+
+	auto id = proc_->get_id();
```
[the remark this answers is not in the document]
kacf: Because `Wait()` sets `proc_` to null, so we have to fetch it before we call `Wait()`.

common/processes_test.cpp
```diff
@@ -150,3 +155,82 @@ TEST_F(ProcessesTests, SpawnFailGenerateLineDataTest) {
 	EXPECT_EQ(proc.GetExitStatus(), 1);
 	EXPECT_EQ(ex_line_data.value().size(), 0);
 }
+
+TEST_F(ProcessesTests, StartInBackground) {
+	mtesting::TemporaryDirectory tmpdir;
+
+	string testfile = tmpdir.Path() + "/testfile";
```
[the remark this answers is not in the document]
kacf: Yes, hmm, weird location though, `mender::common::conf::paths`. But I won't change that now, for now I'll just use it.

common/setup/platform/posix/setup.cpp
kacf: When the reading end of a pipe closes its end, and the other end tries to write to it. This is used in the tests, and is expected to come up if an Update Module crashes in the Download state. In this scenario the client is writing to the process through a pipe, so it's important that the signal is ignored so that we can get an error code instead.

I assume that the other signals will be handled elsewhere, because they may need more targeted treatment. Or maybe some of it can be added here, not sure yet. But it's anyway something I won't handle now, because it's not as clear cut as `SIGPIPE`.

common/events_io_test.cpp
kacf: I believe it's calling the right constructor. `data` is a pointer, so it would have complained about types otherwise. But I admit this looks confusingly like a list of elements, so I'll change it (in many places) to make it clearer.

common/setup/platform/posix/setup.cpp
oleorhagen: I saw it for the tests. Didn't know we actually did it for the `Download` stat, but now that you mention it :)

common/events_io_test.cpp
```diff
@@ -0,0 +1,386 @@
...
+	vector<uint8_t> to_send {data, data + sizeof(data)};
+	vector<uint8_t> to_receive;
+	to_receive.resize(to_send.size());
+
+	bool in_read {false};
+	bool in_write {false};
+
+	auto err = reader.AsyncRead(
+		to_receive.begin(), to_receive.end(), [&loop, &in_read](size_t n, error::Error err) {
+			EXPECT_EQ(err, error::NoError);
```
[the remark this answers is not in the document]
kacf: Maybe. Doesn't save that much though, I think I'll leave it alone for now.

common/processes_test.cpp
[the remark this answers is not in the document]
oleorhagen: Hah, yeh. This has come up in multiple reviews now. 

I'll see if I'll find the time to move it today :)

common/events_io_test.cpp
kacf: Well, kind of, but its primary purpose is still to be an event loop, the timeout is just a fallback. Wouldn't it look funny with `timeout.Run()` to run an event loop?

common/events/platform/boost/events_io.cpp
```diff
@@ -0,0 +1,155 @@
...
+	cancelled_(make_shared<bool>(false)) {
+}
+
+error::Error AsyncFileDescriptorReader::Open(const string &path) {
+	int fd = open(path.c_str(), O_RDONLY);
+	if (fd < 0) {
+		int err = errno;
+		return error::Error(generic_category().default_error_condition(err), "Cannot open " + path);
+	}
+	pipe_.assign(fd);
```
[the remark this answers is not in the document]
kacf: Good observation. The documentation is pretty bad and doesn't say anything. But I checked the source code, and it looks like throws an exception. So let's close it before assigning a new descriptor.

common/events/platform/boost/events_io.cpp
```diff
@@ -0,0 +1,155 @@
...
+	int flags = O_WRONLY | O_CREAT;
+	switch (append) {
+	case Append::Disabled:
+		flags |= O_TRUNC;
+		break;
+	case Append::Enabled:
+		flags |= O_APPEND;
+		break;
+	}
+	int fd = open(path.c_str(), flags, 0644);
```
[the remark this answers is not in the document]
kacf: I was debating this myself as well. The reason I didn't include it as an argument, is that it uses the `mode_t` type, which is a POSIX type, and therefore not part of C++11. I figured that we probably want `0644` in most cases, and if not, then the caller can open the descriptor themself.

common/events_io.hpp
```diff
@@ -0,0 +1,88 @@
...
+	error::Error AsyncRead(
+		vector<uint8_t>::iterator start,
+		vector<uint8_t>::iterator end,
+		mender::common::io::AsyncIoHandler handler) override;
+	void Cancel() override;
+
+private:
+#ifdef MENDER_USE_BOOST_ASIO
+	asio::posix::stream_descriptor pipe_;
+	shared_ptr<bool> cancelled_;
```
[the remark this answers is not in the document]
kacf: Hmm, I didn't think about the destruction of `pipe_`. I know that the ASIO framework can call the handler after having called `cancel()`, which is the motivation for this variable. I'll do some testing. Possibly you may be right, and we can use a reference instead. However, it's also possible that ASIO keeps a stale reference to `pipe_`, and will still call the handler, and in that case we need to keep the shared pointer, and make `pipe_` into a shared pointer as well, to keep it from being destroyed before the handler is called.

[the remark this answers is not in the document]
kacf: @lukfi: Looks like we do need it. The handler can still be called after `pipe_` is destroyed, so we need to keep a shared reference. However, it looks like ASIO copies the data it needs from `pipe_`, so we don't need a shared pointer for that one (I tested it with Valgrind and it was happy).

Check out the new commits, I added a test for this scenario.

kacf:
Updated!

@oleorhagen: Not sure if you want to look again, but this is your chance to try `git range-diff`! 🙂

oleorhagen:
> Updated!
> 
> @oleorhagen: Not sure if you want to look again, but this is your chance to try `git range-diff`! slightly_smiling_face

Yes, giving it a go :)

common/setup/platform/posix/setup.cpp
```diff
@@ -15,13 +15,15 @@
...
 #include <signal.h>
+#include <cstring>
 
 namespace mender {
 namespace common {
 namespace setup {
 
 void GlobalSetup() {
 	struct sigaction action;
+	memset(&action, 0, sizeof(action));
```
oleorhagen, line 26: Makes me sleep a little better at least :)

PR #1215, opened 2023-04-12
Header metadata parser

oleorhagen:
This adds the meta-data parser from MEN-6424, as well as a few touches to our Json implementation, notably:

* Fix: Align the `GetInt` function with the underlying Nlohmann integer representation
* feat: Json::GetDouble

Note, this is based on top of https://github.com/mendersoftware/mender/pull/1203 so feel free to ignore those commits obviously :)

### DIFF
--- a/artifact/v3/header/header_test.cpp
+++ b/artifact/v3/header/header_test.cpp
@@ -32,6 +32,7 @@
 using namespace std;
 
 namespace io = mender::common::io;
+namespace json = mender::common::json;
 namespace tar = mender::tar;
 namespace processes = mender::common::processes;
 namespace mendertesting = mender::common::testing;
@@ -243,8 +244,6 @@ TEST_F(HeaderTestEnv, TestHeaderRootfsAllFlagsSetSuccess) {
 		"rootfs_image_checksum");
 	EXPECT_EQ(
 		header.subHeaders.at(0).type_info.clears_artifact_provides.value().at(2), "rootfs-image.*");
-
-	// meta-data
 }
 
 TEST_F(HeaderTestEnv, TestHeaderModuleImageAllFlagsSetSuccess) {
@@ -347,7 +346,9 @@ TEST_F(HeaderTestEnv, TestHeaderModuleImageAllFlagsSetSuccess) {
 		header.subHeaders.at(0).type_info.clears_artifact_provides.value().at(0),
 		"rootfs-image.dummy-update-module.*");
 
-	// meta-data
+	// headers/0000/meta-data
+	ASSERT_TRUE(header.subHeaders.at(0).metadata);
+	// EXPECT_EQ(header.subHeaders.at(0).meta_data.value().Dump(), "rootfs-image.*");
 }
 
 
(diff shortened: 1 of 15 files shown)


common/json/platform/nlohmann/nlohmann_json.cpp
```diff
@@ -152,6 +152,14 @@ bool Json::IsInt() const {
 	return this->n_json.is_number_integer();
 }
 
+bool Json::IsNumber() const {
```
oleorhagen, line 155: I will split this out

artifact/v3/header/header_test.cpp
```diff
@@ -244,6 +245,7 @@ TEST_F(HeaderTestEnv, TestHeaderRootfsAllFlagsSetSuccess) {
 		header.subHeaders.at(0).type_info.clears_artifact_provides.value().at(2), "rootfs-image.*");
 
 	// meta-data
+	// TODO - Fill in
```
oleorhagen, line 248: Will finish this, when we agree on what the structure will be and look like.

artifact/v3/header/header_test.cpp
```diff
@@ -446,3 +448,176 @@ TEST_F(HeaderTestEnv, TestHeaderFilesOutOfOrder) {
...
+
+	auto expected_meta_data = header::meta_data::Parse(sr);
+
+	ASSERT_FALSE(expected_meta_data);
+	EXPECT_EQ(
+		expected_meta_data.error().message,
+		"The meta-data needs to only be strings, ints and arrays of ints and strings");
+}
+
+// TEST_F(HeaderTestEnv, TestHeaderMetaDataIsMinifiedFailure) {
```
oleorhagen, line 523: Still a little confused, should we verify that it is minified or not?

artifact/v3/header/header_test.cpp
```diff
@@ -446,3 +448,176 @@ TEST_F(HeaderTestEnv, TestHeaderFilesOutOfOrder) {
...
+		ASSERT_TRUE(val) << val.error().message;
+		ASSERT_EQ(val.value(), -9007199254740991);
+	}
+}
+
+TEST_F(HeaderTestEnv, TestHeaderMetaDataIs64BitFloatingPointIsRounded) {
+	stringstream meta_data {
+		R"(
+{
+  "one-out-of-64-bit-range": 9007199254740992,
```
oleorhagen, line 595: Well, technically 53-bit range... Fixing.

artifact/v3/header/header_test.cpp
kacf: I suspect we don't need to. It's mender-artifact's responsibility to make sure that it is minified. If the artifact happens to be handcrafted I suppose it might not be minified, but I don't think the client should care.

When we put the meta-data in the Update Module file structure, it needs to be line separated though. This is important due to the "dumb" parsing done by many shell scripts.

artifact/v3/header/meta_data.cpp
```diff
@@ -0,0 +1,103 @@
...
+	//
+	// The rootfs-image currently creates an empty meta-data file, which does not
+	// play nice, since our Json library is correct in assuming that '' is invalid
+	// Json.
+	//
+	// I'm a little unsure of how to handle this. We could filter this out in the tar-reader
+	//
+	// Or we could handle this case especially like below?
+	//
+	// In short it is difficult to get the size from our stream as is now.
```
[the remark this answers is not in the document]
oleorhagen: > And I assume that after you have consumed it, and gotten an error, you can't re-consume it to check for emptyness?

Correct.

> . I guess this breaks the Reader/Consumer boundary

Yes :(

I am actually in favour of checking the error returned from the `json` lib.

It reports that on `line 1, column 1` it can't find a valid value, which is just this case.

What do you think?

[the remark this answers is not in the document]
kacf: I would also be in favor of checking errors from json, but how reliable is it? If you just put an `X` there, will that produce the same error?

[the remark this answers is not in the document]
oleorhagen: No, but a file with an `X` is not empty. What's ur argument here - I don't follow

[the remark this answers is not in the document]
oleorhagen: Please have a look-over the latest diff now :)

artifact/v3/header/meta_data.cpp
```diff
@@ -0,0 +1,100 @@
...
+ExpectedMetaData Parse(io::Reader &reader) {
+	log::Trace("Parsing the header meta-data");
+	auto expected_json = json::Load(reader);
+
+	if (!expected_json) {
+		log::Trace("Received json load error: " + expected_json.error().message);
+		bool is_empty_payload_error =
+			expected_json.error().message.find(empty_json_error_message) != string::npos;
+		if (is_empty_payload_error) {
+			log::Debug("Received an empty Json body. Not treating this as an error");
```
kacf, line 83: I would even consider `Trace` for this. *Every* rootfs-image will be like this.

artifact/v3/header/meta_data.cpp
```diff
@@ -0,0 +1,100 @@
...
+			return json::Json();
```
kacf, line 84: Shouldn't this be tested? Including the malformed but similar cases we mentioned, like a single `{`, or a single space.

oleorhagen: Done

PR #1216, opened 2023-04-13
chore: Make sure Boost library does not trigger warnings.

kacf:


kacf:
@oleorhagen: I think this should fix your issue. But since I can't reproduce the issue, I can't verify it.

oleorhagen (approved):

PR #1217, opened 2023-04-13
Add a mender::common::main_test loop

oleorhagen:
This means we can call setup and teardown functions for the entire test framework.

For now we only set the log-level to `Trace` for all the tests.

I also did some minor fixes to the logger, feel free to reject any of these at will. I can also move it to a separate PR if you want.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -146,7 +146,7 @@ add_custom_target(coverage_no_tests
 # CMake is not clever enough to build the tests before running them so we use
 # the 'check' target below that does both.
 add_custom_target(check
-  COMMAND ${CMAKE_CTEST_COMMAND}
+  COMMAND ${CMAKE_CTEST_COMMAND} --output-on-failure
 )
 
 # CMake doesn't generate the 'uninstall' target.
--- a/common/main_test.cpp
+++ b/common/main_test.cpp
@@ -1,0 +1,37 @@
+// Copyright 2023 Northern.tech AS
+//
+//    Licensed under the Apache License, Version 2.0 (the "License");
+//    you may not use this file except in compliance with the License.
+//    You may obtain a copy of the License at
+//
+//        http://www.apache.org/licenses/LICENSE-2.0
+//
+//    Unless required by applicable law or agreed to in writing, software
+//    distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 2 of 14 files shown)


common/log.hpp
```diff
@@ -105,7 +105,7 @@ class Logger {
 
 	string name_ {};
 
-	LogLevel level_;
+	LogLevel level_ {LogLevel::Info};
```
kacf, line 109: You are reverting [this fix](https://github.com/mendersoftware/mender/commit/e2721c676686ce1a9d401b0733301c1d28b1e520).

common/log.hpp
```diff
@@ -105,7 +105,12 @@ class Logger {
 
 	string name_ {};
 
+#ifdef NDEBUG
 	LogLevel level_ {LogLevel::Info};
+#else
+	LogLevel level_ {LogLevel::Debug};
+#
+#endif
```
kacf, line 113: Good idea!

CMakeLists.txt
```diff
@@ -146,7 +146,7 @@ add_custom_target(coverage_no_tests
 # CMake is not clever enough to build the tests before running them so we use
 # the 'check' target below that does both.
 add_custom_target(check
-  COMMAND ${CMAKE_CTEST_COMMAND}
+  COMMAND ${CMAKE_CTEST_COMMAND} --output-on-failure
```
kacf, line 149: Ah! That was easier than expected. This means that running the executable directly will still output everything, including trace messages, but we can probably get around that by running the right ctest command.

common/main_test.cpp
```diff
@@ -0,0 +1,21 @@
...
+#include <common/log.hpp>
+
+using namespace testing;
+
+namespace log = mender::common::log;
+
+class GlobalTestEnvironment : public ::testing::Environment {
+public:
+	void SetUp() override {
+		log::SetLevel(log::LogLevel::Trace);
```
kacf, line 12: While you're at it, could you also move [this call](https://github.com/mendersoftware/mender/blob/feature-c%2B%2B-client/common/events_io_test.cpp#L39) to here, and remove it from that specific test?

common/log.hpp
oleorhagen: Seems ur are right - TIL!

kacf (approved):

PR #1224, opened 2023-04-19
feat: Add AsyncReaders and Wait interface to the Process class.

kacf:
Plus several other smaller fixes. See individual commits.


common/processes/platform/tiny_process_library/tiny_process_library.cpp
```diff
@@ -17,30 +17,163 @@
...
 using namespace std;
 
 namespace mender::common::processes {
 
+const chrono::seconds MAX_TERMINATION_TIME(10);
+
+namespace io = mender::common::io;
+namespace log = mender::common::log;
+
+class ProcessReaderFunctor {
```
oleorhagen, line 33: Hah, had me confused till I realised that a `functor` in Cpp is simply a function object.

common/processes/platform/tiny_process_library/tiny_process_library.cpp
```diff
@@ -17,30 +17,163 @@
...
 
+#include <common/events_io.hpp>
+#include <common/io.hpp>
+#include <common/log.hpp>
+
 using namespace std;
 
 namespace mender::common::processes {
 
+const chrono::seconds MAX_TERMINATION_TIME(10);
```
oleorhagen, line 28: Q: Since this is in essence making for the uniform behaviour of our process setup (It is part of the default construction). Should it not be a part of the header, like generate_line_data?

common/processes/platform/tiny_process_library/tiny_process_library.cpp
```diff
@@ -123,4 +327,24 @@ void Process::Kill() {
 	}
 }
 
+void ProcessReaderFunctor::operator()(const char *bytes, size_t n) {
+	if (fd_ < 0) {
+		return;
+	}
+
+	size_t written = 0;
+	while (written < n) {
```
oleorhagen, line 336: Q: While is this essentially a retry, instead of an incremental write?

common/processes/platform/tiny_process_library/tiny_process_library.cpp
```diff
@@ -77,23 +215,89 @@ ExpectedLineData Process::GenerateLineData() {
...
+io::ExpectedAsyncReaderPtr Process::GetAsyncStdoutReader(events::EventLoop &loop) {
+	return GetProcessReader(loop, stdout_pipe_);
+}
+
+io::ExpectedAsyncReaderPtr Process::GetAsyncStderrReader(events::EventLoop &loop) {
+	return GetProcessReader(loop, stderr_pipe_);
+}
+
+int Process::EnsureTerminated() {
+	if (proc_) {
```
oleorhagen, line 274: S:
I would invert this I think:

So, `if (!proc) { return exit_status };`

Then, the logic at the lowest level:

```
		log::Info("Sending SIGTERM to PID " + to_string(proc_->get_id()));

		Terminate();

		auto result = future_exit_status_.wait_for(max_termination_time_);
		if (result == future_status::timeout) {
			log::Info("Sending SIGKILL to PID " + to_string(proc_->get_id()));
			Kill();
			result = future_exit_status_.wait_for(max_termination_time_);
			if (result == future_status::ready) {
				// This should not be possible, SIGKILL always terminates.
				log::Error(
					"PID " + to_string(proc_->get_id()) + " still not terminated after SIGKILL.");
				return -1;
			}
		}
		assert(result == future_status::ready);
		exit_status_ = future_exit_status_.get();

		log::Info(
			"PID " + to_string(proc_->get_id()) + " exited with status " + to_string(exit_status_));

		proc_.reset();
```

common/processes/platform/tiny_process_library/tiny_process_library.cpp
```diff
@@ -77,23 +215,89 @@ ExpectedLineData Process::GenerateLineData() {
...
-		return expected::unexpected(MakeError(
-			ProcessesErrorCode::SpawnError,
-			"Failed to spawn '" + (args_.size() >= 1 ? args_[0] : "<null>") + "'"));
+	return ExpectedLineData(ret);
+}
+
+io::ExpectedAsyncReaderPtr Process::GetProcessReader(events::EventLoop &loop, int &pipe_ref) {
+	if (proc_) {
+		return expected::unexpected(
+			MakeError(ProcessAlreadyStartedError, "Cannot get process output"));
```
oleorhagen, line 243: Q: Why can't I get output from an already started process?

common/processes/platform/tiny_process_library/tiny_process_library.cpp
```diff
@@ -77,23 +215,89 @@ ExpectedLineData Process::GenerateLineData() {
...
+	return GetProcessReader(loop, stdout_pipe_);
+}
+
+io::ExpectedAsyncReaderPtr Process::GetAsyncStderrReader(events::EventLoop &loop) {
+	return GetProcessReader(loop, stderr_pipe_);
+}
+
+int Process::EnsureTerminated() {
+	if (proc_) {
+		log::Info("Sending SIGTERM to PID " + to_string(proc_->get_id()));
```
oleorhagen, line 275: Q: Should this really be as high as info?

common/processes_test.cpp
```diff
@@ -260,22 +272,238 @@ hard_sleep() {
...
+	// Cut down a bit on the kill time in tests.
+	max_termination_time = chrono::seconds {1};
+
+	auto pid = ProcessesTestsHelper::GetNativeProc(*proc)->get_id();
+
+	// Kill by destruction.
+	proc.reset();
+
+	auto result = ::kill(pid, 0);
+	ASSERT_EQ(result, -1);
```
oleorhagen, line 323: smrt!

common/processes_test.cpp
```diff
@@ -260,22 +272,238 @@ hard_sleep() {
...
+	int errnum = errno;
+	// No such process.
+	ASSERT_EQ(errnum, ESRCH);
+
+	auto finish_time = chrono::steady_clock::now();
+	auto diff = finish_time - start_time;
+	EXPECT_LT(diff, chrono::seconds {10});
```
oleorhagen, line 330: Nit: Why not use the already hard-coded variable value?

oleorhagen (approved):
Nice, although pretty involved.

common/processes_test.cpp
kacf: The fact that it's 10 is a coincidence. This value is not related to the `MAX_TERMINATION_TIME`, it's related to the `hard_sleep 10` which is in the shell script above.

common/processes/platform/tiny_process_library/tiny_process_library.cpp
kacf: I suppose it's both. I don't think `write` is supposed to ever return 0 for a blocking file descriptor. If the pipe is closed I get `errno == EPIPE` instead. So the retry case should in theory never happen, at least it didn't in my testing. But if it did, it should still work.

common/processes/platform/tiny_process_library/tiny_process_library.cpp
kacf: I think in most cases we expect processes to return by themselves, so logging that we are sending signals is worthwhile, I think.

common/processes/platform/tiny_process_library/tiny_process_library.cpp
kacf: You're right. The block was smaller when I started writing it. 🙂

common/processes/platform/tiny_process_library/tiny_process_library.cpp
kacf: We can make it possible, but what's the use case? I can't think of a case where starting a process output stream mid-way isn't a mistake.

common/processes/platform/tiny_process_library/tiny_process_library.cpp
kacf: Yes, but I wasn't planning to make this configurable, and in fact it could be a constant. The only reason I didn't make it a constant, but included it in construction instead, is so that the test can change its value and cut down on the testing time. If I didn't do this I would have doubled our current test run time!

kacf:
Updated, and added one new commit for the callbacks that @lukfi wanted.

common/processes/platform/tiny_process_library/tiny_process_library.cpp
oleorhagen: Yes, just wondering if there was any other justification :)

common/processes_test.cpp
oleorhagen: Ahh, I see. Fooled me!

PR #1226, opened 2023-04-24
feat(artifact/payload): Support multiple files read in each payload

oleorhagen:
This adds support for reading multiple files from each payload.

The solution is just an extension of the existing `Next()` interface, so that
now:

```
artifact::Parse() -> artifact

artifact.Next() -> Payload

Payload.Next() -> payload::Reader

...

Payload.Next() -> NoMorePayloadFilesError
```

Hence, each payload, and each file in the given payload can be processed in
turn.

The reader verifies the checksum of the given file as it is consumed, and
returns an error if there is a checksum mismatch.

The `payload::Reader` also has accessors for getting the `filename` and the
`size` of the given file.

Ticket: MEN-6452
Changelog: None

### DIFF
--- a/artifact/v3/payload/CMakeLists.txt
+++ b/artifact/v3/payload/CMakeLists.txt
@@ -1,6 +1,8 @@
 # Test the parser
 add_executable(artifact_payload_parser_test EXCLUDE_FROM_ALL
   payload.cpp
+  ${CMAKE_SOURCE_DIR}/artifact/v3/manifest/manifest.cpp
+  ${CMAKE_SOURCE_DIR}/artifact/error.cpp
   payload_test.cpp
 )
 target_link_libraries(artifact_payload_parser_test PRIVATE
--- a/artifact/error.hpp
+++ b/artifact/error.hpp
@@ -31,6 +31,7 @@ enum Code {
 	NoError = 0,
 	ParseError,
 	TypeError,
+	NoMorePayloadFilesError,
 	EOFError,
 };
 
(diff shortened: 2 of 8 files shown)


artifact/v3/payload/CMakeLists.txt
```diff
@@ -1,6 +1,8 @@
 # Test the parser
 add_executable(artifact_payload_parser_test EXCLUDE_FROM_ALL
   payload.cpp
+  ${CMAKE_SOURCE_DIR}/artifact/v3/manifest/manifest.cpp
+  ${CMAKE_SOURCE_DIR}/artifact/error.cpp
```
kacf, line 5: Why not link to the lib instead?

kacf (approved):
Looks good, just one question.

artifact/v3/payload/CMakeLists.txt
oleorhagen: It has not lib. This is the `artifact` error stuff. Or do you mean link to the whole `artifact` target?

kacf: Yes, the whole thing. It will only grab the symbols it needs.

oleorhagen: Yeh, I see what you mean. I will favour against it this time though, as then it will build the whole shebang, when only the error and manifest symbols are required.

kacf: When building only that test, yes. But be aware that this increases the total build time, since you are building those files twice now.

PR #1227, opened 2023-04-25
Empty payload support

oleorhagen:
Add support for empty payload Artifacts


artifact/parser.cpp
```diff
@@ -45,6 +45,35 @@ namespace version = mender::artifact::v3::version;
...
+			return expected::unexpected(parser_error::MakeError(
+				parser_error::Code::ParseError,
+				"This should never happen, we have a payload token / programmer error"));
+		}
+		auto payload = expected_payload.value();
+		auto expected_payload_file = payload.Next();
+		if (expected_payload_file) {
+			return expected::unexpected(parser_error::MakeError(
+				parser_error::Code::ParseError, "Empty Payload Artifacts cannot have a payload"));
+		}
```
kacf, line 72: Should you check that the error is `NoMorePayloadFilesError` and not some other error?

artifact/v3/header/header_info.cpp
```diff
@@ -104,6 +104,9 @@ ExpectedHeaderInfo Parse(io::Reader &reader) {
 	}
 	info.payloads = payloads.value();
 
+	// if (info.payloads.at(0) == Payload::EmptyPayload) {
+	// }
```
kacf, line 108: Remove?

artifact/v3/header/type_info.cpp
```diff
@@ -64,18 +64,24 @@ ExpectedTypeInfo Parse(io::Reader &reader) {
...
-	auto expected_payload = type_info_json.Get("type").and_then(json::ToString);
+	auto expected_payload = type_info_json.Get("type");
 	if (!expected_payload) {
+		return expected::unexpected(parser_error::MakeError(
+			parser_error::Code::ParseError,
+			"Failed to get the type-info payload type JSON: " + expected_payload.error().message));
+	}
+	auto payload_type = expected_payload.value();
+	if (payload_type.IsNull()) {
+		type_info.type = "null";
```
kacf, line 75: This overlaps with the potential real name `null`, no? We should probably use an extra enum or something to track this.

kacf (changes requested):

artifact/v3/header/type_info.cpp
oleorhagen: Discussed f2f, resolving independently

kacf (approved):

PR #1231, opened 2023-05-04
MEN-6317: Download state handling

kacf:
Update Module Download state handling, with many support commits.


oleorhagen (commented):
As far as I got today, continuing tomorrow

common/events_io.hpp
```diff
@@ -85,6 +85,48 @@ class AsyncFileDescriptorWriter :
 #endif // MENDER_USE_BOOST_ASIO
 };
 
+class AsyncReaderFromReader : virtual public mender::common::io::AsyncReader {
```
oleorhagen, line 88: namespace mio = mender::common::io or something similar and shorter. Maybe just namespace common = mender::common at least, so that we get `common::io` as the namespace?

common/events_io_test.cpp
```diff
@@ -437,3 +437,42 @@ TEST(EventsIo, DestroyReaderBeforeHandlerIsCalled) {
...
+	string input {"abcd"};
+
+	io::ReaderPtr reader = make_shared<io::StringReader>(input);
+
+	vector<uint8_t> output;
+	output.resize(100);
+
+	io::WriterPtr writer = make_shared<io::ByteWriter>(output);
+
+	auto areader = make_shared<events::io::AsyncReaderFromReader>(loop, reader);
```
oleorhagen, line 453: Tbh, I don't think the `From Reader` suffix is necessary, like it's already there in the instantiation?

common/events/events_io.cpp
```diff
@@ -0,0 +1,123 @@
...
+AsyncReaderFromReader::~AsyncReaderFromReader() {
+	Cancel();
+}
+
+error::Error AsyncReaderFromReader::AsyncRead(
+	vector<uint8_t>::iterator start,
+	vector<uint8_t>::iterator end,
+	mender::common::io::AsyncIoHandler handler) {
+	if (reader_thread_.joinable()) {
+		// Clean up previous operation. Careful, this can hang if AsyncRead is called twice
```
oleorhagen, line 38: Can we not detect if the handler has been run?

common/events/events_io.cpp
```diff
@@ -0,0 +1,123 @@
...
+		});
+	});
+
+	return error::NoError;
+}
+
+void AsyncWriterFromWriter::Cancel() {
+	*cancelled_ = true;
+	if (writer_thread_.joinable()) {
+		// Need to wait for thread to finish because iterators may be destroyed after this
```
oleorhagen, line 114: Very good point! Make it a `Note` mby?

mender-update/CMakeLists.txt
```diff
@@ -22,18 +22,23 @@ gtest_discover_tests(context_test)
...
   target_sources(update_module PRIVATE update_module/v3/platform/boost_filesystem/fs_operations.cpp)
   target_compile_options(update_module PRIVATE ${PLATFORM_SPECIFIC_COMPILE_OPTIONS})
   target_include_directories(update_module PRIVATE ${Boost_INCLUDEDIR})
-  target_link_libraries(update_module PUBLIC
-    common_log
-    common_conf
-    mender_context
-    artifact
-    ${Boost_LIBRARIES})
+  target_link_libraries(update_module PUBLIC ${Boost_LIBRARIES})
```
oleorhagen, line 41: Q: Why split htem, and keep both public, is it because `Boost` is external?

mender-update/context.hpp
```diff
@@ -68,6 +68,9 @@ class MenderContext {
 	ExpectedProvidesData LoadProvides();
 	expected::ExpectedString GetDeviceType();
 	error::Error CommitArtifactData(const ProvidesData &data);
+	const conf::MenderConfig &GetConfig() const {
```
oleorhagen, line 71: Q: I seem to recall someone having an argument with me about returning const references, or is this different?

mender-update/update_module/v3/platform/boost_filesystem/fs_operations.cpp
```diff
@@ -144,30 +166,31 @@ error::Error UpdateModule::PrepareFileTree(const string &path) {
 	//
 
 	err = CreateDataFile(
-		header_subdir_path, "artifact_group", update_meta_data_.header.artifact_group);
+		header_subdir_path, "artifact_group", payload_meta_data_.header.artifact_group);
```
oleorhagen, line 169: :+1:

common/json/platform/nlohmann/nlohmann_json.cpp
```diff
@@ -79,7 +79,13 @@ ExpectedJson Load(istream &str) {
 
 ExpectedJson Load(io::Reader &reader) {
 	auto str_ptr = reader.GetStream();
-	return Load(*(str_ptr.get()));
+	try {
+		return Load(*str_ptr);
+	} catch (njson::parse_error &e) {
```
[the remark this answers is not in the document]
kacf: I suppose we could catch `njson::exception` instead. It seems to be the top-level exception it uses, and it inherits from `std::exception`. Any higher exceptions that we should not catch I think, because it's not supposed to throw them, so if it happens, something is very wrong with the code, and it's better that we know about it.

This is pre-existing, so I can apply in a new commit.

mender-update/update_module/v3/platform/boost_filesystem/fs_operations.cpp
```diff
@@ -227,6 +250,159 @@ expected::ExpectedStringVector DiscoverUpdateModules(const conf::MenderConfig &c
...
+	auto opener = make_shared<AsyncFifoOpener>(download_.event_loop_);
+	download_.stream_next_opener_ = opener;
+	return opener->AsyncOpen(download_.stream_next_path_, open_handler);
+}
+
+Error UpdateModule::PrepareAndOpenStreamPipe(
+	const string &path, ExpectedWriterHandler open_handler) {
+	auto fs_path = fs::path(path);
+	boost::system::error_code ec;
+	if (!fs::create_directories(fs_path.parent_path(), ec) && ec) {
```
oleorhagen, line 275: Curiosity, I just learnt that `cpp` actually has an `and` keyword :exploding_head:

oleorhagen: Q: Can this ever fail without setting the error code?

mender-update/update_module/v3/platform/boost_filesystem/fs_operations.cpp
```diff
@@ -227,6 +250,159 @@ expected::ExpectedStringVector DiscoverUpdateModules(const conf::MenderConfig &c
...
+		return error::Error(
+			ec.default_error_condition(),
+			"Could not create stream directory at " + fs_path.parent_path().string());
```
oleorhagen, line 278: S: I would add the `error_code` message in here.

mender-update/update_module/v3/platform/boost_filesystem/fs_operations.cpp
```diff
@@ -227,6 +250,159 @@ expected::ExpectedStringVector DiscoverUpdateModules(const conf::MenderConfig &c
...
+	download_.current_stream_opener_ = opener;
+	return opener->AsyncOpen(path, open_handler);
+}
+
+Error UpdateModule::PrepareDownloadDirectory(const string &path) {
+	auto fs_path = fs::path(path);
+	boost::system::error_code ec;
+	if (!fs::create_directories(fs_path, ec) && ec) {
+		return error::Error(
+			ec.default_error_condition(), "Could not create `files` directory at " + path);
```
oleorhagen, line 298: Same as above

mender-update/update_module/v3/platform/boost_filesystem/fs_operations.cpp
```diff
@@ -227,6 +250,159 @@ expected::ExpectedStringVector DiscoverUpdateModules(const conf::MenderConfig &c
...
+	thread_ = thread([this, handler]() {
+		auto writer = make_shared<events::io::AsyncFileDescriptorWriter>(event_loop_);
+		// This will block for as long as there are no FIFO readers.
+		auto err = writer->Open(path_);
+
+		auto cancelled = cancelled_;
+		if (err != error::NoError) {
+			event_loop_.Post([handler, err, cancelled]() {
+				if (!*cancelled) {
+					// Needed because capture is always const, and `unexpected`
```
oleorhagen, line 367: I remember this conundrum :)

mender-update/update_module/v3/platform/boost_filesystem/fs_operations.cpp
```diff
@@ -227,6 +250,159 @@ expected::ExpectedStringVector DiscoverUpdateModules(const conf::MenderConfig &c
...
+	if (fd < 0) {
+		// Should not happen.
+		int errnum = errno;
+		log::Error(string("Cancel::open() returned error: ") + strerror(errnum));
+		assert(fd >= 0);
+	}
+
+	thread_.join();
+
+	::close(fd);
```
oleorhagen, line 403: Q: No need to care about errors here?

common/testing.cpp
```diff
@@ -38,6 +42,46 @@ std::string TemporaryDirectory::Path() {
...
+	json::Json expected_contents = json::Load(expected_content).value();
+	if (contents.Dump() == expected_contents.Dump()) {
+		return ::testing::AssertionSuccess();
+	}
+	return ::testing::AssertionFailure()
+		   << "Expected: '" << contents.Dump() << "' Got: '" << expected_contents.Dump() << "'";
+}
+
+::testing::AssertionResult FilesEqual(const string &filename1, const string &filename2) {
+	processes::Process proc({"diff", "-u", filename1, filename2});
```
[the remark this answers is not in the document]
kacf: For such extremely common options? `diff` doesn't print patches correctly without `-u/--unified`, it's practically part of the command name, and `git diff` doesn't even support non-unified mode. I think I vote against it in this case, since you have to use it in every instance, so it's better to know what it means. 🙂

I wouldn't use `-y` because it's harder to catch small differences when reading side by side, and I also suspect that Gitlab's log output is not wide enough to render them in a readable way.

common/testing.cpp
```diff
@@ -38,6 +42,46 @@ std::string TemporaryDirectory::Path() {
...
+	auto err = proc.Run();
+	if (err == error::NoError) {
+		return ::testing::AssertionSuccess();
+	}
+	auto result = ::testing::AssertionFailure() << filename1 << " and " << filename2 << " differ";
+	if (!result) {
```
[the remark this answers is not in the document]
kacf: Haha, good point. Will fix.

mender-update/update_module/v3/update_module.cpp
```diff
@@ -24,9 +26,29 @@ namespace v3 {
...
+	MenderContext &ctx,
+	artifact::Payload &payload,
+	artifact::PayloadHeaderView &payload_meta_data) :
+	ctx_(ctx),
+	payload_(payload),
+	payload_meta_data_(payload_meta_data) {
+	download_.buffer_.resize(MENDER_BUFSIZE);
+
+	update_module_path_ =
+		path::Join(conf::paths::DefaultModulesPath, payload_meta_data_.header.payload_type);
```
oleorhagen, line 41: Turned out nice I think :)

mender-update/update_module/v3/update_module.hpp
```diff
@@ -48,11 +54,14 @@ enum class RebootAction { No, Automatic, Yes };
 
 using ExpectedRebootAction = expected::expected<RebootAction, Error>;
 
+using ExpectedWriterHandler = function<void(io::ExpectedAsyncWriterPtr)>;
```
oleorhagen, line 57: Q: Should this really be declared here?

mender-update/update_module/v3/update_module.hpp
```diff
@@ -72,8 +81,62 @@ class UpdateModule {
...
+	Error PrepareAndOpenStreamPipe(const string &path, ExpectedWriterHandler open_handler);
+	Error PrepareDownloadDirectory(const string &path);
+	Error DeleteStreamsFiles();
+
+	void StartDownloadProcess();
+
+	void StreamNextOpenHandler(io::ExpectedAsyncWriterPtr writer);
+	void StreamOpenHandler(io::ExpectedAsyncWriterPtr writer);
+
+	void StreamNextWriteHandler(size_t expected_n, size_t written_n, error::Error err);
```
oleorhagen, line 95: Nit: Above you're just using `Error`, while here you namespace it

mender-update/update_module/v3/update_module.hpp
```diff
@@ -72,8 +81,62 @@ class UpdateModule {
...
+	void PayloadReadHandler(size_t n, error::Error err);
```
oleorhagen, line 96: Q: Could this not b econstants?

common/processes/platform/tiny_process_library/tiny_process_library.cpp
```diff
@@ -103,7 +103,24 @@ error::Error Process::Start(OutputCallback stdout_callback, OutputCallback stder
...
-int Process::Wait() {
+error::Error Process::Run() {
+	auto err = Start();
+	if (err != error::NoError) {
+		log::Error(err.String());
+	}
+	return Wait();
+}
+
+static error::Error ErrorBasedOnExitStatus(int exit_status) {
```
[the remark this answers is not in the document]
kacf: Why? This is the same, just shorter.

common/processes/platform/tiny_process_library/tiny_process_library.cpp
```diff
@@ -310,7 +310,7 @@ int Process::EnsureTerminated() {
 		log::Info("Sending SIGKILL to PID " + to_string(proc_->get_id()));
 		Kill();
 		result = future_exit_status_.wait_for(max_termination_time_);
-		if (result == future_status::ready) {
+		if (result != future_status::ready) {
```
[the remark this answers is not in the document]
kacf: I was wondering the same thing!

common/processes/platform/tiny_process_library/tiny_process_library.cpp
```diff
@@ -79,11 +79,17 @@ error::Error Process::Start(OutputCallback stdout_callback, OutputCallback stder
...
+	if (stdout_pipe_ >= 0 || stdout_callback) {
+		maybe_stdout_callback = ProcessReaderFunctor {stdout_pipe_, stdout_callback};
+	}
+	OutputCallback maybe_stderr_callback;
+	if (stderr_pipe_ >= 0 || stderr_callback) {
+		maybe_stderr_callback = ProcessReaderFunctor {stderr_pipe_, stderr_callback};
+	}
+
+	proc_ =
+		make_unique<tpl::Process>(args_, work_dir_, maybe_stdout_callback, maybe_stderr_callback);
```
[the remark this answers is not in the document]
kacf: Oh, good catch! Fixing.

common/events/events_io.cpp
kacf: You mean `Note: Need to...`. Does it have some special meaning?

mender-update/CMakeLists.txt
kacf: It's because of the `if(MENDER_USE_BOOST_FS)`, I wanted to keep it inside the `if`, while our generic packages can go outside. I suppose you could argue that we don't really know if they are all needed when we build for a different platform. But if we keep it outside it's less likely that we forget to include something and break the build on another platform later. Either method has it's pros and cons, so I'm not really strongly arguing for either.

In any case, `PUBLIC` is always needed because it causes other targets that link to this one to include its dependencies as well, which is required when linking.

mender-update/context.hpp
kacf: Yes, it's different. These are generally the recommendations:

* **Return const by reference**: Usually the best choice when you are returning a member variable, because you get the best performance, and the const prevents the caller from modifying your internals. If the caller assigns to a variable which is non-const, then a copy is made, and the caller won't even know that the return was originally a reference.
* **Return non-const by reference**: Also ok, if, and only if, the caller is allowed to modify the object directly. Useful for informational structs, for example, but still a far less common return type than const.
* **Return const by value**: Should never be used. Since it's by value, the object is owned by the caller, and hence it is a bad idea to enforce const, because you don't care what the caller does to it. And in fact, you cannot enforce const, because if the caller wants non-const, it will just copy-construct the object into a new non-const object. So you get an extra copy instead of move semantics. This is the one I pointed out in your PR earlier.
* **Return non-const by value**: The only appropriate choice when you are returning something you constructed in the function being called, in which case references cannot be used.

mender-update/update_module/v3/platform/boost_filesystem/fs_operations.cpp
kacf: > Curiosity, I just learnt that `cpp` actually has an `and` keyword exploding_head

I vaguely recall having heard about that, yes! Interestingly, it's not listed as part of any particular standard. Has it always been there..? :open_mouth: 

I'm fine with either, but `&&` is pretty established, so I'm not going on a replacement spree! 😄

> Q: Can this ever fail without setting the error code?

I don't think so. If so there is a bug in Boost. It's the same kind of scenario as with `errno`, I suppose. Library functions are supposed to set them, and we can't really do anything but trust them.

mender-update/update_module/v3/platform/boost_filesystem/fs_operations.cpp
kacf: Generally no. There is not much that can go wrong when closing a file handle, and the best we could do about it is a log entry. We have not typically handled this in Golang either, even though it also provides an error return.

mender-update/update_module/v3/update_module.hpp
kacf: I can't think of a different use case for it at the moment. I'd leave it here for now, at least until we identify that it's needed elsewhere.

mender-update/update_module/v3/update_module.hpp
kacf: Sure, but so could almost every variable. As long as they're not shared with anyone I think it's fine to omit const, and there's nothing inherently wrong with modifying them, sometimes it's even useful.

common/processes/platform/tiny_process_library/tiny_process_library.cpp
```diff
@@ -64,6 +66,19 @@ error::Error Process::Start(OutputCallback stdout_callback, OutputCallback stder
 		return MakeError(ProcessAlreadyStartedError, "Cannot start process");
 	}
 
+	// Tiny-process-library doesn't give a good error if the command isn't found (just returns
+	// exit code 1). If the path is absolute, it's pretty easy to check if it exists. This won't
+	// cover all errors (non-absolute or unset executable bit, for example), but helps a little,
+	// at least.
+	if (args_.size() > 0 && path::IsAbsolute(args_[0])) {
```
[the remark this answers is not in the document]
kacf: Because relative paths are resolved by the shell by looking through `PATH`, which I can't do here without doing it manually. And I wanted to keep it simple. This is, as mentioned, just a best effort for a slightly better error message, and not a critical thing.

mender-update/update_module/v3/update_module.hpp
kacf: Indeed. The plain `Error` is only used in this file, and I think I prefer the namespaced one for clarity (it's also what we use practically everywhere). So I reverted to namespaced.

common/events/events_io.cpp
kacf: I did even better actually: I moved the joining to the handler, so if it's joinable when we get here, it's an error which we can immediately report.

kacf:
@oleorhagen: Updated!

common/json/platform/nlohmann/nlohmann_json.cpp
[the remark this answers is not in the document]
kacf: @oleorhagen: Well, I need to eat my words here. Turns out JSON *does* throw other exceptions. Or rather, it doesn't catch exceptions from `std`. So on an open error, these escape out of the JSON library.

I'll change this in an upcoming commit.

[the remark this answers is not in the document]
oleorhagen: Good god - my spidey senses end up victorious this time!

common/processes/platform/tiny_process_library/tiny_process_library.cpp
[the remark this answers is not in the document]
oleorhagen: Have you checked the styleguide on this one (I haven't, just curious)

common/events/events_io.cpp
oleorhagen: Most editors have some special highlighting of `Notes` for instance. It's nice :)

kacf: Oh ok, I guess mine doesn't. But I will add this in an upcoming PR. 🙂

PR #1239, opened 2023-05-12
Various fixes, see commits for details

kacf:


artifact/v3/header/CMakeLists.txt
```diff
@@ -1,11 +1,6 @@
 
 # Test the parser
 add_executable(artifact_header_parser_test EXCLUDE_FROM_ALL
-  header.cpp
-  header_info.cpp
-  type_info.cpp
-  meta_data.cpp
-  ${CMAKE_SOURCE_DIR}/artifact/error.cpp
```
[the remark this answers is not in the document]
oleorhagen: Done with development, so doesn't matter much anymore

common/optional.hpp
```diff
@@ -15,6 +15,14 @@
...
 #define MENDER_COMMON_OPTIONAL_HPP
 
+// optional-lite is not binary compatible between C++ versions. This is important, since (as per
+// 2023-05) we build cross-platform files with C++11, and platform files with (optionally) a later
+// version. And then mix them.
+//
+// Luckily, it's possible to force optional-lite to stick with C++11 regardless of what the compiler
+// is using. We don't need any later features in this particular library, so just stick to C++11 all
+// the time.
+#define optional_CPLUSPLUS 201103L
```
oleorhagen, line 25: Q: Should this be a config in `CMake`? :thinking:

oleorhagen (approved):
Sweet!

Did you time it by any chance?

common/optional.hpp
kacf: I don't think we need it. This does not need to change as long as we don't need any post-C++11 features, which is entirely dependent on our code, not on user needs.

In about a decade or so, we can remove this entirely, and rely on C++23! 😄

kacf:
> Did you time it by any chance?

Overall the PR saves about 32 seconds of build time on my laptop.

PR #1243, opened 2023-05-30
feat(common): Add a crypto library

oleorhagen:
For now the library has the ability to sign data

Ticket: MEN-6465
Changelog: None


common/crypto/platform/openssl/crypto.cpp
```diff
@@ -0,0 +1,165 @@
...
+		EVP_PKEY_free(key);
+	}
+};
+auto bio_free_func = [](BIO *bio) {
+	if (bio) {
+		BIO_free(bio);
+	}
+};
+
+auto EncodeBase64(vector<uint8_t> to_encode) -> expected::ExpectedString {
```
kacf, line 62: Is there a reason to use the `auto` for top-level functions? I think we should try to stick to the standard style for best readability.

common/crypto_test.cpp
```diff
@@ -0,0 +1,44 @@
...
+namespace mender {
+namespace common {
+namespace crypto {
+
+TEST(CryptoTest, TestSign) {
+	string data_ {"foobar"};
+	vector<uint8_t> testdata {data_.begin(), data_.end()};
+	string private_key_file = "./private-key.pem";
+	auto expected_signature = crypto::Sign(private_key_file, testdata);
+	ASSERT_TRUE(expected_signature) << "Unexpected: " << expected_signature.error();
```
kacf, line 36: Should test one negative as well, or?

kacf: Ah, only now did I realize that you are not *verifying* the signature, just making it. My idea was to test an invalid signature, but I guess it doesn't make sense here since we don't verify on the client side. All good then.

kacf (approved):

PR #1244, opened 2023-06-01
Fix the builds for older debians

oleorhagen:
We had a couple of minor mishaps in the CMake files that need cleaning up.


kacf (approved):
👍

PR #1246, opened 2023-06-02
Enable initialization of our IO classes with shared pointers

vpodzime:
Important in the async code where instances of these classes outlive the scopes in which the objects they use where created/defined.


vpodzime:
Replaced by #1247

PR #1248, opened 2023-06-05
feat(json): Add a templated Dump function to dump STL structures to JSON

oleorhagen:
Note - the template still needs to be specialized for each invocation of the algorithm used.

Ticket: None
Changelog: None


kacf (approved):
Slightly different use case, but still, great minds [think alike](https://github.com/mendersoftware/mender/pull/1252/commits/4b0135781735e28ac1b4fb5a5bc587f8223a5957)!

vpodzime (approved):
My only concern is that implementing this with some other JSON "backend" might be quite tricky. But not exactly for these two specific cases. So mostly something to keep in mind for potential future expansions.

PR #1249, opened 2023-06-05
Move the streamreader function implementation to the .cpp file and add const argument

oleorhagen:
minor cleanups


common/io.hpp
```diff
@@ -155,11 +155,11 @@ class StringReader : virtual public Reader {
 	unique_ptr<StreamReader> reader_;
 
 public:
-	StringReader(string &str) :
+	StringReader(const string &str) :
 		s_ {str},
 		reader_ {new StreamReader(s_)} {
 	}
-	StringReader(string &&str) :
+	StringReader(const string &&str) :
```
kacf, line 154: I think const rvalue reference is wrong. It is supposed to modify the original, but the const prevents that. The `stringstream` constructor with rvalue reference is anyway C++20, so it wouldn't have been used anyway, but it is indeed non-const, and this needs to be non-const too, if it is ever to be used.

common/io.hpp
```diff
@@ -130,15 +130,7 @@ class StreamReader : virtual public Reader {
...
-	ExpectedSize Read(vector<uint8_t>::iterator start, vector<uint8_t>::iterator end) override {
-		is_->read(reinterpret_cast<char *>(&*start), end - start);
-		if (is_->bad()) {
-			int io_error = errno;
-			return expected::unexpected(
-				Error(std::generic_category().default_error_condition(io_error), ""));
-		}
-		return is_->gcount();
-	}
+	ExpectedSize Read(vector<uint8_t>::iterator start, vector<uint8_t>::iterator end) override;
```
kacf, line 133: 👍

kacf (changes requested):

common/io.hpp
oleorhagen: `stringstream` with `rvalue` ref is `C++11` according to `cppreference`: 

https://en.cppreference.com/w/cpp/io/basic_stringstream/basic_stringstream

And I'm not following, how is it supposed to modify the original? It's a reader?

kacf (approved):

PR #1250, opened 2023-06-05
Minor fixes, and changes to the sha implementation

oleorhagen:
I am also considering moving it to `common` now?

What do you think?

### DIFF
--- a/artifact/sha/platform/openssl/sha.cpp
+++ b/artifact/sha/platform/openssl/sha.cpp
@@ -98,7 +98,7 @@ expected::ExpectedSize Reader::Read(
 			return expected::unexpected(MakeError(
 				ShasumMismatchError,
 				"The checksum of the read byte-stream does not match the expected checksum, (expected): "
-					+ expected_sha_ + " (calculated): " + real_sha.value()));
+					+ expected_sha_ + " (calculated): " + real_sha.value().String()));
 		}
 		this->done_ = true;
 		this->shasum_ = real_sha.value();
@@ -112,7 +112,8 @@ expected::ExpectedSize Reader::Read(
 	return bytes_read.value();
 }
 
-expected::ExpectedString Reader::ShaSum() {
+
+ExpectedSHA Reader::ShaSum() {
 	if (!initialized_) {
 		return expected::unexpected(MakeError(
 			InitializationError,
@@ -136,12 +137,7 @@ expected::ExpectedString Reader::ShaSum() {
 			"SHA of unexpected length: " + std::to_string(hash_length) + " expected length: 32"));
 	}
 
-	std::stringstream ss {};
-	for (unsigned int i = 0; i < hash_length; ++i) {
-		ss << std::hex << std::setw(2) << std::setfill('0') << (int) hash[i];
-	}
-
-	return ss.str();
+	return SHA(hash);
 }
 
 } // namespace sha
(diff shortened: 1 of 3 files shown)


artifact/sha/sha.hpp
```diff
@@ -58,6 +60,43 @@ extern const ErrorCategoryClass KeyValueParserErrorCategory;
 
 error::Error MakeError(ErrorCode code, const string &msg);
 
+class SHA {
+public:
+	SHA() :
+		sha_ {} {};
+	SHA(vector<uint8_t> &v) :
+		sha_ {v} {};
+	operator string() const {
```
kacf, line 69: Especially since this does conversion, I would suggest using a dedicated `String()` function instead of automatic casting. We've already seen how automatic conversion can go wrong, so it might be wise to use it sparingly.

artifact/sha/sha.hpp
```diff
@@ -58,6 +60,43 @@ extern const ErrorCategoryClass KeyValueParserErrorCategory;
...
+		std::stringstream ss {};
+		for (unsigned int i = 0; i < 32; ++i) {
+			ss << std::hex << std::setw(2) << std::setfill('0') << (int) sha_.at(i);
+		}
+		return ss.str();
+	}
+	operator vector<uint8_t>() const {
```
kacf, line 76: For similarity with `std`, I would suggest the same here. Maybe `Data()`?

kacf (commented):
> I am also considering moving it to `common` now?

This is needed for the auth signing request, I take it? If so, yes. If not, it can be left here until we need it, I think.

artifact/sha/platform/openssl/sha.cpp
```diff
@@ -94,7 +94,7 @@ expected::ExpectedSize Reader::Read(
 		if (!real_sha) {
 			return expected::unexpected(real_sha.error());
 		}
-		if (real_sha.value() != expected_sha_) {
+		if (expected_sha_.size() > 0 and real_sha.value() != expected_sha_) {
```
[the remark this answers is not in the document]
oleorhagen: Use case here is for passing the sha to `openssl`. In this case we don't know the `sha` upfront obviously. In this instance this is a strict sha generator.

artifact/sha/sha.hpp
oleorhagen: Hmm, and then wrap `vector<uint8_t>.data()` you mean?

artifact/sha/sha.hpp
oleorhagen: Hmm, but in this case I want the object to be transparently dual.

I want it to cast to `string` when required, and `vector` when required.

kacf: But why do you need it to be transparent? They don't even have the same content. The string conversion converts it to a human-readable string, so it's not really the same object anymore. I think this should be explicit.

Your `string(variable)` would just change to `variable.String()`, but it has the benefit of warning you if you start mixing types (which you probably don't want).

artifact/sha/sha.hpp
kacf: Actually I was thinking

```
vector<uint8_t> Data() const
```

But maybe that's confusing since `vector.data()` already exists. Is `Vector()` a better name?

artifact/sha/sha.hpp
oleorhagen: I liked it tbh, but it's fine. both work. I give

artifact/sha/sha.hpp
kacf: Keep in mind that for example smart pointers, which are also doing this (behaving like pointers), they require you to call `get()` to get the view of the real, raw pointer. It's makes it explicit and searchable.

However, I won't oppose it, if you want to keep it as is. What you're saying makes sense, I just have a preference for being explicit (I think Golang reinforced that habit!).

kacf (approved):

artifact/sha/sha.hpp
oleorhagen: Noted, keeping it as it is for now though

PR #1251, opened 2023-06-05
fix(http): Set the HOST header for all HTTP requests

oleorhagen:
The AWS Loadbalancer will not route requests to our k8s cluster unless the `HOST` header is set in the request.

Therefore, this fix sets the header for all outgoing HTTP requests.

Ticket: None
Changelog: None

### DIFF
--- a/common/http_test.cpp
+++ b/common/http_test.cpp
@@ -1326,8 +1326,8 @@ TEST(HTTPSTest, CorrectDefaultCertificateStoreVerification) {
 		[&client_hit_header](http::ExpectedIncomingResponsePtr exp_resp) {
 			ASSERT_TRUE(exp_resp) << "Error message: " << exp_resp.error().String();
 			auto resp = exp_resp.value();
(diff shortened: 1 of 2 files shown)


common/http_test.cpp
```diff
@@ -1326,8 +1326,8 @@ TEST(HTTPSTest, CorrectDefaultCertificateStoreVerification) {
 		[&client_hit_header](http::ExpectedIncomingResponsePtr exp_resp) {
 			ASSERT_TRUE(exp_resp) << "Error message: " << exp_resp.error().String();
 			auto resp = exp_resp.value();
-			EXPECT_EQ(resp->GetStatusCode(), 200);
-			EXPECT_EQ(resp->GetStatusMessage(), "OK");
+			EXPECT_EQ(resp->GetStatusCode(), 301);
+			EXPECT_EQ(resp->GetStatusMessage(), "Moved Permanently");
```
kacf, line 1330: I don't think we use redirects in any client responses, but I asked in our chat, just in case.

kacf (approved):

PR #1252, opened 2023-06-05
Implement standalone mode

kacf:
Tickets [MEN-6391](https://northerntech.atlassian.net/browse/MEN-6391) and [MEN-6392](https://northerntech.atlassian.net/browse/MEN-6392)


kacf:
On @oleorhagen's request, singled out all the support commits to #1253.

mender-update/update_module/v3/update_module.hpp
```diff
@@ -82,7 +82,9 @@ static std::string StateString[] = {
 	"Cleanup"};
 
 static inline std::string StateToString(State state) {
-	static_assert(sizeof(StateString) / sizeof(*StateString) == static_cast<int>(State::LastState));
+	static_assert(
+		sizeof(StateString) / sizeof(*StateString) == static_cast<int>(State::LastState),
+		"Make sure to keep State and StateString in sync!");
```
vpodzime, line 87: Why not simply do `static std::string StateString[LastState + 1]` like I suggested before?

mender-update/update_module/v3/update_module.hpp
```diff
@@ -161,12 +159,14 @@ class UpdateModule {
...
 
 	context::MenderContext &ctx_;
-	artifact::Payload &payload_;
-	artifact::PayloadHeaderView &payload_meta_data_;
 	string update_module_path_;
 	string update_module_workdir_;
 
-	struct {
+	class DownloadData {
+	public:
```
vpodzime, line 168: What is a class with no function members and public-only members good for? To me it's just confusing that it's not a `struct`.

Also, this change seems to be a bit unrelated to the commit message.

common/events_io.hpp
```diff
@@ -124,6 +124,21 @@ class AsyncWriterFromWriter : virtual public mio::AsyncWriter {
...
+class ReaderFromAsyncReader : virtual public mio::Reader {
+public:
+	// Since the async reader usually needs the event loop object, we need to use this factory
+	// function to create the async reader inside of the constructor.
+	ReaderFromAsyncReader(AsyncReaderFromEventLoopFunc func);
+
+	mio::ExpectedSize Read(vector<uint8_t>::iterator start, vector<uint8_t>::iterator end) override;
+
+private:
+	EventLoop event_loop_;
```
vpodzime, line 140: Where does `event_loop_` come from? I don't see any constructor here and the factory function seems to be relying on it.

common/events_io.hpp
```diff
@@ -124,6 +124,21 @@ class AsyncWriterFromWriter : virtual public mio::AsyncWriter {
 	EventLoop &loop_;
 };
 
+using AsyncReaderFromEventLoopFunc = function<mio::AsyncReaderPtr(EventLoop &loop)>;
+
+class ReaderFromAsyncReader : virtual public mio::Reader {
+public:
+	// Since the async reader usually needs the event loop object, we need to use this factory
+	// function to create the async reader inside of the constructor.
+	ReaderFromAsyncReader(AsyncReaderFromEventLoopFunc func);
```
vpodzime, line 135: Shouldn't a factory function be `static`?

common/error.cpp
```diff
@@ -48,6 +48,19 @@ std::ostream &operator<<(std::ostream &os, const Error &err) {
 	return os;
 }
 
+Error Error::FollowedBy(const Error &err) const {
```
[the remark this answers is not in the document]
vpodzime: Because this is not Go and `FollowedBy()` makes more sense.

common/json.hpp
```diff
@@ -102,6 +102,9 @@ class Json {
 	ExpectedDouble GetDouble() const;
 	ExpectedBool GetBool() const;
 
+	template <typename T>
+	expected::expected<T, error::Error> Get() const;
```
vpodzime, line 106: We should add at least some trivial unit test. If nothing else than to just avoid regressions in the future.

mender-update/context/context.cpp
```diff
@@ -37,6 +37,16 @@ namespace json = mender::common::json;
 namespace kv_db = mender::common::key_value_database;
 namespace path = mender::common::path;
 
+const string MenderContext::artifact_name_key {"artifact-name"};
```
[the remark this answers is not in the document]
vpodzime: Yeah, same question here. I even believe I tried this and there was some problem with it. Did you try adding them as `extern` to the `.hpp` file?

mender-update/update_module/v3/platform/boost_filesystem/fs_operations.cpp
```diff
@@ -128,6 +128,13 @@ error::Error UpdateModule::PrepareFileTree(
 
 	const fs::path file_tree_path {path};
 
+	boost::system::error_code ec;
+	fs::remove_all(file_tree_path, ec);
```
vpodzime, line 132: There was some reason why I avoided using the versions that take the `error_code` and relied on exceptions instead. But I don't recall what it was, something mentioned in the documentation. Please double-check it.

vpodzime (commented):
Went through all but the last commit.

common/events/events_io.cpp
```diff
@@ -123,6 +123,32 @@ void AsyncWriterFromWriter::Cancel() {
 	}
 }
 
+ReaderFromAsyncReader::ReaderFromAsyncReader(AsyncReaderFromEventLoopFunc func) {
+	reader_ = func(event_loop_);
```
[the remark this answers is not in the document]
kacf: I'm thinking that once you start executing code, it should be in implementation files, leaving initializer lists for true constant defaults. That's by no means a hard rule though, I just thought it made sense.

I did however, move it to the constructor initializer section just now (locally).

common/events/events_io.cpp
```diff
@@ -123,6 +123,32 @@ void AsyncWriterFromWriter::Cancel() {
...
+}
+
+mio::ExpectedSize ReaderFromAsyncReader::Read(
+	vector<uint8_t>::iterator start, vector<uint8_t>::iterator end) {
+	size_t read;
+	error::Error err;
```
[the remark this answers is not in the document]
kacf: You're right, I even got confused when I went back now to read it. Changed it to use a dedicated `inner_err` variable which is now captured instead of the `err`.

common/error.cpp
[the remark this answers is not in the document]
kacf: Yes, that was my thinking too. Also, it is not exactly the same as `Wrap`, because `Wrap` encapsulates the whole error, allowing you to unwrap and get the code of the wrapped error(s). `FollowedBy` only applies to the message, and only the outermost code is available for inspection.

common/error.hpp
```diff
@@ -102,6 +103,13 @@ extern const CommonErrorCategoryClass CommonErrorCategory;
 
 Error MakeError(ErrorCode code, const std::string &msg);
 
+// Some parts of standard C++, such as regex, require exceptions. Use this function in cases where
+// an exception cannot be avoided; it will automatically catch exceptions thrown by the given
+// function, and produce an error from it. On platforms where exceptions are disabled, any attempt
+// to throw them will abort instead, so if you use this function, do your utmost to avoid an
+// exception ahead of time (make sure the regex is valid, for instance).
+Error ExceptionToErrorOrAbort(std::function<void()> func);
```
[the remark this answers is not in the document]
kacf: Yes, but I'm assuming people are using lambdas, so they can return anything they want by capture.

common/json/json.cpp
```diff
@@ -49,6 +49,36 @@ error::Error MakeError(JsonErrorCode code, const string &msg) {
...
+	return ToStringVector(*this);
+}
+
+template <>
+expected::expected<string, error::Error> Json::Get<string>() const {
+	return GetString();
+}
+
+template <>
+expected::expected<int, error::Error> Json::Get<int>() const {
```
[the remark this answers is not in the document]
kacf: Ooh, you're right! Fixed.

mender-update/context/context.cpp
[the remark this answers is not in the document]
kacf: To avoid allocating a new instance every time `MenderContext` is initialized. They are all const, so we might as well allocate them once and be done with it. `extern` would probably work, but it's redundant in this case (it is automatically assumed for static class members).

If you're asking about the initializer in the hpp file, then the reason I had to move them is that they don't work for static members. Keeping them would of course be nice for readability, but it's a language limitation.

mender-update/update_module/v3/update_module.hpp
kacf: Unfortunately it allows defining fewer strings than you have allocated, which means some entries could be unintentionally empty. So my check is stricter in that sense, because the count has to be exact.

Unrelated to that, I also realized that since this is `static` and in a header file, we are re-defining and re-storing it in the binary for each file that includes it. So I also removed `static` and moved the list and `StateToString` to the cpp file to save space.

mender-update/update_module/v3/update_module.hpp
kacf: > What is a class with no function members and public-only members good for? To me it's just confusing that it's not a `struct`.

Hmm, yes, I see your point. This is a bit of a corner case. Even though the members are public in this class, they are private as seen from the outside, because of the outer `private` section. It's just that the implementation is handled by the outer class.

Also, it does have a constructor, which is non-trivial (though in the smallest possible sense).

I don't know, I don't have a strong leaning. It seems marginally more correct to me to use `class` if it has at least one function (the constructor), but I could go either way.

> Also, this change seems to be a bit unrelated to the commit message.

Well, it is related since it triggered the introduction of a constructor. But indeed, it's not immediately obvious, so I'll add a note about that in the message.

common/events_io.hpp
kacf: It's just a default-constructed `EventLoop`, which is independent of the main application event loop. This is specifically to make this `Reader` do only one thing: Read from the `AsyncReader`.

But this is indeed quite unobvious. I'll document it next to the `event_loop_` variable.

common/events_io.hpp
kacf: It is the `AsyncReaderFromEventLoopFunc` which is a factory function, not the `ReaderFromAsyncReader`. I clarified this in the comment.

mender-update/update_module/v3/platform/boost_filesystem/fs_operations.cpp
kacf: Hmm, I can't see anything obvious. You mean [this documentation](https://www.boost.org/doc/libs/1_74_0/libs/filesystem/doc/reference.html#remove_all), right? I checked the "Error reporting" section as well.

common/json.hpp
kacf: 👍

common/error.cpp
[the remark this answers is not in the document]
oleorhagen: Alright, I see. `FollowedBy` sounds a little odd to me, but I can't think of anything better either.

mender-update/context/context.cpp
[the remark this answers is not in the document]
oleorhagen: I see makes sense. However, in the grand scheme of things, what difference does this optimization really make?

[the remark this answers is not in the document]
kacf: Not much, since `MenderContext` is typically only constructed once. However, since they are now globals you don't need a `MenderContext` instance at all to refer to them. I'm using this in several places, for example `MenderContext::standalone_state_key`.

mender-update/update_module/v3/update_module.hpp
vpodzime: > Unfortunately it allows defining fewer strings than you have allocated, which means some entries could be unintentionally empty.

Huh? The rest is just silently initialized to `""` or what? No warnings at all?

mender-update/update_module/v3/update_module.hpp
vpodzime: > Which do you prefer?
I prefer `struct` for "passive" objects.

> I'll add a note about that in the message.
Thanks!

mender-update/context/context.cpp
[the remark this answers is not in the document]
vpodzime: > If you're asking about the initializer in the hpp file, then the reason I had to move them is that they don't work for static members. Keeping them would of course be nice for readability, but it's a language limitation.

Right, that's the thing I was hitting and how I ended up putting them into the class as normal members.

mender-update/update_module/v3/platform/boost_filesystem/fs_operations.cpp
vpodzime: Hmm, don't see anything obvious either, but I do recall reading those docs and going with exceptions after that. :sweat_smile: We'll see. :)

mender-update/update_module/v3/update_module.hpp
kacf: Nope, nothing. Just silent empty strings.

vpodzime:
> If you are ok with all the commits before the last massive one, please let me know, and I can merge it separately in #1253.

Added some more (new) comments there. Let's deal with the fixes/improvements first in that PR.

mender-update/update_module/v3/update_module.hpp
kacf: Alright, done!

That means I also removed the message I just added, since now there is no change to that part.

kacf:
@vpodzime, @oleorhagen: All support commits merged, now only the big one remains!

mender-update/context/context.cpp
```diff
@@ -101,27 +118,37 @@ kv_db::KeyValueDatabase &MenderContext::GetMenderStoreDB() {
 }
 
 ExpectedProvidesData MenderContext::LoadProvides() {
+	ExpectedProvidesData data;
+	auto err = mender_store_.ReadTransaction([this, &data](kv_db::Transaction &txn) {
+		data = LoadProvides(txn);
+		if (!data) {
+			return data.error();
+		} else {
```
oleorhagen, line 126: Nit: Unnecessary else

mender-update/context/context.cpp
```diff
@@ -196,54 +223,147 @@ expected::ExpectedString MenderContext::GetDeviceType() {
...
+	// Now add the provides from the new_provides set.
+	for (auto provide : new_provides) {
+		to_modify[provide.first] = provide.second;
 	}
 
-	auto commit_data_to_db = [&](kv_db::Transaction &txn) {
+	return error::NoError;
+}
+
+error::Error MenderContext::CommitArtifactData(
```
oleorhagen, line 275: Holy moly...

oleorhagen: I don't have the stomach for this one today

mender-update/standalone/standalone.cpp
```diff
@@ -0,0 +1,516 @@
...
+			Result::NoUpdateInProgress,
+			context::MakeError(context::NoUpdateInProgressError, "Cannot commit")};
+	}
+
+	update_module::UpdateModule update_module(main_context, data.payload_types[0]);
+
+	return DoCommit(main_context, data, update_module);
+}
+
+ResultAndError Rollback(context::MenderContext &main_context) {
```
oleorhagen, line 291: Observation:

Usually you use your whitespace for grouping, but in this function it's kinda random?

mender-update/standalone/standalone.cpp
```diff
@@ -0,0 +1,516 @@
...
+	StandaloneData data;
+	auto exp_in_progress = LoadStandaloneData(main_context.GetMenderStoreDB(), data);
+	if (!exp_in_progress) {
+		return {Result::FailedNothingDone, exp_in_progress.error()};
+	}
+
```
oleorhagen, line 297: Like here

mender-update/standalone/standalone.cpp
```diff
@@ -0,0 +1,516 @@
...
+	update_module::UpdateModule &update_module) {
+	auto exp_rollback_support = update_module.SupportsRollback();
+	if (!exp_rollback_support) {
+		return {Result::NoRollback, exp_rollback_support.error()};
+	} else {
+		if (exp_rollback_support.value()) {
+			auto err = update_module.ArtifactRollback();
+			if (err == error::NoError) {
+				return {Result::RolledBack, error::NoError};
+			} else {
```
oleorhagen, line 438: same

mender-update/standalone/standalone.cpp
```diff
@@ -0,0 +1,516 @@
...
+				return {Result::RollbackFailed, err};
+			}
+		} else {
```
oleorhagen, line 441: same

mender-update/update_module/v3/update_module.cpp
```diff
@@ -51,9 +51,9 @@ std::string StateToString(State state) {
 
 UpdateModule::UpdateModule(MenderContext &ctx, const string &payload_type) :
 	ctx_(ctx) {
-	update_module_path_ = path::Join(conf::paths::DefaultModulesPath, payload_type);
+	update_module_path_ = path::Join(ctx.modules_path, payload_type);
 	update_module_workdir_ =
-		path::Join(conf::paths::DefaultModulesWorkPath, "payloads", "0000", "tree");
+		path::Join(ctx.modules_work_path, "modules", "v3", "payloads", "0000", "tree");
```
oleorhagen, line 56: Q: Why change this? I feel like the default paths read a lot better?

oleorhagen (commented):
Phew, did a go through, I will do some more on monday

mender-update/standalone.hpp
```diff
@@ -0,0 +1,131 @@
...
+
+namespace artifact = mender::artifact;
+
+namespace context = mender::update::context;
+
+namespace update_module = mender::update::update_module::v3;
+
+// The keys and data, respectively, of the JSON object living under the `standalone_data_key` entry
+// in the database. Be sure to take into account upgrades when changing this.
+struct StandaloneDataKeys {
```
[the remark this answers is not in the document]
kacf: Yes, that is a good point... However removing it results in `StandaloneData` becoming just `Data`, which, even though we are inside the `standalone` namespace, becomes a bit ambiguous, I think.

I could call it `UpdateData` maybe.. But is it better?

I think I would vote for leaving it as is, unless you have a better suggestion than just `Data`.

mender-update/standalone/standalone.cpp
```diff
@@ -0,0 +1,516 @@
...
+
+const string StandaloneDataKeys::version {"Version"};
+const string StandaloneDataKeys::artifact_name {"ArtifactName"};
+const string StandaloneDataKeys::artifact_group {"ArtifactGroup"};
+const string StandaloneDataKeys::artifact_provides {"ArtifactTypeInfoProvides"};
+const string StandaloneDataKeys::artifact_clears_provides {"ArtifactClearsProvides"};
+const string StandaloneDataKeys::payload_types {"PayloadTypes"};
+
+template <typename T>
+expected::expected<T, error::Error> GetEntry(
```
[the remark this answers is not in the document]
kacf: Good idea! Will do.

mender-update/standalone/standalone.cpp
kacf: I think you have to be more specific. It's not random to me.. Just clumping some logical steps together, while separating others which are a bit more unrelated.

mender-update/standalone/standalone.cpp
```diff
@@ -0,0 +1,516 @@
...
+}
+
+ResultAndError DoRollback(
+	context::MenderContext &main_context,
+	StandaloneData &data,
+	update_module::UpdateModule &update_module) {
+	auto exp_rollback_support = update_module.SupportsRollback();
+	if (!exp_rollback_support) {
+		return {Result::NoRollback, exp_rollback_support.error()};
+	} else {
```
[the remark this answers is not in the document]
kacf: Done. I kept one unnecessary else for the `RolledBack`/`NoRollback` pair. Since neither is a "preferred" path, it makes sense to keep them at the same level.

mender-update/update_module/v3/update_module.cpp
kacf: Yes, but the tests need to be able to change it using the `ctx`. While we could change the global default, I think it is better to keep changes localized, even if readability suffers a little.

kacf:
Updated!

mender-update/standalone.hpp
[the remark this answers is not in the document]
oleorhagen: Keeping it is fine. But to be honest, I think `Data` is best. It's what namespaces are for, imo. If we are in `namespace <namespace>`, then `Data` should be one and one thing only.

But just my opinion.

It also make for clean usage from the outside.

`namespace::Data`. There's not much ambiguity there

mender-update/standalone/standalone.cpp
oleorhagen: Maybe I just misunderstood it then :)

mender-update/update_module/v3/update_module.cpp
oleorhagen: Hmm, how exactly are we changing them in the tests? 
During runtime? :thinking:

kacf: Yes, they are being changed [here](https://github.com/mendersoftware/mender/pull/1252/files#diff-2fdfef8757e1cd478baf77bc0e427559352897984a53f53c18825b17ea3758fcR236) by most of the tests. Otherwise it tries to look in `/usr/share/mender`, which won't work while testing.

mender-update/standalone.hpp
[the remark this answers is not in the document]
kacf: Alright, fair enough, I changed it!

kacf:
@oleorhagen: Updated!

oleorhagen (approved):
Beautiful!

First milestone done :tada: :smile: :smile_cat:

mender-update/standalone.hpp
```diff
@@ -0,0 +1,125 @@
...
+// The keys and data, respectively, of the JSON object living under the `standalone_data_key` entry
+// in the database. Be sure to take into account upgrades when changing this.
+struct DataKeys {
+	static const string version;
+	static const string artifact_name;
+	static const string artifact_group;
+	static const string artifact_provides;
+	static const string artifact_clears_provides;
+	static const string payload_types;
+};
```
vpodzime, line 59: Maybe this could be just a singleton variable and not a type?

mender-update/standalone/standalone.cpp
```diff
@@ -0,0 +1,498 @@
...
+	if (!exp_bytes) {
+		auto &err = exp_bytes.error();
+		if (err.code == database::MakeError(database::KeyError, "").code) {
+			return optional::optional<Data>();
+		} else {
+			return expected::unexpected(err);
+		}
+	}
+
+	auto exp_json = json::Load(string(exp_bytes.value().begin(), exp_bytes.value().end()));
```
vpodzime, line 51: We have `common::StringFromByteVector()`

mender-update/standalone/standalone.cpp
```diff
@@ -0,0 +1,498 @@
...
+			ss << "\"" << elem << "\"";
+			first = false;
+		}
+		ss << "]";
+	}
+
+	ss << "}";
+
+	string strdata = ss.str();
+	vector<uint8_t> bytedata(strdata.begin(), strdata.end());
```
vpodzime, line 185: `common::ByteVectorFromString()`

mender-update/standalone/standalone.cpp
```diff
@@ -0,0 +1,498 @@
...
+
+	return db.Write(context::MenderContext::standalone_state_key, bytedata);
+}
+
+error::Error RemoveData(database::KeyValueDatabase &db) {
```
vpodzime, line 190: *RemoveStateData()*?

mender-update/standalone/standalone.cpp
```diff
@@ -0,0 +1,498 @@
...
+	}
+
+	if (src.find("http://") == 0 || src.find("https://") == 0) {
+		return {
+			Result::FailedNothingDone,
+			error::Error(make_error_condition(errc::not_supported), "HTTP not supported yet")};
+	}
+
+	auto stream = make_shared<ifstream>(src);
+	if (!stream->good()) {
```
vpodzime, line 216: Would be nice to use `OpenIfstream()`.

mender-update/standalone/standalone.cpp
```diff
@@ -0,0 +1,498 @@
...
+	update_module::UpdateModule update_module(main_context, header.header.payload_type);
+
+	auto err = update_module.PrepareFileTree(update_module.GetUpdateModuleWorkDir(), header);
+	if (err != error::NoError) {
+		err = err.FollowedBy(update_module.Cleanup());
+		return {Result::FailedNothingDone, err};
+	}
+
+	Data data;
+	DataFromPayloadHeaderView(header, data);
```
vpodzime, line 249: Shouldn't `DataFromPayloadHeaderView` return `Data`?

common/json/json.cpp
```diff
@@ -147,6 +147,37 @@ ExpectedKeyValueMap ToKeyValuesMap(const json::Json &j) {
...
+	const json::Json &json, const string &key, MissingOk missing_ok);
+template expected::expected<string, error::Error> Get(
+	const json::Json &json, const string &key, MissingOk missing_ok);
+template expected::expected<int64_t, error::Error> Get(
+	const json::Json &json, const string &key, MissingOk missing_ok);
+template expected::expected<double, error::Error> Get(
+	const json::Json &json, const string &key, MissingOk missing_ok);
+template expected::expected<bool, error::Error> Get(
+	const json::Json &json, const string &key, MissingOk missing_ok);
+
```
vpodzime, line 180: Just for a confirmation/clarification -- all these above are so that you can write `Get<string>()` instead of `GetString()`? I feel like I may be missing the point here.

vpodzime (commented):
Late to the party, but still some comments I wanted to share.

mender-update/standalone.hpp
kacf: Doesn't really make much difference though, does it?

mender-update/standalone/standalone.cpp
oleorhagen: +1

mender-update/standalone/standalone.cpp
kacf: StateData, of course! @oleorhagen were discussing what to call it, and we didn't think of this simple name! 😂

mender-update/standalone.hpp
vpodzime: No, it doesn't. It just looks a bit weird to me to create an instance of a type only to have access to 4 constants.

common/json/json.cpp
kacf: Uhm, no..? There is exactly one implementation, above these six instantiations.

PR #1253, opened 2023-06-06
Standalone mode support commits

kacf:
On @oleorhagen's request, singled out all the support commits from #1252 to here.

### DIFF
--- a/common/error.cpp
+++ b/common/error.cpp
@@ -48,6 +48,19 @@ std::ostream &operator<<(std::ostream &os, const Error &err) {
 	return os;
 }
 
+Error Error::FollowedBy(const Error &err) const {
+	if (*this == NoError) {
+		return err;
+	}
+	if (err == NoError) {
+		return *this;
+	}
+
+	Error new_err {*this};
+	new_err.message += "; Then followed error: " + err.String();
+	return new_err;
+}
+
 } // namespace error
 } // namespace common
 } // namespace mender
--- a/common/events/events_io.cpp
+++ b/common/events/events_io.cpp
@@ -123,8 +123,8 @@ void AsyncWriterFromWriter::Cancel() {
 	}
 }
 
-ReaderFromAsyncReader::ReaderFromAsyncReader(AsyncReaderFromEventLoopFunc func) {
-	reader_ = func(event_loop_);
+ReaderFromAsyncReader::ReaderFromAsyncReader(AsyncReaderFromEventLoopFunc func) :
+	reader_(func(event_loop_)) {
 }
 
 mio::ExpectedSize ReaderFromAsyncReader::Read(
@@ -145,9 +145,8 @@ mio::ExpectedSize ReaderFromAsyncReader::Read(
 
 	if (inner_err != error::NoError) {
 		return expected::unexpected(inner_err);
-	} else {
-		return read;
 	}
+	return read;
 }
 
 } // namespace io
--- a/common/events/events_io.cpp
+++ b/common/events/events_io.cpp
@@ -123,6 +123,33 @@ void AsyncWriterFromWriter::Cancel() {
 	}
 }
 
+ReaderFromAsyncReader::ReaderFromAsyncReader(AsyncReaderFromEventLoopFunc func) {
+	reader_ = func(event_loop_);
+}
+
+mio::ExpectedSize ReaderFromAsyncReader::Read(
+	vector<uint8_t>::iterator start, vector<uint8_t>::iterator end) {
+	size_t read;
+	error::Error err;
+	error::Error inner_err;
+	err = reader_->AsyncRead(start, end, [this, &read, &inner_err](size_t n, error::Error err) {
+		read = n;
+		inner_err = err;
+		event_loop_.Stop();
+	});
+	if (err != error::NoError) {
+		return expected::unexpected(err);
+	}
+
+	event_loop_.Run();
+
+	if (inner_err != error::NoError) {
+		return expected::unexpected(inner_err);
+	} else {
+		return read;
+	}
+}
+
 } // namespace io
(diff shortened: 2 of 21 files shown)


mender-update/update_module/v3/update_module.hpp
```diff
@@ -82,7 +82,9 @@ static std::string StateString[] = {
 	"Cleanup"};
 
 static inline std::string StateToString(State state) {
-	static_assert(sizeof(StateString) / sizeof(*StateString) == static_cast<int>(State::LastState));
+	static_assert(
+		sizeof(StateString) / sizeof(*StateString) == static_cast<int>(State::LastState),
+		"Make sure to keep State and StateString in sync!");
```
vpodzime: I'd still (also) do the `StateString[LastState + 1]` thing, but YMMV.

common/events/events_io.cpp
```diff
@@ -123,6 +123,32 @@ void AsyncWriterFromWriter::Cancel() {
...
+	reader_ = func(event_loop_);
+}
+
+mio::ExpectedSize ReaderFromAsyncReader::Read(
+	vector<uint8_t>::iterator start, vector<uint8_t>::iterator end) {
+	size_t read;
+	error::Error err;
+	err = reader_->AsyncRead(start, end, [this, &read, &err](size_t n, error::Error inner_err) {
+		read = n;
+		err = inner_err;
```
vpodzime: This double-assignment to `err` is a bit confusing. I understand it does the right thing because we don't quite care if `AsyncRead()` itself failed or if it called the handler with an error, but I'd still use two separate variables for it.

common/events/events_io.cpp
```diff
@@ -123,6 +123,32 @@ void AsyncWriterFromWriter::Cancel() {
 	}
 }
 
+ReaderFromAsyncReader::ReaderFromAsyncReader(AsyncReaderFromEventLoopFunc func) {
```
vpodzime: So this magic is to avoid implementation of a synchronous reader for what we already have an async reader for?

common/error.cpp
```diff
@@ -48,6 +48,19 @@ std::ostream &operator<<(std::ostream &os, const Error &err) {
...
+Error Error::FollowedBy(const Error &err) const {
+	if (*this == NoError) {
+		return err;
+	}
+	if (err == NoError) {
+		return *this;
+	}
+
+	Error new_err {*this};
+	new_err.message += ": Then followed error: " + err.String();
```
vpodzime, line 60: I'd use `;` to separate the error messages because `:` often shows up in the messages themselves.

common/events/events_io.cpp
```diff
@@ -131,9 +131,10 @@ mio::ExpectedSize ReaderFromAsyncReader::Read(
 	vector<uint8_t>::iterator start, vector<uint8_t>::iterator end) {
 	size_t read;
 	error::Error err;
-	err = reader_->AsyncRead(start, end, [this, &read, &err](size_t n, error::Error inner_err) {
+	error::Error inner_err;
+	err = reader_->AsyncRead(start, end, [this, &read, &inner_err](size_t n, error::Error err) {
```
vpodzime, line 135: I'd squash this with the commit that introduces the confusing version.

common/events/events_io.cpp
```diff
@@ -123,8 +123,8 @@ void AsyncWriterFromWriter::Cancel() {
 	}
 }
 
-ReaderFromAsyncReader::ReaderFromAsyncReader(AsyncReaderFromEventLoopFunc func) {
-	reader_ = func(event_loop_);
+ReaderFromAsyncReader::ReaderFromAsyncReader(AsyncReaderFromEventLoopFunc func) :
+	reader_(func(event_loop_)) {
```
vpodzime, line 127: This is actually potentially quite a big change, right?

vpodzime (commented):
A few more (mostly minor) comments.

common/events/events_io.cpp
vpodzime: see my comment below about squashing the improvement for this

mender-update/update_module/v3/update_module.hpp
```diff
@@ -127,6 +127,8 @@ class UpdateModule {
 	error::Error ArtifactFailure();
 	error::Error Cleanup();
 
+	static error::Error GetProcessError(const error::Error &err);
```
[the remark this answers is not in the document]
kacf: It's not just a getter, there is some logic in there too.

common/events/events_io.cpp
kacf: Will do!

common/events/events_io.cpp
kacf: Isn't it doing the same thing?

mender-update/update_module/v3/update_module.hpp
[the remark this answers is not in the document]
vpodzime: Could still be `inline`, right?

common/events/events_io.cpp
vpodzime: Constructing a default object and then assigning to it vs. copy-constructing an object? I mean, in this case it's (probably?) the same.

mender-update/update_module/v3/update_module.hpp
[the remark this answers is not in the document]
kacf: Not while it's in the cpp file. Which it has to be because the list is now there, and the list is static.

common/events/events_io.cpp
kacf: It really should be, otherwise we are creating problems for ourselves.

vpodzime (approved):
ACK

PR #1254, opened 2023-06-06
chore: Construct IO objects from shared_ptr passed by value

vpodzime:
A follow-up to bc4061daf8b725db94433227a11d00e46d89bc05 and 1bb88cdf8007a162b8a8ec3ae668f4b1a0c1b34c that added constructors from shared pointers passed by reference. If we capture shared pointers by copy, they cannot be passed further as references and thus we need constructors from shared pointers passed by value. However, having both variants (value and reference) makes the constructors ambiguous so we have to only provide the by-value constructors.

Ticket: MEN-6494
Changelog: none

### DIFF
--- a/common/io.hpp
+++ b/common/io.hpp
@@ -127,7 +127,7 @@ public:
 		// the object.
 		is_(&stream, [](std::istream *stream) {}) {
 	}
-	StreamReader(shared_ptr<std::istream> &stream) :
+	StreamReader(shared_ptr<std::istream> stream) :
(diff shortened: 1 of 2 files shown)


vpodzime:
This is what happens when compiling the extended test without the change of the `ByteWriter` constructor:
```

kacf (approved):
👍

I've also changed a few references like this in the past. For shared_ptr it's better to just pass by value. It's cheap anyway.

PR #1255, opened 2023-06-07
ci: Add the jobs flag to set the make parallel factor to n-ncpus

oleorhagen:
Just speed things up a little.

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -31,7 +31,7 @@ test:
     - chmod +x /usr/local/bin/mender-artifact
   script:
     - cmake -D COVERAGE=ON .
-    - make --keep-going coverage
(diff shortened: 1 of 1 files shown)


.gitlab-ci.yml
```diff
@@ -31,7 +31,7 @@ test:
     - chmod +x /usr/local/bin/mender-artifact
   script:
     - cmake -D COVERAGE=ON .
-    - make --keep-going coverage
+    - make --jobs=$(nproc --all) --keep-going coverage
```
vpodzime, line 34: I usually do '-j -l$NCPUS`, the load is what matters, `-j$NCPUS` hardly every saturates the available resources.

oleorhagen: Doing a test-run with load set now :) 

Good idea!

PR #1257, opened 2023-06-07
CheckNewDeployments()

vpodzime:


mender-update/deployments.cpp
```diff
@@ -0,0 +1,207 @@
...
+			log::Error("Request to check new deployments failed: " + exp_resp.error().message);
+			APIResponse response = expected::unexpected(exp_resp.error());
+			api_handler(response);
+		}
+		auto resp = exp_resp.value();
+		auto status = resp->GetStatusCode();
+		if ((status == http::StatusOK) || (status == http::StatusNoContent)) {
+			handle_data(status);
+		} else if (status == http::StatusNotFound) {
+			log::Warning(
```
kacf, line 191: I think `Info` is enough here. This will happen on all requests to an Open Source server.

kacf (approved):
Beautiful!

You and @oleorhagen have ended up doing things in very similar ways. Which I suppose is a good thing, the API can't be too shabby then!

mender-update/deployments.cpp
```diff
@@ -0,0 +1,207 @@
...
+#include <common/expected.hpp>
+#include <common/http.hpp>
+#include <common/io.hpp>
+#include <common/json.hpp>
+#include <common/log.hpp>
+#include <common/optional.hpp>
+#include <common/path.hpp>
+#include <mender-update/context.hpp>
+
+#include <mender-update/deployments.hpp>
```
oleorhagen, line 30: Nit: Should be the first import

mender-update/deployments.cpp
```diff
@@ -0,0 +1,207 @@
...
+
+	auto ex_provides = ctx.LoadProvides();
+	if (!ex_provides) {
+		return ex_provides.error();
+	}
+	auto provides = ex_provides.value();
+	if (provides.find("artifact_name") == provides.end()) {
+		return MakeError(InvalidDataError, "Missing artifact name data");
+	}
+	string artifact_name = provides["artifact_name"];
```
oleorhagen, line 95: Nit: const

mender-update/deployments.cpp
```diff
@@ -0,0 +1,207 @@
...
+	// TODO: APIRequest
+	auto v2_req = make_shared<http::OutgoingRequest>();
+	v2_req->SetAddress(path::Join(server_url, v2_uri));
+	v2_req->SetMethod(http::Method::POST);
+	v2_req->SetHeader("Content-Type", "application/json");
+	v2_req->SetHeader("Content-Length", to_string(v2_payload.size()));
+	v2_req->SetHeader("Accept", "application/json");
+	v2_req->SetBodyGenerator(payload_gen);
+
+	string v1_args = "artifact_name=" + http::URLEncode(artifact_name)
```
oleorhagen, line 119: Observation, it'd be nice if `URLEncode` took a dictionary

mender-update/deployments.cpp
```diff
@@ -0,0 +1,207 @@
...
+	auto provides = ex_provides.value();
+	if (provides.find("artifact_name") == provides.end()) {
+		return MakeError(InvalidDataError, "Missing artifact name data");
+	}
+	string artifact_name = provides["artifact_name"];
+
+	stringstream ss;
+	ss << R"({"update_control_map": false, "device_provides": {)";
+	ss << R"("device_type": ")";
+	ss << json::EscapeString(device_type);
```
oleorhagen, line 100: Q: Are we not missing some device provides here?

vpodzime:
> Q: Have you tested it with `hosted-mender` or any other Mender server?

Nope. I'd have to add the auth bits and it's a lot of work for me to set everything up.

mender-update/deployments.cpp
vpodzime: Yeah, this is always a thing in C/C++. But we don't do it much often in our code elsewhere...

mender-update/deployments.cpp
vpodzime: Agree to disagree. It would be nice, but then you have to create the dictionary first and pass it in. And this is not Python/JavaScript :wink:

mender-update/deployments.cpp
vpodzime: @kacf should this add the other pairs from `LoadProvides()`? The docs 
image
are not super helpful for me, no clue what `additionalProperties` actually means.

vpodzime: Just for the record: the answer is yes.

mender-update/deployments.cpp
```diff
@@ -92,14 +92,17 @@ error::Error CheckNewDeployments(
...
 
 	stringstream ss;
 	ss << R"({"update_control_map": false, "device_provides": {)";
 	ss << R"("device_type": ")";
 	ss << json::EscapeString(device_type);
-	ss << R"(", "artifact_name": ")";
-	ss << json::EscapeString(artifact_name);
+
+	for (const auto &kv : provides) {
+		ss << "\", \"" + json::EscapeString(kv.first) + "\": \"";
```
kacf, line 102: Suggest sending it minified (no spaces) simply to save space.

vpodzime: I actually had it that way initially, but it looked ugly in the tests. I'll happily change it back, though :+1:

mender-update/deployments.cpp
vpodzime: And have a rule that every variable containing data we don't want to modify should be declared `const`?

mender-update/deployments.cpp
oleorhagen: There's nothing keeping it from doing both, no?

mender-update/deployments.cpp
kacf: If we mean local variables, I think I'd be against such a rule. Don't get me wrong, using const is good, please do whenever you see fit. But to force it onto all local variables as a rule is too much, IMHO.

At least before doing so we should consider what kind of mistakes we think this is preventing.

kacf: I don't have enough experience with it to say, really.

But what I do know is this: I have never thought a local const variable has saved me from doing something stupid. It's already local, so I can easily see everywhere it's used.

OTOH I have often been annoyed at variables that are const, because I actually need to change them.

This is a democratic decision, so if both of your are in favor, we will do it. I just think a hard rule is a bit much.

oleorhagen: Yes, I agree. A hard rule is probably a bit much, but I do think we should keep it in mind.

Seems the style guide is also not giving us much here:
https://google.github.io/styleguide/cppguide.html#Use_of_const

```Using const on local variables is neither encouraged nor discouraged.```

vpodzime: > That other language like Rust for instance have const as the default doesn't change ur mind?

Not everything Rust does is good and the right way :wink:

vpodzime: > This is a democratic decision, so if both of your are in favor, we will do it. I just think a hard rule is a bit much.

I'm definitely not in favor of a hard rule. Generally, I still believe programming is a craft and the programmer is responsible for their creation and they create it in a way that works for them. `const` local variables sometimes help with readability and to emphasize something, but if used everywhere, it's just noise, I think.

kacf: Ok, it sounds like we pretty much agree. Local variable const is good to keep in mind; use it where it makes sense, but no hard rule.

PR #1258, opened 2023-06-08
Http fixes

kacf:


kacf:
@vpodzime: In the end I decided to keep the debug level for the "ignored body" message. Although it *can* be a mistake, it isn't necessarily so if there is no need for the body (for example for 404 messages). So I'm thinking debug is good enough.

vpodzime (approved):
Looks good to me. Thanks! Just one thing -- the commits talk about *client*, but they all modify code handling the `stream_active_` variable. Maybe it could have better name?

kacf:
Good point. I'll rename it to `client_active_`, which is really what it is.

common/http/platform/beast/http.cpp
```diff
@@ -91,19 +91,24 @@ error::Error OutgoingResponse::AsyncReply(ReplyFinishedHandler reply_finished_ha
...
-	// pass an uninitialized `optional` to mean unlimited, but they do not check for
-	// `has_value()` in their code, causing their subsequent comparison operation to
-	// misbehave. So pass highest possible value instead.
-	http_response_parser_.body_limit(numeric_limits<uint64_t>::max());
+	ssl_ctx_.set_verify_mode(ssl::verify_peer);
+
+	beast::error_code ec {};
+	ssl_ctx_.set_default_verify_paths(ec); // Load the default CAs
+	if (ec) {
+		log::Error("Failed to load the SSl default directory");
```
vpodzime, line 104: SS**L**

common/http/platform/beast/http.cpp
```diff
@@ -199,9 +206,23 @@ void Client::ResolveHandler(
...
+	// Note: There is a bug in Beast here (tested on 1.74): One is supposed to be able to
+	// pass an uninitialized `optional` to mean unlimited, but they do not check for
+	// `has_value()` in their code, causing their subsequent comparison operation to
+	// misbehave. So pass highest possible value instead.
+	http_response_parser_->body_limit(numeric_limits<uint64_t>::max());
+
 	weak_ptr<Client> weak_client(stream_active_);
 
-	beast::get_lowest_layer(this->stream_)
+	beast::get_lowest_layer(*this->stream_)
```
vpodzime, line 225: I'd prefer parentheses here.

common/http.hpp
```diff
@@ -285,9 +285,8 @@ class ClientConfig {
 	ClientConfig(string server_cert_path);
 	~ClientConfig();
 
-#ifdef MENDER_USE_BOOST_BEAST
-	ssl::context ctx_ {ssl::context::tls_client};
-#endif // MENDER_USE_BOOST_BEAST
+public:
+	string server_cert_path;
```
vpodzime, line 289: Time to turn `ClientConfig` into a `struct`? I mean, it should, by definition be a passive entity.

vpodzime (approved):
Looks good to me as is, but see suggestions below.

common/http/platform/beast/http.cpp
kacf: Preexisting, but fixed.

common/http/platform/beast/http.cpp
kacf: I just removed `this` altogether. It was the only instance which looked like this, all the others use implicit `this`.

common/http.hpp
kacf: Indeed, maybe it wouldn't have even happened then. Fixed.

PR #1259, opened 2023-06-08
chore(crypto): harvest the OpenSSL error message for all failed calls

oleorhagen:
Since the OpenSSL errors are stored in a queue, and harvested in a FIFO like fashion, it is important that we do harvest the messages upon every call which fails, otherwise the extraction will generate the wrong error messages.

### DIFF
--- a/common/crypto/platform/openssl/crypto.cpp
+++ b/common/crypto/platform/openssl/crypto.cpp
@@ -66,6 +66,34 @@ auto bio_free_all_func = [](BIO *bio) {
 	}
 };
 
+// NOTE: GetOpenSSLErrorMessage should be called upon all OpenSSL errors, as
+// the errors are queued, and if not harvested, the FIFO structure of the
+// queue will mean that if you just get one, you might actually get the wrong
+// one.
+string GetOpenSSLErrorMessage() {
+	const auto sysErrorCode = errno;
+	auto sslErrorCode = ERR_get_error();
+
+	std::string errorDescription {};
+	while (sslErrorCode != 0) {
+		if (!errorDescription.empty()) {
+			errorDescription += '\n';
+		}
+		errorDescription += ERR_error_string(sslErrorCode, nullptr);
+		sslErrorCode = ERR_get_error();
+	}
+	if (sysErrorCode != 0) {
+		if (!errorDescription.empty()) {
+			errorDescription += '\n';
+		}
+		errorDescription += "System error, code=" + std::to_string(sysErrorCode);
+		errorDescription += ", ";
+		errorDescription += strerror(sysErrorCode);
+	}
+	return errorDescription;
+}
+
+
 expected::ExpectedString EncodeBase64(vector<uint8_t> to_encode) {
 	// Predict the len of the decoded for later verification. From man page:
 	// For every 3 bytes of input provided 4 bytes of output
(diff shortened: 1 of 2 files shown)


lluiscampos:
> I am also pondering extracting the entire message queue every time we gall `GetOpenSSLError`. Which is probably safer.
> 
> What do you think?

Yes, we should. This function should be as defensive as possible so that. Locally I had the following code:
```
diff --git a/common/crypto/platform/openssl/crypto.cpp b/common/crypto/platform/openssl/crypto.cpp
index 7e5a763f..2ae3f3da 100644
--- a/common/crypto/platform/openssl/crypto.cpp
+++ b/common/crypto/platform/openssl/crypto.cpp
@@ -129,11 +129,15 @@ sha::ExpectedSHA Shasum(vector<uint8_t> data) {
 /* TODO: harvest all errors pending and concatenate them */
 string GetOpenSSLErrorMessage() {
        const auto sysErrorCode = errno;
-       const auto sslErrorCode = ERR_get_error();
+       auto sslErrorCode = ERR_get_error();
 
        std::string errorDescription;
-       if (sslErrorCode != 0) {
-               errorDescription = ERR_error_string(sslErrorCode, nullptr);
+       while (sslErrorCode != 0) {
+               if (!errorDescription.empty()) {
+                       errorDescription += '\n';
+               }
+               errorDescription += ERR_error_string(sslErrorCode, nullptr);
+               sslErrorCode = ERR_get_error();
        }
        if (sysErrorCode != 0) {
                if (!errorDescription.empty()) {
```

common/crypto/platform/openssl/crypto.cpp
```diff
@@ -66,6 +66,43 @@ auto bio_free_all_func = [](BIO *bio) {
...
+			if (!errorDescription.empty()) {
+				errorDescription += '\n';
+			}
+			errorDescription += "System error, code=" + std::to_string(sysErrorCode);
+			errorDescription += ", ";
+			errorDescription += strerror(sysErrorCode);
+		}
+		return errorDescription;
+	}
+	return "";
```
lluiscampos, line 102: Is it possible that we don't have an ssl error but we do have a system error? If so the function would return `"`.

lluiscampos (changes requested):
Thanks for taking up this again!

Some fine tuning still required, I think.

common/crypto/platform/openssl/crypto.cpp
oleorhagen: Good point! I guess it _could happen_. Let's just say we will never forget logging the error :p

oleorhagen:
> Thanks for taking up this again!
> 
> Some fine tuning still required, I think.

New day, new opportunity @lluiscampos

common/crypto/platform/openssl/crypto.cpp
```diff
@@ -66,6 +66,36 @@ auto bio_free_all_func = [](BIO *bio) {
...
+	std::string errorDescription {};
+	if (sslErrorCode != 0) {
+		while (sslErrorCode != 0) {
+			if (!errorDescription.empty()) {
+				errorDescription += '\n';
+			}
+			errorDescription += ERR_error_string(sslErrorCode, nullptr);
+			sslErrorCode = ERR_get_error();
+		}
+	}
```
lluiscampos, line 86: Now that the second condition is out of this `if` block, the code could be simplified with:
```suggestion
	while (sslErrorCode != 0) {
		if (!errorDescription.empty()) {
			errorDescription += '\n';
		}
		errorDescription += ERR_error_string(sslErrorCode, nullptr);
		sslErrorCode = ERR_get_error();
	}
```

lluiscampos (approved):
A minor suggestion for improvement, but looks good now!

PR #1261, opened 2023-06-09
fix: Don't crash if the handler owns `Client` by capture.

kacf:
Changelog: None
Ticket: None

### DIFF
--- a/common/http/platform/beast/http.cpp
+++ b/common/http/platform/beast/http.cpp
@@ -169,6 +169,15 @@ error::Error Client::AsyncCall(
 	return error::NoError;
 }
 
+void Client::CallHandler(ResponseHandler handler) {
+	// This function exists to make sure we have a copy of the handler we're calling (in the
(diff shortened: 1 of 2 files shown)


common/http/platform/beast/http.cpp
```diff
@@ -169,6 +169,15 @@ error::Error Client::AsyncCall(
 	return error::NoError;
 }
 
+void Client::CallHandler(ResponseHandler handler) {
+	// This function exists to make sure we have a copy of the handler we're calling. This is
```
vpodzime, line 173: I'd emphasize that the argument is actually the copy (if I understand it correctly) because it's not entirely obvious.

vpodzime (approved):
Thanks for the fix.

common/http/platform/beast/http.cpp
kacf: Good point, fixed!

PR #1262, opened 2023-06-09
fix: Fix race condition in AsyncWait for quick processes.

kacf:
If the process finishes so quickly that `SetupAsyncWait` completes before the caller has a chance to call `AsyncWait`, then the handler is never called. With this commit, this scenario is detected and the handler is scheduled right away if we find that the process has already ended.

This showed up on my system occasionally as a very rare hang in the Update Module state calls tests.

Changelog: None
Ticket: None

### DIFF
--- a/common/processes/platform/tiny_process_library/tiny_process_library.cpp
+++ b/common/processes/platform/tiny_process_library/tiny_process_library.cpp
@@ -168,6 +168,13 @@ error::Error Process::AsyncWait(events::EventLoop &loop, AsyncWaitHandler handle
 	async_wait_data_->event_loop = &loop;
 	async_wait_data_->handler = handler;
 
+	if (async_wait_data_->process_ended) {
+		// The process has already ended. Schedule the handler immediately.
+		auto &async_wait_data = async_wait_data_;
+		async_wait_data->event_loop->Post(
+			[async_wait_data, this]() { AsyncWaitInternalHandler(async_wait_data); });
+	}
+
 	return error::NoError;
 }
 
@@ -176,6 +183,7 @@ void Process::Cancel() {
 
 	async_wait_data_->event_loop = nullptr;
 	async_wait_data_->handler = nullptr;
+	async_wait_data_->process_ended = false;
 }
 
 void Process::SetupAsyncWait() {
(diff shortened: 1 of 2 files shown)


common/processes/platform/tiny_process_library/tiny_process_library.cpp
```diff
@@ -168,6 +168,14 @@ error::Error Process::AsyncWait(events::EventLoop &loop, AsyncWaitHandler handle
 	async_wait_data_->event_loop = &loop;
 	async_wait_data_->handler = handler;
 
+	if (async_wait_data_->process_ended) {
+		// The process has already ended. Schedule the handler immediately.
+		auto async_wait_data = async_wait_data_;
```
oleorhagen, line 173: Hmm, why copy it here, and in the lambda capture?

oleorhagen (approved):

common/processes/platform/tiny_process_library/tiny_process_library.cpp
```diff
@@ -192,31 +201,39 @@ void Process::SetupAsyncWait() {
...
 			});
 		}
 
+		async_wait_data->process_ended = true;
+
 		return status;
 	});
 }
 
+void Process::AsyncWaitInternalHandler(shared_ptr<AsyncWaitData> async_wait_data, Process &proc) {
```
vpodzime, line 214: Should this even be a member function? It doesn't work with any private members and in practice it actually takes `*this` as the second argument (which is a bit confusing and deserves a comment, actually).

vpodzime (commented):
Looks good to me otherwise.

common/processes/platform/tiny_process_library/tiny_process_library.cpp
kacf: Because it's not possible to capture member variables, and we have to capture by value in order to keep our own, independent pointer (so we can't use `this`).

It looks like it works to make this a reference though, and it can still capture by value below, so I switched to that.

common/processes/platform/tiny_process_library/tiny_process_library.cpp
kacf: It was meant to make it clearer that our own `async_wait_data` must be used over the one in the member variables. But you're right, it's not clearer, so I'll just change it to a member function and explain `async_wait_data` separately.

PR #1266, opened 2023-06-13
Implement bootstrap artifact support + legacy upgrades

kacf:


mender-update/standalone/standalone.cpp
```diff
@@ -445,6 +451,22 @@ ResultAndError DoRollback(
...
+
+	auto err = main_context.CommitArtifactData(
+		data.artifact_name,
+		data.artifact_group,
+		data.artifact_provides,
+		data.artifact_clears_provides,
+		[](database::Transaction &txn) { return error::NoError; });
+	if (err != error::NoError) {
+		return {Result::InstalledButFailedInPostCommit, err};
+	} else {
```
oleorhagen, line 465: Nit: Unneeded else

oleorhagen (approved):
Nice

mender-update/standalone/standalone.cpp
```diff
@@ -272,6 +273,14 @@ ResultAndError Commit(context::MenderContext &main_context) {
 
 	update_module::UpdateModule update_module(main_context, data.payload_types[0]);
 
+	if (data.payload_types[0] == "rootfs-image") {
```
vpodzime, line 284: IMHO, these payload type strings should be constants somewhere, not just magic string values scattered in multiple places in the code.

oleorhagen: Some of them can, but not all

oleorhagen: They can be user defined

vpodzime: Sure, but the string literals in the code can all be constants defined somewhere, right?

oleorhagen: Like I said, some of them can :)

vpodzime: Well, the ones that are string literals are not user defined and can be constants.

mender-update/update_module/v3/platform/boost_filesystem/fs_operations.cpp
```diff
@@ -207,6 +218,31 @@ error::Error UpdateModule::PrepareFileTree(
...
 }
 
+error::Error UpdateModule::EnsureRootfsImageFileTree(const string &path) {
+	// Historical note: Versions of the client prior to 4.0 had the rootfs-image module built
+	// in. Because of this it has no Update Module File Tree. So if we are upgrading, we might
+	// hit an on-going upgrade without a File Tree. It's too late to create a complete one with
+	// all the artifact content by the time we get here, but at least we can create one which
+	// has the current Provides information, as well as a folder for the Update Module to run
+	// in.
+	ifstream payload_type(path::Join(path, "header", "payload_type"));
```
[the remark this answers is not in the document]
kacf: It doesn't help in this particular case because we are not handling the error, we are just checking for the existence of any error at all. So the original is actually shorter, since we don't need `expected` handling.

mender-update/standalone/standalone.cpp
kacf: Hmm, yes, I see your point. There are a lot of these in the code though, and they are not always easily replaced by a constant because they are combined with other strings:

```
mender-update/cli/cli_test.cpp:	EXPECT_TRUE(VerifyProvides(tmpdir.Path(), R"(rootfs-image.version=test
mender-update/cli/cli_test.cpp:rootfs-image.checksum=f2ca1bb6c7e907d06dafe4687e579fce76b37e4e93b7605022da52e6ccc26fd2
mender-update/cli/cli_test.cpp:	string update_module = path::Join(tmpdir.Path(), "rootfs-image");
...
mender-update/update_module/v3/update_module_test.cpp:	      "type": "rootfs-image"
mender-update/update_module/v3/update_module_test.cpp:	    "rootfs-image.checksum":
mender-update/update_module/v3/update_module_test.cpp:	    "rootfs-image.version": "test-artifact"
mender-update/update_module/v3/update_module_test.cpp:	    "rootfs-image.*"
```

This is just a handful of examples, there are many more.

I don't think they should all be replaced with a constant, because it would actually make it harder to see what the code is about.

The `rootfs-image` type is essentially an API string at this point. It is not going to change. So I think we might as well leave it alone for readability.

PR #1272, opened 2023-06-15
MEN-6203: Artifact manifest signature parse and verification

lluiscampos:


lluiscampos (commented):
@kacf @oleorhagen @vpodzime This is not ready for review :rainbow: Looking forward to your feedback.

artifact/parser.cpp
```diff
@@ -121,6 +121,13 @@ ExpectedArtifact Parse(io::Reader &reader, config::ParserConfig config) {
 	tok = lexer.Next();
 	optional::optional<ManifestSignature> signature;
 	if (tok.type == token::Type::ManifestSignature) {
+		// Early check for keys
+		if (config.artifact_verify_keys.size() == 0) {
+			return expected::unexpected(parser_error::MakeError(
+				parser_error::Code::ParseError,
+				"Missing public key(s) to verify manifest signature"));
```
kacf, line 128: Missing keys is not an error. It's allowed to install a signed artifact onto a device which doesn't have signatures enabled, but not vice versa.

artifact/parser.cpp
```diff
@@ -129,6 +136,19 @@ ExpectedArtifact Parse(io::Reader &reader, config::ParserConfig config) {
...
+		auto expected_verified = manifest_sig::VerifySignature(
+			*signature, manifest.GetRaw(), config.artifact_verify_keys);
+		if (!expected_verified) {
+			return expected::unexpected(parser_error::MakeError(
+				parser_error::Code::ParseError,
+				"Failed to verify the manifest signature: " + expected_verified.error().message));
+		}
+		if (!expected_verified.value()) {
+			return expected::unexpected(parser_error::MakeError(
+				parser_error::Code::ParseError, "Wrong manifest signature or wrong key"));
```
kacf, line 150: Should it be `ParseError` or a dedicated error for signature?

artifact/v3/manifest/manifest.cpp
```diff
@@ -84,14 +84,25 @@ ExpectedManifest Parse(mender::common::io::Reader &reader) {
 
 	auto str = reader.GetStream();
 	string line {};
+
+	// auto start_pos = str.get()->tellg();
```
kacf, line 88: Remove?

artifact/v3/manifest/manifest.cpp
```diff
@@ -84,14 +84,25 @@ ExpectedManifest Parse(mender::common::io::Reader &reader) {
...
 	while (getline(*str.get(), line, '\n')) {
 		auto manifest_line = Tokenize(line);
 		if (!manifest_line) {
 			return expected::unexpected(manifest_line.error());
 		}
 
 		m.map_[manifest_line->entry_name] = manifest_line->shasum;
+
+		std::copy_n(line.data(), line.length(), std::back_inserter(m.raw_copy_));
+		m.raw_copy_.push_back('\n');
```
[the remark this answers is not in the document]
kacf: Hmm, I think I need a f2f on this one.

artifact/v3/manifest_sig/manifest_sig.cpp
```diff
@@ -39,8 +41,21 @@ ExpectedManifestSignature Parse(io::Reader &reader) {
...
+		auto e_verify_sign = crypto::VerifySign(key, raw_data, signature.value);
+		if (e_verify_sign && e_verify_sign.value()) {
+			return expected::ExpectedBool(true);
+		}
+		if (!e_verify_sign) {
+			err = err.FollowedBy(e_verify_sign.error());
+		}
+	}
+	if (err == error::NoError) {
+		return expected::ExpectedBool(false);
```
kacf, line 56: Hmm, this will fail signature verification if one key isn't found for example, even if one of the other keys work. I could see this happening with a broken symlink for example. Is it better to let a valid verification trump everything else? I don't see how it would be possible to "accidentally" verify something. If it's verified, it's verified, no matter the state of the rest of the keys.

kacf (changes requested):

common/crypto/platform/openssl/crypto.cpp
```diff
@@ -231,6 +260,76 @@ expected::ExpectedString Sign(string private_key_path, vector<uint8_t> raw_data)
...
+	// verify signature
+	ret = EVP_PKEY_verify(
+		pkey_signer_ctx.get(), signature.data(), signature.size(), shasum.data(), shasum.size());
+	if (ret <= 0) {
+		return expected::unexpected(MakeError(
+			SetupError,
+			"Failed to verify signature. OpenSSL PKEY verify failed: " + GetOpenSSLErrorMessage()));
+	}
+
+	return expected::ExpectedBool(ret == 1);
```
oleorhagen, line 310: S: Just return `ret == 1`

common/crypto/platform/openssl/crypto.cpp
```diff
@@ -231,6 +260,76 @@ expected::ExpectedString Sign(string private_key_path, vector<uint8_t> raw_data)
...
+	log::Debug("Shasum is: " + shasum.String());
+
+	// signature: decode base64
+	auto exp_decoded_signature = DecodeBase64(signature);
+	if (!exp_decoded_signature) {
+		return expected::unexpected(exp_decoded_signature.error());
+	}
+	auto decoded_signature = exp_decoded_signature.value();
+
+	return VerifySignData(public_key_path, shasum, decoded_signature);
```
oleorhagen, line 330: Nice function all in all :+1:

artifact/parser_test.cpp
```diff
@@ -246,11 +242,79 @@ TEST_F(ParserTestEnv, TestParseTopLevelSigned) {
 
 	io::StreamReader sr {fs};
 
-	auto expected_artifact = mender::artifact::parser::Parse(sr);
+	mender::artifact::config::ParserConfig cfg = {
+		.artifact_verify_keys = {tmpdir->Path() + "/public.key"},
```
oleorhagen, line 246: Optional: We do have a `path::Join` function

artifact/v3/manifest_sig/manifest_sig.cpp
```diff
@@ -39,8 +41,21 @@ ExpectedManifestSignature Parse(io::Reader &reader) {
 }
 
 expected::ExpectedBool VerifySignature(
-	ManifestSignature &signature, vector<string> &artifact_verify_keys) {
-	return expected::ExpectedBool(true);
+	ManifestSignature &signature, vector<uint8_t> raw_data, vector<string> &artifact_verify_keys) {
+	error::Error err;
+	for (auto key : artifact_verify_keys) {
```
oleorhagen, line 46: S: `const auto &`

artifact/v3/manifest/manifest.cpp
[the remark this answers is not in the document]
oleorhagen: What about a `TeeReader` similar to what we do in `Golang`?

oleorhagen (commented):
I think it looks like it is coming together real nice!

Easy 3 `sp` :sweat_smile:

common/crypto/platform/openssl/crypto.cpp
```diff
@@ -231,6 +260,76 @@ expected::ExpectedString Sign(string private_key_path, vector<uint8_t> raw_data)
...
+	if (ret <= 0) {
+		return expected::unexpected(MakeError(
+			SetupError,
+			"Failed to set the OpenSSL signature to sha256" + GetOpenSSLErrorMessage()));
+	}
+
+	// verify signature
+	ret = EVP_PKEY_verify(
+		pkey_signer_ctx.get(), signature.data(), signature.size(), shasum.data(), shasum.size());
+	if (ret <= 0) {
```
vpodzime, line 304: `0` is returned if the verification fails not because of an error, but because the signature is not valid or there was some mismatch. So this should be `< 0`, I believe.

artifact/v3/manifest_sig/manifest_sig.hpp
```diff
@@ -0,0 +1,54 @@
...
+namespace io = mender::common::io;
+namespace error = mender::common::error;
+
+class ManifestSignature {
+public:
+	ManifestSignature(const std::string &sig) :
+		value(sig) {
+	}
+	std::string value;
+};
```
vpodzime, line 41: What is this class good for? It's basically the same as `using ManifestSignature = string`.

vpodzime (commented):
My 2 cents.

common/crypto.hpp
```diff
@@ -45,8 +45,15 @@ error::Error MakeError(CryptoErrorCode code, const string &msg);
...
 
+expected::ExpectedString EncodeBase64(vector<uint8_t> to_encode);
+
+expected::ExpectedBytes DecodeBase64(string to_decode);
+
 expected::ExpectedString Sign(const string private_key_path, const vector<uint8_t> raw_data);
 
+expected::ExpectedBool VerifySign(
+	const string public_key_path, vector<uint8_t> raw_data, string signature);
+
```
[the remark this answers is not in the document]
lluiscampos: Right. Why was not the case already for `Sign`?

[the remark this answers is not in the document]
oleorhagen: Laziness?

artifact/parser.cpp
lluiscampos: Good catch. Thanks

.gitlab-ci.yml
```diff
@@ -59,7 +55,7 @@ test:
         gnupg-agent
         software-properties-common
     - curl -fsSL https://downloads.mender.io/repos/debian/gpg >> /etc/apt/trusted.gpg.d/mender.asc
-    - echo "deb [arch=$(dpkg --print-architecture)] https://downloads.mender.io/repos/debian ubuntu/jammy/stable main"
+    - echo "deb [arch=$(dpkg --print-architecture)] https://downloads.mender.io/repos/debian ubuntu/jammy/experimental main"
```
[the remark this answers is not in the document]
lluiscampos: No, this is on a different job (template, really) which is only present in this branch. I needed to update it due to hitting [this bug](https://github.com/mendersoftware/mender-artifact/pull/498), not related to the changes you did for `test:modules-artifact-gen`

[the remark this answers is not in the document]
kacf: Ah ok 👍

common/crypto/platform/openssl/crypto.cpp
```diff
@@ -81,6 +84,32 @@ expected::ExpectedString EncodeBase64(vector<uint8_t> to_encode) {
...
+			Base64Error,
+			"The predicted (" + std::to_string(predicted_len) + ") and the actual ("
+				+ std::to_string(output_len) + ") length differ"));
+	}
+
+	// Clear padding
+	// TODO Is this the right thing to do?
+	while (buffer.back() == 0) {
+		buffer.pop_back();
+	}
```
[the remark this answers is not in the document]
kacf: Hmm, this is tricky indeed. Isn't `\0` a valid output byte from base64? So trailing null bytes can't be distinguished from real data unless the function tells you the length. Are you sure that `EVP_DecodeBlock` doesn't return only the valid data, without the trailing null bytes?

artifact/v3/manifest/manifest.cpp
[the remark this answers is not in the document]
lluiscampos: Discussed face to face. Rework coming up.

common/crypto/platform/openssl/crypto.cpp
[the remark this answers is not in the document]
lluiscampos: I am sure. [This test](https://github.com/mendersoftware/mender/pull/1272/files#diff-b7462d7f6f9d4e4511cb0bfe6d7d52593fd3faf6eff32e83b49ec1cc89fd6430R76) fails if I remove this piece of code that we are discussing.

That test is arbitrary, the real data for us to encode/decode (the signature) will be multiple of `4` , but if we aim for this library method to be generic we should handle this somehow.

I will look into examples online using this function and see what is the common practice :+1:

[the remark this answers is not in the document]
lluiscampos: I have implemented [this](https://github.com/openssl/openssl/issues/17197#issuecomment-986482576) solution. 

And, while at it, I found another bug in the code that was hidden due this aggressive clean-up I was doing :sweat_smile: 

Coming up in next push.

artifact/parser.cpp
lluiscampos: Okey, I created `SignatureVerificationError`

artifact/parser.hpp
```diff
@@ -70,24 +72,25 @@ class Artifact {
...
 		lexer_ {lexer},
 		version {version},
 		manifest {manifest},
-		header {header},
-		manifest_signature {manifest_signature} {
+		header {header} {
 	}
+
+	bool is_signed() const {
+		return manifest_signature.has_value();
```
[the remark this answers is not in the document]
lluiscampos: I think I am going to remove this method altogether, it is only used by tests.

Do we need to know if the artifact is signed out of the Artifact parsing itself?

artifact/v3/manifest/manifest.cpp
lluiscampos: Removed now - this was part of the conversation below.

artifact/v3/manifest_sig/CMakeLists.txt
```diff
@@ -0,0 +1,20 @@
...
+add_executable(artifact_manifest_sig_parser_test EXCLUDE_FROM_ALL
+  manifest_sig.cpp
+  ${CMAKE_SOURCE_DIR}/artifact/error.cpp
+  manifest_sig_test.cpp
+)
+target_link_libraries(artifact_manifest_sig_parser_test PRIVATE
+  common_log
+  common_io
+  common_error
+  GTest::gtest_main
```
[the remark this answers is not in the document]
lluiscampos: I don't understand what you mean :smiley:

artifact/v3/manifest_sig/manifest_sig.cpp
lluiscampos: Mmmm. I didn't intend that behavior, but maybe I expressed it incorrectly.

So inside the for loop we'll return `true` when the first key is found (parsed, etc) and validates correctly. We explicitly check it with `if (e_verify_sign && e_verify_sign.value())`. Otherwise we just append errors (if exist).

Finally, this part that you are pointing out (which is outside of the loop) will either return just a false if there were no errors but still no correct verification, or the concatenated errors (line below).

Does it make sense?

artifact/v3/manifest_sig/manifest_sig.hpp
```diff
@@ -0,0 +1,51 @@
...
+//    Unless required by applicable law or agreed to in writing, software
+//    distributed under the License is distributed on an "AS IS" BASIS,
+//    WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+//    See the License for the specific language governing permissions and
+//    limitations under the License.
+
+#ifndef MENDER_ARTIFACT_V3_MANIFEST_SIG_PARSER_HPP
+#define MENDER_ARTIFACT_V3_MANIFEST_SIG_PARSER_HPP
+
+#include <string>
```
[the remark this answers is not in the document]
lluiscampos: Which include is missing?

artifact/v3/manifest_sig/manifest_sig.hpp
```diff
@@ -0,0 +1,51 @@
...
+using namespace std;
+
+namespace expected = mender::common::expected;
+namespace io = mender::common::io;
+namespace error = mender::common::error;
+
+class ManifestSignature {
+public:
+	ManifestSignature(const std::string &sig) :
+		value(sig) {
```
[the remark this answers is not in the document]
lluiscampos: > We have been pretty consistent with {} initialization in the artifact code

Could this be enforced by style?

artifact/v3/manifest_sig/manifest_sig.hpp
lluiscampos: @oleorhagen @vpodzime I'll convert into struct :+1:

common/crypto/platform/openssl/crypto.cpp
```diff
@@ -81,6 +84,32 @@ expected::ExpectedString EncodeBase64(vector<uint8_t> to_encode) {
 	return string(buffer.begin(), buffer.end() - 1); // Remove the last zero byte
 }
 
+expected::ExpectedBytes DecodeBase64(string to_decode) {
+	const auto predicted_len = 3 * ((to_decode.size() + 3) / 4);
```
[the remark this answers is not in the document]
lluiscampos: I'll add also a similar comment to existing `EncodeBase64` len prediction

common/crypto/platform/openssl/crypto.cpp
```diff
@@ -164,15 +193,15 @@ expected::ExpectedString ExtractPublicKey(const string &private_key_path) {
 }
 
 expected::ExpectedBytes SignData(const string private_key_path, const vector<uint8_t> digest) {
-	auto bio_public_key = unique_ptr<BIO, void (*)(BIO *)>(
+	auto bio_private_key = unique_ptr<BIO, void (*)(BIO *)>(
```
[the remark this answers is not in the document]
lluiscampos: After final review, I'll squash everything and split out the few things that are fixes to existing code :+1:

common/crypto/platform/openssl/crypto.cpp
```diff
@@ -231,6 +260,76 @@ expected::ExpectedString Sign(string private_key_path, vector<uint8_t> raw_data)
...
+	// load key
+	auto bio_key =
+		unique_ptr<BIO, void (*)(BIO *)>(BIO_new_file(public_key_path.c_str(), "r"), bio_free_func);
+	if (!bio_key.get()) {
+		return expected::unexpected(
+			MakeError(SetupError, "Failed to open the public key file" + GetOpenSSLErrorMessage()));
+	}
+	auto pkey = unique_ptr<EVP_PKEY, void (*)(EVP_PKEY *)>(
+		PEM_read_bio_PUBKEY(bio_key.get(), nullptr, nullptr, nullptr), pkey_free_func);
+	if (pkey == nullptr) {
```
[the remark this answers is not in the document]
lluiscampos: Thanks for striving for consistency :smiley: I'll fix

common/crypto/platform/openssl/crypto.cpp
```diff
@@ -231,6 +260,76 @@ expected::ExpectedString Sign(string private_key_path, vector<uint8_t> raw_data)
...
+	}
+
+	// prepare context
+	auto pkey_signer_ctx = unique_ptr<EVP_PKEY_CTX, void (*)(EVP_PKEY_CTX *)>(
+		EVP_PKEY_CTX_new(pkey.get(), nullptr), pkey_ctx_free_func);
+
+	auto ret = EVP_PKEY_verify_init(pkey_signer_ctx.get());
+	if (ret <= 0) {
+		return expected::unexpected(MakeError(
+			SetupError, "Failed to initialize the OpenSSL signer " + GetOpenSSLErrorMessage()));
```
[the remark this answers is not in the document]
lluiscampos: I fixed these everywhere. Thanks

common/crypto_test.cpp
```diff
@@ -62,6 +62,36 @@ TEST(CryptoTest, TestPublicKeyExtractionError) {
...
+	auto expected_signature = crypto::Sign(private_key_file, testdata);
+	ASSERT_TRUE(expected_signature) << "Unexpected: " << expected_signature.error();
+
+	auto signature = expected_signature.value();
+
+	// vector<uint8_t> testsig {signature.begin(), signature.end()};
+	string public_key_file = "./public-key.pem";
+	auto expected_verify_signature = crypto::VerifySign(public_key_file, testdata, signature);
+	ASSERT_TRUE(expected_verify_signature) << "Unexpected: " << expected_verify_signature.error();
+	ASSERT_TRUE(expected_verify_signature.value());
```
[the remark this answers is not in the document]
lluiscampos: I added two invalid signatures tests now

common/crypto/platform/openssl/crypto.cpp
```diff
@@ -231,6 +260,76 @@ expected::ExpectedString Sign(string private_key_path, vector<uint8_t> raw_data)
...
+			SetupError,
+			"Failed to set the OpenSSL signature to sha256" + GetOpenSSLErrorMessage()));
+	}
+
+	// verify signature
+	ret = EVP_PKEY_verify(
+		pkey_signer_ctx.get(), signature.data(), signature.size(), shasum.data(), shasum.size());
+	if (ret <= 0) {
+		return expected::unexpected(MakeError(
+			SetupError,
```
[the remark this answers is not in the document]
lluiscampos: I created `VerificationError` :+1:

common/crypto/platform/openssl/crypto.cpp
lluiscampos: Yes, you are completely correct. On invalid signature we should not return an error, but just `false`. I'll fix

artifact/v3/manifest_sig/manifest_sig.hpp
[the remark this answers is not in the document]
oleorhagen: Hmm, not sure :thinking:  I think it'd be nice but

lluiscampos:
@kacf @oleorhagen @vpodzime Ready for a second round :eyes: 

As the changes were scattered all across, I pushed a single patch with all the changes. Before merge, I will split out the small fixes to existing code into separate commits and squash together everything else as the main feature :+1:

artifact/v3/manifest_sig/manifest_sig.hpp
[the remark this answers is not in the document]
oleorhagen: `vector`

artifact/parser.hpp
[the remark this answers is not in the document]
oleorhagen: No

artifact/sha/platform/openssl/sha.cpp
```diff
@@ -140,5 +142,22 @@ ExpectedSHA Reader::ShaSum() {
 	return SHA(hash, SHA_256_digest_length);
 }
 
+ExpectedSHA Shasum(const vector<uint8_t> &data) {
+	string in {data.begin(), data.end()};
+
+	io::StringReader is {in};
+
+	Reader r {is, ""};
```
oleorhagen, line 150: Observation, I don't think you need to pass in a string to the constructor at all, if you don't expect a shasum check

artifact/v3/manifest/manifest.hpp
```diff
@@ -36,12 +37,12 @@ namespace error = mender::common::error;
 class Manifest {
 public:
 	string Get(const string &key);
-	vector<uint8_t> GetRaw() const {
-		return raw_copy_;
+	mender::sha::SHA GetShaSum() const {
```
oleorhagen, line 40: S: Maybe namespace this so the reference isn't so long?

common/crypto.hpp
```diff
@@ -49,10 +51,13 @@ expected::ExpectedString EncodeBase64(vector<uint8_t> to_encode);
...
 
-expected::ExpectedString Sign(const string private_key_path, const vector<uint8_t> raw_data);
+expected::ExpectedString Sign(const string &private_key_path, const mender::sha::SHA &shasum);
+
+expected::ExpectedString SignRawData(
+	const string &private_key_path, const vector<uint8_t> &raw_data);
 
 expected::ExpectedBool VerifySign(
-	const string public_key_path, vector<uint8_t> raw_data, string signature);
+	const string &public_key_path, const mender::sha::SHA &shasum, const string &signature);
```
oleorhagen, line 60: S: Same, `namespace sha = mender::sha`

artifact/v3/manifest_sig/manifest_sig.cpp
kacf: You're right, I misread it. Looks good! 👍

common/crypto/platform/openssl/crypto.cpp
```diff
@@ -69,9 +69,15 @@ auto bio_free_all_func = [](BIO *bio) {
...
 expected::ExpectedString EncodeBase64(vector<uint8_t> to_encode) {
+	// Predict the len of the decoded for later verification. From man page:
+	// For every 3 bytes of input provided 4 bytes of output
+	// data will be produced. If n is not divisible by 3 (...)
+	// the output is padded such that it is always divisible by 4.
 	const auto predicted_len = 4 * ((to_encode.size() + 2) / 3);
 
-	auto buffer {vector<unsigned char>(predicted_len + 1)}; // Add space for a NUL terminator
+	// Add space for a NUL terminator. From man page:
+	// Additionally a NUL terminator character will be added
```
oleorhagen, line 79: :+1:

common/crypto/platform/openssl/crypto.cpp
```diff
@@ -101,40 +111,23 @@ expected::ExpectedBytes DecodeBase64(string to_decode) {
 				+ std::to_string(output_len) + ") length differ"));
 	}
 
-	// Clear padding
-	// TODO Is this the right thing to do?
-	while (buffer.back() == 0) {
+	// Subtract padding bytes. Inspired by internal OpenSSL code from:
+	// https://github.com/openssl/openssl/blob/ff88545e02ab48a52952350c52013cf765455dd3/crypto/ct/ct_b64.c#L46
+	for (auto it = to_decode.crbegin(); *it == '='; it++) {
```
oleorhagen, line 116: Nice

oleorhagen (approved):
Nice!

artifact/parser.hpp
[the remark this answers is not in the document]
kacf: > Q: Now we don't default initialize `manifest_signature`. Is this always safe?

All non-intrinsic types (everything except `char`, the `int` and `float` types, and raw pointers) are always default initialized. So using `{}` is equivalent to specifying nothing.

artifact/parser_test.cpp
```diff
@@ -238,12 +237,12 @@ TEST_F(ParserTestEnv, TestParseEmptyPayloadArtifact) {
 }
 
 TEST_F(ParserTestEnv, TestParseTopLevelSigned) {
-	std::fstream fs {tmpdir->Path() + "/test-artifact-signed.mender"};
+	std::fstream fs {mender::common::path::Join(tmpdir->Path(), "/test-artifact-signed.mender")};
```
kacf, line 240: Beware that absolute path in the second argument won't work with the upcoming https://github.com/mendersoftware/mender/pull/1286, and is bad style nevertheless. Applies to many locations.

kacf (approved):
I spotted one fix which is needed, but it's easy, so pre-approving.

artifact/parser_test.cpp
kacf: Btw, you could also shorten to `path::Join` by inserting `namespace path = mender::common::path;` near the top.

artifact/sha/platform/openssl/sha.cpp
lluiscampos: Right. The problem here is that from this compile unit `Reader r {}` is ambiguous because both declared constructors `Reader(io::Reader &reader)` and `Reader(io::Reader &reader, const std::string &expected_sha = "")` are invoked the same way. This was not a problem when this function was part of `crypto` library, from the outside there is no ambiguity (`.hpp` declaration is with no default value).

I'll remove the default value in the constructor :+1:

artifact/parser_test.cpp
lluiscampos: Thank you.

lluiscampos:
Ready to press the :green_circle: button. The final change-set looks like:

* chore: Drain out existing errors from OpenSSL
    OpenSSL stores errors in a FIFO and we want to return to the user all
    pending errors.
* chore: Rename variable to represent the actual key being handled
* chore: Remove default argument value for `sha::Reader` constructor
    It cannot be used outside this compile unit (the declaration at the
    header does not contain the default argument) and it creates ambiguity
* chore: Move helper function `Shasum` to `artifact/sha` library
    So that it can be reused outside of `crypto` library
* ci: Install master mender-artifact for static tests
    And get the rid of the redundant installation via direct download
* style: Return boolean directly for ExpectedBool
    So that this file, and the project as a whole, keeps consistency.
    with the other constructor.
* chore: Extend `crypto` library to support RSA signature verification
    The library is extended with the new method `VerifySign` that verifies a
    signature given the public key and the digest of the message.
    To keep the interface consistent, the existing function `Sign` which took
    the raw message as an input is moved to `SignRawData` and a new `Sign`
    function is implemented for signing pre-computed digests. Pass also
    const reference in these functions where possible.
    Additionally:
    Functions to encode and decode Base64 are implemented/exposes
    Docstring is added to `EncodeBase64` to clarify the length prediction
    The `mender::sha::SHA` class now knows the length of the digest
* feat: Artifact manifest RSA signature parse and verification
    Implement artifact/v3/manifest_sig to parse and hold the optional
    signature of an Artifact. It is verified by the main Artifact parser at
    `Parse` time if public keys were provided by the configuration.
    The `Manifest` class has been modified to parse and hold a SHA256 digest
    of the contents so that the signature can be verified later on.

PR #1274, opened 2023-06-15
Inventory API

vpodzime:


common/api.hpp
```diff
@@ -0,0 +1,45 @@
...
+#include <string>
+#include <vector>
+
+#include <common/common.hpp>
+#include <common/expected.hpp>
+#include <common/json.hpp>
+
+namespace mender {
+namespace common {
+namespace api {
```
vpodzime, line 27: Feel free to suggest a better name/place for this (and things like `OutgoingAPIRequest` in the future)

oleorhagen: First thing which struck me is: Should it not be under `http::`?

common/api.hpp
```diff
@@ -0,0 +1,45 @@
...
+
+namespace common = mender::common;
+namespace expected = mender::common::expected;
+
+static inline expected::ExpectedString ErrorMsgFromErrorResponse(const vector<uint8_t> &response) {
+	auto ex_j = json::Load(StringFromByteVector(response));
+	if (!ex_j) {
+		return expected::unexpected(ex_j.error());
+	} else {
```
oleorhagen, line 36: Nit: Unnecessary `else` 
(I'm writing this a lot these days)

common/api.hpp
```diff
@@ -0,0 +1,45 @@
...
+		return ex_j.value().Get("error").and_then(json::ToString);
```
oleorhagen, line 37: What a beauty :princess: :sunglasses: :wink:

mender-update/inventory.cpp
```diff
@@ -0,0 +1,171 @@
...
+//    WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+//    See the License for the specific language governing permissions and
+//    limitations under the License.
+
+#include <mender-update/inventory.hpp>
+
+#include <sstream>
+#include <string>
+
+#include <common/api.hpp>
```
oleorhagen, line 20: Then this would be `common/http/api`?

mender-update/inventory.cpp
```diff
@@ -0,0 +1,171 @@
...
+	http::Client &client,
+	events::EventLoop &loop,
+	APIResponseHandler api_handler) {
+	auto ex_inv_data = inv_parser::GetInventoryData(inventory_generators_dir);
+	if (!ex_inv_data) {
+		return ex_inv_data.error();
+	}
+
+	stringstream top_ss;
+	top_ss << "[";
```
oleorhagen, line 84: S: No need to do this manually, we have functionality for this:

```
diff --git a/mender-update/inventory.cpp b/mender-update/inventory.cpp
index fd34ba82..bdf9d583 100644
--- a/mender-update/inventory.cpp
+++ b/mender-update/inventory.cpp
@@ -14,7 +14,6 @@
 
 #include <mender-update/inventory.hpp>
 
-#include <sstream>
 #include <string>
 
 #include <common/api.hpp>
@@ -80,31 +79,11 @@ error::Error PushInventoryData(
 		return ex_inv_data.error();
 	}
 
-	stringstream top_ss;
-	top_ss << "[";
-	auto inv_data = ex_inv_data.value();
-	for (const auto &kv : inv_data) {
-		top_ss << R"({"name":")";
-		top_ss << json::EscapeString(kv.first);
-		top_ss << R"(","value":)";
-		if (kv.second.size() == 1) {
-			top_ss << "\"" + json::EscapeString(kv.second[0]) + "\"";
-		} else {
-			stringstream items_ss;
-			items_ss << "[";
-			for (const auto &str : kv.second) {
-				items_ss << "\"" + json::EscapeString(str) + "\",";
-			}
-			auto items_str = items_ss.str();
-			// replace the trailing comma with the closing square bracket
-			items_str[items_str.size() - 1] = ']';
-			top_ss << items_str;
-		}
-		top_ss << R"(},)";
+	auto exp_payload {json::Dump(ex_inv_data.value())};
+	if (!exp_payload) {
+		return exp_payload.error();
 	}
-	auto payload = top_ss.str();
-	// replace the trailing comma with the closing square bracket
-	payload[payload.size() - 1] = ']';
+	auto payload = exp_payload.value();
 
 	http::BodyGenerator payload_gen = [payload]() {
 		return make_shared<io::StringReader>(payload);
```

mender-update/inventory.cpp
```diff
@@ -0,0 +1,171 @@
...
+			}
+
+			auto resp = exp_resp.value();
+			auto status = resp->GetStatusCode();
+			if (status == http::StatusOK) {
+				api_handler(error::NoError);
+			} else {
+				auto ex_err_msg = api::ErrorMsgFromErrorResponse(*(received_body.get()));
+				string err_str;
+				if (ex_err_msg) {
```
oleorhagen, line 156: Hmm, perhaps not a bad idea actually, doing it unconditionally :thinking:

common/api.hpp
vpodzime: Yeah, until you think too much about what it does. I was actually wondering how the potential error from `Get()` gets wrapped into `expected::unexpected`. Better not think too much...

common/api.hpp
vpodzime: You could argue for both -- does this code have to care it's HTTP? And also, will there be any other Mender API used here?

mender-update/inventory.cpp
vpodzime: That doesn't convert one-item vectors to plain string values, right?

mender-update/inventory.cpp
vpodzime: Do what unconditionally?

mender-update/inventory.cpp
vpodzime: I had `http_api.hpp` for a while, but I guess it should use a subfolder, yes.

mender-update/inventory.cpp
vpodzime: You mean no matter what the status code is?

common/api.hpp
oleorhagen: Other than `HTTP` you mean?

mender-update/inventory.cpp
oleorhagen: > That doesn't convert one-item vectors to plain string values, right?

Good point, this is true!

> Plus it doesn't add the "name" and "value" things. The input key-value data doesn't have anything like that

Nope, it doesn't, and yeh, maybe ur right. That this is the way :sigh:

mender-update/inventory.cpp
vpodzime: What should the code do then? If the status code is 200:Success, but the response payload contains `"error": "Some error"`. File a bug? :wink:

common/api.hpp
kacf: How about simply `mender::api`? Then we can put `mender::api::auth` under that. I think this is better to not overbloat the already very crowded `common` namespace. Plus, `api` is already pretty generic, but still a bit narrower than `common`.

I think I vote for not putting under `http`, because the transport protocol isn't really important here IMHO. And we could add other transport protocols with similar semantics for other platforms later.

mender-update/inventory.cpp
oleorhagen: I was commenting on what you are doing here, which is doing it unconditionally, for all error values. Also undocumented ones.

kacf (approved):
Nice work!

I commented on the namespace/folder naming, but ultimately I'm fine with whatever you decide.

common/api.hpp
oleorhagen: I think I `mender::api` is a good idea.  But it wouldn't hurt if it internally had `http`, and then the `mender::api::auth` could internally use `http::`

oleorhagen: I did this in the `auth` client now: https://github.com/mendersoftware/mender/pull/1278#pullrequestreview-1487616296

mender-update/inventory.cpp
vpodzime: Oh! That's why it's best-effort -- it tries to extract the error and if it's not there, it uses the `StatusMessage` instead.

oleorhagen: Yes, exactly :)

common/api.hpp
vpodzime: The same in #1282. Adapting this PR now.

common/api.hpp
vpodzime: Usually I'm for not using `else` because it introduces an unnecessary indentation for the code inside it. But here it reads well and there's just a different (**else**) `return` in it.

PR #1275, opened 2023-06-19
feat(standalone): Add support for installing from HTTP + support commits

kacf:
Ticket: MEN-6393

### DIFF
--- a/common/events/events_io.cpp
+++ b/common/events/events_io.cpp
@@ -123,8 +123,18 @@ void AsyncWriterFromWriter::Cancel() {
 	}
 }
 
-ReaderFromAsyncReader::ReaderFromAsyncReader(AsyncReaderFromEventLoopFunc func) :
-	reader_(func(event_loop_)) {
+ReaderFromAsyncReader::ReaderFromAsyncReader() {
+}
+
+mio::ExpectedReaderPtr ReaderFromAsyncReader::Construct(AsyncReaderFromEventLoopFunc func) {
+	shared_ptr<ReaderFromAsyncReader> reader(new ReaderFromAsyncReader);
+	auto async_reader = func(reader->event_loop_);
+	if (!async_reader) {
+		return expected::unexpected(async_reader.error());
+	}
+
+	reader->reader_ = async_reader.value();
+	return reader;
 }
 
 mio::ExpectedSize ReaderFromAsyncReader::Read(
--- a/common/events/platform/boost/events.cpp
+++ b/common/events/platform/boost/events.cpp
@@ -23,6 +23,7 @@ namespace events {
 namespace asio = boost::asio;
 
 void EventLoop::Run() {
+	ctx_.restart();
 	ctx_.run();
 }
 
--- a/common/events_io.hpp
+++ b/common/events_io.hpp
@@ -126,18 +126,21 @@ private:
 	EventLoop &loop_;
 };
 
-using AsyncReaderFromEventLoopFunc = function<mio::AsyncReaderPtr(EventLoop &loop)>;
+using AsyncReaderFromEventLoopFunc = function<mio::ExpectedAsyncReaderPtr(EventLoop &loop)>;
 
 class ReaderFromAsyncReader : virtual public mio::Reader {
 public:
 	// Since the async reader usually needs the event loop object, we need to pass in the
 	// AsyncReaderFromEventLoopFunc factory function to create the async reader inside of the
-	// constructor.
-	ReaderFromAsyncReader(AsyncReaderFromEventLoopFunc func);
+	// constructor. Because this can sometimes fail, use a factory function which can return
+	// errors, so that we don't have to use exceptions.
+	static mio::ExpectedReaderPtr Construct(AsyncReaderFromEventLoopFunc func);
 
 	mio::ExpectedSize Read(vector<uint8_t>::iterator start, vector<uint8_t>::iterator end) override;
 
 private:
+	ReaderFromAsyncReader();
+
 	// Default-constructed EventLoop. This loop is used for one, and only one thing, namely to
 	// read from the AsyncReader. If you want to use an EventLoop which also processes other
 	// events, and you cannot use the AsyncReader directly (for example because some layer only
--- a/common/http.hpp
+++ b/common/http.hpp
@@ -345,20 +345,19 @@ private:
 
 	void CallHandler(ResponseHandler handler);
 	void CallErrorHandler(
-		const error_code &err, const OutgoingRequestPtr &req, ResponseHandler handler);
+		const error_code &ec, const OutgoingRequestPtr &req, ResponseHandler handler);
 	void CallErrorHandler(
 		const error::Error &err, const OutgoingRequestPtr &req, ResponseHandler handler);
-	void ResolveHandler(
-		const error_code &err, const asio::ip::tcp::resolver::results_type &results);
-	void ConnectHandler(const error_code &err, const asio::ip::tcp::endpoint &endpoint);
-	void HandshakeHandler(const error_code &err, const asio::ip::tcp::endpoint &endpoint);
-	void WriteHeaderHandler(const error_code &err, size_t num_written);
-	void WriteBodyHandler(const error_code &err, size_t num_written);
+	void ResolveHandler(const error_code &ec, const asio::ip::tcp::resolver::results_type &results);
+	void ConnectHandler(const error_code &ec, const asio::ip::tcp::endpoint &endpoint);
+	void HandshakeHandler(const error_code &ec, const asio::ip::tcp::endpoint &endpoint);
+	void WriteHeaderHandler(const error_code &ec, size_t num_written);
+	void WriteBodyHandler(const error_code &ec, size_t num_written);
 	void PrepareBufferAndWriteBody();
 	void WriteBody();
-	void ReadHeaderHandler(const error_code &err, size_t num_read);
+	void ReadHeaderHandler(const error_code &ec, size_t num_read);
 	void ReadHeader();
-	void ReadBodyHandler(const error_code &err, size_t num_read);
+	void ReadBodyHandler(const error_code &ec, size_t num_read);
 #endif // MENDER_USE_BOOST_BEAST
 };
 
@@ -431,13 +430,13 @@ private:
 
 	void AcceptHandler(const error_code &ec);
 	void ReadHeader();
-	void ReadHeaderHandler(const error_code &err, size_t num_read);
-	void ReadBodyHandler(const error_code &err, size_t num_read);
+	void ReadHeaderHandler(const error_code &ec, size_t num_read);
+	void ReadBodyHandler(const error_code &ec, size_t num_read);
 	void AsyncReply(ReplyFinishedHandler reply_finished_handler);
-	void WriteHeaderHandler(const error_code &err, size_t num_written);
+	void WriteHeaderHandler(const error_code &ec, size_t num_written);
 	void PrepareBufferAndWriteBody();
 	void WriteBody();
-	void WriteBodyHandler(const error_code &err, size_t num_written);
+	void WriteBodyHandler(const error_code &ec, size_t num_written);
 	void CallBodyHandler();
 	void FinishReply();
 #endif // MENDER_USE_BOOST_BEAST
--- a/common/http.hpp
+++ b/common/http.hpp
@@ -56,6 +56,8 @@ namespace expected = mender::common::expected;
 namespace io = mender::common::io;
 namespace log = mender::common::log;
 
+class Client;
+
 class HttpErrorCategoryClass : public std::error_category {
 public:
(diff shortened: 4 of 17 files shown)


mender-update/standalone/standalone.cpp
```diff
@@ -195,6 +197,66 @@ error::Error RemoveStateData(database::KeyValueDatabase &db) {
 	return db.Remove(context::MenderContext::standalone_state_key);
 }
 
+static expected::expected<pair<http::ClientPtr, io::ReaderPtr>, error::Error>
+ClientAndReaderFromUrl(const string &src) {
+	http::ClientPtr http_client;
+	auto maybe_artifact_reader = events::io::ReaderFromAsyncReader::Construct(
```
kacf, line 203: I'm not super-happy with how long this lamdba is, and I considered turning it into a reusable convenience function when using direct Readers. OTOH, I suspect we need to revisit this area when we start using the parser in daemon mode, and maybe this part will change, so maybe it's good enough for now.

I'll let you guys judge how bad it is.

common/events/events_io.cpp
```diff
@@ -123,8 +123,18 @@ void AsyncWriterFromWriter::Cancel() {
 	}
 }
 
-ReaderFromAsyncReader::ReaderFromAsyncReader(AsyncReaderFromEventLoopFunc func) :
-	reader_(func(event_loop_)) {
+ReaderFromAsyncReader::ReaderFromAsyncReader() {
+}
+
+mio::ExpectedReaderPtr ReaderFromAsyncReader::Construct(AsyncReaderFromEventLoopFunc func) {
+	shared_ptr<ReaderFromAsyncReader> reader(new ReaderFromAsyncReader);
```
oleorhagen, line 130: Q: `make_shared`?

common/io.cpp
```diff
@@ -234,6 +234,17 @@ ExpectedSize StreamReader::Read(vector<uint8_t>::iterator start, vector<uint8_t>
 	return is_->gcount();
 }
 
+expected::ExpectedSize FileSize(const string &path) {
+	// Probably not as efficient as stat(), but portable.
+	ifstream is(path, ifstream::ate | ifstream::binary);
+	if (is.bad()) {
```
oleorhagen, line 240: Q: Is this the same as `(!is)`?

common/events/platform/boost/events.cpp
```diff
@@ -23,6 +23,7 @@ namespace events {
 namespace asio = boost::asio;
 
 void EventLoop::Run() {
+	ctx_.restart();
```
oleorhagen, line 26: Yes! I had this issue

oleorhagen (commented):
I only had the brains for the small commits today

common/io.cpp
kacf: No. But maybe it's better to use it, since it catches more errors. It's annoying that the bool operator does not have a named equivalent.

common/events/events_io.cpp
kacf: The constructor is private, so I can't use `make_shared` here.

common/io.cpp
kacf: Ok, fixed this, and the one above, which is the one I took it from. I made this into a new commit, since I'm still not 100% sure about it and I'd like @vpodzime to comment.

kacf:
Added one more commit to fix a thread sanitizer issue.

common/events_io.hpp
```diff
@@ -126,18 +126,21 @@ class AsyncWriterFromWriter : virtual public mio::AsyncWriter {
...
 
 class ReaderFromAsyncReader : virtual public mio::Reader {
 public:
 	// Since the async reader usually needs the event loop object, we need to pass in the
 	// AsyncReaderFromEventLoopFunc factory function to create the async reader inside of the
-	// constructor.
-	ReaderFromAsyncReader(AsyncReaderFromEventLoopFunc func);
+	// constructor. Because this can sometimes fail, use a factory function which can return
+	// errors, so that we don't have to use exceptions.
+	static mio::ExpectedReaderPtr Construct(AsyncReaderFromEventLoopFunc func);
```
vpodzime, line 137: Now it actually, finally, makes sense to me! With what I imagine being *a factory function*.

common/http.hpp
```diff
@@ -233,20 +235,39 @@ class IncomingRequest : public Request {
...
+		~BodyAsyncReader();
+
+		error::Error AsyncRead(
+			vector<uint8_t>::iterator start,
+			vector<uint8_t>::iterator end,
+			io::AsyncIoHandler handler) override;
+		void Cancel() override;
+
+		weak_ptr<Client> client_;
+		bool done_ {false};
```
vpodzime, line 264: I guess the two above should be under `private:`?

common/http/platform/beast/http.cpp
```diff
@@ -494,22 +495,64 @@ void Client::ReadHeaderHandler(const error_code &ec, size_t num_read) {
...
+	}
+}
+
+void Client::AsyncReadNextBodyPart(
+	vector<uint8_t>::iterator start, vector<uint8_t>::iterator end, io::AsyncIoHandler handler) {
+	reader_buf_start_ = start;
+	reader_buf_end_ = end;
+	reader_handler_ = handler;
+	size_t read_size = end - start;
+	size_t smallest = (body_buffer_.size() < read_size) ? body_buffer_.size() : read_size;
```
vpodzime, line 549: `std::min()`

common/http/platform/beast/http.cpp
```diff
@@ -536,14 +579,34 @@ void Client::ReadHeaderHandler(const error_code &ec, size_t num_read) {
...
+	if (ec == http::make_error_code(http::error::need_buffer)) {
+		// This can be ignored. We always reset the buffer between reads anyway.
+		ec = error_code();
+	}
+
+	if (response_->body_async_reader_) {
+		assert(reader_handler_);
+
+		size_t buf_size = reader_buf_end_ - reader_buf_start_;
+		size_t smallest = (num_read < buf_size) ? num_read : buf_size;
```
vpodzime, line 599: Also `std::min()`

common/io.cpp
vpodzime: Yeah, the bool operator covers more cases, AFAICT. I keep forgetting the gore details, though.

common/testing.cpp
```diff
@@ -89,7 +89,14 @@ const string HttpFileServer::serve_address_ {"http://127.0.0.1:53272"};
...
 	server_(http::ServerConfig {}, loop_) {
-	thread_ = thread([this]() {
+	// The reason we need this synchronization is because of the thread sanitizer and
+	// logging. AsyncServeUrl uses the logger internally, and the log level is also set by
+	// certain tests. Since these things happen in two separate threads, we need to make sure
+	// that AsyncServeUrl has returned before we leave this function.
+	promise<bool> running;
+	auto maybe_running = running.get_future();
+
+	thread_ = thread([this, &running]() {
```
vpodzime, line 99: Why do we even need a thread here? It should be possible to use the same loop for `AsyncServeUrl()` and for the clients, just like we do in the other HTTP tests not using this new class, right?

vpodzime (commented):
Looks `promise<ing>`

common/http/platform/beast/http.cpp
```diff
@@ -536,14 +579,34 @@ void Client::ReadHeaderHandler(const error_code &ec, size_t num_read) {
...
 		logger_.Trace("Read " + to_string(num_read) + " bytes of body data from stream.");
+		response_body_read_ += num_read;
+	}
+
+	if (ec == http::make_error_code(http::error::need_buffer)) {
+		// This can be ignored. We always reset the buffer between reads anyway.
+		ec = error_code();
+	}
+
+	if (response_->body_async_reader_) {
```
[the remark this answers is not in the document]
kacf: It is possible for it to return a character count and also an error, if it managed to partially read something. This is something we apparently don't handle in [our sync I/O API](https://github.com/mendersoftware/mender/blob/feature-c%2B%2B-client/common/io.hpp#L51) due to `ExpectedSize`, so I guess it's somewhat theoretical. But no reason to forbid it I think.

common/http_test.cpp
```diff
@@ -1395,6 +1395,181 @@ TEST(HttpTest, DestroyClientBeforeRequestComplete) {
...
+
+			auto content_length = resp->GetHeader("Content-Length");
+			ASSERT_TRUE(content_length);
+			ASSERT_EQ(content_length.value(), to_string(BodyOfXes::TARGET_BODY_SIZE));
+
+			auto body_writer = make_shared<io::ByteWriter>(received_body);
+			body_writer->SetUnlimited(true);
+			auto reader = resp->MakeBodyAsyncReader();
+			// Lambdas can't be recursive, so we need to use an explicit function object
+			// to refer to ourselves.
```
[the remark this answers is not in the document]
kacf: You mean as a sort of `AsyncRepeatingReader`, which instead of calling the callback once, calls it until the stream is exhausted? That's a neat idea, and I think it should work, but only for readers, not writers, since only readers have a predictable termination condition (`err != error::NoError || read == 0`).

I don't want to spend too much time on it, but I can see how hard it is to make a convenience interface for it.

common/io.cpp
```diff
@@ -234,6 +234,17 @@ ExpectedSize StreamReader::Read(vector<uint8_t>::iterator start, vector<uint8_t>
 	return is_->gcount();
 }
 
+expected::ExpectedSize FileSize(const string &path) {
+	// Probably not as efficient as stat(), but portable.
```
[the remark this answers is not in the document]
kacf: Windows? 😄

But seriously, I don't really know. I think Zephyr has optional support for a file system? I just prefer to stick to C++11 if possible, and we can always optimize this later.

common/testing.cpp
kacf: Yes, it's possible. However, it's very tricky in the case where I'm using it, because we are calling the binary's `Main()` function. Which is long before an `EventLoop` has been created. And fishing it out during the running of `Main()` would also be tricky, and would need test callback of some sort. So the thread just makes it easier with a fully standalone server. It can almost be considered a separate process, semantically speaking.

common/http_test.cpp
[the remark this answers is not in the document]
kacf: Ah, no wait, the bool. You're right, that can be the termination condition.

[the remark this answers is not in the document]
kacf: @vpodzime: This did indeed help. Check out the last commit. I didn't add a corresponding one for Writer, mostly because then I'd have to test it, but it can easily be done later in the same manner.

Rest of the smaller fixes were squashed into existing commits.

common/io.cpp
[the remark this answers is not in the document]
vpodzime: [Windows is covered](https://learn.microsoft.com/en-us/cpp/c-runtime-library/reference/stat-functions?view=msvc-170).

I'm just leaning towards using `stat()` where possible and fallback to something less efficient, but potentially more portable where `stat()` is not available. Just like I suggested with `syncfs()`.

vpodzime (approved):
ACK

common/io.cpp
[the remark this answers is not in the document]
kacf: I think portability is more important than performance currently, at least until a performance problem has been demonstrated. One important reason is that we want to try to code-cover as many lines as possible in Linux, where we can test most effectively. So if it's possible, it's better if they share code.

Syncfs is different, because there is no alternative available in C++11.

What do you think, @oleorhagen and @lluiscampos?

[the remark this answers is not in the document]
oleorhagen: I vote for leaning on the `stdlib` everywhere we can :+1:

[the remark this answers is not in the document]
oleorhagen: I don't see performance being a problem anywhere in our client atm, except for I/O, and this we haven't spent much time on either.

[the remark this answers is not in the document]
vpodzime: What bugs me more here is that we have some (potentially fake/artificial) suspicion that `stat()` won't be available somewhere and so we don't use it. That's not a good practice in general.

In this particular case, there might be other issues because there is quite a bit of a difference between reading FS metadata and opening a file and seeking to its end. A trivial example is when this code is given a (named) pipe, for example. But there could be many more.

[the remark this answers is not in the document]
kacf: Ok, let's the drop the C++11 variant then.

But oh man, once I started pulling at this thread, a lot of things unraveled. I started by looking at C++17, which we have said is ok to use for our POSIX platforms. And they have `filesystem::file_size()`. In fact, they have everything that `boost::filesystem` has! So we don't need `boost::filesystem` at all! We still depend on it indirectly because of Boost Log, but we can remove all our own references and replace with C++17.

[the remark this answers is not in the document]
kacf: See https://github.com/mendersoftware/mender/pull/1286.

PR #1278, opened 2023-06-19
refac(mender-auth/client): Add a namespace around the client implementation

oleorhagen:


mender-auth/client/client.cpp
```diff
@@ -27,6 +27,11 @@
 #include <common/expected.hpp>
 #include <common/identity_parser.hpp>
 
+namespace mender {
+namespace auth {
+namespace http_client {
```
kacf, line 32: Maybe `http_client` is a bit much? We might use other transport protocols later. See also [this comment](https://github.com/mendersoftware/mender/pull/1274#discussion_r1234914575).

kacf (approved):

mender-auth/client/client.cpp
oleorhagen: Yeh, you're right, I'll do as we spoke about there, and subspace http, and use it in `auth`.

PR #1279, opened 2023-06-19
Add the authentication main loop to mender-auth

oleorhagen:


mender-auth/ipc/client.hpp
```diff
@@ -0,0 +1,99 @@
...
+
+	string GetServerURL() {
+		return this->server_url_;
+	}
+	string GetJWTToken() {
+		return this->jwt_token_;
+	}
+	void CacheAPIResponse(auth_client::APIResponse resp) {
+		mlog::Trace("Handle JWT Token callback");
+		if (resp) {
```
oleorhagen, line 73: Not quite sure how this should be handled?

mender-auth/ipc/client_test.cpp
```diff
@@ -0,0 +1,226 @@
...
+namespace http = mender::http;
+namespace io = mender::common::io;
+namespace path = mender::common::path;
+namespace mtesting = mender::common::testing;
+
+namespace ipc = mender::ipc;
+
+using TestEventLoop = mender::common::testing::TestEventLoop;
+
+class ListenClientTests : public testing::Test {
```
oleorhagen, line 47: The tests here are a little different.

1. is run in threads
2. The other has all the functionality in one loop, which I expect to be racy?

How should we ideally handle tests like this with multiple servers, setup, timing and ordering requirements?

mender-auth/main.cpp
```diff
@@ -12,21 +12,52 @@
...
 namespace error = mender::common::error;
+namespace events = mender::common::events;
+namespace mlog = mender::common::log;
+
+namespace ipc = mender::ipc;
+
+void SetLogLevel(conf::MenderConfig &config) {
+	auto expected_log_level {mlog::StringToLogLevel(config.daemon_log_level)};
+	if (!expected_log_level) {
+		// TODO - set to Info
```
kacf, line 49: Don't forget.

mender-auth/ipc/client_test.cpp
kacf: You are using separate objects in each thread, so I can't immediately see why it should be racy. To me it looks ok (we are doing something similar in the Timers test to cut down on test time, btw).

If you need ordering, something like what I did [here](https://github.com/mendersoftware/mender/commit/1bbb3856e5c0ef433dc0953e93fdb26d5457f1a2) is a good way.

mender-auth/ipc/client.cpp
```diff
@@ -0,0 +1,162 @@
...
+
+#include <common/common.hpp>
+#include <common/error.hpp>
+#include <common/events_io.hpp>
+#include <common/http.hpp>
+#include <common/log.hpp>
+#include <common/path.hpp>
+
+namespace mender {
+namespace ipc {
```
kacf, line 29: Being under the `mender-auth` folder, I think that the namespace should also be `mender::auth::ipc`. At least that's how we've done it in mender-update.

mender-auth/ipc/client.hpp
```diff
@@ -0,0 +1,121 @@
...
+	error::Error ListenForEvents(
+		events::EventLoop &loop,
+		http::Server &server,
+		const string &server_url,
+		http::Client &client,
+		const conf::MenderConfig &config,
+		const string &private_key_path,
+		const string &identity_script_path);
+
+	void CacheAPIResponse(const string &server_url, auth_client::APIResponse resp) {
```
kacf, line 76: AFAICS, this should only be called from within the handler. Should it be private?

mender-auth/ipc/client.cpp
```diff
@@ -0,0 +1,162 @@
...
+					auto err = auth_client::FetchJWTToken(
+						client,
+						config.server_url,
+						private_key_path,
+						identity_script_path,
+						std::bind(
+							&Caching::CacheAPIResponse,
+							this,
+							config.server_url,
+							std::placeholders::_1),
```
kacf, line 113: I'd argue that a lambda with a direct call inside would be more straightforward and easier to understand than `bind`.

mender-auth/ipc/client.cpp
```diff
@@ -0,0 +1,162 @@
...
+		return "Unknown";
+	}
+}
+
+error::Error MakeError(IPCErrorCode code, const string &msg) {
+	return error::Error(error_condition(code, IPCErrorCategory), msg);
+}
+
+namespace type {
+namespace webserver {
```
kacf, line 66: Should maybe the file also be called `webserver.cpp` instead of `client.cpp`?

oleorhagen: I understand the argument. My original thinking was that it could have multiple types, so that we have:

`client::type::<sometype>`

But mby ur right.

kacf: Yes, that's good thinking, but we'd anyway want those to be in different files, no? I'm considering that later we may have even more types than webserver and dbus, it might be "zephyr_ipc", or something like that. So it makes sense to split them, IMHO.

mender-auth/ipc/client.cpp
```diff
@@ -0,0 +1,162 @@
...
+
+error::Error ListenForEvents(events::EventLoop &loop, const conf::MenderConfig &config) {
+	mlog::Trace("Setting up the webserver");
+
+	auto client_config {http::ClientConfig(config.server_certificate)};
+	http::Client client {client_config, loop};
+
+	const string server_url {"http://127.0.0.1:8001"};
+	http::ServerConfig server_config;
+	http::Server server(server_config, loop);
```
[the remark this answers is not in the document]
oleorhagen: Oh, shite..

mender-auth/ipc/client.cpp
oleorhagen: hahaha, nooo.. I was so proud of learning `bind` and using it

mender-auth/ipc/client.cpp
oleorhagen: I called it `server.cpp` for now?

mender-auth/ipc/server.cpp
```diff
@@ -70,93 +71,85 @@ auto NoOpHeaderHandler = [](http::ExpectedIncomingRequestPtr exp_req) {};
...
+namespace server {
+error::Error Listen(events::EventLoop &loop, const conf::MenderConfig &config) {
+	log::Trace("Setting up the webserver");
 
 	const string server_url {"http://127.0.0.1:8001"};
-	http::ServerConfig server_config;
-	http::Server server(server_config, loop);
 
-	auto ipc_server {type::webserver::Caching {}};
+	type::webserver::Caching ipc_server {loop, config.server_certificate, config};
```
kacf, line 142: I think there is still the same problem here: `Caching` is being constructed on the stack, and then the function returns, destroying it, and everything inside it. I suspect the whole API needs to be a class.

kacf (changes requested):

mender-auth/ipc/client.cpp
kacf: 👍

mender-auth/ipc/server.cpp
oleorhagen: lol, ur right of course. weak

mender-auth/ipc/client.cpp
```diff
@@ -0,0 +1,162 @@
...
+				resp->SetHeader("X-MEN-JWTTOKEN", this->GetJWTToken());
+				resp->SetHeader("X-MEN-SERVERURL", this->GetServerURL());
+			} else if (req->GetPath() == "/fetchjwttoken") {
+				loop.Post([this, &config, &client, private_key_path, identity_script_path]() {
+					auto err = auth_client::FetchJWTToken(
+						client,
+						config.server_url,
+						private_key_path,
+						identity_script_path,
+						std::bind(
```
vpodzime: Great, now we have all the goods that people took from from JS to C++ :sweat_smile:

mender-auth/ipc/client.cpp
```diff
@@ -0,0 +1,162 @@
...
+				});
+			} else {
+				resp->SetStatusCodeAndMessage(404, "Not Found");
+				resp->AsyncReply([](error::Error err) {
+					if (err != error::NoError) {
+						mlog::Error("Failed to reply: " + err.message);
+					};
+				});
+			}
+			resp->SetStatusCodeAndMessage(200, "Success");
```
vpodzime: This will run even in the `404` case.

mender-auth/ipc/client.cpp
```diff
@@ -0,0 +1,162 @@
...
+						private_key_path,
+						identity_script_path,
+						std::bind(
+							&Caching::CacheAPIResponse,
+							this,
+							config.server_url,
+							std::placeholders::_1),
+						config.tenant_token);
+					if (err != error::NoError) {
+						mlog::Error("Failed to set up the fetch request: " + err.message);
```
vpodzime: There shouldn't be `200 -- Success` after this error, right?

mender-auth/ipc/client.cpp
```diff
@@ -0,0 +1,162 @@
...
+				resp->SetHeader("X-MEN-SERVERURL", this->GetServerURL());
+			} else if (req->GetPath() == "/fetchjwttoken") {
+				loop.Post([this, &config, &client, private_key_path, identity_script_path]() {
+					auto err = auth_client::FetchJWTToken(
+						client,
+						config.server_url,
+						private_key_path,
+						identity_script_path,
+						std::bind(
+							&Caching::CacheAPIResponse,
```
vpodzime: Just `CacheAPIResponse` should be enough here, AFAICT (even without the `&`).

mender-auth/ipc/client.cpp
```diff
@@ -0,0 +1,162 @@
...
+				mlog::Error("Failed to create the response: " + exp_resp.error().message);
+				return;
+			}
+			auto resp = exp_resp.value();
+
+			mlog::Debug("Got path: " + req->GetPath());
+
+			if (req->GetPath() == "/getjwttoken") {
+				resp->SetHeader("X-MEN-JWTTOKEN", this->GetJWTToken());
+				resp->SetHeader("X-MEN-SERVERURL", this->GetServerURL());
```
vpodzime: So there's no response data/payload only headers?

vpodzime (commented):
My two cents...

mender-auth/ipc/client.hpp
```diff
@@ -0,0 +1,121 @@
...
+		this->server_url_ = url;
+	}
+
+	string GetServerURL() {
+		return this->server_url_;
+	}
+
+	string GetJWTToken() {
+		return this->jwt_token_;
+	}
```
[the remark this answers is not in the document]
oleorhagen: I was pondering both approaches. No particular good reason for going with the one, except making it look like Java :clown_face: .

I can do the vars public if you like

mender-auth/ipc/client.cpp
oleorhagen: Yes. I figured it was the easiest approach, but you being the consumer, feel free to tell me otherwise

mender-auth/ipc/client.hpp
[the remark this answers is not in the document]
vpodzime: But does it make sense to set one but not the other? Changing `server_url` without (re)setting `jwt_token` doesn't make sense, right?

mender-auth/ipc/client.cpp
vpodzime: I thought this was part of some public/documented API.

oleorhagen: No, the public/documented API is DBus

mender-auth/ipc/client.hpp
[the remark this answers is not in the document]
oleorhagen: No, it's only set in one place, so we could have a `Set` function, which just sets both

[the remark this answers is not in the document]
oleorhagen: I changed it to a `Cache` method, which takes both options.

kacf (approved):
One improvement suggestion, looks good otherwise!

mender-auth/main.cpp
```diff
@@ -36,11 +55,29 @@ int main(int argc, char *argv[]) {
...
 	} else {
 		auto err = config.LoadDefaults();
-		if (error::NoError != err) {
+		if (err != error::NoError) {
 			cerr << "Failed to process command line options: " + err.String() << endl;
 			return 1;
 		}
 	}
 
+	SetLogLevel(config);
```
vpodzime, line 64: This is done by `ProcessCmdlineArgs()`.

mender-auth/main.cpp
```diff
@@ -36,11 +55,29 @@ int main(int argc, char *argv[]) {
...
+
+	auto err = ipc_server.Listen(server_url);
+	if (err != error::NoError) {
+		cerr << "Failed to start the listen loop" << endl;
+		cerr << "Error: " << err.message << endl;
+		return 1;
+	}
+
+	loop.Run();
+	cout << "Finished running the loop" << endl;
```
vpodzime, line 80: Why not `log::info()`? (and `log::error()` above)

vpodzime (commented):
Looks good to me otherwise.

mender-auth/main.cpp
oleorhagen: No reason. Fixing

mender-auth/ipc/server.cpp
```diff
@@ -0,0 +1,140 @@
...
+	const string &server_url, const string &private_key_path, const string &identity_script_path) {
+	return this->server_.AsyncServeUrl(
+		server_url,
+		NoOpHeaderHandler,
+		[this, private_key_path, identity_script_path](http::ExpectedIncomingRequestPtr exp_req) {
+			if (!exp_req) {
+				log::Error("Expected request failure: " + exp_req.error().message);
+				return;
+			}
+			auto req = exp_req.value();
```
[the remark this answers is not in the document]
oleorhagen: Yeh, I have no idea. Tbh, in cases like this, I'd expect the compiler to be pretty good at understanding where this is used and needed, so that the optimiser handles it in any cases - but might be I'm dreaming...

mender-auth/ipc/server.cpp
```diff
@@ -0,0 +1,140 @@
...
+				resp->SetHeader("X-MEN-SERVERURL", this->GetServerURL());
+				resp->SetStatusCodeAndMessage(200, "Success");
+				resp->AsyncReply([](error::Error err) {
+					if (err != error::NoError) {
+						log::Error("Failed to reply: " + err.message);
+					};
+				});
+			} else if (req->GetPath() == "/fetchjwttoken") {
+				this->loop_.Post([this, private_key_path, identity_script_path]() {
+					auto err = auth_client::FetchJWTToken(
```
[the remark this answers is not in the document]
oleorhagen: Good, point! I think ur right here actually. Let me give it a spin.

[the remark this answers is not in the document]
oleorhagen: Ur right! Good catch

PR #1284, opened 2023-06-22
Various small fixes discovered while integrating with Yocto

kacf:


lluiscampos (approved):

support/modules/rootfs-image
```diff
@@ -69,21 +85,21 @@ EOF
 
     ArtifactVerifyReboot)
         ensure_correct_root_mounted
-        if test "$(${PRINTENV} upgrade_available)" != "upgrade_available=1"; then
+        if test "$upgrade_available" != 1; then
```
oleorhagen, line 105: Nit, if right side is not quoted, then either should the left side have to be?

support/modules/rootfs-image
```diff
@@ -30,15 +33,28 @@ else
...
-    if [ "$(stat -L -c %02t%02T $active)" != "$(stat -L -c %04D /)" ]; then
-        echo "Mounted root does not match boot loader environment!" 1>&2
-        exit 1
+    if [ "$(stat -L -c %02t%02T $active)" = "$(stat -L -c %04D /)" ]; then
+        return 0
     fi
+
+    # If an upgrade is in progress then it can be on either partition, except in the Download
+    # and ArtifactInstall states (not rebooted yet).
+    if [ "$upgrade_available" = 1 ] && \
```
oleorhagen, line 63: function?

support/modules/rootfs-image
```diff
@@ -2,10 +2,13 @@
...
 set -ue
 
-# Is expected to contain two variable definitions with device paths:
-# - MENDER_ROOTFS_PART_A
-# - MENDER_ROOTFS_PART_B
-. /etc/mender/rootfs-image-v2.conf
+STATE="$1"
+FILES="$2"
+
+CONF_FILE="/var/lib/mender/mender.conf"
```
oleorhagen, line 14: Q: Do we not support overriding `/var/lib/` with an environment variable?

Also, you're not gonna agree here, but in _theory_, this could be overridden in the local config file

support/modules/rootfs-image
```diff
@@ -30,15 +33,28 @@ else
     passive_num=$MENDER_ROOTFS_PART_A_NUMBER
 fi
 passive_num_hex=$(printf '%x' $passive_num)
+upgrade_available="$(${PRINTENV} upgrade_available)"
+upgrade_available="${upgrade_available#upgrade_available=}"
```
oleorhagen, line 54: I like it, but could just as well be done on the above line, no?

support/modules/rootfs-image
```diff
@@ -30,15 +33,28 @@ else
...
 
 ensure_correct_root_mounted() {
-    if [ "$(stat -L -c %02t%02T $active)" != "$(stat -L -c %04D /)" ]; then
-        echo "Mounted root does not match boot loader environment!" 1>&2
-        exit 1
+    if [ "$(stat -L -c %02t%02T $active)" = "$(stat -L -c %04D /)" ]; then
```
oleorhagen, line 57: Also know you don't agree, but `--dereference`, and `--format`

common/error.cpp
```diff
@@ -61,6 +61,14 @@ Error Error::FollowedBy(const Error &err) const {
 	return new_err;
 }
 
+Error Error::WithContext(const std::string &context) const {
```
oleorhagen, line 64: Nice!

support/modules/rootfs-image
```diff
@@ -5,10 +5,27 @@ set -ue
 STATE="$1"
 FILES="$2"
 
+if [ -n "$(which jq >/dev/null 2>&1)" ]; then
```
oleorhagen, line 8: Hmm, safeguarding for weird which behaviour?

Why not use `command`?

support/modules/rootfs-image
```diff
@@ -5,10 +5,27 @@ set -ue
...
+    JQ_AVAILABLE=1
+else
+    JQ_AVAILABLE=0
+fi
+
 CONF_FILE="/var/lib/mender/mender.conf"
 
-MENDER_ROOTFS_PART_A="$(jq -r .RootfsPartA < "$CONF_FILE")"
-MENDER_ROOTFS_PART_B="$(jq -r .RootfsPartB < "$CONF_FILE")"
+if [ "$JQ_AVAILABLE" = 1 ]; then
```
oleorhagen, line 16: No need to quote

support/modules/rootfs-image
```diff
@@ -5,10 +5,27 @@ set -ue
...
+    MENDER_ROOTFS_PART_A="$(jq -r .RootfsPartA < "$CONF_FILE")"
+    MENDER_ROOTFS_PART_B="$(jq -r .RootfsPartB < "$CONF_FILE")"
+else
+    # Fall back to line based parsing. Vulnerable to weird JSON nesting, as well as unexpected
+    # newlines, although it is unlikely with a regular configuration file.
+
+    # Poor man's case insensitive match.
```
oleorhagen, line 23: :laughing:

oleorhagen: Is it needed though? Our format is specified? Client won't accept but one format here?

support/CMakeLists.txt
```diff
@@ -152,15 +152,3 @@ add_custom_target(uninstall-rootfs-image-module
   COMMAND ${CMAKE_COMMAND} -D CMAKE_INSTALL_COMPONENT=rootfs-image-module -P ${CMAKE_BINARY_DIR}/cmake_uninstall.cmake
   WORKING_DIRECTORY ${CMAKE_BINARY_DIR}
 )
-
-install(FILES ${CMAKE_SOURCE_DIR}/support/version
```
oleorhagen, line 156: Q: Why not just revert the commit?

oleorhagen (commented):
Nice, sry I'm late. But left a few comments. Do with them as you please

support/modules/rootfs-image
kacf: A variable should always be quoted, because it could have unexpected content, unintentionally. The right side could also be quoted, of course, but it's a constant, so there is no need. No strong preference on whether it should be, I just thought it looked nicer.

support/modules/rootfs-image
kacf: I think it's better without a function actually, since it relies on a variable. If we make a function, then either we have to pass it in, or use the global variable which then makes it less obvious where the variable is used.

support/modules/rootfs-image
kacf: Yes, but we didn't support this in the old client, so I don't think we should start adding it now. At least it's not something to prioritize here.

support/modules/rootfs-image
kacf: I couldn't find a way to combine the `#` operator with a command substitution. I think it requires that it already is in a variable.

support/modules/rootfs-image
kacf: Actually, here I kind of do agree, at least with `--format` which is very unintuitive and uncommon. This is preexisting though, I just extended it.

support/modules/rootfs-image
kacf: > Hmm, safeguarding for weird which behaviour?

Something like:

```json
{
  "MyOwnStuff": {
    "RootfsPartA": "wrong-stuff",
    "RootfsPartB": "more-wrong-stuff"
  },
  "RootfsPartA": "/dev/sda2",
  "RootfsPartB": "/dev/sda3"
}
```

But as you can imagine, it's a pretty theoretical possibility. Why would anyone do that? 🙂  

> Why not use `command`?

`command` is not available in busybox.

support/modules/rootfs-image
oleorhagen: but does it have to be done via variable substitution? :p

support/modules/rootfs-image
oleorhagen: Since we're now renaming it, is it not the opportune moment?

support/modules/rootfs-image
kacf: I can't find if the JSON standard itself is case insensitive, but at least [the Golang parser is](https://pkg.go.dev/encoding/json#Unmarshal).

support/modules/rootfs-image
oleorhagen: I'm aware, I just struck me that we're essentially doing string comparison with one string (I know sh don't care what's there)

support/modules/rootfs-image
kacf: Ah, but here also, busybox doesn't support long options. I think many of the long options are extensions, and POSIX only specifies the short options.

support/modules/rootfs-image
oleorhagen: > command is not available in busybox.

hah, what? Really

oleorhagen: Which version of `sh` do they run?

support/CMakeLists.txt
kacf: Actually, I didn't notice that this was done in one commit. And indeed I missed the removal of the `version` file itself. I'll revert what I did, and then revert the original instead. 😄

support/modules/rootfs-image
oleorhagen: On the same note, we should also use `{}` for variables then

support/modules/rootfs-image
kacf: Busybox *is* `sh`. 🙂

support/modules/rootfs-image
kacf: Wait, are you talking about the path, or the config file name?

The config file name I hadn't even thought about... But yeah, in a new major release, it is indeed a good opportunity to rename it. However, this is a pretty big change, if we want to do it I think it's better to add a task for it and fix it later in the project.

Does it sound like a good idea to split the file into `mender-auth.conf` and `mender-update.conf`?

support/modules/rootfs-image
oleorhagen: `JSON` is case sensitive, coming from Ecma

support/modules/rootfs-image
kacf: I'm sure there are other ways, but what are you suggesting? 🙂

support/modules/rootfs-image
oleorhagen: > Does it sound like a good idea to split the file into mender-auth.conf and mender-update.conf?

support/modules/rootfs-image
oleorhagen: Ahh, shite. checkmate.

support/CMakeLists.txt
oleorhagen: Not strictly needed ofc, but do as you see best :)

support/modules/rootfs-image
oleorhagen: uno pipo was what struck me, but it's not necessary. Just leave it be

support/modules/rootfs-image
kacf: AFAIK, it has no practical effect, it's more of a style issue. I don't mind it, but I'm not going to change the whole file now.

support/modules/rootfs-image
kacf: Hmm ok. I think we should probably do whatever Golang does though, unless you think it's a bad idea.

oleorhagen: Agree

support/modules/rootfs-image
kacf: Principally I agree as well. I suspect it's quite a lot of work to do this though, due to all the moving pieces that interact with `mender.conf`. For now I have added it as a client team meeting topic.

PR #1286, opened 2023-06-23
Followups after #1275.

kacf:
Followups plus the discovery and subsequent removal of Boost::filesystem, since we don't actually need it.


lluiscampos (approved):
I don't understand the first two commits (I probably missed the background).

The other two look good to me, nice to get the rid of one boos library in favor of the std library.

kacf:
@oleorhagen and I already discussed the revert, so I think I can say that it's approved (basically I made a mistake when removing his earlier code, so I reverted my commit, and then reverted his original introduction of the code instead).

oleorhagen (approved):
I guess a case could have been made for something like `url::Join` or similar

kacf:
It could. OTOH there is no alternative for URLs like there is for platform paths.

oleorhagen:
> It could. OTOH there is no alternative for URLs like there is for platform paths.

It's usually nice when it's present to handle double `/` etc

vpodzime (approved):
I'd squash the first two commits. Regarding `join()` and absolute path as the second argument -- it's how Python's `os.path.join()` works as well so I think it's quite a common practice to ignore the first argument in such cases. And :+1: for not using it for URLs, I was actually planning to change that too.

kacf:
> I'd squash the first two commits

Normally I'd do that too, but for reverts I think it's worth keeping them separate for historical reasons (so that they match 1-to-1).

kacf:
> > It could. OTOH there is no alternative for URLs like there is for platform paths.
> 
> It's usually nice when it's present to handle double `/` etc

I'll add this.

I also forgot to move the files to different directories (out of `boost_filesystem`), so I'll do that too.

vpodzime (approved):
ACK

kacf:
> I guess a case could have been made for something like `url::Join` or similar

Submitted in https://github.com/mendersoftware/mender/pull/1291.

PR #1291, opened 2023-06-27
chore: Make `http::JoinUrl` function.

kacf:


kacf:
I decided not to make a full namespace just for this one function. Especially because we already have URL handling in the `http` namespace.

vpodzime (approved):
ACK

oleorhagen:
> I decided not to make a full namespace just for this one function. Especially because we already have URL handling in the `http` namespace.

I think this was the wrong choice tbh, but what have you

PR #1292, opened 2023-06-27
Compatibility fixes for Debian Buster.

kacf:


common/http/platform/beast/http.cpp
```diff
@@ -98,8 +97,6 @@ Client::Client(ClientConfig &client, events::EventLoop &event_loop) :
 	cancelled_(make_shared<bool>(false)),
 	resolver_(GetAsioIoContext(event_loop)),
 	body_buffer_(HTTP_BEAST_BUFFER_SIZE) {
-	response_buffer_.reserve(body_buffer_.size());
```
kacf, line 101: @oleorhagen: I was wrong. This is safe, looks like the buffer is automatically resized on demand.

kacf: However, it does degrade performance significantly, because it uses the smallest buffer size available. I dug through the source code and found a way to reserve using a different approach instead.

common/CMakeLists.txt
```diff
@@ -20,6 +22,12 @@ target_sources(common_path PRIVATE path/platform/c++17/path.cpp)
 target_compile_options(common_path PRIVATE ${PLATFORM_SPECIFIC_COMPILE_OPTIONS})
 target_include_directories(common_path PRIVATE ${CMAKE_SOURCE_DIR} ${CMAKE_BINARY_DIR})
 
+check_cxx_source_compiles("#include <filesystem>
+  int main() { std::filesystem::path p(\"/\"); }" STD_FILESYSTEM_AVAILABLE)
```
vpodzime, line 26: Would [`CheckSymbolExists()`](https://cmake.org/cmake/help/latest/module/CheckSymbolExists.html) work here?

vpodzime (commented):
Interesting stuff.

common/CMakeLists.txt
kacf: It claims that it can be unreliable when the symbol has more than one overload, and they recommend using `CheckCXXSourceCompiles`. [[1]](https://cmake.org/cmake/help/latest/module/CheckCXXSymbolExists.html#command:check_cxx_symbol_exists)

common/CMakeLists.txt
```diff
@@ -20,6 +22,12 @@ target_sources(common_path PRIVATE path/platform/c++17/path.cpp)
...
+if(NOT "${STD_FILESYSTEM_AVAILABLE}" STREQUAL 1)
```
oleorhagen, line 27: Q: Is this not the same as `if(STD_FILESYSTEM_AVAILABLE)`?

common/http/platform/beast/http.cpp
oleorhagen: What did you do?

common/http/platform/beast/http.cpp
```diff
@@ -697,7 +699,10 @@ Stream::Stream(Server &server) :
 	logger_("http"),
 	socket_(server_.GetAsioIoContext(server_.event_loop_)),
 	body_buffer_(HTTP_BEAST_BUFFER_SIZE) {
-	request_buffer_.reserve(body_buffer_.size());
+	// This is equivalent to:
+	//   request_buffer_.reserve(body_buffer_.size());
+	// but compatible with Boost 1.67.
```
oleorhagen, line 704: Alright, nice!

common/http/platform/beast/http.cpp
```diff
@@ -235,7 +236,8 @@ void Client::ResolveHandler(
 
 	weak_ptr<Client> weak_client(client_active_);
 
-	beast::get_lowest_layer(*stream_).async_connect(
+	asio::async_connect(
```
oleorhagen, line 239: As long as we're just goign one layer deep - all good!

oleorhagen (approved):

common/http/platform/beast/http.cpp
kacf: I just looked at what `reserve` itself does. 🙂

PR #1296, opened 2023-07-12
chore: Add state machine framework.

kacf:
Not adding any tests now, I still need to decide what should be tested directly, and indirectly as part of usage.

Note that this is not going to the main development branch yet. I still want to validate a bit more by implementing the real work in each state.

### DIFF
--- a/common/state_machine.hpp
+++ b/common/state_machine.hpp
@@ -1,0 +1,233 @@
+// Copyright 2023 Northern.tech AS
+//
+//    Licensed under the Apache License, Version 2.0 (the "License");
+//    you may not use this file except in compliance with the License.
+//    You may obtain a copy of the License at
+//
+//        http://www.apache.org/licenses/LICENSE-2.0
+//
+//    Unless required by applicable law or agreed to in writing, software
+//    distributed under the License is distributed on an "AS IS" BASIS,
+//    WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+//    See the License for the specific language governing permissions and
+//    limitations under the License.
+
+#ifndef MENDER_COMMON_STATE_MACHINE_HPP
+#define MENDER_COMMON_STATE_MACHINE_HPP
+
+#include <queue>
+#include <unordered_map>
+#include <unordered_set>
+
+#include <common/events.hpp>
+#include <common/log.hpp>
+
+namespace mender {
+namespace common {
+namespace state_machine {
+
+using namespace std;
+
+namespace events = mender::common::events;
+namespace log = mender::common::log;
+
+template <typename ContextType, typename EventType>
+class StateMachineRunner;
+
+template <typename EventType>
+class EventPoster {
+public:
+	virtual ~EventPoster() {
+	}
+
+	virtual void PostEvent(EventType event) = 0;
+};
+
+template <typename ContextType, typename EventType>
+class State {
+public:
+	virtual ~State() {
+	}
+
+	virtual void OnEnter(ContextType &ctx, EventPoster<EventType> &poster) = 0;
+};
+
+enum class TransitionFlag {
+	Immediate,
+	Deferred,
+};
+
+template <typename ContextType, typename EventType>
+class StateMachine {
+public:
+	StateMachine() = default;
+	StateMachine(StateMachine &) = delete;
+
+private:
+	struct TransitionCondition {
+		// Note: Comparing address-of states. We don't want to rely on comparison operators
+		// in the states themselves, and we just want to know if they are different
+		// instances.
+		State<ContextType, EventType> *state;
+		EventType event;
+
+		bool operator==(const TransitionCondition &t) const {
+			return state == t.state && event == t.event;
+		}
+	};
+
+	class Hasher {
+	public:
+		size_t operator()(const TransitionCondition &obj) const {
+			return std::hash<State<ContextType, EventType> *>()(obj.state)
+				   ^ std::hash<int>()(static_cast<int>(obj.event));
+		}
+	};
+
+	State<ContextType, EventType> *current_state_;
+	unordered_map<TransitionCondition, State<ContextType, EventType> *, Hasher> transitions_;
+	unordered_set<EventType> deferred_events_;
+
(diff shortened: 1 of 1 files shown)


common/state_machine.hpp
```diff
@@ -0,0 +1,231 @@
...
+class StateMachineRunner;
+
+template <typename EventType>
+class EventPoster {
+public:
+	virtual ~EventPoster() {
+	}
+
+	virtual void PostEvent(EventType event) = 0;
+};
```
vpodzime, line 44: Will there ever be multiple classes implementing this interface? If not, I'd save those lines of code, dynamic dispatch and whatever else for other things.

common/state_machine.hpp
```diff
@@ -0,0 +1,231 @@
...
+	ContextType &ctx_;
+
+	shared_ptr<bool> cancelled_;
+	vector<StateMachine<ContextType, EventType> *> machines_;
+
+	queue<EventType> event_queue_;
+
+	// Would be nice with optional<EventLoop &> reference here, for better crash messages, but
+	// optional doesn't support references.
+	events::EventLoop *event_loop_;
```
vpodzime, line 224: What about `weak_ref`?

common/state_machine.hpp
```diff
@@ -0,0 +1,231 @@
...
+		ctx_(ctx) {
+	}
+	StateMachineRunner(StateMachineRunner &) = delete;
+
+	void PostEvent(EventType event) override {
+		event_queue_.push(event);
+		PostToEventLoop();
+	}
+
+	void RunOne() {
```
vpodzime, line 120: Should this really be `public`?

common/state_machine.hpp
```diff
@@ -0,0 +1,231 @@
...
+			return;
+		}
+
+		size_t size = event_queue_.size();
+		vector<State<ContextType, EventType> *> to_run;
+
+		for (size_t count = 0; count < size; count++) {
+			bool deferred = false;
+			auto event = event_queue_.front();
+			event_queue_.pop();
```
vpodzime, line 131: I find it weird that there is no `pop_front()`. Hit this took when I wanted to use `queue` for the `Authenticator::pending_actions_`.

common/state_machine.hpp
```diff
@@ -0,0 +1,231 @@
...
+			to_run.clear();
+
+			for (auto machine : machines_) {
+				typename StateMachine<ContextType, EventType>::TransitionCondition cond {
+					machine->current_state_, event};
+				if (machine->deferred_events_.find(event) != machine->deferred_events_.end()) {
+					deferred = true;
+				}
```
vpodzime, line 139: This can be done further below only if there's a match.

common/state_machine.hpp
```diff
@@ -0,0 +1,231 @@
...
+				if (deferred) {
+					// Put back in the queue to try later. This won't be tried
+					// again during this run, due to only making `size`
+					// attempts in the for loop.
+					event_queue_.push(event);
+				} else {
+					log::Warning(
+						"State machine event " + to_string(static_cast<int>(event))
+						+ " was not handled by any state. This is a bug and could hang the state machine.");
+					assert(!to_run.empty());
```
vpodzime, line 162: Might as well do `assert(false)` here because it's inside an `if` with an exactly opposite condition.

common/state_machine.hpp
```diff
@@ -0,0 +1,231 @@
...
+		event_queue_.push(event);
+		PostToEventLoop();
+	}
+
+	void RunOne() {
+		if (event_queue_.empty()) {
+			return;
+		}
+
+		size_t size = event_queue_.size();
```
vpodzime, line 125: This one I would personally mark as `const`, but we've discussed this before :)

vpodzime (commented):
Looks really nice to me so far! :+1:

common/state_machine.hpp
kacf: It's pretty unlikely, but this interface is not about that. It's about restricting the view of the objects that the states get as arguments. They should not be allowed to tinker with the state runner. All they should see is an interface where they can post events, and that's it.

common/state_machine.hpp
kacf: I take it you mean `weak_ptr`? It's not really what we need, because it's not the ownership we are tracking. What we are actually doing is:

```
bool event_loop_is_set_;
events::EventLoop *event_loop_;
```

but instead of a separate bool, we are taking `nullptr` to mean that the bool is false.

But I agree it could be improved with something similar. I'm changing it to take a shared pointer with a null deleter instead. That means we still don't own the object, but it will give us better messages if we ever use the null pointer by mistake.

common/state_machine.hpp
kacf: It made sense to me to have it just as an alternative if not used with an event loop, but I suppose we could hide it for now, until we actually need it. Made it private.

common/state_machine.hpp
kacf: I know. Especially since some of the other structures which are more flexible have both. So it would mirror the interfaces better if `queue` also had it.

common/state_machine.hpp
kacf: I know, I left it in mostly as "documentation". It's debug mode only anyway, so I think it's fine to leave it.

common/state_machine.hpp
kacf: No, this is important to do also when there is no match. The first match is broader, and matches if the event is marked as deferred in *any* transition. The next match requires exact transition match.

And if there is no transition match, then `to_run` will be empty, and the behavior is different based on whether we have a `deferred` event or not. If it's deferred, then we save it for later processing, if it's not, then it's a potential error.

common/state_machine.hpp
kacf: I might as well make this one const, since we are always supposed to process the number of events that were in the queue when we entered the function, never more.

common/state_machine.hpp
vpodzime: Right, `deferred` is defined in the outer loop, not the inner one that `continue` advances. :+1:

vpodzime (approved):
ACK

PR #1297, opened 2023-07-13
Deployment status API

vpodzime:
The first commit is cherry-picked from #1289, whichever gets merged first will pull it in.

### DIFF
--- a/mender-update/deployments.cpp
+++ b/mender-update/deployments.cpp
@@ -71,15 +71,15 @@ error::Error MakeError(DeploymentsErrorCode code, const string &msg) {
 	return error::Error(error_condition(code, DeploymentsErrorCategory), msg);
 }
 
-const string v1_uri = "/api/devices/v1/deployments/device/deployments/next";
-const string v2_uri = "/api/devices/v2/deployments/device/deployments/next";
+static const string check_updates_v1_uri = "/api/devices/v1/deployments/device/deployments/next";
+static const string check_updates_v2_uri = "/api/devices/v2/deployments/device/deployments/next";
 
 error::Error CheckNewDeployments(
 	context::MenderContext &ctx,
 	const string &server_url,
 	http::Client &client,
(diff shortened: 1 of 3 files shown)


mender-update/deployments.cpp
```diff
@@ -222,6 +222,94 @@ error::Error CheckNewDeployments(
 	return client.AsyncCall(v2_req, header_handler, v2_body_handler);
 }
 
+static const string
+	deployment_status_strings[static_cast<int>(DeploymentStatus::AlreadyInstalled) + 1] = {
```
kacf, line 226: Suggest adding a `DeploymentStatus::End` to the list and using that instead (without `+ 1`). It's a bit more future proof than using one of the elements.

mender-update/deployments_test.cpp
```diff
@@ -659,3 +659,225 @@ TEST_F(DeploymentsTests, TestV1APIFallbackWithError) {
...
+			EXPECT_EQ(req->GetMethod(), http::Method::PUT);
+			EXPECT_EQ(common::StringFromByteVector(received_body), expected_request_data);
+
+			auto result = req->MakeResponse();
+			ASSERT_TRUE(result);
+			auto resp = result.value();
+
+			resp->SetHeader("Content-Length", to_string(response_data.size()));
+			resp->SetBodyReader(make_shared<io::StringReader>(response_data));
+			resp->SetStatusCodeAndMessage(403, "Unauthorized");
```
kacf, line 859: 403 is actually "Forbidden". Not that it matters here, but maybe change the text just to avoid confusion with the real "401 Unauthorized"? Which would trigger a different code path in production.

kacf (approved):
Just a couple of suggestions. Looks great!

mender-update/deployments_test.cpp
vpodzime: Yeah, doesn't matter here, but better not seed confusion.

PR #1299, opened 2023-07-14
Various supporting commits for daemon state machine

kacf:


common/events/events_io.cpp
```diff
@@ -31,43 +30,23 @@ AsyncReaderFromReader::~AsyncReaderFromReader() {
...
-				// This should always be true, but let's use this as a safeguard
-				// anyway.
-				assert(reader_thread_.joinable());
-				if (reader_thread_.joinable()) {
-					reader_thread_.join();
-				}
-				handler(result);
-			}
-		});
+		in_progress_ = false;
```
vpodzime, line 37: Wasn't this actually quite the same even before? Spawning a new thread and then doing `loop.Post()` with a blocking `join()` would block the loop until the `read()` is done anyway. Or am I missing something?

vpodzime (approved):
Looks good to me, but there seems to be something ThreadSan is not happy about.

common/events/events_io.cpp
kacf: No, the `Post()` was not done until the `Read()` had finished, so the sequence would be this:

1. `Read()` is called in thread 2, while thread 1 returns and continues processing other events.
2. `Read()` finishes in thread 2, and `Post()` is called.
3. The event loop is running only on thread 1, so it picks up the `Post()`ed task and runs the handler on thread 1.

But now we block instead. And the reason this is ok is, as we talked about, that deep inside the `Read()` we start processing events again, inside the call stack of the tar reader callback.

kacf: And, just to be clear, that last part with processing inside the callback: I have not added it to this PR yet. I want to test it a bit more first. But it's coming.

kacf:
> Looks good to me, but there seems to be something ThreadSan is not happy about.

Right. I'm just going to skip that test for now. It will almost certainly be fixed by my upcoming event handling fixes.

common/events/events_io.cpp
vpodzime: > No, the `Post()` was not done until the `Read()` had finished

Right! Not so easy to see this in the diff.

PR #1300, opened 2023-07-17
Deployment logs pushing

vpodzime:


mender-update/deployments.cpp
```diff
@@ -314,6 +315,93 @@ error::Error PushStatus(
...
 }
 
+static const string logs_uri_suffix = "/log";
+
+error::Error PushLogs(
+	const string &deployment_id,
+	LogMessages messages,
+	const string &server_url,
+	http::Client &client,
+	events::EventLoop &loop,
```
kacf, line 325: Should remove this, see https://github.com/mendersoftware/mender/pull/1301.

mender-update/deployments.cpp
```diff
@@ -314,6 +315,93 @@ error::Error PushStatus(
...
+	LogsAPIResponseHandler api_handler) {
+	if (messages.size() == 0) {
+		// nothing to do
+		return error::NoError;
```
kacf, line 329: I would argue that returning no error, but then not calling `api_handler`, goes against our typical async behavior, and could be surprising. I have three suggestions:

1. Return a "nothing-to-do" kind of error.
2. Send a log anyway, consisting of one message, "No messages".
3. Call `api_handler` right away with success. In this case it can probably be called directly, instead of using `Post` since there is no waiting involved.

I don't have a strong preference, but when deciding, keep in mind that this is extremely unlikely to happen in practice, because even just the client itself will produce messages before an update even gets going.

mender-update/deployments.cpp
```diff
@@ -314,6 +315,93 @@ error::Error PushStatus(
 		});
 }
 
+static const string logs_uri_suffix = "/log";
+
+error::Error PushLogs(
+	const string &deployment_id,
+	LogMessages messages,
```
kacf, line 322: One thing which only occurred to me now, is that messages are stored in JSON format in the deployment log files, aka [MEN-6578](https://northerntech.atlassian.net/browse/MEN-6578), which you are now working on. And some logs can get pretty big, so I wonder if this should take a reader instead of a vector, and that you can simply assume that the content is line-by-line valid JSON (and if it's not, you could insert the broken line into its own message, perhaps, with the same timestamp as the previous valid line; useful if the device crashed).

Sorry, I should have thought of this sooner.

kacf: I also updated https://northerntech.atlassian.net/browse/MEN-6578.

mender-update/deployments.cpp
vpodzime: Yup! Too much copy pasta going all the way from the original `GetJWTToken()` I used for inspiration :sweat_smile:

mender-update/deployments.cpp
vpodzime: > not calling `api_handler`, goes against our typical async behavior, and could be surprising

100% agree, I've been trying to keep an eye on this myself. But briefly forgot about it when I did this optimization/simplification. I like the 3rd option the best, it makes most sense to me.

mender-update/deployments.cpp
```diff
@@ -313,6 +315,160 @@ error::Error PushStatus(
...
+	}
+}
+
+const vector<uint8_t> JsonLogMessagesReader::header_ = {
+	'{', '"', 'm', 'e', 's', 's', 'a', 'g', 'e', 's', '"', ':', '['};
+const vector<uint8_t> JsonLogMessagesReader::closing_ = {']', '}'};
+
+ExpectedSize JsonLogMessagesReader::Read(
+	vector<uint8_t>::iterator start, vector<uint8_t>::iterator end) {
+	if (header_rem_ > 0) {
```
kacf, line 345: Nice three-split solution with header/raw/closing here! 👍

kacf (approved):
Some small nitpicks/suggestions, but looks good overall!

common/io/io.cpp
```diff
@@ -227,6 +227,19 @@ ExpectedIfstream OpenIfstream(const string &path) {
 	return ExpectedIfstream(std::move(is));
 }
 
+ExpectedSharedIfstream OpenSharedIfstream(const string &path) {
+	auto is = make_shared<ifstream>();
```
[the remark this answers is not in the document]
vpodzime: Looks like something that should work! :+1: Let me see.

common/io/io.cpp
```diff
@@ -227,6 +227,19 @@ ExpectedIfstream OpenIfstream(const string &path) {
...
+	errno = 0;
+	is->open(path);
+	if (!(*is.get())) {
```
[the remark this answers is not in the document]
vpodzime: :+1: copy-pasta

common/io/io.cpp
```diff
@@ -289,6 +289,26 @@ ExpectedSize StreamReader::Read(vector<uint8_t>::iterator start, vector<uint8_t>
...
+		is_ = ex_is.value();
+	}
+	if (!(*is_.get())) {
+		return Error(std::error_condition(std::errc::io_error), "Bad stream, cannot rewind");
+	}
+	is_->seekg(0, ios::beg);
+	if (!(*is_.get())) {
+		return Error(
+			std::error_condition(std::errc::io_error),
+			"Failed to seek to the beginning of the stream");
```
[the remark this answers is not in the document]
vpodzime: I'm not quite sure and it's hard to test, but let's give it a try.

PR #1301, opened 2023-07-18
chore: Remove unneeded `EventLoop::Post` in `CheckNewDeployments`.

kacf:
Doing this is unnecessary because the `AsyncCall()` that it performs is already an asynchronous call which will be posted on the event loop, so there is no need to post twice. This has the convenient side effect of getting rid of the `EventLoop` argument.

Also add error handling for the `AsyncCall()`, which was missing, and remove the loop argument from many API functions that don't actually use it.


vpodzime (approved):
ACK

PR #1302, opened 2023-07-18
Daemon mode and state machine support commits

kacf:
More support commits for daemon mode and state machine.

### DIFF
--- a/common/events/events_io.cpp
+++ b/common/events/events_io.cpp
@@ -75,33 +75,44 @@ void AsyncWriterFromWriter::Cancel() {
 	assert(!in_progress_);
 }
 
-ReaderFromAsyncReader::ReaderFromAsyncReader() {
+ReaderFromAsyncReader::ReaderFromAsyncReader(EventLoop &event_loop, mio::AsyncReaderPtr reader) :
+	event_loop_(event_loop),
+	reader_(reader) {
 }
 
-mio::ExpectedReaderPtr ReaderFromAsyncReader::Construct(AsyncReaderFromEventLoopFunc func) {
-	shared_ptr<ReaderFromAsyncReader> reader(new ReaderFromAsyncReader);
-	auto async_reader = func(reader->event_loop_);
-	if (!async_reader) {
-		return expected::unexpected(async_reader.error());
-	}
-
-	reader->reader_ = async_reader.value();
-	return reader;
+ReaderFromAsyncReader::ReaderFromAsyncReader(EventLoop &event_loop, mio::AsyncReader &reader) :
+	event_loop_(event_loop),
+	// For references, just use a destructor-less pointer.
+	reader_(&reader, [](mio::AsyncReader *) {}) {
 }
 
 mio::ExpectedSize ReaderFromAsyncReader::Read(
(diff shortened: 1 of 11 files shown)


mender-update/update_module/v3/platform/c++17/update_module_call.cpp
```diff
@@ -61,35 +61,57 @@ error::Error UpdateModule::StateRunner::AsyncCallState(
...
+			} else {
+				content.append("{...}");
+			}
+			auto lines = mender::common::SplitString(content, "\n");
+			for (auto line : lines) {
+				log::Info(prefix + line);
+			}
+		}
+		string prefix;
+	} stderr_handler {"Update Module output (stderr): "};
```
vpodzime, line 86: This could be just a lambda assigned to a variable, can't it?

common/events/events_io.cpp
```diff
@@ -30,12 +30,16 @@ AsyncReaderFromReader::~AsyncReaderFromReader() {
...
 	vector<uint8_t>::iterator start, vector<uint8_t>::iterator end, mio::AsyncIoHandler handler) {
-	// Simple, "cheating" implementation, we just do it synchronously.
-	in_progress_ = true;
-	loop_.Post([this, start, end, handler]() {
-		auto result = reader_->Read(start, end);
-		in_progress_ = false;
-		handler(result);
+	cancelled_ = make_shared<bool>(false);
+	auto cancelled = cancelled_;
+	loop_.Post([this, cancelled, start, end, handler]() {
```
vpodzime, line 35: Wouldn't `cancelled_` (i.e. captured as a copy) be the same here?

vpodzime (approved):
Looks good to me otherwise.

common/events/events_io.cpp
kacf: That would make perfect sense, but C++ doesn't allow you to capture member variables directly, you have to make them into local variables first.

However, I realize now that I can make the local variable a reference, so that I avoid at least one copy (still one in the capture, but that's intentional).

mender-update/standalone/standalone.cpp
```diff
@@ -197,64 +197,56 @@ error::Error RemoveStateData(database::KeyValueDatabase &db) {
...
+		return expected::unexpected(err);
+	}
+	error::Error inner_err;
+	io::AsyncReaderPtr reader;
+	err = http_client.AsyncCall(
+		req,
+		[&loop, &inner_err, &reader](http::ExpectedIncomingResponsePtr exp_resp) {
+			if (!exp_resp) {
+				inner_err = exp_resp.error();
+				return;
```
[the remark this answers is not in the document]
kacf: Indeed it should! Good catch!

mender-update/update_module/v3/platform/c++17/update_module_call.cpp
kacf: Hmm, I don't think so.. I'm reusing it below with a different variable. Do you see a way to do both with just one lambda?

vpodzime: Oh, I didn't see the `stdout_handler` instance. This class definition combined with an instance definition confused me.

PR #1319, opened 2023-07-24
fix: do not scan bootstrap artifact if it does not exists.

merlin-northern:
fix: do not scan bootstrap artifact if it does not exists.

Changelog: Title
Ticket: MEN-6618

### DIFF
--- a/cli/cli.go
+++ b/cli/cli.go
@@ -1,16 +1,16 @@
-// Copyright 2022 Northern.tech AS
+// Copyright 2023 Northern.tech AS
 //
-//    Licensed under the Apache License, Version 2.0 (the "License");
-//    you may not use this file except in compliance with the License.
-//    You may obtain a copy of the License at
+//	Licensed under the Apache License, Version 2.0 (the "License");
+//	you may not use this file except in compliance with the License.
+//	You may obtain a copy of the License at
 //
-//        http://www.apache.org/licenses/LICENSE-2.0
+//	    http://www.apache.org/licenses/LICENSE-2.0
 //
-//    Unless required by applicable law or agreed to in writing, software
-//    distributed under the License is distributed on an "AS IS" BASIS,
-//    WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
-//    See the License for the specific language governing permissions and
-//    limitations under the License.
+//	Unless required by applicable law or agreed to in writing, software
+//	distributed under the License is distributed on an "AS IS" BASIS,
+//	WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+//	See the License for the specific language governing permissions and
+//	limitations under the License.
 package cli
 
 import (
--- a/cli/commands.go
+++ b/cli/commands.go
@@ -1,16 +1,16 @@
 // Copyright 2023 Northern.tech AS
 //
-//    Licensed under the Apache License, Version 2.0 (the "License");
-//    you may not use this file except in compliance with the License.
-//    You may obtain a copy of the License at
+//	Licensed under the Apache License, Version 2.0 (the "License");
+//	you may not use this file except in compliance with the License.
+//	You may obtain a copy of the License at
 //
-//        http://www.apache.org/licenses/LICENSE-2.0
+//	    http://www.apache.org/licenses/LICENSE-2.0
 //
-//    Unless required by applicable law or agreed to in writing, software
-//    distributed under the License is distributed on an "AS IS" BASIS,
-//    WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
-//    See the License for the specific language governing permissions and
-//    limitations under the License.
+//	Unless required by applicable law or agreed to in writing, software
+//	distributed under the License is distributed on an "AS IS" BASIS,
+//	WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+//	See the License for the specific language governing permissions and
+//	limitations under the License.
 package cli
 
 import (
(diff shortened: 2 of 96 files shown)


merlin-northern:
Kristian: I figured: if the file does not exist, why bother parsing? does that make any sense?

cli/cli.go
```diff
@@ -1,16 +1,16 @@
...
-//    Unless required by applicable law or agreed to in writing, software
-//    distributed under the License is distributed on an "AS IS" BASIS,
-//    WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
-//    See the License for the specific language governing permissions and
-//    limitations under the License.
+//	Unless required by applicable law or agreed to in writing, software
+//	distributed under the License is distributed on an "AS IS" BASIS,
+//	WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+//	See the License for the specific language governing permissions and
+//	limitations under the License.
```
kacf, line 13: Please put this in the other commit.

cli/cli.go
```diff
@@ -508,13 +508,21 @@ func (runOptions *runOptionsType) handleCLIOptions(ctx *cli.Context) error {
...
 		}
+	case "show-provides":
+		_, err = os.Stat(config.BootstrapArtifactFile)
+		if err == nil {
+			log.Debug("calling doHandleBootstrapArtifact")
+			err = doHandleBootstrapArtifact(config, runOptions)
+			if err != nil {
+				log.Errorf("Error while handling bootstrap Artifact, continuing: %s", err.Error())
+			}
+		}
```
kacf, line 525: This won't work correctly, you are skipping the entire handling, including the handling for when the bootstrap artifact is not present. With this the client would not get any name if the artifact is not present, and would probably misbehave on the first update.

The problem is likely somewhere in [this function](https://github.com/mendersoftware/mender/blob/master/app/mender.go#L667), but it still has to do its job.

kacf (changes requested):

cli/cli.go
merlin-northern: okey, I need some education on the bootstrap artifact.

merlin-northern: my reasoning was: if the bootstrap artifact file does not exist then there is no point in calling `doHandleBootstrapArtifact`, but apparently I missed something?

kacf: Yes, that case is handled inside the function I mentioned. It then needs to commit the `unknown` name, and prevent the bootstrap artifact from being considered later by writing an entry into the database.

My gut feeling is that the speed has something to do with the database loading, and especially `LoadStateData`, which requires a write transaction, even when reading (long story). This could be optimized, but it would be useful to know whether this actually is the problem.

Perhaps you could time a few key points in that function? The most likely candidates are `LoadStateData` and `LoadProvides`, but there could of course be something else as well.

merlin-northern: ok, but if the file does not exist, `HandleBootstrapArtifact` returns nil [at mender.go:749](https://github.com/mendersoftware/mender/blob/master/app/mender.go#L749) and does nothing.

merlin-northern: okey, discussed privately. I am on it in other words.

cli/cli.go
merlin-northern: sure. moved.

cli/cli.go
merlin-northern: okey, @kacf what do you think now? I want to add tests, make it prettier, move some functions a bit, but the general idea is here.

merlin-northern:
@mender-test-bot start pipeline please honey

cli/cli.go
merlin-northern: correction: I messed up: @kacf I will ping you for a review, I just found a new problem.

cli/commands.go
```diff
@@ -184,8 +164,41 @@ func commonInit(
...
+	dirstore, dbstore, err := commonInitDatabase(opts)
+	if err != nil {
+		dbstore.Close()
+		return err
+	}
+
+	_, err = os.Stat(config.BootstrapArtifactFile)
+	if err != nil {
+		return datastore.InitDatabase(dbstore, "unknown")
+	}
```
merlin-northern, line 199: ok I messed up here, fix coming up.

cli/cli.go
merlin-northern: @kacf ok, the current idea is much more simple, than what I did earlier today. main principle:
* when the `BootstrapArtifactFile` does not point to an existing file, let's only do a database initialization in `commonInit`, for the show-provides case. the rest stays exactly as it was. ready for review.

cli/commands.go
```diff
@@ -185,7 +192,12 @@ func commonInit(
 }
 
 func doHandleBootstrapArtifact(config *conf.MenderConfig, opts *runOptionsType) error {
-	controller, mp, err := commonInit(config, opts)
+	_, err := os.Stat(config.BootstrapArtifactFile)
+	initDbOnly := false
+	if err != nil {
+		initDbOnly = true
+	}
```
kacf, line 199: Making it dependent on the bootstrap file is not needed, IMHO, you can just pass `true` unconditionally in this function. The operations that need a full init do it again later anyway.

kacf (changes requested):

cli/commands.go
merlin-northern: sure. fixed.

tranchitella (approved):
It looks good to me.

cli/commands.go
merlin-northern: and also I tested manually to verify and be sure all is fine. the following scenarios:
* no db
* empty db
* db with only artifact_name unknown
* full db from real device
```
real	0m0.424s
user	0m0.380s
sys	0m0.044s
```
in all the cases.  (and yes, I have some ideas for the tests, just let me get some spare time after the cherry-picks)

kacf (approved):
👍 Thanks!

merlin-northern:
I am not running the full pipeline again. thanks for reviews.

merlin-northern:
hey @mender-test-bot  cherry-pick to:
 * 3.5.x

please

PR #1324, opened 2023-07-26
Deployment logger

vpodzime:


common/io/io.cpp
```diff
@@ -235,10 +235,10 @@ ExpectedSharedIfstream OpenSharedIfstream(const string &path) {
 	return make_shared<ifstream>(std::move(exp_is.value()));
 }
 
-ExpectedOfstream OpenOfstream(const string &path) {
+ExpectedOfstream OpenOfstream(const string &path, bool append) {
 	ofstream os;
 	errno = 0;
-	os.open(path);
+	os.open(path, append ? ios::app : ios::out);
```
kacf, line 241: Doesn't this need to be `append ? (ios::app | ios::out) : ios::out`?

mender-update/deployments/platform/boost_log/deployments.cpp
```diff
@@ -0,0 +1,197 @@
...
+			 << json::EscapeString(boost::posix_time::to_iso_extended_string(val.get())) << "\",";
+	}
+
+	auto level = logging::extract<mlog::LogLevel>("Severity", rec);
+	if (level) {
+		string lvl = mlog::ToStringLogLevel(level.get());
+		strm << R"("level":")" << json::EscapeString(lvl) << "\",";
+	}
+
+	strm << R"("message":")" << json::EscapeString(*rec[expr::smessage]) << "\"}";
```
kacf, line 65: `*rec[expr::smessage]`, the magic string! 😃

mender-update/deployments/platform/boost_log/deployments.cpp
```diff
@@ -0,0 +1,197 @@
...
+}
+
+static const size_t kMaxExistingLogs = 5;
+static const uintmax_t kLogsFreeSpaceRequired = 100 * 1024; // 100 KiB
```
kacf, line 69: Why not `size_t`?

mender-update/deployments/platform/boost_log/deployments.cpp
```diff
@@ -0,0 +1,197 @@
...
+
+error::Error DeploymentLog::PrepareLogDirectory() {
+	fs::path dir_path(data_store_dir_);
+	bool created = fs::create_directories(dir_path);
+	if (created) {
+		// should basically never happen, but if we happened to create the
+		// directory, it's empty and thus well-prepared
+		return error::NoError;
```
kacf, line 77: Lol, I had to read this multiple times to understand it. "Bail out.. on success??". But yeah, it makes sense!

mender-update/deployments/platform/boost_log/deployments.cpp
```diff
@@ -0,0 +1,197 @@
...
+		}
+
+		string file_name = file_path.filename().string();
+
+		if (file_name == "deployments.0000." + id_ + ".log") {
+			// this log file will be (re)used, leave it alone
+			continue;
+		}
+
+		if ((file_name.find("deployments.") != 0)
```
kacf, line 95: Should actually use `substr` here also, or? Otherwise, won't this file match: `my-deployments.0000.ID.log` ?

We also aren't matching the format between `deployments.` and `.log`, which could have implications below. Since indexes are always four characters, and the deployment ID is always a UUID, you know the expected filename length up front, as well as the position of every expected character, which could make matching easier.

mender-update/deployments/platform/boost_log/deployments.cpp
```diff
@@ -0,0 +1,197 @@
...
+	std::sort(old_logs.begin(), old_logs.end());
+
+	error_code ec;
+	fs::space_info space_info = fs::space(dir_path, ec);
+	if (ec) {
+		return error::Error(
+			ec.default_error_condition(), "Failed to check free space for log files");
+	}
+
+	while ((old_logs.size() > 0)
```
kacf, line 111: This could end up deleting *all* files, can't it? But we don't want that in the "same ID / append" case.

mender-update/deployments/platform/boost_log/deployments.cpp
```diff
@@ -0,0 +1,197 @@
...
+				return error::Error(
+					ec.default_error_condition(), "Failed to check free space for log files");
+			}
+		}
+	}
+
+	// now let's make sure old logs have an increasing index starting with 0001
+	for (ssize_t i = old_logs.size() - 1; i >= 0; i--) {
+		// "deployments.".size() == 12
+		auto second_dot_pos = old_logs[i].find('.', 12);
```
kacf, line 133: This especially, is susceptible to the `my-deployment...` issue I pointed out above.

mender-update/deployments/platform/boost_log/deployments.cpp
```diff
@@ -0,0 +1,197 @@
...
+		auto last_dot_pos = old_logs[i].find_last_of('.');
+
+		// should never happen due the filter above when populating old_logs and
+		// due to how these files are named
+		assert(second_dot_pos != string::npos);
+		assert(last_dot_pos != string::npos);
```
kacf, line 139: Since we haven't matched what's between `deployments.` and `.log`, I suspect these might be vulnerable. Especially that `second_dot_pos` and `last_dot_pos` could be the same character.

common/io/io.cpp
vpodzime: Not sure, the docs don't say. But it works with `ios::app` alone.

mender-update/deployments/platform/boost_log/deployments.cpp
vpodzime: Because 
image

mender-update/deployments/platform/boost_log/deployments.cpp
vpodzime: > Should actually use `substr` here also, or? Otherwise, won't this file match: `my-deployments.0000.ID.log` ?

`find()` returns the position of the first match and I check that it's `0` so basically the above is `file_name.starts_with("deployments.")`  (which is only available in C++20)

mender-update/deployments/platform/boost_log/deployments.cpp
vpodzime: I can add a simple check of the file size to ensure there is something between the dots.

mender-update/deployments/platform/boost_log/deployments.cpp
vpodzime: That file is excluded from `old_logs`.

mender-update/deployments/platform/boost_log/deployments.cpp
kacf: Aaaah, of course. I missed that. That probably invalidates many of my other assumptions.

mender-update/deployments/platform/boost_log/deployments.cpp
kacf: I guess this can be ignored.

mender-update/deployments/platform/boost_log/deployments.cpp
kacf: Fair enough, but it can still fail in a case like this, can't it: `deployments.0000-ID.log` (note the dash instead of dot)? Since the second dot might actually be the last dot.

mender-update/deployments/platform/boost_log/deployments.cpp
```diff
@@ -0,0 +1,197 @@
...
+			space_info = fs::space(dir_path, ec);
+			if (ec) {
+				return error::Error(
+					ec.default_error_condition(), "Failed to check free space for log files");
+			}
+		}
+	}
+
+	// now let's make sure old logs have an increasing index starting with 0001
+	for (ssize_t i = old_logs.size() - 1; i >= 0; i--) {
```
kacf, line 131: Come to think of it, after you've done deletions, if, and only if, you are in the append-case (same deployment ID), you can probably skip this step, since no rotation happens when you just append to an existing file.

oleorhagen (commented):
I have a couple of questions regarding the overarching design:

* Why not keep the log specific stuff in `common/log` ? That way, if we change (when) `boost/log` then at least there is just one place in which we have to change it
* How is this envisioned to be used in our code? I see we manually call `Begin` and `Finish`. But is there no better way to capture context? Before going through this code, I was imagining either a logger option, or parameter, or constructing a new deployment logger or similar.

mender-update/deployments/platform/boost_log/deployments.cpp
```diff
@@ -0,0 +1,197 @@
...
+	auto level = logging::extract<mlog::LogLevel>("Severity", rec);
+	if (level) {
+		string lvl = mlog::ToStringLogLevel(level.get());
+		strm << R"("level":")" << json::EscapeString(lvl) << "\",";
+	}
+
+	strm << R"("message":")" << json::EscapeString(*rec[expr::smessage]) << "\"}";
+}
+
+static const size_t kMaxExistingLogs = 5;
```
oleorhagen, line 68: Q: Should this not be at the very top?

mender-update/deployments/platform/boost_log/deployments.cpp
```diff
@@ -0,0 +1,197 @@
...
+	bool created = fs::create_directories(dir_path);
+	if (created) {
+		// should basically never happen, but if we happened to create the
+		// directory, it's empty and thus well-prepared
+		return error::NoError;
+	}
+
+	vector<string> old_logs;
+	fs::directory_iterator end_itr; // default construction yields past-the-end
+	for (fs::directory_iterator itr(dir_path); itr != end_itr; ++itr) {
```
oleorhagen, line 82: S: 

Use the auto iterator

```
	for (auto const &dir: fs::directory_iterator {dir_path}) {
```

mender-update/deployments/platform/boost_log/deployments.cpp
```diff
@@ -0,0 +1,197 @@
...
+		fs::path file_path = itr->path();
+		if (!fs::is_regular_file(file_path)) {
+			continue;
+		}
+
+		string file_name = file_path.filename().string();
+
+		if (file_name == "deployments.0000." + id_ + ".log") {
```
oleorhagen, line 90: Q: Lots and lots of string comparisons. Why not pattern/regex match it in some way?

mender-update/deployments/platform/boost_log/deployments.cpp
```diff
@@ -0,0 +1,197 @@
...
+			return error::Error(
+				ec.default_error_condition(),
+				"Failed to rename old log file '" + old_logs[i] + "'");
+		}
+	}
+
+	return error::NoError;
+}
+
+error::Error DeploymentLog::BeginLogging() {
```
oleorhagen, line 166: S: 

Why not call it just `Begin`. the variable is already something deployment logger, right?

mender-update/deployments/platform/boost_log/deployments.cpp
kacf: I would try to avoid regex if possible. It's typically a big dependency, and I suspect we have to remove it when aiming for smaller platforms. It is used in a few other places already, but let's try to avoid increasing its usage unless the alternative is very messy.

vpodzime: Yeah, I remember you had issues with it before so I tried to avoid it.

mender-update/deployments/platform/boost_log/deployments.cpp
vpodzime: It's called `DeploymentLog` and it actually begins logging when you call this function. So to me it makes sense to call it `BeginLogging()`.

mender-update/deployments/platform/boost_log/deployments.cpp
vpodzime: I don't know. I like it being close to where it's used, but can move it if we have some rule for this.

mender-update/deployments/platform/boost_log/deployments.cpp
```diff
@@ -0,0 +1,197 @@
...
+
+namespace fs = std::filesystem;
+
+namespace error = mender::common::error;
+namespace io = mender::common::io;
+namespace json = mender::common::json;
+namespace mlog = mender::common::log;
+namespace path = mender::common::path;
+
+static void JsonLogFormatter(logging::record_view const &rec, logging::formatting_ostream &strm) {
```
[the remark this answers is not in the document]
vpodzime: They are not supported by https://docs.mender.io/api/#device-api-deployments-schemas-deploymentlog AFAICT

vpodzime:
> * How is this envisioned to be used in our code? I see we manually call `Begin` and `Finish`. But is there no better way to capture context? Before going through this code, I was imagining either a logger option, or parameter, or constructing a new deployment logger or similar.

I don't see any better way. We don't want to log things twice -- to the global logger and to some deployment log logger -- and we want everything that goes to the global log between some two points in time to go to this special log with special fields and format. To me that matches a new sink nicely because it does exactly what we want. And determining the times when to start and stop this special logging is out of scope of the implementation here. Thus it has a nice explicit way to start and stop and a destructor that stops automatically (if this is part of some other `Deployment` object).

mender-update/deployments/platform/boost_log/deployments.cpp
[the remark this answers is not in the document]
oleorhagen: And `level` is actually `required`... Damn

mender-update/deployments/platform/boost_log/deployments.cpp
oleorhagen: Just check the style-guide

mender-update/deployments/platform/boost_log/deployments.cpp
oleorhagen: Yeh, just a suggestion.

But doesn't:

`deployment_logger.Begin()` 

make sense?

oleorhagen:
> > * How is this envisioned to be used in our code? I see we manually call `Begin` and `Finish`. But is there no better way to capture context? Before going through this code, I was imagining either a logger option, or parameter, or constructing a new deployment logger or similar.
> 
> I don't see any better way. We don't want to log things twice -- to the global logger and to some deployment log logger -- and we want everything that goes to the global log between some two points in time to go to this special log with special fields and format. To me that matches a new sink nicely because it does exactly what we want.
> 
> And determining the times when to start and stop this special logging is out of scope of the implementation here. Thus it has a nice explicit way to start and stop and a destructor that stops automatically (if this is part of some other `Deployment` object).

Yes, I don't agree with the sink. It's the natural approach. But wouldn't actually adding the sink to the logger be nice, something like:

```
log.AddSink(...)
```

or similar?

oleorhagen:
> > * Why not keep the log specific stuff in `common/log` ? That way, if we change (when) `boost/log` then at least there is just one place in which we have to change it
> 
> If I understand it correctly, it's not a common logging code, it's specific to `mender-update`.

But maintenance would be easier, no?

vpodzime:
> But maintenance would be easier, no?

Sure, but that applies to many other things were we decided to separate things for the sake of clarity (or whatever).

vpodzime:
> But wouldn't actually adding the sink to the logger be nice, something like:
> 
> ```
> log.AddSink(...)
> ```
> 
> or similar?

That's exactly what this is doing. Except that it wraps it into an API that hides the details and takes care of things like storing the sink reference/address to that it can be removed later.

mender-update/deployments/platform/boost_log/deployments.cpp
vpodzime: What if there is `deployments.0000.some_other_id.log` as well as `deployments.0000.matching_id.log`? I know it should never happen, but I don't think the extra rotation (i.e. fixing the names) will hurt.

mender-update/deployments/platform/boost_log/deployments.cpp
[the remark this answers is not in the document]
vpodzime: Yeah, but it's probably better to have an entry without it in the file and then report it being incomplete somewhere else because the actual message is the most important bit here.

mender-update/deployments/platform/boost_log/deployments.cpp
vpodzime: Haven't found any such rule.

oleorhagen (approved):

kacf (approved):

PR #1325, opened 2023-07-26
State saving, recovery, status submissions, and the big state test

kacf:
The line count looks scary here, but keep in mind that the huge `state_test.cpp` file is almost entirely copied from Golang, and just tweaked to fit C++, so it can mostly be ignored as far as reviewing goes.


mender-update/daemon/states.cpp
```diff
@@ -542,6 +542,14 @@ void EndOfDeploymentState::OnEnter(Context &ctx, sm::EventPoster<StateEvent> &po
 	poster.PostEvent(StateEvent::Success);
 }
 
+ExitState::ExitState(events::EventLoop &event_loop) :
+	event_loop_(event_loop) {
+}
+
+void ExitState::OnEnter(Context &ctx, sm::EventPoster<StateEvent> &poster) {
+	event_loop_.Stop();
+}
```
vpodzime, line 624: Both of these can easily be in the `.hpp` file.

mender-update/daemon/states.cpp
```diff
@@ -284,6 +284,59 @@ void UpdateDownloadState::ParseArtifact(Context &ctx, sm::EventPoster<StateEvent
...
+		}
+	}
+
+	auto err = ctx.deployment_client->PushStatus(
+		ctx.deployment.state_data->update_info.id,
+		status,
+		"",
+		ctx.mender_context.GetConfig().server_url,
+		ctx.http_client,
+		[result_handler](error::Error err) { result_handler(err); });
```
vpodzime, line 332: Can't you just use `result_handler` directly as the last argument here?

vpodzime (approved):
Looks good to me otherwise.

mender-update/daemon/states.cpp
```diff
@@ -363,29 +414,62 @@ void UpdateFailureState::OnEnterUpdateState(Context &ctx, sm::EventPoster<StateE
 			ctx.event_loop, DefaultStateHandler {poster}));
 }
 
-void UpdateSaveArtifactDataState::OnEnterUpdateState(
-	Context &ctx, sm::EventPoster<StateEvent> &poster) {
+static string AddInconsistentSuffix(const string &str) {
+	const auto &suffix = main_context::MenderContext::broken_artifact_name_suffix;
+	// `string::ends_with` is C++20... grumble
```
oleorhagen, line 511: Q: Why not add an `ends_with` and have it default to `std` if `ends_with` is present, like we have done with `unexpected` etc?

mender-update/daemon/states.cpp
```diff
@@ -399,6 +483,75 @@ void UpdateCleanupState::OnEnterUpdateState(Context &ctx, sm::EventPoster<StateE
...
+
+	poster.PostEvent(StateEvent::Success);
+}
+
+void StateLoopState::OnEnter(Context &ctx, sm::EventPoster<StateEvent> &poster) {
+	assert(ctx.deployment.state_data);
+	auto &artifact = ctx.deployment.state_data->update_info.artifact;
+
+	// Mark update as inconsistent.
+	artifact.artifact_name = AddInconsistentSuffix(artifact.artifact_name);
```
oleorhagen, line 595: Q: Why not make a new string?

mender-update/daemon/state_machine.cpp
```diff
@@ -140,9 +146,85 @@ StateMachine::DeploymentTracking::DeploymentTracking() :
...
+		deployment_tracking_.states_.SetState(deployment_tracking_.failure_state_);
+
+	} else {
+		// All other states trigger a rollback.
+		main_states_.SetState(update_check_rollback_state_);
+		deployment_tracking_.states_.SetState(deployment_tracking_.failure_state_);
+	}
+
+	auto &payload_types = ctx_.deployment.state_data->update_info.artifact.payload_types;
+	assert(payload_types.size() == 1);
```
oleorhagen, line 266: I would be careful with this I think. Why not do a check and log ? This should be handled in the `parser` and nowhere else imio

oleorhagen (commented):
Got like halfway through today.

This is gigantic!

mender-update/daemon/states.cpp
kacf: I know, but due to the sheer number of states, I prefer to keep all the code in the `.cpp` file so that it's easier to navigate.

mender-update/daemon/states.cpp
kacf: Indeed I can!

Documentation/update-modules-v3-file-api.md
```diff
@@ -152,40 +152,31 @@ if `Download` fails.
 
 #### `ArtifactRollback` state
 
-`ArtifactRollback` is only considered in some circumstances. After the
-`Download` state, Mender calls the update module with:
+`ArtifactRollback` is only considered in some circumstances. At some point,
```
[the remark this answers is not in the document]
kacf: I'd like to keep due to the example it refers to next. I changed it to

> When a rollback is being considered, Mender calls the update module

How's that?

common/testing.cpp
```diff
@@ -162,17 +162,19 @@ void HttpFileServer::Serve(http::ExpectedIncomingRequestPtr exp_req) {
...
-	}
 
-	auto exp_size = io::FileSize(file_path);
-	if (!exp_size) {
-		log::Warning("HttpFileServer: " + exp_size.error().String());
+		auto exp_size = io::FileSize(file_path);
+		if (!exp_size) {
+			log::Warning("HttpFileServer: " + exp_size.error().String());
+		}
+		resp->SetHeader("Content-Length", to_string(exp_size.value()));
```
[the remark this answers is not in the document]
kacf: Good point, fixed.

mender-update/daemon/states.cpp
kacf: Unfortunately it's a class member, so we can't add it the way we have done for standalone classes. But I can add it to `common` instead.

mender-update/daemon/states.cpp
kacf: We are..? I'm not sure I understand you.

mender-update/daemon/state_machine.cpp
kacf: Actually, this is exactly why I added an assert and nothing else. I don't expect this to ever be triggered, because it is handled in the parser. But if we ever add support to the parser, then this will need adapting too, and at that point it's good to have this assert as a reminder.

kacf:
> Got like halfway through today.
> 
> This is gigantic!

I know, sorry about that! It was difficult to make it smaller because the imported test was guiding my coding, and only at the end, when all things were passing, was I sure of the correctness of everything.

mender-update/daemon/states.cpp
oleorhagen: I meant not store it back in the artifact structure

Documentation/update-modules-v3-file-api.md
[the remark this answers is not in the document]
oleorhagen: Much better imo

common/state_machine.hpp
```diff
@@ -202,6 +204,7 @@ class StateMachineRunner : virtual public EventPoster<EventType> {
 
 		if (!to_run.empty()) {
 			for (auto &state : to_run) {
+				log::Trace("Entering state " + common::BestAvailableTypeName(*state));
```
oleorhagen, line 207: Thanks!

However, why not just give the states a named string?

mender-update/daemon/states.cpp
```diff
@@ -48,6 +48,7 @@ class DefaultStateHandler {
 
 static void DefaultAsyncErrorHandler(sm::EventPoster<StateEvent> &poster, const error::Error &err) {
 	if (err != error::NoError) {
+		log::Error(err.String());
```
oleorhagen, line 51: Q: Do the states add enough context here, or do you think we should add a note on what's up here?

oleorhagen (commented):
I have to admit, the state-machine turned out really nice I think.

Kinda jealous I didn't help writing it!

mender-update/daemon/states.cpp
kacf: Ah, now I see what you mean. Right, because it's hiding the original name... Yes, I think you're right, it's better to keep the original information and make a new string. I'll change it.

mender-update/daemon/states.cpp
```diff
@@ -284,6 +284,59 @@ void UpdateDownloadState::ParseArtifact(Context &ctx, sm::EventPoster<StateEvent
...
+			case FailureMode::RetryThenFail: // MEN-6573: Handle this
+				poster.PostEvent(StateEvent::Failure);
+				return;
+			}
+		}
+
+		poster.PostEvent(StateEvent::Success);
+	};
+
+	deployments::DeploymentStatus status;
```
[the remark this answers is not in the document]
kacf: `status_.value()` will throw an exception if you don't check the condition first.

common/state_machine.hpp
kacf: Yeah, I did consider this too. The reason I opted for this method is because the states tend to be copy-pasted a lot, and you'd have to make sure that every copy-paste was followed by a change to this string. Very easy to miss. I also couldn't find any way to static-assert that two classes don't have the same strings, so I chose this method simply because it's less error prone.

But honestly, both methods suck, and I implemented this quite late in the process, because I was trying to avoid it. I just really needed to see the flow.

Very open to suggestions here, I just considered this the "least evil" method.

mender-update/daemon/states.cpp
kacf: All of the Update Module calls already include the state as context in the error, so I think this is more or less covered.

mender-update/daemon/state_test.cpp
```diff
@@ -0,0 +1,2810 @@
...
+	for (auto &state : states) {
+		if (state.find("_Enter") == state.npos && state.find("_Leave") == state.npos
+			&& state.find("_Error") == state.npos) {
+			ret.push_back(state);
+		}
+	}
+	return ret;
+}
+
+TEST_P(StateDeathTest, StateTransitionsTest) {
```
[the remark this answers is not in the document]
kacf: There was some quite creative regex-search-and-replace to get it C++ compatible, I assure you!

mender-update/daemon/states.cpp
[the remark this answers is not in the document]
oleorhagen: I was thinking of `status`, not `status_`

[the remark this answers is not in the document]
kacf: I know, but the value assigned to `status` is `status_.value()`, which I can't access until I have checked the condition. Unless I do the other if-block first, but this just makes it more confusing, IMHO.

mender-update/daemon/states.cpp
kacf: Handled in https://github.com/mendersoftware/mender/pull/1331.

PR #1326, opened 2023-07-27
chore: Treat events with no state transitions as fatal.

kacf:
This was discussed with the team members. Since an unhandled event is
almost guaranteed to hang the state machine, then it's better to
terminate and let systemd try to restart us, in the hopes that
recovery will still work.


kacf:
@vpodzime: As we discussed.

vpodzime (approved):
ACK

PR #1327, opened 2023-07-28
Add tpm2tss engine support with openssl

jeffreysfllo24:
```
feature: Added tpm2tss engine support with openssl load key functionality

Changelog: tpm2tss engine support
Ticket: None
```
### Description

Currently, mender supports only pkcs11 format for loading keys with openssl engine. Adding functionality to load from engine `tpm2tss`.

### DIFF
--- a/client/client.go
+++ b/client/client.go
@@ -35,6 +35,7 @@ import (
 	log "github.com/sirupsen/logrus"
 
 	"github.com/mendersoftware/openssl"
+	"github.com/mendersoftware/mender/utils"
 )
 
 const (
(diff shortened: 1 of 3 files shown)


kacf:
Hey @jeffreysfllo24, thanks for this! Unfortunately we no longer accept feature additions to the Golang client. See our [contribution guide](https://github.com/mendersoftware/mender/blob/feature-c%2B%2B-client/CONTRIBUTING.md#contribution-coordination-during-client-rewrite-to-c) as well as [this announcement](https://hub.mender.io/t/mender-to-rewrite-client-using-c-and-retain-go-for-its-backend/5332).

It's very likely that we would accept something similar to the new C++ codebase, but this is work-in-progress, so not very stable at the moment, and not usable for production yet. So I propose:

1. Keep the patch in your own tree for now.
2. If you want to look into implementing the same support for the C++ client, take a look at the `feature-c++-client` branch.

PR #1328, opened 2023-07-31
chore: Use a range-based for loop on a directory iterator in inventory parser

vpodzime:
Nicer syntax and one line shorter!

Ticket: none
Changelog: none

### DIFF
--- a/common/inventory_parser/platform/c++17/inventory_parser.cpp
+++ b/common/inventory_parser/platform/c++17/inventory_parser.cpp
@@ -41,9 +41,8 @@ kvp::ExpectedKeyValuesMap GetInventoryData(const string &generators_dir) {
 		return kvp::ExpectedKeyValuesMap(data);
 	}
 
-	fs::directory_iterator end_itr; // default construction yields past-the-end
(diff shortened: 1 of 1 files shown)


oleorhagen (approved):

common/inventory_parser/platform/c++17/inventory_parser.cpp
```diff
@@ -41,9 +41,8 @@ kvp::ExpectedKeyValuesMap GetInventoryData(const string &generators_dir) {
 		return kvp::ExpectedKeyValuesMap(data);
 	}
 
-	fs::directory_iterator end_itr; // default construction yields past-the-end
-	for (fs::directory_iterator itr(dir_path); itr != end_itr; ++itr) {
-		fs::path file_path = itr->path();
+	for (auto const &entry : fs::directory_iterator {dir_path}) {
```
oleorhagen, line 44: Hmm, I don't fully remember this, but with const used like this, should it not be :+1: 
`const auto const&` ?

vpodzime: Why? The `entry` itself is not constant, you assign different values to it. And as we discussed before, we are not so strict with `const`...

PR #1331, opened 2023-08-01
Final test fixes for state machine.

kacf:


mender-update/daemon/states.cpp
```diff
@@ -194,7 +194,10 @@ void UpdateDownloadState::OnEnter(Context &ctx, sm::EventPoster<StateEvent> &pos
 		[&poster](http::ExpectedIncomingResponsePtr exp_resp) {
 			if (!exp_resp) {
 				log::Error(exp_resp.error().String());
-				poster.PostEvent(StateEvent::Failure);
```
oleorhagen, line 197: Not sure I get it, but I'll have to trust you here

kacf: The artifact reader also produces an error, which posts an event. So if we also post it here, it will be posted twice, most likely bringing the state machine into a bad state.

mender-update/daemon/state_test.cpp
```diff
@@ -2691,6 +2729,65 @@ class TestDeploymentClient : virtual public deployments::DeploymentAPI {
 	deployments::DeploymentStatus fail_status_report_status_;
 };
 
+// Normal DB, but all writes fail.
```
vpodzime, line 2905: This is not true for the code below.

vpodzime (approved):
Looks good to me otherwise.

kacf:
@oleorhagen: Updated, and I also tacked on the fix for [this](https://github.com/mendersoftware/mender/pull/1325#discussion_r1279015372) to the end.

mender-update/daemon/state_test.cpp
kacf: Indeed, it was only true when I started writing it!

PR #1339, opened 2023-08-02
Process timeout

kacf:


common/processes/platform/tiny_process_library/tiny_process_library.cpp
```diff
@@ -178,7 +178,33 @@ error::Error Process::AsyncWait(events::EventLoop &loop, AsyncWaitHandler handle
...
 
+error::Error Process::AsyncWait(
+	events::EventLoop &loop, AsyncWaitHandler handler, chrono::nanoseconds timeout) {
+	timeout_timer_.reset(new events::Timer(loop));
+
+	auto err = AsyncWait(loop, handler);
+	if (err != error::NoError) {
+		return err;
+	}
+
```
lluiscampos, line 189: Why aren't we using here the our private `async_wait_data_` (and related lock, etc)? Do we support parallel calls to the two overloads of `AsyncWait` (with and without timer)?

kacf: It's because only the members inside the `async` call in `SetupAsyncWait` are executed in a different thread, everything else is in the main thread. What's executed inside the `Post()` is also on the main thread, since it is executing on the event loop. So we don't need to worry about any members not explicitly mentioned inside the `async`.

kacf: > Do we support parallel calls to the two overloads of `AsyncWait` (with and without timer)?

No, the locks are only there because we internally use a thread. Parallel calls are not allowed.

common/processes/platform/tiny_process_library/tiny_process_library.cpp
```diff
@@ -178,7 +178,33 @@ error::Error Process::AsyncWait(events::EventLoop &loop, AsyncWaitHandler handle
...
+	timeout_timer_->AsyncWait(timeout, [this, handler](error::Error err) {
+		// Move timer here so it gets destroyed after this handler.
+		auto timer = std::move(timeout_timer_);
```
oleorhagen, line 192: Q: Are we always _sure_ that it gets moved?

oleorhagen (approved):

common/processes/platform/tiny_process_library/tiny_process_library.cpp
kacf: Yes, `unique_ptr` doesn't allow copying at all.

PR #1340, opened 2023-08-02
Add status report retrying.

kacf:


common/http/http.cpp
```diff
@@ -324,5 +326,29 @@ void OutgoingResponse::SetBodyReader(io::ReaderPtr body_reader) {
...
+ExponentialBackoff::ExpectedInterval ExponentialBackoff::NextInterval() {
+	iteration_++;
+
+	if (try_count_ > 0 && iteration_ > try_count_) {
+		return expected::unexpected(MakeError(MaxRetryError, "Exponential backoff"));
+	}
+
+	chrono::milliseconds current_interval = smallest_interval_;
+	for (int count = 0; count < iteration_; count++) {
+		if (count != 0 && count % 3 == 0) {
```
vpodzime, line 338: I'm pretty sure the `for` loop can be tweaked a bit to avoid this `if`.

common/http.hpp
```diff
@@ -533,6 +534,59 @@ class Server : public events::EventLoopObject {
 #endif // MENDER_USE_BOOST_BEAST
 };
 
+class ExponentialBackoff {
+public:
+	ExponentialBackoff(chrono::milliseconds max_interval) {
```
vpodzime, line 540: I think this should allow setting `try_count_` as well.

common/http.hpp
```diff
@@ -533,6 +534,59 @@ class Server : public events::EventLoopObject {
 #endif // MENDER_USE_BOOST_BEAST
 };
 
+class ExponentialBackoff {
```
vpodzime, line 538: Might be worth it to change this into a template so the unit doesn't always have to be milliseconds.

common/http/http.cpp
```diff
@@ -324,5 +326,29 @@ void OutgoingResponse::SetBodyReader(io::ReaderPtr body_reader) {
...
+			auto new_interval = current_interval * 2;
+			if (new_interval > max_interval_) {
+				new_interval = max_interval_;
+			}
+			if (new_interval == current_interval) {
```
vpodzime, line 343: This condition is  worth a comment, I think. It's not entirely clear at the first sight that no change means maximum reached.

vpodzime (commented):
Looks good to me otherwise.

mender-update/daemon/states.cpp
```diff
@@ -289,24 +289,87 @@ void UpdateDownloadState::ParseArtifact(Context &ctx, sm::EventPoster<StateEvent
...
+							log::Error(
+								"Unexpected error in SendStatusUpdateState wait timer: "
+								+ err.String());
+							poster.PostEvent(StateEvent::Failure);
+							return;
+						}
+
+						// Try again.
+						DoStatusUpdate(ctx, poster);
+					});
```
lluiscampos, line 372: I would add a default case here with a programming error

mender-update/daemon/states.cpp
```diff
@@ -289,24 +289,87 @@ void UpdateDownloadState::ParseArtifact(Context &ctx, sm::EventPoster<StateEvent
...
+
 			switch (mode_) {
 			case FailureMode::Ignore:
 				break;
-			case FailureMode::Fail:
-			case FailureMode::RetryThenFail: // MEN-6573: Handle this
-				poster.PostEvent(StateEvent::Failure);
+			case FailureMode::RetryThenFail:
+				if (err.code
+					== deployments::MakeError(deployments::DeploymentAbortedError, "").code) {
```
lluiscampos, line 338: Can't we do:
```suggestion
				if (err.code == deployments::DeploymentAbortedError) {
```
? 

Maybe I am missing something

common/events.hpp
```diff
@@ -75,6 +75,10 @@ class Timer : public EventLoopObject {
 	}
 
 #ifdef MENDER_USE_BOOST_ASIO
+	Timer(Timer &&other) :
```
oleorhagen, line 78: S: Just use the default:

```
Timer(Timer &&) = default
```

oleorhagen: Then move it out of the `ifdef`

oleorhagen (approved):

common/http/http.cpp
kacf: I'll just document the function better in general. I also discovered a bug while writing it: A low `try_count_` was respected, but not a high one.

common/http.hpp
kacf: Makes sense. But since all coarser timers are already compatible with it, and we don't need more precision, maybe we can keep it non-templated just to save code space?

common/http.hpp
```diff
@@ -533,6 +535,59 @@ class Server : public events::EventLoopObject {
...
```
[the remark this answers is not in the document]
kacf: Yes!

common/http_test.cpp
```diff
@@ -1681,3 +1681,127 @@ TEST(HttpsTest, CorrectDefaultCertificateStoreVerification) {
...
+		ASSERT_TRUE(exp_interval) << exp_interval.error().String();
+		EXPECT_EQ(exp_interval.value(), chrono::minutes(1)) << duration_fmt(exp_interval.value());
+
+		exp_interval = backoff.NextInterval();
+		ASSERT_TRUE(exp_interval) << exp_interval.error().String();
+		EXPECT_EQ(exp_interval.value(), chrono::minutes(1)) << duration_fmt(exp_interval.value());
+
+		exp_interval = backoff.NextInterval();
+		ASSERT_FALSE(exp_interval);
+		EXPECT_EQ(exp_interval.error().code, http::MakeError(http::MaxRetryError, "").code);
```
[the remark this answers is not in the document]
kacf: No, this is a subtlety with what `code` actually is. It is an `error_condition` which consists of a code and an error category. You can't compare the error code without also comparing the category, since there is overlap in the error code values, so for example `2` could mean many different things. So `http::MaxRetryError` cannot by itself be compared with anything without wrapping it in inside the right kind of error.

Strictly speaking we could have constructed an `error_condition` directly here, and used that instead, but because of our convenience `MakeError` functions, this is easier to write.

So I guess `code` is a somewhat misleading name, but we've become used to it.

common/events.hpp
kacf: Oh good point.

common/http.hpp
vpodzime: Yup, I didn't realize `chrono::` took care of the unit conversions.

vpodzime (approved):
ACK

PR #1341, opened 2023-08-02
feat: Add MenderContext::MatchesArtifactDepends()

vpodzime:
Using a helper function ArtifactMatchesContext() to make testing
easier.

Ticket: MEN-6511
Changelog: none

### DIFF
--- a/mender-update/context.hpp
+++ b/mender-update/context.hpp
@@ -18,6 +18,7 @@
 #include <string>
 #include <unordered_map>
 
+#include <artifact/artifact.hpp>
 #include <common/conf.hpp>
 #include <common/error.hpp>
 #include <common/expected.hpp>
@@ -34,6 +35,7 @@ namespace mender {
 namespace update {
 namespace context {
 
+namespace artifact = mender::artifact;
 namespace conf = mender::common::conf;
 namespace error = mender::common::error;
 namespace expected = mender::common::expected;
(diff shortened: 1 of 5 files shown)


vpodzime:
@kacf to me it would make more sense to add this as a member function of `MenderContext`, but you suggested it to be separate for easier testing,... Maybe I could add that member function that will simply get the data from `this` and call this new function with it?

oleorhagen (approved):
Looks good :)

mender-update/context.hpp
```diff
@@ -158,6 +162,13 @@ class MenderContext {
 	conf::MenderConfig &config_;
 };
 
+// Only here to make testing easier, use MenderContext::MatchesArtifactDepends().
```
oleorhagen, line 165: Can't you then hide it in a namespace?

vpodzime: I could, but it wouldn't make much difference.

mender-update/context/context.cpp
```diff
@@ -371,6 +376,59 @@ error::Error MenderContext::CommitArtifactData(
...
+	}
+	auto ex_provides = LoadProvides();
+	if (!ex_provides) {
+		return expected::unexpected(ex_provides.error());
+	}
+	auto &provides = ex_provides.value();
+	if (provides.find("artifact_name") == provides.end()) {
+		return expected::unexpected(MakeError(ValueError, "Missing artifact_name value"));
+	}
+	if (provides.find("artifact_group") == provides.end()) {
```
kacf, line 392: `artifact_group` can be (and usually is) missing. In fact, it has rarely been used, because we introduced it early, but never used it for its intended purpose. But since it's in the database we have to keep supporting it.

`artifact_name` is guaranteed to be present though.

PR #1343, opened 2023-08-03
chore: Switch to use authenticated server requests.

kacf:
Ticket: MEN-6582

### DIFF
--- a/mender-update/daemon/context.hpp
+++ b/mender-update/daemon/context.hpp
@@ -27,6 +27,9 @@
 
 #include <artifact/artifact.hpp>
 
+#include <api/auth.hpp>
+#include <api/client.hpp>
+
 #include <mender-update/context.hpp>
 #include <mender-update/deployments.hpp>
 #include <mender-update/update_module/v3/update_module.hpp>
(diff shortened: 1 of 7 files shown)


oleorhagen (approved):

mender-update/daemon/context.hpp
```diff
@@ -145,8 +151,10 @@ class Context {
 	mender::update::context::MenderContext &mender_context;
 	events::EventLoop &event_loop;
 
+	auth::Authenticator authenticator;
```
vpodzime, line 154: Only slightly related -- shouldn't this and some of the other members above and below be `private`? I don't think any code should be allowed to do `context.authenticator.Something` on a `Context` instance.

kacf: Yes, I suppose you're right. Most members are meant to be used directly, but I guess this is an exception since it's just a prerequisite for one of the other members. I can follow up on this.

kacf: I think it only applies to `Authenticator` though. All the others are used at some point.

PR #1344, opened 2023-08-03
chore: Produce error message when trying to call two endpoints at once.

kacf:
Instead of auto-cancelling the first one, which is quite confusing to debug.


oleorhagen (approved):

vpodzime (approved):
ACK

PR #1346, opened 2023-08-03
feat: Added private key abstraction

larsewi:

the form below:

```

<OPTIONAL LONGER DESCRIPTION>

```

### Description

This PR adds a private key abstraction and implements the `--passphrase-file` option according to ticket [MEN-6545](https://northerntech.atlassian.net/browse/MEN-6545). As the `--passphrase-file` option belongs to the `daemon` command, I also implemented some skeleton code required for parsing it as a command line argument. The logic for the `daemon` command will need to be implemented in a future PR.

The command `echo "secret" | sudo mender-auth daemon --passphrase-file=-` would now load the private key from file, using the passphrase `secret`. The private key will be available for use in the respective action pointer when developing the "daemon" logic later on.

[MEN-6545]: https://northerntech.atlassian.net/browse/MEN-6545?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


common/crypto.hpp
```diff
@@ -37,6 +37,22 @@ enum CryptoErrorCode {
 	VerificationError,
 };
 
+class PrivateKey;
+using ExpectedPrivateKey = expected::expected<PrivateKey, error::Error>;
+
+class PrivateKey {
+public:
+	static ExpectedPrivateKey LoadFromPEM(const string &private_key_path, const string &passphrase);
```
oleorhagen, line 45: Nice! 

But why `static`?

common/crypto.hpp
```diff
@@ -37,6 +37,22 @@ enum CryptoErrorCode {
...
+	static ExpectedPrivateKey LoadFromPEM(const string &private_key_path);
+#ifdef MENDER_CRYPTO_OPENSSL
+	unique_ptr<EVP_PKEY, void (*)(EVP_PKEY *)> pkey;
+
+private:
+	PrivateKey(unique_ptr<EVP_PKEY, void (*)(EVP_PKEY *)> &private_key) : pkey(std::move(private_key)){};
```
oleorhagen, line 51: Observation:

A couple of things going on here.

1. Read about `move` constructors: https://en.cppreference.com/w/cpp/language/move_constructor

I guess this made it compile for you, but it looks a little odd. Unique_ptrs handle what you need in a sense.

common/crypto.hpp
```diff
@@ -37,6 +37,22 @@ enum CryptoErrorCode {
...
+	static int CapturePrivateKeyPassphrase(char *buf, int size, int rwflag, void *u);
```
oleorhagen, line 52: Q:

It's fine, but it doesn't necessarily need to be a part of the class?

common/crypto/platform/openssl/crypto.cpp
```diff
@@ -93,6 +93,45 @@ string GetOpenSSLErrorMessage() {
 	return errorDescription;
 }
 
+// We need our own custom callback routine, as the default one will prompt for a
+// passphrase.
+int PrivateKey::CapturePrivateKeyPassphrase(char *buf, int size, int rwflag, void *u) {
+	// We'll only use this callback for reading passphrases, not for writing
+	// them.
+	assert(rwflag == 0);
```
oleorhagen, line 101: Observation:

This is fine. It would also be possible to wrap this function also, so that it's not possible to pass in anything other than `0`

common/crypto/platform/openssl/crypto.cpp
```diff
@@ -93,6 +93,45 @@ string GetOpenSSLErrorMessage() {
...
 
+// We need our own custom callback routine, as the default one will prompt for a
+// passphrase.
+int PrivateKey::CapturePrivateKeyPassphrase(char *buf, int size, int rwflag, void *u) {
+	// We'll only use this callback for reading passphrases, not for writing
+	// them.
+	assert(rwflag == 0); 
+
+	// Copy passphrase to out, out is not expected to be null terminated.
+	char *pass = (char *)u;
```
oleorhagen, line 104: Question:

Why not `static_cast` here?

common/crypto/platform/openssl/crypto.cpp
```diff
@@ -93,6 +93,45 @@ string GetOpenSSLErrorMessage() {
...
+ExpectedPrivateKey PrivateKey::LoadFromPEM(const string &private_key_path, const string &passphrase) {
+	auto private_bio_key = unique_ptr<BIO, void (*)(BIO *)>(
+		BIO_new_file(private_key_path.c_str(), "r"), bio_free_func);
+	if (!private_bio_key.get()) {
+		return expected::unexpected(MakeError(
+			SetupError, "Failed to open the private key file: " + GetOpenSSLErrorMessage()));
+	}
+
+	// Maybe there is a better way to convert a string reference into a mutable
+	// C-string?
```
oleorhagen, line 118: Yes :)

`Cpp` strings have the `c_str()` function :)

common/crypto/platform/openssl/crypto.cpp
```diff
@@ -93,6 +93,45 @@ string GetOpenSSLErrorMessage() {
...
+	memcpy(c_str, passphrase.c_str(), sizeof(c_str) + 1);
+
+	auto private_key = unique_ptr<EVP_PKEY, void (*)(EVP_PKEY *)>(
+		PEM_read_bio_PrivateKey(private_bio_key.get(), nullptr, &CapturePrivateKeyPassphrase, c_str), pkey_free_func);
+	if (private_key == nullptr) {
+		return expected::unexpected(
+			MakeError(SetupError, "Failed to load the key: " + GetOpenSSLErrorMessage()));
+	}
+
+	return ExpectedPrivateKey(PrivateKey(private_key));
```
oleorhagen, line 129: S:

No need to explicitly create the expected value, it is done implicitly if you leave it out :)

common/crypto/platform/openssl/crypto.cpp
```diff
@@ -93,6 +93,45 @@ string GetOpenSSLErrorMessage() {
...
+}
+
+ExpectedPrivateKey PrivateKey::LoadFromPEM(const string &private_key_path) {
+	return ExpectedPrivateKey(PrivateKey::LoadFromPEM(private_key_path, ""));
```
oleorhagen, line 133: Same as above :)

oleorhagen (commented):
It's a good start I think.

I left some quick observations.

However, I would like to see it used before we merge it, so that we can see if the interface makes sense.

common/crypto.hpp
kacf: Indeed, `&private_key` should be `&&private_key, and `move` should not be used here, because the `&&` should already make it movable.

common/crypto.hpp
kacf: Isn't it because it's a factory function?

common/crypto/platform/openssl/crypto.cpp
```diff
@@ -93,6 +93,45 @@ string GetOpenSSLErrorMessage() {
...
+	// Copy passphrase to out, out is not expected to be null terminated.
+	char *pass = (char *)u;
+	strncpy(buf, pass, size);
+	return (static_cast<int>(strlen(pass)) < size) ? static_cast<int>(strlen(pass)) : size;
+}
+
+ExpectedPrivateKey PrivateKey::LoadFromPEM(const string &private_key_path, const string &passphrase) {
+	auto private_bio_key = unique_ptr<BIO, void (*)(BIO *)>(
+		BIO_new_file(private_key_path.c_str(), "r"), bio_free_func);
+	if (!private_bio_key.get()) {
```
kacf, line 112: In general we try to avoid `.get()` because it sidesteps the safety of unique_ptr when debugging is turned on. In this particular case, it doesn't matter, because it's just a boolean test, but it would matter for example here:

```c++
void myfunc(char &c);
...
unique_ptr<char> p;
// Passes invalid reference, might crash sometime long after.
myfunc(*p.get());
// Crashes immediately.
myfunc(*p);
```

So I think we should just stay away from it unless we are forced to use it.

common/crypto/platform/openssl/crypto.cpp
kacf: Yes, but it's not mutable. I'd use `vector<char>`, and it's `data()` function, which is mutable.

common/crypto/platform/openssl/crypto.cpp
```diff
@@ -93,6 +93,45 @@ string GetOpenSSLErrorMessage() {
...
+			SetupError, "Failed to open the private key file: " + GetOpenSSLErrorMessage()));
+	}
+
+	// Maybe there is a better way to convert a string reference into a mutable
+	// C-string?
+	char c_str[passphrase.size() + 1];
+	memcpy(c_str, passphrase.c_str(), sizeof(c_str) + 1);
+
+	auto private_key = unique_ptr<EVP_PKEY, void (*)(EVP_PKEY *)>(
+		PEM_read_bio_PrivateKey(private_bio_key.get(), nullptr, &CapturePrivateKeyPassphrase, c_str), pkey_free_func);
```
kacf, line 123: Nice use of `unique_ptr`!

Good example of where you *do* need to use `.get()`.

common/crypto.hpp
```diff
@@ -37,6 +37,22 @@ enum CryptoErrorCode {
...
 
+class PrivateKey;
+using ExpectedPrivateKey = expected::expected<PrivateKey, error::Error>;
+
+class PrivateKey {
+public:
+	static ExpectedPrivateKey LoadFromPEM(const string &private_key_path, const string &passphrase);
+	static ExpectedPrivateKey LoadFromPEM(const string &private_key_path);
+#ifdef MENDER_CRYPTO_OPENSSL
+	unique_ptr<EVP_PKEY, void (*)(EVP_PKEY *)> pkey;
```
kacf, line 48: Should this be private with a getter instead? I'm a bit undecided...

In any case, `pkey` should be `pkey_`, following member naming style.

common/crypto/platform/openssl/crypto.cpp
kacf: You should also probably use [`copy`](https://en.cppreference.com/w/cpp/algorithm/copy) then, since again, it's safer than its C-counterpart.

common/crypto.hpp
oleorhagen: But is it needed? :thinking:

common/crypto/platform/openssl/crypto.cpp
oleorhagen: True, didn't think about the mutable part

common/crypto.hpp
kacf: Constructors can't return errors (without exceptions), so to me it looks like it's needed. But feel free to correct me, @larsewi, I'm just guessing at your motivation.

common/crypto.hpp
larsewi: This concept is totally new for me, and requires some practice on my hands. Please check if what I have now is better :)

common/crypto/platform/openssl/crypto.cpp
larsewi: Thanks for the complements, but I just copied this code from a function below 😜

larsewi:
> However, I would like to see it used before we merge it, so that we can see if the interface makes sense.

@oleorhagen, I will convert this PR to draft and continue developing on top of it. But please feel free to review it along the way.

common/crypto/platform/openssl/crypto.cpp
kacf: You know what they say: Mediocre music copies ideas from others. Great music is just copying that hasn't been detected yet!

common/crypto.hpp
```diff
@@ -37,6 +37,22 @@ enum CryptoErrorCode {
...
 
+class PrivateKey;
+using ExpectedPrivateKey = expected::expected<PrivateKey, error::Error>;
+
+class PrivateKey {
+public:
+	static ExpectedPrivateKey LoadFromPEM(const string &private_key_path, const string &passphrase);
+	static ExpectedPrivateKey LoadFromPEM(const string &private_key_path);
+#ifdef MENDER_CRYPTO_OPENSSL
+	const unique_ptr<EVP_PKEY, void (*)(EVP_PKEY *)> &pkey_;
```
kacf, line 48: You generally don't want to use `unique_ptr` and a reference (`&`) together. While not inherently wrong, it's confusing to determine ownership, and it's usually better to use one or the other, so that you are either definitely the owner, or definitely not the owner.

`unique_ptr`s are exactly that, unique, so they only allow moving pointers into them (not copying), but that should not be a problem since your constructor takes an rvalue-reference (movable reference, in layman's terms).

common/crypto/platform/openssl/crypto.cpp
```diff
@@ -93,6 +93,50 @@ string GetOpenSSLErrorMessage() {
...
+	auto callback = [](char *buf, int size, int rwflag, void *u) {
+		// We'll only use this callback for reading passphrases, not for
+		// writing them.
+		assert(rwflag == 0);
+
+		if (u == nullptr) {
+			return 0;
+		}
+
+		// NB: out is not expected to be null terminated.
```
kacf, line 119: By `out` you mean `buf`?

common/crypto/platform/openssl/crypto.cpp
larsewi: Ref. our F2F discussion, this is resolved

common/crypto.hpp
larsewi: Removed the reference operator, but then I must use move in order for it to compile. Don't know why though 🤔

kacf: Yeah, @oleorhagen and I had a long discussion about it yesterday. It's one of the dark corners of C++, and I didn't know either until we looked into it. Come talk to me if you want to know the reason! 🙂

mender-auth/main.cpp
```diff
@@ -31,14 +35,54 @@ namespace http = mender::http;
...
+		vector<struct conf::ExtraOption> extra = {
+			{"--passphrase-file",
+			 [&passphrase_file](const string &option, const string &value) {
+				 if (option == "--passphrase-file") {
+					 passphrase_file = value;
+				 }
+			 },
+			 true}};
+
+		auto success = config.ProcessCmdlineArgs(args.begin(), args.end(), extra);
```
kacf, line 85: With this the option is being added to the top level, correct? It should only be added to the `daemon` command, so IOW:

```bash
mender-client --passphrase-file myfile daemon # WRONG
mender-client daemon --passphrase-file myfile # Correct
```

If it's done like this, then I believe you don't need any changes to `conf.cpp` at all, because you can just do the same thing which is done [in mender-update's `cli.cpp`](https://github.com/mendersoftware/mender/blob/feature-c%2B%2B-client/mender-update/cli/cli.cpp). Notice how `ProcessCmdlineArgs` is called without any extra arguments, and then extra arguments are parsed afterwards in `ParseUpdateArguments` because the former stops parsing after the command name.

**Update**: Ok, so after writing this, I realized that we *don't have* a `daemon` command, because [MEN-6639](https://northerntech.atlassian.net/browse/MEN-6639) hasn't been done yet! Sorry, this little detail slipped my mind. But that ticket should be very easy, perhaps you can just combine the two so that we do it right the first time?

mender-auth/main.cpp
```diff
@@ -51,6 +95,22 @@ int main(int argc, char *argv[]) {
 		}
 	}
 
+	auto ex_passphrase = GetPassphraseFromFile(passphrase_file);
```
kacf, line 98: Remember to make it conditional, it's legal not to have a key.

larsewi:
Output from manual test using no passphrase, passphrase from file, and passphrase from stdin:
```
vagrant@hub:/ntech/mendersoftware/mender$ sudo mender-auth/mender-auth daemon
Failed to process command line options: Error during crypto library setup: Failed to load the key: error:1C800064:Provider routines::bad decrypt
error:11800074:PKCS12 routines::pkcs12 cipherfinal error
error:1C800064:Provider routines::bad decrypt
error:11800074:PKCS12 routines::pkcs12 cipherfinal error
vagrant@hub:/ntech/mendersoftware/mender$ sudo mender-auth/mender-auth daemon --passphrase-file=/home/vagrant/secret
^C
vagrant@hub:/ntech/mendersoftware/mender$ echo "secret" | sudo mender-auth/mender-auth daemon --passphrase-file=-
^C
vagrant@hub:/ntech/mendersoftware/mender$ 
```
Similarly when the private key is not encrypted, you don't have to use this option. And if you do, it is simply ignored:
```
vagrant@hub:/ntech/mendersoftware/mender$ sudo mender-auth/mender-auth daemon
^C
vagrant@hub:/ntech/mendersoftware/mender$ sudo mender-auth/mender-auth daemon --passphrase-file=/home/vagrant/secret
^C
vagrant@hub:/ntech/mendersoftware/mender$ 
```

common/crypto.hpp
larsewi: Yes I want to know !

mender-auth/main.cpp
larsewi: Giving the empty password (i.e., `""`) as an argument to `PrivateKey::LoadFromPEM()` is same as no password

kacf (approved):
Code looks good to me now, but I guess Gitlab isn't happy?

PR #1347, opened 2023-08-04
chore: Clean up access flags of daemon::Context a little bit.

kacf:


kacf:
@vpodzime: Due to constructor requirements I can't change the order, so I can't group them at the bottom, unfortunately.

vpodzime (approved):
ACK

PR #1350, opened 2023-08-08
chore: Respect type info depends in MenderContext::MatchesArtifactDepends()

vpodzime:
Actually done in the ArtifactMatchesContext() helper function which needs to check values from artifact::HeaderView::type_info (if any).

Also handle cases when `artifact_group` is missing in the local provides data gracefully (the value is not required to be present).

Ticket: MEN-6511
Changelog: none

### DIFF
--- a/common/common.hpp
+++ b/common/common.hpp
@@ -22,6 +22,7 @@
 
 #include <algorithm>
 #include <string>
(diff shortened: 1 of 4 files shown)


kacf (approved):
Perfect, thanks!

PR #1351, opened 2023-08-09
fix(paths): Get default variables through functions

oleorhagen:
This way, the variables can be added to the environment during runtime.

Ticket: MEN-6673
Changelog: None

-------


vpodzime (commented):
Is the thing that this allows really desirable? Regarding the ticket -- the `MenderConfig::data_store_dir` can be used to set the value which `MenderContext` then uses. Shouldn't that variable be respected/propagated? Using environment variables instead of global variables is cheap cheating. :wink:

kacf (changes requested):
Indeed, `data_store_dir` needs to be respected, otherwise this doesn't solve the ticket. That value probably needs to be stored in a non-const global variable which is changable, so that command line arguments can change it.

common/io/io.cpp
```diff
@@ -245,7 +245,7 @@ ExpectedOfstream OpenOfstream(const string &path, bool append) {
 			generic_category().default_error_condition(io_errno),
 			"Failed to open '" + path + "' for writing")));
 	}
-	return ExpectedOfstream(std::move(os));
+	return os;
```
vpodzime, line 248: This is the same, the original code is just more explicit which I like better. But YMMV.

common/conf.hpp
```diff
@@ -92,9 +92,128 @@ class CmdlineOptionsIterator {
 	ArgumentsMode mode_ {ArgumentsMode::RejectBareArguments};
 };
 
+// NOTE - When updating this class - either adding or removing variables. Be
+// sure to update the transitive dependencies also.
+class DefaultConf {
+private:
+	string DefaultPathConfDir = conf::GetEnv("MENDER_CONF_DIR", path::Join("/etc", "mender"));
```
vpodzime, line 99: 4 things here, all YMMV-style, but I'd:
- call this `Paths`
- make it a `struct`
- not call things `Default` when they actually hold the current value that no longer necessarily matches the actual/original default
- move this into `MenderConfig` because there should not be any (other) instances outside of it

vpodzime (commented):
My 2 cents again.

common/io/io.cpp
kacf: I don't like the original because it makes it look more like an error return. They are visually easier to distinguish when they look like this, I think:

```c++
return os;

# as opposed to

return ExpectedOfstream(std::move(os));
```

I think it also expresses intent better, since `Expected` is kind of an implementation detail in this context. What you really want is `os`.

That being said, I am glad that `expected::unexpected` is required in error context, because of the type ambiguity it can lead to otherwise.

mender-auth/ipc/server.hpp
```diff
@@ -54,13 +54,14 @@ class Caching {
...
-		server_ {server_config_, loop} {};
+		server_ {server_config_, loop},
+		default_identity_script_path {config.default_config.GetDefaultIdentityScript()} {};
 
 	Caching(Caching &&) = default;
 
 	error::Error Listen(
 		const string &server_url,
-		const string &identity_script_path = conf::paths::DefaultIdentityScript,
+		const string &identity_script_path = "",
```
kacf, line 64: This shouldn't have a default if we can't provide one, or? Unless you want to use `default_identity_script_path`, but it seems unused.

mender-auth/ipc/server.hpp
```diff
@@ -98,6 +99,7 @@ class Caching {
 	http::Client client_;
 	http::ServerConfig server_config_;
 	http::Server server_;
+	string default_identity_script_path;
```
kacf, line 102: You probably meant to use this in the logic somewhere, but I can't see that it is..?

kacf (commented):
Nice, sometimes verbose is just the right way to go!

common/conf.hpp
kacf: Agree with 1 and 3. I don't think it should be a struct because some of the setters are doing non-trivial things, and all the members are private for that reason.

Number 4.. I mean we could, but I don't see high value in it, and I suppose keeping it outside is easier to test. The namespace isn't too crowded anyway.

common/io/io.cpp
vpodzime:
```
return ExpectedOfstream(std::move(os));
```
These two look different enough for me, but like I said YMMV. And this is not at all important. Both things use RVO and result in the same assembly I believe/hope.

common/conf.hpp
```diff
@@ -92,9 +92,128 @@ class CmdlineOptionsIterator {
...
+		path::Join(PathDataDir, "identity", "mender-device-identity");
+
+	string DataStore =
+		conf::GetEnv("MENDER_DATASTORE_DIR", path::Join("/var/lib", "mender"));
+	string ArtScriptsPath = path::Join(DataStore, "scripts");
+	string ModulesWorkPath = path::Join(DataStore, "modules/v3");
+	string BootstrapArtifactFile = path::Join(DataStore, "bootstrap.mender");
+	string FallbackConfFile = path::Join(DataStore, "mender.conf");
+
+	string KeyFile = "mender-agent.pem";
```
vpodzime, line 115: These should all end with `_` because they are private. But I'd be the first one to break the rule for the sake of better readability, so not really suggesting a change here. :grin:

common/conf.hpp
```diff
@@ -92,9 +92,128 @@ class CmdlineOptionsIterator {
...
+	}
+	void SetBootstrapArtifactFile(const string &BootstrapArtifactFile) {
+		this->BootstrapArtifactFile = BootstrapArtifactFile;
+	}
+};
+
 class MenderConfig : public cfg_parser::MenderConfigFromFile {
 public:
-	string data_store_dir = paths::DefaultDataStore;
+	Paths default_config {};
```
vpodzime, line 216: Why call it `default_config` and not `paths`?

mender-auth/ipc/server.hpp
vpodzime, line 102: missing `_` at the end of the name

vpodzime (approved):
I trust you that you will give this a nice final polish :)

common/conf.hpp
kacf: Right, they should all have `variable_name_style_` actually, I didn't even think about that.

common/conf.hpp
oleorhagen: moved to `paths`

mender-auth/ipc/server.hpp
oleorhagen: Discussed, ignored for readability this time around.

PR #1352, opened 2023-08-09
feat: Add signal handling to mender-update::deamon::StateMachine

vpodzime:
Handling of SIGUSR1 and SIGUSR2 triggering update check and inventory push respectively and SIGTERM, SIGINT and SIGQUIT triggering clean shutdown (stopping the main loop).

Ticket: MEN-6564
Changelog: none


mender-update/daemon/state_machine.cpp
```diff
@@ -280,6 +285,16 @@ error::Error StateMachine::Run() {
...
 
+	// Register signal handlers
+	check_update_handler_.RegisterHandler({SIGUSR1}, [this](events::SignalNumber signum) {
+		runner_.PostEvent(StateEvent::DeploymentPollingTriggered);
+	});
+	inventory_update_handler_.RegisterHandler({SIGUSR2}, [this](events::SignalNumber signum) {
+		runner_.PostEvent(StateEvent::InventoryPollingTriggered);
+	});
+	termination_handler_.RegisterHandler(
+		{SIGTERM, SIGINT, SIGQUIT}, [this](events::SignalNumber signum) { event_loop_.Stop(); });
```
kacf, line 296: Can you move this into its own method, and put that inside a `platform` file? First of all, not all the signals are actually [present in the C++ standard](https://en.cppreference.com/w/cpp/header/csignal), and just intuitively, we know this isn't going to be used on embedded platforms. 🙂

kacf (changes requested):
Being trivial certainly isn't bad when it works!

You asked about testing in the chat, and I agree that some can be done with integration tests. But there is one quite easy test you could do. Just initialize the state machine, run it, then send a SIGTERM and check that the loop exits again. You can grab the initialization code [from here](https://github.com/mendersoftware/mender/blob/c854e8dd41fc7a376bd3c0108f0c1dea1769c76a/mender-update/daemon/state_test.cpp#L3037-L3070), just remove all the stuff not relevant for this test.

vpodzime:
> Just initialize the state machine, run it, then send a SIGTERM and check that the loop exits again.

Easy. I just used `SIGQUIT` instead of `SIGTERM` for a lower chance of collision with the test framework.

mender-update/daemon/state_test.cpp
```diff
@@ -3207,6 +3209,28 @@ TEST_P(StateDeathTest, StateTransitionsTest) {
...
+	conf::MenderConfig config {
+		.data_store_dir = tmpdir.Path(),
+	};
+	context::MenderContext main_context {config};
+	auto err = main_context.Initialize();
+	mtesting::TestEventLoop event_loop {chrono::seconds {3}};
+	Context ctx {main_context, event_loop};
+
+	events::Timer signal_timer {event_loop};
+	signal_timer.AsyncWait(chrono::seconds {1}, [](error::Error err) { kill(getpid(), SIGQUIT); });
```
kacf, line 3223: Could use `raise` instead, which doesn't require `unistd.h` and `getpid`.

mender-update/daemon/state_test.cpp
```diff
@@ -3207,6 +3209,28 @@ TEST_P(StateDeathTest, StateTransitionsTest) {
...
+
+	StateMachine state_machine {ctx, event_loop};
+	err = state_machine.Run();
+	ASSERT_EQ(err, error::NoError);
+
+	// Nothing more to check here, either SIGQUIT is handled properly and
+	// terminates the loop or the TestEventLoop's timer kicks in and marks this
+	// test as timing out and thus failing.
```
kacf, line 3231: 👍

kacf (approved):

mender-update/daemon/state_test.cpp
vpodzime: Nice! I didn't know about `raise()`. However, it's man page says:
> If the signal causes a handler to be called, raise() will return only after the signal handler has returned
Not sure if that's what we want :)

kacf: I expect that it will be fine, because the signal handler in this context is the async one in POSIX. While I don't know for sure about the Boost internals, almost certainly they are only using this handler to set some flag and trigger a condition somewhere, and then the real work is carried out later in the event loop.

vpodzime: Well, it should be like that because the handler registration probably happens before the loop is started. Let's give it a try...

PR #1354, opened 2023-08-10
test: Fix race condition due to exit handlers.

kacf:
It's a lot easier to view this with whitespace turned off.


vpodzime (approved):
ACK

PR #1355, opened 2023-08-10
chore: Explicitly set golang version to 1.17

lluiscampos:
As agreed in the client team, we shall stick to the golang version present in the latest release of yocto LTS.

This commit is only setting the version and running `go mod tidy`.


lluiscampos:
Custom build @ [Pipeline 961795492](https://gitlab.com/Northern.tech/Mender/mender-qa/-/pipelines/961795492)

oleorhagen (approved):

PR #1356, opened 2023-08-10
chore: Expire token before committing, so that credentials are verified.

kacf:
While we don't test a scenario with broken credentials exactly, it is indirectly tested because what would happen is that the status update would fail. This is already tested in other ways, so let's leave it at that.

Ticket: MEN-6582


lluiscampos (approved):
Reasonable :+1:

vpodzime (approved):
ACK

oleorhagen (approved):

PR #1358, opened 2023-08-14
chore: Update base image for `mender-client-docker`

lluiscampos:
Previously used Ubuntu 20.04 uses golang 1.13 which is very old at this point. Ubuntu 22.04 runs golang 1.18.

This image is for internal use only and should not be the blocker for dependencies updates.


oleorhagen (approved):

lluiscampos:

PR #1359, opened 2023-08-14
feat: Check artifact's depends before installing it

vpodzime:
Ticket: MEN-6574
Changelog: none

### DIFF
--- a/mender-update/daemon/state_test.cpp
+++ b/mender-update/daemon/state_test.cpp
@@ -81,6 +81,7 @@ struct StateTransitionsTestCase {
 	int do_schema_update_at_invocation {-1};
 	int use_non_writable_db_after_n_writes {-1};
 	bool empty_payload_artifact;
+	bool device_type_mismatch {false};
 };
 
 vector<StateTransitionsTestCase> GenerateStateTransitionsTestCases() {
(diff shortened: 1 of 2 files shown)


oleorhagen (approved):

mender-update/daemon/state_test.cpp
```diff
@@ -3038,6 +3054,12 @@ void StateTransitionsTestSubProcess(
 			artifact_url = http::JoinUrl(server.GetBaseUrl(), path::BaseName(artifact_path));
 		}
 
+		if (test.GetParam().device_type_mismatch) {
+			ofstream f(path::Join(tmpdir, "device_type"));
+			f << "device_type=mismatch-type\n";
+			ASSERT_IN_DEATH_TEST(f.good()) << "Failed to write mismatching device type";
+		}
```
lluiscampos, line 3061: I would add another test case where an artifact explicitly depends on something that the devices does not provide.

lluiscampos (approved):

mender-update/daemon/state_test.cpp
vpodzime: We have [separate unit tests](https://github.com/mendersoftware/mender/blob/feature-c%2B%2B-client/mender-update/context_test.cpp#L519) for checking the artifact's depends vs. what the device provides with pretty good coverage. This test is really about how the mismatch is handled in the state machine, it doesn't matter what doesn't match.

lluiscampos: Okey :+1:

kacf (approved):
Beautiful!

PR #1360, opened 2023-08-14
MEN-6580: Implement deployment logging in deployments

kacf:
Plus some supporting commits and fixes.

### DIFF
--- a/common/state_machine.hpp
+++ b/common/state_machine.hpp
@@ -228,7 +228,7 @@ private:
 
 		auto cancelled = cancelled_;
 		event_loop_->Post([cancelled, this]() {
-			if (!*cancelled_) {
+			if (!*cancelled) {
 				RunOne();
 			}
 		});
--- a/common/state_machine.hpp
+++ b/common/state_machine.hpp
@@ -194,9 +194,15 @@ private:
 					// attempts in the for loop.
 					event_queue_.push(event);
 				} else {
+					string states = common::BestAvailableTypeName(*machines_[0]->current_state_);
+					for (size_t i = 1; i < machines_.size(); i++) {
+						states += ", ";
+						states += common::BestAvailableTypeName(*machines_[i]->current_state_);
+					}
 					log::Fatal(
 						"State machine event " + StateEventToString(event)
-						+ " was not handled by any state. This is a bug and an irrecoverable error. "
+						+ " was not handled by any transition. Current states: " + states
+						+ ". This is a bug and an irrecoverable error. "
 						+ "Aborting in the hope that restarting will help.");
 				}
 			}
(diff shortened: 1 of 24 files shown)


common/state_machine.hpp
```diff
@@ -228,7 +228,7 @@ class StateMachineRunner : virtual public EventPoster<EventType> {
 
 		auto cancelled = cancelled_;
 		event_loop_->Post([cancelled, this]() {
-			if (!*cancelled_) {
+			if (!*cancelled) {
```
kacf, line 231: @lluiscampos: Very relevant for what you're doing.

mender-update/daemon/context.cpp
```diff
@@ -521,6 +521,28 @@ expected::ExpectedBool Context::LoadDeploymentStateData(StateData &state_data) {
 	}
 }
 
+void BeginDeploymentLogging(Context &ctx) {
```
vpodzime, line 524: Why not make this a member function of `Context`?

vpodzime (approved):
Looks good to me otherwise.

mender-update/daemon/state_test.cpp
```diff
@@ -3140,6 +3140,13 @@ TEST_P(StateDeathTest, StateTransitionsTest) {
 
 	string status_log_path = path::Join(tmpdir.Path(), "status.log");
 
+	{
+		// We don't care about existence, only content, so pre-create these files in order
+		// to make FileContains* functions happy.
+		ofstream state_log(state_log_path);
```
oleorhagen, line 3158: ?

mender-update/daemon/state_machine/state_machine.cpp
```diff
@@ -291,6 +298,8 @@ error::Error StateMachine::Run() {
 		return err;
 	}
 
+	log::Info("Running Mender client " + conf::kMenderVersion);
```
oleorhagen, line 301: Q: Wrong commit?

oleorhagen (approved):

lluiscampos (approved):

mender-update/daemon/state_test.cpp
kacf: What's the question?

mender-update/daemon/state_machine/state_machine.cpp
kacf: No, it's correct. We want each deployment log, as well as after every reboot, to start with the version, for debugging reasons in the field, like we do [in the Golang client](https://github.com/mendersoftware/mender/blob/3.5.1/app/deployment_logger.go#L144).

mender-update/daemon/state_test.cpp
oleorhagen: Never mind

PR #1361, opened 2023-08-14
docs: Document "file size" extension to Update Module protocol.

kacf:
Needed for Flash based updates. This will be implemented in MEN-6273.

Ticket: MEN-6272
Changelog: None

### DIFF
--- a/Documentation/update-modules-v3-file-api.md
+++ b/Documentation/update-modules-v3-file-api.md
@@ -38,6 +38,7 @@ action, but which just gather information:
 * `SupportsAugmentedArtifacts`
 * `ListSupportedOriginalTypes`
 * `PermittedAugmentedHeaders`
+* `ProvidePayloadFileSizes`
 
 `SupportsRollback` is described under [the `ArtifactRollback`
 state](#artifactrollback-state), `NeedsArtifactReboot` under [the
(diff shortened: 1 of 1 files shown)


oleorhagen (commented):
Hmm, why can't they be provided in `Download` in any case? :thinking:

kacf:
> Hmm, why can't they be provided in `Download` in any case? thinking

@oleorhagen: The idea behind this is: If an old client is used with a new Update Module, then `ProvidePayloadFileSizes` will never be called, and `Download` will provide files without sizes, which the Update Module will then not be prepared for.

There are a few ways to deal with this:

1. The Update Module can save a file somewhere when `ProvidePayloadFileSizes` is called, indicating that it is running with a recent enough client. In the `Download` state it can check that this file exists and produce an error if it does not (too old client).
2. The Update Module can use the format of `stream-next` (where the file sizes may or may not be) to decide, by providing a robust parser which handles both.
3. We can have two separate states, which means that entering `Download` is an immediate failure, because you know the client is too old right away.

Then I tried to stick to the principles that:

* The Update Module should be as simple as possible.
* It should be as hard as possible to make mistakes in an Update Module.

And from that I concluded that method 3 is the least likely to fail. Does that reasoning make sense?

Related to that, if we keep method 3, maybe I should explicitly recommend in the documentation to return failure in the `Download` state if you expect to get file sizes. Unless you can see a better way to produce an error when using an old client.

lluiscampos (approved):
Thanks for the reasoning.

> explicitly recommend in the documentation to return failure in the Download state if you expect to get file sizes

I vote for adding this - it helps showing what is the motivation on this API extension.

PR #1362, opened 2023-08-16
Add config.h and make the constructor public

oleorhagen:


common/crypto.hpp
```diff
@@ -49,7 +49,6 @@ class PrivateKey {
 #ifdef MENDER_CRYPTO_OPENSSL
 	unique_ptr<EVP_PKEY, void (*)(EVP_PKEY *)> pkey_;
 
-private:
```
vpodzime, line 50: Why? I think this is exactly what private constructors are for.

oleorhagen: I don't follow. 

What's the benefit of a private constructor if the variable is `public`?

vpodzime: I guess the benefit is that is shows us the real bug here -- the variable should be `private`. It even follows the naming convention for private members...

larsewi: The motivation for encapsulating the constructor was to hide the exact details of how you load a private key from the rest of the world. For the member variable, a getter is probably the "correct" way to do things, and it should only be there when `MENDER_CRYPTO_OPENSSL` is defined.

vpodzime: Why do we need a getter? Shouldn't this take care of all the operations done with the private key?

lluiscampos:
@oleorhagen Superseded by #1429 ?

oleorhagen:
Yeh

PR #1363, opened 2023-08-16
Actually push inventory from mender-update

vpodzime:
TODO:
- [x] extend the new test to check periodic data submission

### DIFF
--- a/common/common.hpp
+++ b/common/common.hpp
@@ -90,6 +90,16 @@ static inline bool MapContainsStringKey(
 	return map.find(str) != map.end();
 }
 
+template <typename KeyType, typename ValueType>
+static inline vector<KeyType> GetMapKeyVector(const unordered_map<KeyType, ValueType> &map) {
+	vector<KeyType> ret;
+	ret.reserve(map.size());
+	for (const auto &kv : map) {
+		ret.push_back(kv.first);
+	}
+	return ret;
+}
+
 } // namespace common
 } // namespace mender
 
--- a/mender-update/inventory.cpp
+++ b/mender-update/inventory.cpp
@@ -14,6 +14,7 @@
 
 #include <mender-update/inventory.hpp>
 
+#include <functional>
 #include <sstream>
 #include <string>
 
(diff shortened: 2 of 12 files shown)


common/common.hpp
```diff
@@ -90,6 +90,16 @@ static inline bool MapContainsStringKey(
 	return map.find(str) != map.end();
 }
 
+template <typename KeyType, typename ValueType>
+static inline vector<KeyType> GetMapKeyVector(const unordered_map<KeyType, ValueType> &map) {
+	vector<KeyType> ret;
+	ret.reserve(map.size());
```
oleorhagen, line 96: Suggestion: Reserve the size in the initialiser?

mender-update/inventory.cpp
```diff
@@ -105,11 +110,16 @@ error::Error PushInventoryData(
 	// replace the trailing comma with the closing square bracket
 	payload[payload.size() - 1] = ']';
 
+	size_t payload_hash = std::hash<string> {}(payload);
```
oleorhagen, line 113: Had to look this one up, so we call a function on the constructed object to hash..

oleorhagen (approved):
Looks good imo

mender-update/inventory.cpp
```diff
@@ -82,16 +85,18 @@ error::Error PushInventoryData(
 	stringstream top_ss;
 	top_ss << "[";
 	auto inv_data = ex_inv_data.value();
-	for (const auto &kv : inv_data) {
+	auto key_vector = common::GetMapKeyVector(inv_data);
+	std::sort(key_vector.begin(), key_vector.end());
```
kacf, line 89: Could also just switch over to `std::map`, which is sorted from the get-go, unlike `unordered_map`.

kacf (approved):

common/common.hpp
vpodzime: I actually tried that. I did `vector<KeyType> ret(map.size())`. But that doesn't reserve space, it actually creates a vector of that size with default-initialized items.

mender-update/inventory.cpp
vpodzime: Yeah, it's a bit confusing.

mender-update/inventory.cpp
vpodzime: Yes, but to me it makes more sense to keep the data unordered and then only sort it in the place where we need it sorted.

PR #1364, opened 2023-08-16
MEN-6498: feat: Implement `http::DownloadResumer

lluiscampos:
Tentative code for the download resume feature using HTTP Range requests.

Missing:
- [ ] Schedule further HTTP requests with exponential backoff functionality
- [ ] Stream data to the user as we get it, not only at the end
- [ ] More unit tests
- [ ] Remove `TestRangeHeaderParse` and the declaration of the struct in the `.hpp` (it will be tested indirectly with the new tests)


common/io/io.cpp
```diff
@@ -92,6 +92,21 @@ Error Copy(Writer &dst, Reader &src, vector<uint8_t> &buffer) {
 	}
 }
 
+ExpectedSize ByteReader::Read(vector<uint8_t>::iterator start, vector<uint8_t>::iterator end) {
+	if (bytes_read_ == emitter_->size()) {
+		return 0;
+	}
+	assert(end > start);
```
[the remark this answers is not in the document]
lluiscampos: I don't think so. The caller cannot ask for "read 0 bytes, thank you" :thinking: 

I took it from [here](https://github.com/mendersoftware/mender/blob/096a1e9ee8614df9bf66c28734d5617d2ac2f281/common/io/io.cpp#L101).

common/http/platform/beast/http_resumer.cpp
```diff
@@ -0,0 +1,454 @@
...
+		client_.logger_.Error(exp_resp.error().String());
+		client_.Cancel();
+		user_header_handler_(exp_resp);
+		return;
+	}
+	auto resp = exp_resp.value();
+
+	if (resp->GetStatusCode() != mender::http::StatusOK) {
+		client_.logger_.Error("Unexpected status code " + to_string(resp->GetStatusCode()));
+		client_.Cancel();
```
kacf, line 247: Hmm, won't this terminate the connection? I understand that for the resumer to handle other status codes would be difficult, but maybe it's better to simply disable the resumer if an unexpected status code is received, and just continue as if it was a normal Client. That way, callers can still consume the bodies of other status codes, without resume support.

Granted, this isn't of great importance right now, but the behavior might be a bit surprising if this is extended to something else later, such as 201 Created (which can return UUIDs in the body). I'll let you decide if it's worth doing based on effort level.

If you choose not to do it, then maybe we should treat this as an error, because the caller will expect the body handler to be called after this. And it won't be here, will it? Since the client is now cancelled?

common/http/platform/beast/http_resumer.cpp
```diff
@@ -0,0 +1,454 @@
...
+
+	io::WriterPtr user_writer = resp->GetBodyWriter();
+	if (!user_writer) {
+		auto err = MakeError(
+			DownloadResumerError, "The user did not set a BodyWriter to write the data into");
+		client_.logger_.Error(err.String());
+		client_.Cancel();
+		return;
+	}
+	client_.state_.user_body_writer = user_writer;
```
kacf, line 289: Unfortunately this section will need some care. In hindsight, using a body writer was a mistake (*). Later the `MakeBodyAsyncReader` was added, but I kept `SetBodyWriter` because it is handy for small requests and tests. But I kind of regret it now. Maybe we should discuss this face to face, because I'm not completely sure what's the best solution here, and I'm also deep into http readers and writers as part of the http forwarding.

(*) First of all, [error reporting is not possible](https://github.com/mendersoftware/mender/commit/5251a1a6fdb6051b2ddb5958f89019a8c1fc1db8), and we also need to make it an async interface in order to be able to multitask operations, which we use for many things, dbus, update module process interaction, soon http forwarding, etc.

common/http/platform/beast/http_resumer.cpp
```diff
@@ -0,0 +1,454 @@
...
+		client_.Cancel();
+		client_.state_.err = err;
+	} else if (content_range.range_start < client_.state_.offset) {
+		// The server resumed from an earlier offset, but we can still use the data
+		auto buffer = client_.state_.buffer.get();
+		auto overriden = client_.state_.offset - content_range.range_start;
+		client_.logger_.Warning(
+			"HTTP server returned a range with an earlier start. Expected "
+			+ to_string(client_.state_.offset) + ", got " + to_string(content_range.range_start)
+			+ ". The last " + to_string(overriden) + " bytes will be overriden");
```
kacf, line 354: You mean "The first x bytes", not the last, or? Also, I would maybe say "discarded", rather than overridden.

common/http/platform/beast/http_resumer.cpp
```diff
@@ -0,0 +1,454 @@
...
+		buffer->erase(buffer->end() - overriden, buffer->end());
```
kacf, line 355: What if `overridden` is larger than the buffer size? If this is difficult to handle, then I think producing error here is perfectly fine. There are limits to how far we should go in tolerating badly behaving servers.

common/http/platform/beast/http_resumer.cpp
```diff
@@ -0,0 +1,454 @@
...
+			resumer_next_body_handler.user_body_handler_(expected::unexpected(err_user));
+			return;
+		}
+
+		auto next_call_err = resumer_next_body_handler.client_.client_.AsyncCall(
+			range_req, resumer_next_header_handler, resumer_next_body_handler);
+		if (next_call_err != error::NoError) {
+			auto err_user = MakeError(
+				DownloadResumerError,
+				"Failed to copy data to schedule the next resumer call: " + next_call_err.String());
```
kacf, line 383: Not sure if I understand this error message. What does "copy data" have to do with a failure from `AsyncCall`?

common/http/platform/beast/http_resumer.cpp
```diff
@@ -0,0 +1,454 @@
...
+	}
+
+	// We can resume the download if either:
+	// * there is a short read (partial_message error) or
+	// * successful read with status code Partial Content and there is still data missing
+	const bool is_short_read =
+		!exp_resp && exp_resp.error().code == http::make_error_code(http::error::partial_message);
+	const bool is_range_response =
+		exp_resp && exp_resp.value().get()->GetStatusCode() == mender::http::StatusPartialContent;
+	const auto buffer_size = (long long int) client_.state_.buffer.get()->size();
```
kacf, line 407: Is `size_t` better?

kacf (commented):
It's hard to follow every detail going on here; pretty advanced; but I got the impression that the design is logical and well thought out. I hadn't thought of the `HandlerFunctor` pattern beforehand, but to me it seems like a good solution to the problem.

Many smaller comments, but the trickiest bit here is the question of `GetBodyWriter`, which is arguably "my fault". 🙂 See below for more details.

common/testing.hpp
```diff
@@ -67,7 +67,7 @@ class TemporaryDirectory {
 // An event loop which automatically times out after a given amount of time.
 class TestEventLoop : public mender::common::events::EventLoop {
 public:
-	TestEventLoop(chrono::seconds seconds = chrono::seconds(5)) :
+	TestEventLoop(chrono::seconds seconds = chrono::seconds(65)) :
```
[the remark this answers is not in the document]
lluiscampos: Of course not. Thank you for the obvious suggestion :+1:

common/http/platform/beast/http_resumer.cpp
lluiscampos: Bad copy paste, it is meant to say `Failed to schedule the next resumer call`

common/http.hpp
```diff
@@ -486,6 +491,70 @@ class Client : virtual public ClientInterface {
...
+		shared_ptr<vector<uint8_t>> buffer;
+		shared_ptr<io::Writer> user_body_writer;
+		void init() {
+			ongoing = true;
+			err = error::NoError;
+			offset = 0;
+			content_length = 0;
+			user_body_writer.reset();
+			buffer.reset();
+			buffer = make_shared<vector<uint8_t>>();
```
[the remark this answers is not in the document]
lluiscampos: Thank you.

(I keep thinking about structs as `C` structs instead of basically a `C++` class with everything public...)

common/http/platform/beast/http_resumer.cpp
```diff
@@ -0,0 +1,454 @@
...
+	if (start_and_end.size() != 2) {
+		return expected::unexpected(
+			MakeError(NoSuchHeaderError, "Invalid Content-Range returned by server: " + content));
+	}
+
+	auto exp_range_start = common::StringToLongLong(start_and_end[0]);
+	auto exp_range_end = common::StringToLongLong(start_and_end[1]);
+	if (!exp_range_start || !exp_range_end) {
+		return expected::unexpected(MakeError(
+			NoSuchHeaderError, "Content-Range contains invalid number: " + start_and_end[0]));
```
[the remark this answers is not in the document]
lluiscampos: Yes, you are right and it is a typo.

I am going to print the whole thing in all NaN - for a human user will be clear which of the fields is wrong. Examples:
`Content-Range contains invalid number: 1000-2000/3abc`
`Content-Range contains invalid number: 1ab0-2000/3000`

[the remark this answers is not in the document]
lluiscampos: Once we are fine-tuning error messages, there is one thing that annoyed me in the legacy client, that some errors are `Unexpected Content-Range received from server` while others are `Invalid Content-Range returned by server`. I am going to unify to _invalid_.

common/http/platform/beast/http_resumer.cpp
```diff
@@ -0,0 +1,454 @@
...
+		if (err != error::NoError) {
+			client_.logger_.Error(err.String());
+			client_.Cancel();
+			user_body_handler_(expected::unexpected(err));
+		}
+		return;
+	}
+	auto resp = exp_resp.value();
+
+	auto exp_content_range_header = resp->GetHeader("Content-Range");
```
oleorhagen, line 308: Suggestion:

```
	auto exp_content_range = resp->GetHeader("Content-Range").and_then(ParseRangeHeader);
	if (!exp_content_range) {
		client_.logger_.Error(exp_content_range.error().String());
		client_.Cancel();
		client_.state_.err = exp_content_range.error();
		return;
	}

```

common/http/platform/beast/http_resumer.cpp
kacf: As a primer for this discussion, I just opened #1371 which finalizes the async support. Perhaps you want to review / learn that one for more context, and we can take the discussion afterwards.

common/http/platform/beast/http_resumer.cpp
```diff
@@ -0,0 +1,454 @@
...
+struct RangeHeader {
+	long long int range_start;
+	long long int range_end;
+	long long int size;
+};
+using ExpectedRangeHeader = expected::expected<RangeHeader, error::Error>;
+ExpectedRangeHeader ParseRangeHeader(string resp);
+
+ExpectedRangeHeader ParseRangeHeader(string header) {
+	RangeHeader range_header {.range_start = 0, .range_end = 0, .size = 0};
```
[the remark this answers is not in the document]
kacf: I'm a bit on the fence. Your version is obviously simpler, but I don't find it *that* bad. Long, but not difficult.

Of course, if I wasn't worried about having to rip out regex later because of size constraints on embedded devices, it would be a no-brainer: yours is clearly better. But the regex argument is really just a hunch, so I can't use it as a strong argument.

How about this suggestion: Make a comment in the code that a non-regex version exists in this PR (or some other suitable branch), and switch to regex. Then our code will be nice and clean, but at least we can get the non-regex version back if we need to, without having to rewrite it.

common/http.hpp
```diff
@@ -486,6 +491,70 @@ class Client : virtual public ClientInterface {
...
+	error::Error AsyncCall(
+		OutgoingRequestPtr req,
+		ResponseHandler header_handler,
+		ResponseHandler body_handler) override;
+
+	void Cancel() override {
+		state_.ongoing = false;
+		retry_.backoff.Reset();
+		// TODO: When to cancel http::Client?
+		// client_.Cancel();
```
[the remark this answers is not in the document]
lluiscampos: Related to the other comment (terminate the connection) - let's focus there the discussion.

oleorhagen:
Looks nice I think. Well organised, and logical.

I will finish off my review tomorrow however.

common/http/platform/beast/http_resumer.cpp
lluiscampos: (Relates to the `TODO` in the hpp that we need to address)

First please note that this `client_` there is the resumer, which as you can see in the header is not calling `Cancel()` in the next client and hence not terminating the connection.

The logic is not the cleanest, but it works as you describe. Any non 200 will "cancel" the resumer functionality and in the resumer body handler it will just propagate the call to the user handler (first three lines).

However, the issue is still where to cancel the underlying client. I cannot do it in `DownloadResumerClient::Cancel` exactly for this case, but on a user initiated `Cancel` I should indeed propagate it so that he connection terminates.

Update: As I write and read this, I realize that the issue is that `DownloadResumerClient::Cancel` is serving to purposes, to cancel the channel and to (internally) disable the functionality. An explicit (and private) `Disable` method should solve the issue! I'll give it a try.

common/http/platform/beast/http_resumer.cpp
lluiscampos: Face to face on Monday, please :)

common/http/platform/beast/http_resumer.cpp
lluiscampos: I mean the last x bytes of what already has been received in previous HTTP responses. Overridden is better in the sense that the new content is overriding the last x bytes that were there.

This is a bit different than the [golang implementation](https://github.com/mendersoftware/mender/blob/d9010526d35d3ac861ea1e4210d36c2fef748ef8/client/update_resumer.go#L157) because here I don't have control of the incoming stream to "consume" the first x bytes (it is controlled by the underlying client) so instead I remove the last x bytes of my buffer and concatenate the buffer from there.

Once clarified, do you have a suggestion on how to better rewrite the warning message?

common/http/platform/beast/http_resumer.cpp
lluiscampos: It should not be possible for it to be bigger (it would mean that we got a range start < 0 !) but you have a point that I should sanity check it and error accordingly. I'll do.

common/http/platform/beast/http_resumer.cpp
```diff
@@ -0,0 +1,454 @@
...
+
+// Represents the parts of a Content-Range HTTP header
+// See https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Range
+struct RangeHeader {
+	long long int range_start;
+	long long int range_end;
+	long long int size;
+};
+using ExpectedRangeHeader = expected::expected<RangeHeader, error::Error>;
+ExpectedRangeHeader ParseRangeHeader(string resp);
```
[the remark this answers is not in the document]
lluiscampos: Leftovers from reworking this part.

common/http/platform/beast/http_resumer.cpp
kacf: Ah yes, ok. The naming here is a bit confusing, since the same variable exists in `DownloadResumerClient`, but has a different meaning there. Maybe you could call it `resumer_`, since it is a fully qualified subclass instance?

The `Disable` method sounds good.

common/http/platform/beast/http_resumer.cpp
kacf: Yeah, I see what you mean now. You are always operating within one buffer block, so you can actually overwrite bytes you already received. Then the message is good. 👍

common/http/platform/beast/http_resumer.cpp
kacf: > It should not be possible for it to be bigger (it would mean that we got a range start < 0 !)

Can't it? Maybe I'm misreading it, but to me it seems like `client_.state_.offset` can be *much* larger than buffer, and if the `range_start` is far behind, then the `overridden` result can easily be larger than buffer. But maybe I'm misunderstanding `offset`. Isn't it the full download offset? So it could be hundreds of megabytes.

lluiscampos: I rewrote quite a bit this part. Have a second look in the next round :eyes:

common/http/platform/beast/http_resumer.cpp
```diff
@@ -0,0 +1,454 @@
...
+	if (!client_.IsOngoing()) {
+		if (client_.state_.err != error::NoError) {
+			user_body_handler_(expected::unexpected(client_.state_.err));
+		}
+		return;
+	}
+
+	// We can resume the download if either:
+	// * there is a short read (partial_message error) or
+	// * successful read with status code Partial Content and there is still data missing
```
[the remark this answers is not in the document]
lluiscampos: Well spotted. For the header handler we resume on any error, while for the body we only handle this specific error (and there is no reason to do so).

Thanks.

common/io_test.cpp
```diff
@@ -251,6 +251,17 @@ TEST(IO, TestStringReader) {
...
 
+TEST(IO, TestByteReader) {
+	vector<uint8_t> buffer {1, 2, 3, 4, 5, 6, 5, 4, 3, 2, 1};
+	auto byte_reader = io::ByteReader(buffer);
+
+	auto discard_writer = io::Discard {};
+
+	auto err = Copy(discard_writer, byte_reader);
+
+	ASSERT_EQ(error::NoError, err);
```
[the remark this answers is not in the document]
lluiscampos: And how can I do that?

common/io.hpp
```diff
@@ -266,6 +268,19 @@ class ByteWriter : virtual public Writer {
 		vector<uint8_t>::const_iterator start, vector<uint8_t>::const_iterator end) override;
 };
 
+class ByteOffsetWriter : public ByteWriter {
+public:
+	ByteOffsetWriter(vector<uint8_t> &receiver, Vsize offset) :
+		ByteWriter(receiver) {
```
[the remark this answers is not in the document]
lluiscampos: I guess that could be said to virtually all readers and writers. Similarly, I thought about having a "AppendWriter" for example that would extend existing buffer with new data (which is really what I do in the code using `ByteOfffsetWriter`, to use an offset of the current buffer size).

I think I'll leave it as is.

common/http/platform/beast/http_resumer.cpp
```diff
@@ -0,0 +1,454 @@
...
+
+namespace mender {
+namespace http {
+
+namespace common = mender::common;
+
+// Represents the parts of a Content-Range HTTP header
+// See https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Range
+struct RangeHeader {
+	long long int range_start;
```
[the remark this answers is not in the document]
lluiscampos: I think it is best to leave it signed so that we could handle if a server sends a negative number (?). Also for existing code parsing for example content-length we use `common::StringToLongLong` so it is better to treat them all the same.

mender-update/daemon/context.hpp
```diff
@@ -162,7 +162,7 @@ class Context {
 	// For polling, and for making status updates.
 	api::Client http_client;
 	// For the artifact download.
-	http::Client download_client;
+	http::DownloadResumerClient download_client;
```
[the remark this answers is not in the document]
lluiscampos: You are right., that was the whole point. I'll use the interface here and create a `DownloadResumerClient` instance in the constructor

common/http/platform/beast/http_resumer.cpp
lluiscampos: TIL `.and_then()` :rocket:

vpodzime:
@lluiscampos please ping me when the existing comments are taken care of here. I don't think me stepping into the discussions now would make much sense. I'll wait for the nice and polished result and then take a look.

common/http/platform/beast/http_resumer.cpp
lluiscampos: As discussed on Monday, I will keep the PR on hold until you submit some rework on the `http::Client` and I can adapt my code to address this point.

common/http/platform/beast/http_resumer.cpp
lluiscampos: I went back and forth with this and I don't know what is actually best. To use `size_t` here (and not cast again back later to `long long` for the [comparison](https://gitlab.com/Northern.tech/Mender/mender/-/jobs/4976660957)) I should then also set the fields of `struct RangeHeader` also to `size_t`.

I chose there signed version thinking that maybe a misbehaving server could send negative numbers. And also it was convenient and coherent to use `common::StringToLongLong` with we already used to parse `Content-Lenght` header.

lluiscampos:
@vpodzime You can have a look now :eyes: :tea: 

@kacf @oleorhagen  You can also have a look at the rework, although as discussed with Kristian we will put this PR on hold until we can rework a bit the `http::Client` and `IncomingResponse` interfaces.

vpodzime:
> @vpodzime You can have a look now 👀 🍵

I think I'll wait for the version adapted to the changes from #1375

common/http/platform/beast/http_resumer.cpp
lluiscampos: This discussion is obsolete now.

common/io_test.cpp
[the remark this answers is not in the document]
lluiscampos: I am removing `io::ByteReader` and `io::ByteOffsetWriter` from final PR as they turned out to be unused after the rework(s).

lluiscampos:
@kacf @oleorhagen @vpodzime Ready for final review! :crossed_fingers:

:warning: The test `UpdateModuleTests.DownloadProcessOpensStreamNextButDoesntRead` [failed](https://gitlab.com/Northern.tech/Mender/mender/-/jobs/5150905212) for threadsan but then [passed](https://gitlab.com/Northern.tech/Mender/mender/-/jobs/5150950584) on re-run. Is this a known unstable test? I cannot see a relation between such a test and my changes

kacf:
Hmm, no, that failure is new to me. But I also don't see the relation. If you don't see it again, I'd sweep this one under the rug for now.

mender-update/http_resumer/http_resumer.cpp
```diff
@@ -0,0 +1,482 @@
...
+			}
+		});
+}
+
+DownloadResumerClient::DownloadResumerClient(
+	const http::ClientConfig &config, events::EventLoop &event_loop) :
+	client_(config, event_loop, "http_resumer:client"),
+	logger_ {"http_resumer:client"},
+	cancelled_ {make_shared<bool>(true)},
+	retry_ {.backoff = http::ExponentialBackoff(chrono::seconds(1), 10), .wait_timer = event_loop} {
```
kacf, line 345: Should the backoff parameters be passed in?

In any case, the minimum interval in the original client is 1 minute.

mender-update/http_resumer/http_resumer.cpp
```diff
@@ -0,0 +1,482 @@
...
+	resumer_state_ = make_shared<DownloadResumerClientState>();
+}
+
+DownloadResumerClient::~DownloadResumerClient() {
+	if (!*cancelled_) {
+		logger_.Warning("DownloadResumerClient destroyed while request is still active!");
+	}
+	client_.Cancel();
```
kacf, line 353: As mentioned in the reader, it might be dangerous to call this here without disabling handler calling first. Maybe calling `Disable()` is enough, and the rest can be handled by the destructors in each object?

kacf (changes requested):

mender-update/http_resumer.hpp
```diff
@@ -0,0 +1,179 @@
...
+enum class DownloadResumerUserHandlersStatus {
+	None,
+	HeaderHandlerCalled,
+	BodyHandlerCalled,
+};
+
+struct DownloadResumerClientState {
+	bool initialized {false};
+	long long content_length {0};
+	long long offset {0};
```
vpodzime, line 49: Why not `size_t`?

mender-update/http_resumer.hpp
```diff
@@ -0,0 +1,179 @@
...
+	http::ResponseHandler user_body_handler_;
+	http::OutgoingRequestPtr user_request_;
+
+	struct {
+		http::ExponentialBackoff backoff;
+		events::Timer wait_timer;
+	} retry_;
+
+	// Parameters from the last time that user called AsyncRead.
+	// They are are re-used when resuming the download
```
vpodzime, line 162: double *are*

mender-update/http_resumer/http_resumer.cpp
```diff
@@ -0,0 +1,482 @@
...
+			"HTTP server returned an different range than requested. Requested "
+				+ to_string(resumer_client->resumer_state_->offset) + "-"
+				+ to_string(resumer_client->resumer_state_->content_length - 1) + ", got "
+				+ to_string(content_range.range_start) + "-" + to_string(content_range.range_end));
+		resumer_client->DisableWithError(bad_range_err);
+		resumer_client->CallUserHandler(expected::unexpected(bad_range_err));
+		return;
+	}
+
+	// Get the the reader for the new response
```
vpodzime, line 226: double *the*

mender-update/http_resumer/http_resumer.cpp
```diff
@@ -0,0 +1,482 @@
...
+	auto resumer_client = resumer_client_.lock();
+	if (!resumer_client || *resumer_client->cancelled_) {
+		return;
+	}
+
+	// We resume the download if either:
+	// * there is any error or
+	// * successful read with status code Partial Content and there is still data missing
+	const bool is_range_response =
+		exp_resp && exp_resp.value().get()->GetStatusCode() == mender::http::StatusPartialContent;
```
vpodzime, line 258: Should (be possible to) drop `get()` here.

mender-update/http_resumer/http_resumer.cpp
vpodzime, line 345: The `.wait_timer = event_loop` part is hurting my eyes. Can we use the `Timer` constructor, please?

vpodzime (commented):
All I could find. Quite a beast, I can tell you that!

common/http.hpp
```diff
@@ -381,8 +381,23 @@ static inline bool AtLeast(TransactionStatus status, TransactionStatus expected_
 	return static_cast<int>(status) >= static_cast<int>(expected_status);
 }
 
-// Object which manages one connection, and its requests and responses (one at a time).
-class Client : public events::EventLoopObject, virtual public io::Canceller {
+// Interface which manages one connection, and its requests and responses (one at a time).
+class ClientInterface : public events::EventLoopObject {
```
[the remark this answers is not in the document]
lluiscampos: It should be moved. I'll take care of it :+1:

common/http.hpp
```diff
@@ -395,6 +412,9 @@ class ClientInterface : public events::EventLoopObject {
 	// Use this to get an async reader for the body. If there is no body, it returns a
 	// `BodyMissingError`; it's safe to continue afterwards, but without a reader.
 	virtual io::ExpectedAsyncReaderPtr MakeBodyAsyncReader(IncomingResponsePtr resp) = 0;
+
+	// Returns the next client in the stack, which eventually should return a fully featured Client
+	virtual ClientInterface &GetNextClient() = 0;
```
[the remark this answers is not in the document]
lluiscampos: Thanks.

mender-update/http_resumer/http_resumer.cpp
```diff
@@ -0,0 +1,482 @@
...
+
+	std::regex content_range_regexp {R"(bytes\s+(\d+)\s?-\s?(\d+)?\s?\/?\s?(\d+|\*)?)"};
+
+	std::smatch range_matches;
+	if (!regex_match(header, range_matches, content_range_regexp)) {
+		return expected::unexpected(http::MakeError(
+			http::NoSuchHeaderError, "Invalid Content-Range returned from server: " + header));
+	}
+
+	if (!range_matches[0].matched || !range_matches[1].matched) {
```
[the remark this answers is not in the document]
lluiscampos: I made the second group also unconditional and removed this if clause.

mender-update/http_resumer/http_resumer.cpp
lluiscampos: Thanks.

mender-update/http_resumer.hpp
lluiscampos: `offset` could be `size_t`, but `content_length` I would still keep it signed (related discussion [here](https://github.com/mendersoftware/mender/pull/1364#discussion_r1309279151)) for easy parsing from the header. If we accept than, then is better that both are signed for comparison between then.

I really don't have an strong opinion. If you insist I will do it: I can very well parse content length from the header directly to `size_t` and propagate this also to the regular client's content length parsing.

mender-update/http_resumer/http_resumer.cpp
```diff
@@ -0,0 +1,482 @@
...
+		}
+	} else {
+		// Update headers with the last received server response. When resuming has taken place,
+		// the user will get different headers on header and body handlers, representing (somehow)
+		// what the resumer has been doing in its behalf.
+		auto resp = exp_resp.value();
+		resumer_client->response_->status_code_ = resp->GetStatusCode();
+		resumer_client->response_->status_message_ = resp->GetStatusMessage();
+		for (auto header : resp->GetHeaders()) {
+			resumer_client->response_->headers_[header.first] = header.second;
```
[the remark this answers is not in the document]
lluiscampos: Yes. Thank you.

mender-update/http_resumer/http_resumer.cpp
```diff
@@ -0,0 +1,482 @@
...
+		resumer_client->DisableWithWarning(
+			"Content-Length contains invalid number: " + exp_header.value());
+		resumer_client->CallUserHandler(exp_resp);
+		return;
+	}
+
+	// Prepare state
+	resumer_client->resumer_state_ = make_shared<DownloadResumerClientState>();
+	resumer_client->resumer_state_->initialized = true;
+	resumer_client->resumer_state_->content_length = exp_length.value();
```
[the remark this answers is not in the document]
lluiscampos: Thanks for questioning this. It is per request, so it doesn't need to be created in constructor.

Also, by looking at it now with fresh eyes, we don't need the `initialized` field at all as we can deduce if the resuming functionality is active or not by just checking if `resumer_state` is not `nullptr` (iow there is state). I will rework it a bit :+1:

mender-update/http_resumer.hpp
vpodzime: There's also `ssize_t`. :wink:

mender-update/http_resumer/http_resumer.cpp
```diff
@@ -0,0 +1,482 @@
...
+
+		// Finished, call the user handler \o/
+		resumer_client->Disable();
+		resumer_client->CallUserHandler(resumer_client->response_);
+		resumer_client->logger_.Debug("Download resumed and completed successfully");
+	}
+}
+
+DownloadResumerAsyncReader::~DownloadResumerAsyncReader() {
+	Cancel();
```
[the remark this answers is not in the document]
lluiscampos: Thanks. Removed.

mender-update/http_resumer/http_resumer.cpp
lluiscampos: I reworked all of these (and removed `Disable` ambiguous method).

mender-update/http_resumer/http_resumer.cpp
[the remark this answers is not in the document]
lluiscampos: So my take off after looking back and forth on this.

The `DownloadResumerClientState` is not going to be per request, but tied to `DownloadResumerClient` life.

Then, this `initialized` needs to be an (explicit) state like None/Inactive/Resuming. The inactive could be understood as "passive" or "pass-through" where the `DownloadResumerClient` encountered a non-resumable response so it only forwards the handlers from `http::Client` to the user.

Coming up in the next update :+1:

lluiscampos:
@kacf Please have a look at the rework, specially [this separate commit](https://github.com/mendersoftware/mender/pull/1364/commits/ef8578abda6d57662bbd11337a55d00158afaa0c).

Three other small things still to work on (ref non resolved conversations :point_up:).

mender-update/http_resumer/http_resumer.cpp
```diff
@@ -108,52 +108,52 @@ class BodyHandlerFunctor {
...
 	}
 }
 
 void HeaderHandlerFunctor::HandleFirstResponse(
 	const shared_ptr<DownloadResumerClient> &resumer_client,
 	http::ExpectedIncomingResponsePtr exp_resp) {
-	// The first response shall call the user header callback. On errors, we log at Warning level,
-	// disable the functionality, and call the user handler with the original response
+	// The first response shall always call the user header callback. On resumable responses, we
+	// create a ouw own incoming response and call the user header handler. On errors, we log a
```
kacf, line 123: "create **our** own", I suppose?

mender-update/http_resumer/http_resumer.cpp
lluiscampos: I updated it to 1 minute :+1: It is not parametrized in the legacy client, though, so I think I'll let this one pass.

mender-update/http_resumer/http_resumer.cpp
```diff
@@ -265,16 +277,12 @@ void BodyHandlerFunctor::operator()(http::ExpectedIncomingResponsePtr exp_resp)
 		resumer_client->response_->headers_ = resp->GetHeaders();
 
 		// Finished, call the user handler \o/
-		resumer_client->Disable();
-		resumer_client->CallUserHandler(resumer_client->response_);
 		resumer_client->logger_.Debug("Download resumed and completed successfully");
+		resumer_client->DoCancel();
```
kacf, line 281: I don't think you need this. The connection is already in a cancelled state after a successful handler call.

mender-update/http_resumer/http_resumer.cpp
```diff
@@ -420,7 +427,8 @@ error::Error DownloadResumerClient::ScheduleNextResumeRequest() {
 				auto err_user = http::MakeError(
 					http::DownloadResumerError,
 					"Failed to schedule the next resumer call: " + next_call_err.String());
-				DisableWithError(err_user);
+				logger_.Error(err_user.String());
+				Cancel();
```
kacf, line 431: This cancels the whole request, doesn't it? Shouldn't it rather wait and retry here?

mender-update/http_resumer/http_resumer_test.cpp
```diff
@@ -861,3 +861,217 @@ TEST_F(DownloadResumerTest, SmallIntervalsErrorOnFirstRead) {
...
+	// Client
+	http::ClientConfig client_config;
+	shared_ptr<http_resumer::DownloadResumerClient> client =
+		make_shared<http_resumer::DownloadResumerClient>(client_config, loop);
+	client->SetSmallestWaitInterval(chrono::seconds(2));
+
+	client->AsyncCall(
+		req,
+		[](http::ExpectedIncomingResponsePtr exp_resp) {},
+		[&client](http::ExpectedIncomingResponsePtr exp_resp) { client->Cancel(); });
```
kacf, line 1071: It should already be cancelled ("finished" is perhaps a better word), once the body handler is called, so this is, in theory, a no-op.

mender-update/http_resumer/http_resumer_test.cpp
```diff
@@ -861,3 +861,217 @@ TEST_F(DownloadResumerTest, SmallIntervalsErrorOnFirstRead) {
...
+	shared_ptr<http_resumer::DownloadResumerClient> client =
+		make_shared<http_resumer::DownloadResumerClient>(client_config, loop);
+	client->SetSmallestWaitInterval(chrono::seconds(2));
+
+	client->AsyncCall(
+		req,
+		[&client](http::ExpectedIncomingResponsePtr exp_resp) { client->Cancel(); },
+		[](http::ExpectedIncomingResponsePtr exp_resp) {
+			ASSERT_FALSE(exp_resp);
+			EXPECT_EQ(exp_resp.error().code, make_error_condition(errc::operation_canceled));
```
kacf, line 1025: Can't you stop the loop here instead?

kacf (commented):
Reads much easier!

mender-update/http_resumer/http_resumer.cpp
kacf: Oops, my mistake, I misread this. I though you were calling it on the real client. Never mind this comment.

mender-update/http_resumer/http_resumer.cpp
kacf: 👍

As long as the tests run fast enough, it's fine.

mender-update/http_resumer/http_resumer.cpp
lluiscampos: Actually, this one I need.

Note that it is calling `DoCancel` (and not `Cancel`) to set our `cancelled_` pointer to true. Without this the client object is not reusable and subsequent calls to `AsyncCall` will error with `HTTP call already ongoing`

mender-update/http_resumer/http_resumer.cpp
lluiscampos: This is while re-scheduling and an error on `http::Client.AsyncCall` -> I consider it unrecoverable. In which cases would a re-schedule help?

mender-update/http_resumer.hpp
lluiscampos: Using `ssize_t` now.

mender-update/http_resumer/http_resumer_test.cpp
lluiscampos: To be honest, I copied these two tests almost verbatim from [http_test.cpp](https://github.com/mendersoftware/mender/blob/b6b72ae7f0ccfa90ac1ed9bbf2bd572aab8d1401/common/http_test.cpp#L929).

I see that this test doesn't do much - I repurpose to do stream actual data and cancel half way from a timer.

mender-update/http_resumer/http_resumer_test.cpp
lluiscampos: This test I also copied (see above) and interestingly stopping from the loop doesn't work (which is a bug in itself). But it also doesn't work correctly on `http_test.cpp` so that we need to follow-up separately.

lluiscampos: Just to be clear. The following patch for `http_test.cpp` fails the test
```
$ git diff common/http_test.cpp
diff --git a/common/http_test.cpp b/common/http_test.cpp
index 4ba364ec..ccf1b3f6 100644
--- a/common/http_test.cpp
+++ b/common/http_test.cpp
@@ -915,13 +915,14 @@ TEST(HttpTest, TestClientCancelInHeaderHandler) {
        client.AsyncCall(
                req,
                [&client](http::ExpectedIncomingResponsePtr exp_resp) { client.Cancel(); },
-               [](http::ExpectedIncomingResponsePtr exp_resp) {
+               [&loop](http::ExpectedIncomingResponsePtr exp_resp) {
                        ASSERT_FALSE(exp_resp);
                        EXPECT_EQ(exp_resp.error().code, make_error_condition(errc::operation_canceled));
+                       loop.Stop();
                });
 
-       events::Timer timer(loop);
-       timer.AsyncWait(chrono::milliseconds(500), [&loop](error::Error err) { loop.Stop(); });
+       // events::Timer timer(loop);
+       // timer.AsyncWait(chrono::milliseconds(500), [&loop](error::Error err) { loop.Stop(); });
 
        loop.Run();
 }
```

mender-update/http_resumer/http_resumer_test.cpp
lluiscampos: This was quite useful. It revealed an interesting bug: the resumer was _unstoppable_! After a `Cancel()` it would just interpret the call to body handler (with operation cancelled error) as a generic error and resume again :laughing: 

Fixed now in the push coming up.

mender-update/http_resumer.hpp
```diff
@@ -0,0 +1,179 @@
...
+	void CallUserHandler(http::ExpectedIncomingResponsePtr exp_resp);
+
+	void DisableWithWarning(string reason);
+	void DisableWithError(error::Error err);
+	void Disable();
+
+	shared_ptr<DownloadResumerClientState> resumer_state_;
+	shared_ptr<DownloadResumerAsyncReader> resumer_reader_;
+
+	http::Client client_;
```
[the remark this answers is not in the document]
lluiscampos: This am sorry but I have to leave out. The code gets very messy - how should I then implement `DownloadResumerClient::GetHttpClient` :thinking:

lluiscampos:
Addressed the pending discussions, squashed the commits, and rebased on top of latest master :sparkles: :sparkles:

mender-update/http_resumer/http_resumer.cpp
kacf: Yep, looks correct to me!

mender-update/http_resumer/http_resumer.cpp
kacf: You're probably right.. Probably..

But could it happen for example if all interfaces are down? I have no idea if Boost would report that immediately or through the handler. Maybe it makes sense to be fault tolerant also here?

mender-update/http_resumer/http_resumer_test.cpp
kacf: Haha, nice! The unstoppable resumer! :muscle:

mender-update/http_resumer/http_resumer.cpp
lluiscampos: Okey, we'll retry also then.

mender-update/http_resumer.hpp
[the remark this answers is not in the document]
kacf: You're right, not important.

mender-update/http_resumer/http_resumer_test.cpp
lluiscampos: > This test I also copied (see above) and interestingly stopping from the loop doesn't work (which is a bug in itself). 

This I fixed in the previous force-push. The sleep is not part of this test anymore. Sorry for not being explicit.

kacf (approved):
:tada: :tada: :tada:

Thanks for doing the rebase on top of the protocol switching also!

PR #1365, opened 2023-08-16
chore: Implement pushing of logs to the server.

kacf:
Ticket: MEN-6581

### DIFF
--- a/mender-update/daemon/states.cpp
+++ b/mender-update/daemon/states.cpp
@@ -383,7 +383,10 @@ void SendStatusUpdateState::DoStatusUpdate(Context &ctx, sm::EventPoster<StateEv
 							return;
 						}
 
-						// Try again.
+						// Try again. Since both status and logs are sent
+						// from here, there's a chance this might resubmit
+						// the status, but there's no harm in it, and it
+						// won't happen often.
 						DoStatusUpdate(ctx, poster);
 					});
 				return;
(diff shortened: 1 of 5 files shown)


kacf:
Only the two last commits are new, the rest is from #1360.

mender-update/daemon/states.cpp
```diff
@@ -401,13 +407,33 @@ void SendStatusUpdateState::DoStatusUpdate(Context &ctx, sm::EventPoster<StateEv
...
 		ctx.deployment.state_data->update_info.id,
 		status,
 		"",
 		ctx.mender_context.GetConfig().server_url,
 		ctx.http_client,
-		result_handler);
+		[result_handler, &ctx](error::Error err) {
+			// If the deployment succeeded, we don't submit logs, so there's nothing
+			// more to do.
+			if (err != error::NoError || !ctx.deployment.failed) {
```
lluiscampos, line 420: :thinking: Shouldn't this be:
```suggestion
			if (err == error::NoError && !ctx.deployment.failed) {
```
?

mender-update/daemon/states.cpp
```diff
@@ -401,13 +407,33 @@ void SendStatusUpdateState::DoStatusUpdate(Context &ctx, sm::EventPoster<StateEv
...
+				return;
+			}
+
+			// Push logs.
+			err = ctx.deployment_client->PushLogs(
+				ctx.deployment.state_data->update_info.id,
+				ctx.deployment.logger->LogFilePath(),
+				ctx.mender_context.GetConfig().server_url,
+				ctx.http_client,
+				result_handler);
```
lluiscampos, line 431: Pushing the logs will be tried only once? (Probably covered somewhere else, I am new to the state machine)

lluiscampos (changes requested):

lluiscampos:
I only reviewed the last two commints

oleorhagen (approved):

vpodzime (approved):
ACK

mender-update/daemon/states.cpp
kacf: Maybe the comment is a bit misleading, how is this clarification:

> If there is an error, we don't submit logs now, but call the handler, which may schedule a retry later. If there is no error, and the deployment as a whole was successful, then also call the handler here, since we don't need to submit logs at all then.

lluiscampos: It is much better. Please update the comment in the code :)

lluiscampos (approved):

mender-update/daemon/states.cpp
kacf: The `result_handler` has retry logic inside it, so it's initiated from there.

PR #1366, opened 2023-08-22
chore: Fix mender-auth CLI interface.

kacf:
Require `daemon` argument to actually run the daemon, and produce
failure when no argument is given.

Ticket: MEN-6639

### DIFF
--- a/mender-auth/cli/actions.cpp
+++ b/mender-auth/cli/actions.cpp
@@ -30,7 +30,7 @@ namespace cli {
 using namespace std;
 
 namespace events = mender::common::events;
-namespace mlog = mender::common::log;
+namespace log = mender::common::log;
 namespace path = mender::common::path;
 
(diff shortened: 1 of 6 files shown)


mender-auth/cli/actions.cpp
```diff
@@ -17,16 +17,23 @@
...
+#include <mender-auth/ipc/server.hpp>
+
 namespace mender {
 namespace auth {
 namespace cli {
 
 using namespace std;
 
+namespace events = mender::common::events;
+namespace mlog = mender::common::log;
```
oleorhagen, line 33: Nit: When not in the top-level namespace we can use `log` instead of `mlog` just fine

Just so it's consistent with the rest of the codebase

oleorhagen (approved):
Nice - thx!

:smile_cat:

lluiscampos (approved):

PR #1367, opened 2023-08-23
Add a synchronous script runner

oleorhagen:
Also change the interface to the script runner a little.

Along with some minor fixes to the existing code.

### DIFF
--- a/artifact/v3/scripts/executor.cpp
+++ b/artifact/v3/scripts/executor.cpp
@@ -143,18 +143,13 @@ string ScriptRunner::Name() {
 
 ScriptRunner::ScriptRunner(
 	events::EventLoop &loop,
-	State state,
-	Action action,
 	chrono::seconds state_script_timeout,
 	const string &artifact_script_path,
 	const string &rootfs_script_path,
 	mender::common::processes::OutputCallback stdout_callback,
 	mender::common::processes::OutputCallback stderr_callback) :
 	loop_ {loop},
-	is_artifact_script_ {IsArtifactScript(state)},
 	state_script_timeout_ {state_script_timeout},
-	state_ {state},
-	action_ {action},
 	artifact_script_path_ {artifact_script_path},
 	rootfs_script_path_ {rootfs_script_path},
 	stdout_callback_ {stdout_callback},
@@ -234,8 +229,10 @@ Error ScriptRunner::Execute(
 		this->state_script_timeout_);
 }
 
-Error ScriptRunner::AsyncRunScripts(HandlerFunction handler) {
-	if (this->is_artifact_script_) {
+Error ScriptRunner::AsyncRunScripts(State state, Action action, HandlerFunction handler) {
+	this->action_ = action;
+	this->state_ = state;
+	if (IsArtifactScript(state)) {
 		// Verify the version in the version file (OK if no version file present)
 		auto version_file_error {
 			CorrectVersionFile(path::Join(this->artifact_script_path_, "version"))};
@@ -266,6 +263,20 @@ Error ScriptRunner::AsyncRunScripts(HandlerFunction handler) {
 	auto scripts_iterator {this->collected_scripts_.begin()};
 	auto scripts_iterator_end {this->collected_scripts_.end()};
 	return Execute(scripts_iterator, scripts_iterator_end, handler);
+}
+
+
+Error ScriptRunner::RunScripts(State state, Action action) {
+	auto run_err {error::NoError};
+	auto err = AsyncRunScripts(state, action, [this, &run_err](Error error) {
+		run_err = error;
+		this->loop_.Stop();
+	});
+	if (err != error::NoError) {
+		return err;
+	}
+	this->loop_.Run();
+	return run_err;
 }
 
 } // namespace executor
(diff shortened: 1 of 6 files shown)


artifact/v3/scripts/executor.cpp
```diff
@@ -270,6 +265,22 @@ Error ScriptRunner::AsyncRunScripts(State state, Action action, HandlerFunction
...
+
+Error ScriptRunner::RunScripts(State state, Action action) {
+	auto run_err {error::NoError};
+	auto err {AsyncRunScripts(state, action, [this, &run_err](Error error) {
+		log::Info("Finished running sync script");
+		run_err = error;
+		this->loop_.Stop();
+	})};
+	if (err != error::NoError) {
+		this->loop_.Stop();
```
kacf, line 277: Should not call `Stop` when `Run` hasn't been called yet. I think you can just remove it.

artifact/v3/scripts/executor.hpp
```diff
@@ -107,15 +107,15 @@ class ScriptRunner {
...
-	Action action_;
 	string artifact_script_path_;
 	string rootfs_script_path_;
 	mender::common::processes::OutputCallback stdout_callback_;
 	mender::common::processes::OutputCallback stderr_callback_;
 	Error error_script_error_;
 	vector<string> collected_scripts_;
 	unique_ptr<mender::common::processes::Process> script_;
+	State state_;
+	Action action_;
```
lluiscampos, line 118: I think you meant to move these in the last commit, not in the first one :thinking:

artifact/v3/scripts/executor.cpp
```diff
@@ -270,6 +265,22 @@ Error ScriptRunner::AsyncRunScripts(State state, Action action, HandlerFunction
 	return Execute(scripts_iterator, scripts_iterator_end, handler);
 }
 
+
+Error ScriptRunner::RunScripts(State state, Action action) {
+	auto run_err {error::NoError};
+	auto err {AsyncRunScripts(state, action, [this, &run_err](Error error) {
```
lluiscampos, line 271: Same here wrt using explicit `=`. But again, I might be too late and this is now all across our code base?

kacf (approved):
Some minor stuff, looks good otherwise! Pre-approving.

artifact/v3/scripts/executor_test.cpp
```diff
@@ -57,13 +57,12 @@ TEST_F(ArtifactScriptTestEnv, VersionFileDoesNotExist_Success) {
...
 		loop,
-		executor::State::ArtifactInstall,
-		executor::Action::Enter,
 		chrono::seconds {10},
 		path::Join(tmpdir.Path(), "scripts"),
 		path::Join(tmpdir.Path(), "scripts")};
 	auto handler_func = [](error::Error err) { EXPECT_EQ(err, error::NoError) << err.String(); };
-	auto err = runner.AsyncRunScripts(handler_func);
+	auto err = runner.AsyncRunScripts(
+		executor::State::ArtifactInstall, executor::Action::Enter, handler_func);
```
lluiscampos, line 65: The tests fixes should also be on the first commit.

lluiscampos (approved):
Some nit picks wrt the contents of the first and last commits and one _rant_. See below.

artifact/v3/scripts/executor.cpp
lluiscampos: Right. By looking at the other PR from @kacf I can see that this is [common place](https://github.com/mendersoftware/mender/pull/1366/commits/42455fe465df6e47fd3c9f1c93f97680682b3f92#diff-cfe2c3235321dfa58992c067e1a040136acfb859fd842096917ce75a9209f58aR56). Too late, then.

artifact/v3/scripts/executor.cpp
```diff
@@ -104,7 +104,12 @@ bool isValidStateScript(const string &file, State state, Action action) {
 
 function<bool(const string &)> Matcher(State state, Action action) {
 	return [state, action](const string &file) {
-		return path::IsExecutable(file) and isValidStateScript(file, state, action);
+		auto exp_executable {path::IsExecutable(file)};
```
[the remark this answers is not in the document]
lluiscampos: Actually, the Google C++ style guide summarizes very well the issue (I hadn't read it, sorry for that). Thanks.

Okey, let's stick to `{` as much a possible and just be aware of the downside (namely types that have `std::initializer_list` constructor). 

:+1:

artifact/v3/scripts/executor.cpp
kacf: Initialization wasn't really the point I was trying to make though. I fully agree with that, based on the style guide and subtle effects people have described about it.

The thing which triggered me was that much more than initialization is being done here. `AsyncRunScripts` is full of side effects.

So it's more about readability and expectations, but maybe I'm just being old fashioned and being used to the `err = SomeBigOperation()`. Since you disagree and there are valid reasons for brace initialization, I won't object to it (and might even use it, we'll see 🙂).

lluiscampos: That is the distinction that I was also trying to make (in my other comment): out of the constructors world, it just reads so much better to use the `=` sign.

(I also feel I am being old fashioned here :sweat: )

PR #1374, opened 2023-08-30
feat: Introduce an ASIO+libdbus based DBusClient

vpodzime:
Just the very basic implementation only usable for methods that take no arguments and only return one string.

Ticket: MEN-6650
Changelog: none


common/dbus/platform/asio_libdbus/dbus_client.cpp
```diff
@@ -0,0 +1,307 @@
...
+	delete handler;
+}
+
+error::Error DBusClient::CallMethod(
+	const string &destination,
+	const string &path,
+	const string &iface,
+	const string &method,
+	DBusCallReplyHandler handler) {
+	// TODO: with a mutex to make this thread-safe?
```
kacf, line 100: I would perhaps suggest that it is documented as not thread safe instead, and the caller should use a separate instance if you need to utilize several threads. I'm not even sure how the API would work if it could be accessed from several threads simultaneously..

common/dbus.hpp
```diff
@@ -0,0 +1,103 @@
...
+#endif // MENDER_USE_ASIO_LIBDBUS
+
+private:
+	events::EventLoop &loop_;
+
+#ifdef MENDER_USE_ASIO_LIBDBUS
+	// Cannot initialize this in the constructor to a real connection because
+	// the connecting can fail.
+	unique_ptr<DBusConnection, decltype(&dbus_connection_unref)> dbus_conn_ {
+		nullptr, dbus_connection_unref};
```
kacf, line 93: Nice! I thought the deleter had to be passed in on every reset, so TIL.

common/dbus/platform/asio_libdbus/dbus_client.cpp
```diff
@@ -0,0 +1,307 @@
...
+	return error::NoError;
+}
+
+void HandleDispatch(DBusConnection *conn, DBusDispatchStatus status, void *data) {
+	DBusClient *client = static_cast<DBusClient *>(data);
+	if (status == DBUS_DISPATCH_DATA_REMAINS) {
+		// This must give other things in the loop a chance to run because
+		// dbus_connection_dispatch() below can cause this to be called again.
+		client->loop_.Post([conn, status]() {
+			while (dbus_connection_get_dispatch_status(conn) == DBUS_DISPATCH_DATA_REMAINS) {
```
kacf, line 139: Why does the first `DBUS_DISPATCH_DATA_REMAINS` cause a `Post()`, but all subsequent ones are done in a loop? Based on your comment, I would have assumed that they would repeatedly be calling `Post()` if necessary.

common/dbus/platform/asio_libdbus/dbus_client.cpp
```diff
@@ -0,0 +1,307 @@
...
+	// followed by ToggleDBusWatch() to enable the watch (see below). We
+	// simplify things for ourselves by ignoring disabled watches and only
+	// actually adding them when ToggleDBusWatch() is called.
+	if (!dbus_watch_get_enabled(w)) {
+		return TRUE;
+	}
+
+	DBusClient *client = static_cast<DBusClient *>(data);
+	asio::posix::stream_descriptor *sd =
+		new asio::posix::stream_descriptor {DBusClient::GetAsioIoContext(client->loop_)};
```
kacf, line 158: Should use `unique_ptr` (evidence: looks like a memory leak at the nearest return below!). If you need to get the ownership back, to pass it to C, you can use `release()`.

common/dbus/platform/asio_libdbus/dbus_client.cpp
```diff
@@ -0,0 +1,307 @@
...
+	return TRUE;
+}
+
+static void RemoveDBusTimeout(DBusTimeout *t, void *data) {
+	asio::steady_timer *timer = static_cast<asio::steady_timer *>(dbus_timeout_get_data(t));
+	if (timer != NULL) {
+		timer->cancel();
+		delete timer;
+	}
+	dbus_timeout_set_data(t, NULL, NULL);
```
kacf, line 243: Should you call this first? Probably it doesn't matter in a single threaded context, but to me it seems more natural with the opposite order: Remove from list first, then delete, so that there is never an invalid object in there.

common/dbus/platform/asio_libdbus/dbus_client.cpp
```diff
@@ -0,0 +1,307 @@
...
+	if (dbus_timeout_get_enabled(t)) {
+		AddDBusTimeout(t, data);
+	} else {
+		RemoveDBusTimeout(t, data);
+	}
+}
+
+static void HandleReply(DBusPendingCall *pending, void *data) {
+	DBusCallReplyHandler *handler = static_cast<DBusCallReplyHandler *>(data);
+	DBusMessage *reply = dbus_pending_call_steal_reply(pending);
```
kacf, line 256: Could use `unique_ptr` with unref deleters for both `pending` and `reply`, which would get rid of several cleanups below.

common/dbus.hpp
```diff
@@ -0,0 +1,103 @@
...
+#include <functional>
+#include <memory>
+#include <string>
+
+#include <common/error.hpp>
+#include <common/expected.hpp>
+#include <common/events.hpp>
+
+#ifdef MENDER_USE_ASIO_LIBDBUS
+#include <dbus/dbus.h>
```
kacf, line 29: Include order (see other comment).

kacf (approved):
Without knowing the details of the DBus API, it's difficult to follow the flow exactly, but it looks good to me from a high level. All my comments are just small observations.

common/CMakeLists.txt
```diff
@@ -264,3 +264,28 @@ gtest_discover_tests(crypto_test
...
+find_package(Boost 1.35 REQUIRED)
+pkg_check_modules(dbus REQUIRED dbus-1)
+add_library(common_dbus STATIC dbus/dbus.cpp dbus/platform/asio_libdbus/dbus_client.cpp)
+target_link_libraries(common_dbus PUBLIC
+  common_log
+  common_error
+  common_events
+  ${Boost_LIBRARIES}
+  ${dbus_LDFLAGS}
+  OpenSSL::SSL    # required by ASIO
```
oleorhagen, line 279: Q: Shouldn't this be provided by linking to `common_events` then?

common/dbus.hpp
```diff
@@ -0,0 +1,103 @@
...
+error::Error MakeError(DBusErrorCode code, const string &msg);
+
+// TODO: template <typename reply_type> or
+//       DBusCallReplyHandler<DBusString> or
+//       DBusCallStringReplyHandler?
+using DBusCallReplyHandler = function<void(expected::ExpectedString)>;
+
+class DBusClient : events::EventLoopObject {
+public:
+	DBusClient(events::EventLoop &loop) :
```
oleorhagen, line 66: Nit: explicit?

common/dbus/platform/asio_libdbus/dbus_client.cpp
```diff
@@ -0,0 +1,307 @@
...
+	if (!dbus_conn_) {
+		return MakeError(
+			ConnectionError,
+			string("Failed to get connection to system bus: ") + dbus_error.message + "["
+				+ dbus_error.name + "]");
+	}
+
+	dbus_connection_set_exit_on_disconnect(dbus_conn_.get(), FALSE);
+	if (!dbus_connection_set_watch_functions(
+			dbus_conn_.get(), AddDBusWatch, RemoveDBusWatch, ToggleDBusWatch, this, NULL)) {
```
oleorhagen, line 75: Q: Why `NULL` as opposed to `nullptr`? 

I guess it makes not difference really

common/dbus/platform/asio_libdbus/dbus_client.cpp
```diff
@@ -0,0 +1,307 @@
...
+	asio::posix::stream_descriptor *sd =
+		new asio::posix::stream_descriptor {DBusClient::GetAsioIoContext(client->loop_)};
+	boost::system::error_code ec;
+	sd->assign(dbus_watch_get_unix_fd(w), ec);
+	if (ec) {
+		log::Error("Failed to assign DBus FD to ASIO stream descriptor");
+		return FALSE;
+	}
+
+	unsigned int flags = dbus_watch_get_flags(w);
```
oleorhagen, line 166: S: Use `unsigned int flags {dbus_watch_get_flags(w)};` here, and we will have this checked at compile time

oleorhagen (commented):
Looks good I think!

common/dbus.hpp
```diff
@@ -0,0 +1,103 @@
...
+		const string &destination,
+		const string &path,
+		const string &iface,
+		const string &method,
+		DBusCallReplyHandler handler);
+
+#ifdef MENDER_USE_ASIO_LIBDBUS
+	// These take an instance of this class as the *data argument and need to
+	// access its private members. But they cannot be private member functions,
+	// we need them to be plain C functions.
```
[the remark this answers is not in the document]
vpodzime: > Neither approach is more correct, I'm just mentioning it as an alternative to `friend`.

I'll just stick to `friend` and keeping them obviously separate independent (C) functions.

common/dbus/platform/asio_libdbus/dbus_client.cpp
vpodzime: :+1: Actually, thread-safety and (true) absence of race conditions was among the things often mentioned by the authors of the other DBus implementations not based on *libdbus*.

common/dbus/platform/asio_libdbus/dbus_client.cpp
vpodzime: The docs say:
> If the dispatch status is [DBUS_DISPATCH_DATA_REMAINS](https://dbus.freedesktop.org/doc/api/html/group__DBusConnection.html#ggabf250a32ff740b2a8c99136e0142d8d2adf8e8a594ab338cb9cf7855b19e37e93), then [dbus_connection_dispatch()](https://dbus.freedesktop.org/doc/api/html/group__DBusConnection.html#ga66ba7df50d75f4bda6b6e942430b81c7) needs to be called to process incoming messages. However, [dbus_connection_dispatch()](https://dbus.freedesktop.org/doc/api/html/group__DBusConnection.html#ga66ba7df50d75f4bda6b6e942430b81c7) MUST NOT BE CALLED from inside the DBusDispatchStatusFunction. Indeed, almost any reentrancy in this function is a bad idea. Instead, the DBusDispatchStatusFunction should simply save an indication that messages should be dispatched later, when the main loop is re-entered.

And I found this when looking at an implementation combining *libdbus* with *libevent* so I did it the same way. It seems to work fine. I believe we only need the initial `Post()` to make sure that when the `HandleDispatch()` function is called, execution of the caller continues and finishes before the actual dispatch happens.

common/dbus.hpp
vpodzime: I wish there was a `-fno-fucking-implicit-anything` compiler option. But yes, we shouldn't (I guess?!) allow things like `DBusClient client = loop` or using `loop` in places where an actual `DBusClient` instance is expected.

common/dbus/platform/asio_libdbus/dbus_client.cpp
vpodzime: There are implicit conversions from `nullptr` to `NULL`, but I tried to stick with `NULL` when passing the value to a C API.

common/dbus/platform/asio_libdbus/dbus_client.cpp
vpodzime: Or `-fno-fucking-implicit-anything` :grin:

common/dbus/platform/asio_libdbus/dbus_client.cpp
kacf: Right, that makes sense.

PR #1375, opened 2023-08-31
Change HTTP API to be more subclass friendly.

kacf:


kacf:
@lluiscampos: There is a lot of noise here, but the important part is the removal of API in the various Request/Response types, and the corresponding introduction of virtual methods in `Server` and `Client`. For your case, only `Client` is needed I suppose, but I wanted to keep both the same.

common/http.hpp
```diff
@@ -236,90 +237,62 @@ class OutgoingRequest : public Request {
...
+	public Response,
+	virtual public io::Canceller,
+	public enable_shared_from_this<IncomingResponse> {
 public:
+	void Cancel() override;
+
 	// Use these after receiving the headers, if appropriate.
 	void SetBodyWriter(io::WriterPtr body_writer);
-	io::AsyncReaderPtr MakeBodyAsyncReader();
+	io::ExpectedAsyncReaderPtr MakeBodyAsyncReader();
```
lluiscampos, line 277: Rephrase the `if appropriate` part  to reflect that these are not optional anymore (the user needs to use one of the two methods below).

lluiscampos: This applies to`IncomingRequest` and other places too.

common/http/platform/beast/http.cpp
```diff
@@ -565,26 +615,37 @@ void Client::ReadHeaderHandler(const error_code &ec, size_t num_read) {
...
+		CallErrorHandler(MakeError(BodyIgnoredError, ""), request_, body_handler_);
 	}
 }
 
 void Client::AsyncReadNextBodyPart(
 	vector<uint8_t>::iterator start, vector<uint8_t>::iterator end, io::AsyncIoHandler handler) {
+	if (body_status_ == BodyReadingStatus::None
+		|| body_status_ == BodyReadingStatus::ReaderCreated) {
+		body_status_ = BodyReadingStatus::InProgress;
+	}
```
lluiscampos, line 628: When can we transition from `None` to `InProgress`? I thought that calling `MakeBodyAsyncReader` is mandatory before starting the read here.

common/http/platform/beast/http.cpp
```diff
@@ -565,26 +615,37 @@ void Client::ReadHeaderHandler(const error_code &ec, size_t num_read) {
...
+
+	if (body_status_ != BodyReadingStatus::InProgress) {
+		handler(0);
+		if (body_status_ == BodyReadingStatus::ReachedEnd) {
```
lluiscampos, line 632: I don't get it. Otherwise is an error and we should also call the user body handler, right?

common/http_test.cpp
```diff
@@ -1656,6 +1649,38 @@ TEST(HttpTest, TestRequestBodyReaderFailure) {
 	EXPECT_TRUE(got_read_error);
 }
 
+TEST(HttpTest, TestRequestBodyIgnored) {
```
lluiscampos, line 1665: This is testing the server ignoring the body. We should have a similar test for the client ignoring the body.

lluiscampos (commented):
The API change looks good to me. I should be able to make my client work now, thanks!

The last implementation details got me a bit confused with the `BodyReadingStatus`. Please help me understand before I can properly approve the PR :)

common/http_test.cpp
```diff
@@ -48,6 +48,19 @@ class TestInspector {
 		return server.streams_;
 	}
 };
+
+class TestServer : public Server {
+public:
+	using Server::Server;
```
vpodzime, line 54: What does this `using` do?

vpodzime (commented):
Lluis' comments/questions seem legit to me. Otherwise this looks good to my, honestly, limited knowledge/understanding of this machinery.

common/http/platform/beast/http.cpp
kacf: Good catch, this was a mistake, it should have been testing for `ReaderCreated` only.

common/http/platform/beast/http.cpp
kacf: This is called by the reader, which should always return `0` even if called multiple times. The body handler however, should only be called once, which is why we have this block to determine whether we have already done so or not.

common/http_test.cpp
kacf: It reuses the constructors from the original class, since I don't need any modifications in the sub class.

kacf:
@lluiscampos: Updated!

lluiscampos (approved):
Looks good to me now :+1:

common/http_test.cpp
vpodzime: I thought that was the default

mender-update/CMakeLists.txt
```diff
@@ -154,7 +154,9 @@ target_link_libraries(mender_update_cli_test PUBLIC
   gmock
 )
 target_compile_options(mender_update_cli_test PRIVATE ${PLATFORM_SPECIFIC_COMPILE_OPTIONS})
-gtest_discover_tests(mender_update_cli_test)
+# Use NO_PRETTY_VALUES to avoid very long byte strings in the output due to
+# parametrized tests that have binary objects as input.
+gtest_discover_tests(mender_update_cli_test NO_PRETTY_VALUES)
```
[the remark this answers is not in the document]
kacf: Trust me, what it prints is just garbage. Structs turned into a raw byte stream with no structure. It's not really necessary because it still prints the name, so you can identify the test case.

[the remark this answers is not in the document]
kacf: Or is this related to the `<<` operator you made? That would be prettier, obviously, but I think it's way too verbose to printed for every test.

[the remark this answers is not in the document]
oleorhagen: yes

PR #1382, opened 2023-09-05
chore: Clean-up `gmock` use in tests

lluiscampos:
Many tests had the header and include path while only using GTest.

Completely unnecessary clean-up...


oleorhagen (approved):

vpodzime (approved):
ACK. I recall it was needed for something quite unexpected like `EXPECT_THAT()`. But as long as CI is green :+1:

lluiscampos:
> ACK. I recall it was needed for something quite unexpected like `EXPECT_THAT()`. But as long as CI is green 👍

Yes, that is correct. I removed it from the tests that don't use it.

PR #1383, opened 2023-09-06
test: Fix and add the daemon state script tests

oleorhagen:
This adds the expected functionality for the daemon state scripts.

Still some things I'm wondering, have a close look at:

* LoadStateFromDB
* How to handle DB errors when running state scripts, and saving state
* The updated tests

### DIFF
--- a/mender-update/daemon/context.cpp
+++ b/mender-update/daemon/context.cpp
@@ -408,16 +408,22 @@ static error::Error UnmarshalJsonStateData(const json::Json &json, StateData &st
 	auto exp_string_vector = json_artifact.Get("CompatibleDevices").and_then(json::ToStringVector);
 	SetOrReturnIfError(artifact.compatible_devices, exp_string_vector);
 
+	exp_string = json_artifact.Get("ArtifactName").and_then(json::ToString);
+	SetOrReturnIfError(artifact.artifact_name, exp_string);
+
 	exp_string_vector = json_artifact.Get("PayloadTypes").and_then(json::ToStringVector);
 	SetOrReturnIfError(artifact.payload_types, exp_string_vector);
+	// It's possible for there not to be an initialized update,
+	// if the deployment failed before we could successfully parse the artifact.
+	if (artifact.payload_types.size() == 0 and artifact.artifact_name == "") {
+		return error::NoError;
+	}
 	if (artifact.payload_types.size() != 1) {
 		return error::Error(
 			make_error_condition(errc::not_supported),
-			"Only exactly one payload type is supported");
+			"Only exactly one payload type is supported. Got: "
+				+ to_string(artifact.payload_types.size()));
 	}
-
-	exp_string = json_artifact.Get("ArtifactName").and_then(json::ToString);
-	SetOrReturnIfError(artifact.artifact_name, exp_string);
 
 	exp_string = json_artifact.Get("ArtifactGroup").and_then(json::ToString);
 	SetOrReturnIfError(artifact.artifact_group, exp_string);
--- a/mender-update/daemon/state_machine.hpp
+++ b/mender-update/daemon/state_machine.hpp
@@ -62,6 +62,8 @@ private:
 	// Main states
 	///////////////////////////////////////////////////////////////////////////////////////////
 
+	InitState init_state_;
+
 	IdleState idle_state_;
 	SubmitInventoryState submit_inventory_state_;
 	PollForDeploymentState poll_for_deployment_state_;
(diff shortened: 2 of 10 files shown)


mender-update/daemon/state_test.cpp
```diff
@@ -2644,6 +2768,20 @@ vector<StateTransitionsTestCase> GenerateStateTransitionsTestCases() {
 	};
 }
 
+std::ostream &operator<<(std::ostream &os, const StateTransitionsTestCase &test_case) {
```
oleorhagen, line 2771: So that we get something, rather than the ugly byte output.

Format can be debated though

oleorhagen (commented):
Enjoy!

mender-update/daemon/state_test.cpp
```diff
@@ -486,7 +612,7 @@ vector<StateTransitionsTestCase> GenerateStateTransitionsTestCases() {
 					"ArtifactReboot_Error_00",
 					"ArtifactRollback_Enter_00",
 					"ArtifactRollback",
-					"ArtifactRollback_Leave_00",
+					"ArtifactRollback_Error_00",
```
kacf, line 615: No error scripts on the error path, as described [here](https://github.com/mendersoftware/mender/pull/1368#discussion_r1305213570).

mender-update/daemon/state_test.cpp
```diff
@@ -1067,9 +1195,31 @@ vector<StateTransitionsTestCase> GenerateStateTransitionsTestCases() {
...
+					"ArtifactInstall",
+					"ArtifactInstall_Leave_00",
+					"ArtifactReboot_Enter_00",
+					"ArtifactReboot",
+					"ArtifactVerifyReboot",
+					"ArtifactReboot_Leave_00",
+					"ArtifactCommit_Enter_00",
+					"ArtifactCommit",
+					"ArtifactCommit_Leave_00",
+					"Cleanup",
```
kacf, line 1211: This is wrong, it should not run any installation steps.

mender-update/daemon/state_test.cpp
```diff
@@ -2189,66 +2355,24 @@ vector<StateTransitionsTestCase> GenerateStateTransitionsTestCases() {
...
+					"ArtifactRollbackReboot",    "ArtifactVerifyRollbackReboot",
+					"ArtifactRollbackReboot",    "ArtifactVerifyRollbackReboot",
+					"ArtifactRollbackReboot",    "ArtifactVerifyRollbackReboot",
+					"ArtifactRollbackReboot",    "ArtifactVerifyRollbackReboot",
+					"ArtifactRollbackReboot",    "ArtifactVerifyRollbackReboot",
+					"ArtifactRollbackReboot",    "ArtifactVerifyRollbackReboot",
+					"ArtifactRollbackReboot",    "ArtifactVerifyRollbackReboot",
+					"ArtifactRollbackReboot",    "ArtifactVerifyRollbackReboot",
+					"ArtifactRollbackReboot",    "ArtifactVerifyRollbackReboot",
+					// TODO - Missing ArtifactRollback_Leave_00
```
kacf, line 2375: Why? It is listed above.

mender-update/daemon/state_test.cpp
```diff
@@ -2537,11 +2650,20 @@ vector<StateTransitionsTestCase> GenerateStateTransitionsTestCases() {
...
 			.state_chain =
 				{
+					"Download_Enter_00",
 					"Download",
+					"Download_Leave_00",
+					"ArtifactInstall_Error_00",
+					"ArtifactRollback_Enter_00",
 					"ArtifactRollback",
+					"ArtifactRollback_Leave_00",
+					// "ArtifactRollbackReboot_Enter_00", // Fails, so not ran
```
kacf, line 2660: Hmm, I didn't understand this. Why?

mender-update/daemon/state_test.cpp
```diff
@@ -2608,7 +2732,7 @@ vector<StateTransitionsTestCase> GenerateStateTransitionsTestCases() {
 					"Download_Leave_00",
 					"ArtifactInstall_Enter_00",
 					"ArtifactInstall",
-					"ArtifactInstall_Leave_00",
+					"ArtifactInstall_Error_00",
```
kacf, line 2735: This one can be debated. Strictly speaking the check is unrelated to the install and happens "afterwards". But I'm fine with leaving it like this, it is an extremely unlikely error path anyway, since it's a static query.

mender-update/daemon/state_test.cpp
kacf: Sure, but is it for debugging? It's not clear to me where it is used.

mender-update/daemon/state_test.cpp
```diff
@@ -2693,6 +2831,15 @@ class StateDeathTest : public testing::TestWithParam<StateTransitionsTestCase> {
 	mtesting::TemporaryDirectory tmpdir_;
 };
 
+INSTANTIATE_TEST_SUITE_P(
+	Regular_Non_Update_State_Tests,
+	StateDeathTest,
+	::testing::ValuesIn(idle_and_sync_test_cases),
```
kacf, line 2837: Why is this a separate data set?

mender-update/daemon/state_test.cpp
oleorhagen: It's as opposed to the output:

GetParam() = 0x64 64 46 42 32 64 64 

beauty

mender-update/daemon/state_test.cpp
```diff
@@ -87,6 +87,132 @@ struct StateTransitionsTestCase {
 	bool device_type_mismatch {false};
 };
 
+
+vector<StateTransitionsTestCase> idle_and_sync_test_cases {
+	StateTransitionsTestCase {
+		.case_name = "Normal_flow_Idle_Sync_Idle",
+		.state_chain =
+			{
+				"Idle_Enter_00",           "Idle_Leave_00",
```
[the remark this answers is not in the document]
oleorhagen: Yeh, I agree, it's retarded.

mender-update/daemon/state_test.cpp
```diff
@@ -87,6 +87,132 @@ struct StateTransitionsTestCase {
...
+			{
+				"Idle_Enter_00",
+				"Idle_Leave_00",
+				"Sync_Enter_00",
+				"Sync_Leave_00",
+				"Sync_Error_00",
+				"Idle_Enter_00",
+				"Idle_Leave_00",
+				"Sync_Enter_00",
+				"Sync_Leave_00",
```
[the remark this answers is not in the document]
oleorhagen: It only fails in sync leave the first time:

```
				"Sync_Leave_00", <- here
				"Sync_Error_00",
				"Idle_Enter_00",
				"Idle_Leave_00",
				"Sync_Enter_00",
				"Sync_Leave_00",
				"Download_Enter_00",
```

mender-update/daemon/state_test.cpp
```diff
@@ -87,6 +87,132 @@ struct StateTransitionsTestCase {
...
+
+vector<StateTransitionsTestCase> idle_and_sync_test_cases {
+	StateTransitionsTestCase {
+		.case_name = "Normal_flow_Idle_Sync_Idle",
+		.state_chain =
+			{
+				"Idle_Enter_00",           "Idle_Leave_00",
+				"Sync_Enter_00",           "Sync_Leave_00",
+				"Idle_Enter_00",           "Idle_Leave_00",
+				"Sync_Enter_00",           "Sync_Leave_00",
```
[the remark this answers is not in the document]
oleorhagen: It's sending inventory first :p

mender-update/daemon/state_test.cpp
oleorhagen: I figured cause it's a separate part of the state-machine, and it makes running them separately easier, but can merge them if you prefer.

mender-update/daemon/state_test.cpp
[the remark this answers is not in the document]
kacf: Ah, of course.

mender-update/daemon/state_test.cpp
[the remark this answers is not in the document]
kacf: Ah, right, it only does it once, I forgot about that detail.

mender-update/daemon/state_test.cpp
kacf: No, it's fine. I was just a bit confused since you added this, but didn't add a test, so I just wondered if you had forgotten something.

mender-update/daemon/state_test.cpp
oleorhagen: Hmm, but when `Download_Enter_00` is rerun after having been killed, it succeeds, how can it then know it should not continue down the successful path?

mender-update/daemon/state_test.cpp
oleorhagen: I was thinking of after the last `ArtifactVerifyRollbackReboot`

mender-update/daemon/state_test.cpp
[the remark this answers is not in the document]
oleorhagen: I will add a comment in the test

mender-update/daemon/state_test.cpp
kacf: Hmm, this is not correct I think. It should still run if it's able, and it's on the error path. After all, `ArtifactFailure_Enter_00` runs, and it's also a SaveState.

oleorhagen: Then we have to change the `SaveState` functionality.

mender-update/daemon/state_test.cpp
oleorhagen: So this is literally what we want?

```
					"Download_Enter_00",
					"Cleanup",
```

kacf: I think that cleanup might be missing from the record because it fails so early, so the Update Module isn't initialized yet. It still runs internally, but you probably won't get it in the list. You'll have to see.

oleorhagen: Ur right! Was looking into why it wasn't showing just now. Got tired, and checked my mail :laughing:

mender-update/daemon/state_test.cpp
oleorhagen: Probably wasn't a save state as I made the test

mender-update/daemon/state_test.cpp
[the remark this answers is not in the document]
kacf: I think this should be fixed before submission. In addition to reading, it makes copy-pasting/rearranging very tricky. If you can find an option for it in the clang-format it would be great. But feel free to slap a `clang-format off` on it if you don't feel like dealing with it now.

mender-update/daemon/state_test.cpp
kacf: So do we need this? Even with a pretty format, I think it's too long, the name should be enough.

An alternative could be to print it at the start of the test. Then it won't be visible when the test passes, but it will be visible if it fails.

kacf (commented):
The test code looks good now, apart from the things I pointed out below.

But what's the status of the rest? Originally this was a test-only PR, and that is what I have reviewed. If the rest is also to be reviewed, I need a bit more time.

mender-update/daemon/state_test.cpp
oleorhagen: Yeh, can change it to be name only.

> So de we need this?

It's only called when `googletest` tries to print `GetParam()`, so I don't see it hurting.

kacf (commented):
Really nice work! I have a lot of comments, but practically all of them are just nitpicks. I'm very happy with how this is solved!

mender-update/daemon/state_machine/state_machine.cpp
```diff
@@ -54,130 +59,222 @@ StateMachine::StateMachine(Context &ctx, events::EventLoop &event_loop) :
...
-	dt.states_.AddTransition(dt.failure_state_,                      se::DeploymentEnded,            dt.idle_state_,                       tf::Immediate);
+ 	main_states_.AddTransition(update_rollback_successful_state_,       se::Success,                     ss.failure_enter_,                       tf::Immediate);
 
-	dt.states_.AddTransition(dt.rollback_attempted_state_,           se::Failure,                    dt.rollback_failed_state_,            tf::Immediate);
-	dt.states_.AddTransition(dt.rollback_attempted_state_,           se::DeploymentEnded,            dt.idle_state_,                       tf::Immediate);
+ 	main_states_.AddTransition(update_failure_state_,                   se::Success,                     ss.failure_leave_update_save_provides_,  tf::Immediate);
+ 	main_states_.AddTransition(update_failure_state_,                   se::Failure,                     ss.failure_leave_update_save_provides_,  tf::Immediate);
+ 	main_states_.AddTransition(update_failure_state_,                   se::StateLoopDetected,           ss.failure_leave_state_loop_state_,      tf::Immediate);
```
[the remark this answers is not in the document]
oleorhagen: Depends if we want to call the `Failure Leave` scripts or not. If we do, which I believe, then this is needed.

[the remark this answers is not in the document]
kacf: None of the others are doing this, so I vote for going straight to `state_loop_state_`. Otherwise the execution of `Failure_Leave` depends on whether you hit the counter here, or someplace earlier.

mender-update/daemon/context.cpp
```diff
@@ -410,10 +410,16 @@ static error::Error UnmarshalJsonStateData(const json::Json &json, StateData &st
 
 	exp_string_vector = json_artifact.Get("PayloadTypes").and_then(json::ToStringVector);
 	SetOrReturnIfError(artifact.payload_types, exp_string_vector);
+	// It's possible for there not to be an initialized update,
+	// if the deployment failed before we could successfully parse the artifact.
+	if (artifact.payload_types.size() == 0) {
+		return error::NoError;
+	}
```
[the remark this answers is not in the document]
oleorhagen: Correct.

We could consider making `payload_types` default value something else, to clearly separate it from `zero` payloads, and `not parsed yet` maybe?

mender-update/daemon/state_machine/state_machine.cpp
```diff
@@ -54,130 +59,222 @@ StateMachine::StateMachine(Context &ctx, events::EventLoop &event_loop) :
...
+  main_states_.AddTransition(ss.download_leave_save_provides,         se::Success,                     update_save_provides_state_,             tf::Immediate);
+  main_states_.AddTransition(ss.download_leave_save_provides,         se::Failure,                     ss.download_error_,                      tf::Immediate);
 
-	main_states_.AddTransition(poll_for_deployment_state_,           se::Success,                    send_download_status_state_,          tf::Immediate);
-	main_states_.AddTransition(poll_for_deployment_state_,           se::NothingToDo,                idle_state_,                          tf::Immediate);
-	main_states_.AddTransition(poll_for_deployment_state_,           se::Failure,                    idle_state_,                          tf::Immediate);
+ 	main_states_.AddTransition(update_download_state_,                  se::StateLoopDetected,           state_loop_state_,                       tf::Immediate);
 
-	// Cannot fail due to FailureMode::Ignore.
-	main_states_.AddTransition(send_download_status_state_,          se::Success,                    update_download_state_,               tf::Immediate);
```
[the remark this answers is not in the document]
oleorhagen: I have another idea, which I wanted to discuss with you, how about we give it a `AlwaysSuccess` event, and we don't need the comments?

mender-update/daemon/state_machine/state_machine.cpp
```diff
@@ -54,130 +59,222 @@ StateMachine::StateMachine(Context &ctx, events::EventLoop &event_loop) :
...
 
+	auto &ss = state_scripts_;
+
 	// When updating the table below, make sure that the initial states are in sync as well, in
 	// LoadStateFromDb().
 
 	// clang-format off
-	main_states_.AddTransition(idle_state_,                          se::DeploymentPollingTriggered, poll_for_deployment_state_,           tf::Deferred );
-	main_states_.AddTransition(idle_state_,                          se::InventoryPollingTriggered,  submit_inventory_state_,              tf::Deferred );
+ 	main_states_.AddTransition(init_state_,                             se::Started,                     ss.idle_enter_,                          tf::Deferred);
```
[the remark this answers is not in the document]
oleorhagen: Yes, I will leave it as an exercise for the reader to figure out why :smiley:

mender-update/daemon/state_machine/state_machine.cpp
```diff
@@ -54,130 +59,222 @@ StateMachine::StateMachine(Context &ctx, events::EventLoop &event_loop) :
...
+ 	main_states_.AddTransition(ss.sync_error_,                          se::Failure,                     ss.idle_enter_,                          tf::Immediate);
+
+ 	main_states_.AddTransition(ss.idle_leave_inv_,                      se::Success,                     ss.sync_enter_inventory_,                tf::Immediate);
+ 	main_states_.AddTransition(ss.idle_leave_inv_,                      se::Failure,                     ss.sync_enter_inventory_,                tf::Immediate);
+
+ 	main_states_.AddTransition(ss.sync_enter_inventory_,                se::Success,                     submit_inventory_state_,                 tf::Immediate);
+ 	main_states_.AddTransition(ss.sync_enter_inventory_,                se::Failure,                     ss.sync_error_,                          tf::Immediate);
+
+ 	main_states_.AddTransition(submit_inventory_state_,                 se::Success,                     ss.sync_leave_,                          tf::Immediate);
+ 	main_states_.AddTransition(submit_inventory_state_,                 se::Failure,                     ss.sync_leave_,                          tf::Immediate);
```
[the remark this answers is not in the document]
oleorhagen: Hmm, didn't we agree `Sync` was one of the states which ignored `Error`? :thinking:

mender-update/daemon/context.cpp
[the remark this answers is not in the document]
kacf: Yes, or use something else which is even more unambiguously invalid in an artifact. What about `state_data->update_info.id == ""`?

mender-update/daemon/state_machine/state_machine.cpp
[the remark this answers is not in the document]
kacf: Oh, that's a good idea, go for it!

mender-update/daemon/state_machine/state_machine.cpp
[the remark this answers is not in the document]
kacf: No [Sync error scripts are called](https://github.com/mendersoftware/mender/pull/1368#discussion_r1305213570). Besides, it's already being called elsewhere, so not doing it here would be inconsistent.

mender-update/daemon/context.cpp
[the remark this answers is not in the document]
oleorhagen: I'd rather have an initialised, semantic value tbh

mender-update/daemon/state_machine/state_machine.cpp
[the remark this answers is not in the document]
oleorhagen: We agreed that it should be called :see_no_evil: 

:military_helmet:

mender-update/daemon/context.cpp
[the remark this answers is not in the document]
kacf: Sure, just remember that if you introduce a new value, then it's a database schema update. I think it should be fine though, because all of this happens before the download has started, so there should be no risk of an old client trying to read it.

mender-update/daemon/state_test.cpp
oleorhagen: I still get this after rebasing on `master`:
```
[  FAILED  ] 1 test, listed below:
[  FAILED  ] Regular_Non_Update_State_Tests/StateDeathTest.StateTransitionsTest/Idle_Leave_Error, where GetParam() = 200-byte object <10-E8 15-B2 B9-55 00-00 10-00 00-00 00-00 00-00 10-00 00-00 00-00 00-00 6F-6E 5F-55 70-64 61-74 50-DF 15-B2 B9-55 00-00 10-E2 15-B2 B9-55 00-00 10-E2 15-B2 B9-55 00-00 60-E3 15-B2 B9-55 00-00 ... 00-00 00-00 00-00 00-00 00-00 00-00 00-00 00-00 00-00 00-00 00-00 00-00 00-00 00-00 00-00 00-00 00-00 00-00 00-00 00-00 00-00 00-00 00-00 00-00 00-00 00-00 FF-FF FF-FF FF-FF FF-FF 00-00 01-00>
```

mender-update/daemon/context.cpp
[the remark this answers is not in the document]
oleorhagen: Hmm, didn't think of that.. I'll ponder it a bit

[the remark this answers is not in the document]
oleorhagen: Alright, going with `update-info.id` :+1:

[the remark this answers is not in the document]
oleorhagen: Decided to use the combination of missing `payload` and `artifact` name as the empty marker

mender-update/daemon/state_machine.hpp
```diff
@@ -107,6 +109,293 @@ class StateMachine {
...
+				rootfs_script_path),
+			reboot_enter_(
+				loop,
+				script_executor::State::ArtifactReboot,
+				script_executor::Action::Enter,
+				retry_interval,
+				artifact_script_path,
+				rootfs_script_path,
+				Context::kUpdateStateArtifactReboot,
+				true),
```
kacf, line 241: You added `true` here in the latest update. Was it intentional, or were you just a bit quick when adding it to `rollback_reboot_enter` (where it indeed should be)?

kacf (approved):
Nice!

Beware of one possible copy-paste error, looks good otherwise!

mender-update/daemon/state_machine.hpp
oleorhagen: Uhh! Good catch!

PR #1384, opened 2023-09-07
DBusClient followup

vpodzime:
Two fixes + signal handling.

### DIFF
--- a/common/dbus/platform/asio_libdbus/dbus_client.cpp
+++ b/common/dbus/platform/asio_libdbus/dbus_client.cpp
@@ -64,10 +64,12 @@ error::Error DBusClient::InitializeConnection() {
 	dbus_error_init(&dbus_error);
 	dbus_conn_.reset(dbus_bus_get(DBUS_BUS_SYSTEM, &dbus_error));
 	if (!dbus_conn_) {
-		return MakeError(
+		auto err = MakeError(
 			ConnectionError,
 			string("Failed to get connection to system bus: ") + dbus_error.message + "["
 				+ dbus_error.name + "]");
+		dbus_error_free(&dbus_error);
+		return err;
 	}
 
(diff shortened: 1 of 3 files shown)


kacf (approved):
👍

common/dbus/platform/asio_libdbus/dbus_client.cpp
```diff
@@ -64,10 +64,12 @@ error::Error DBusClient::InitializeConnection() {
 	dbus_error_init(&dbus_error);
 	dbus_conn_.reset(dbus_bus_get(DBUS_BUS_SYSTEM, &dbus_error));
 	if (!dbus_conn_) {
-		return MakeError(
```
oleorhagen, line 67: If we'd wrapped it in a `unique_ptr` to begin with, this would not have happened:

```
error::Error DBusClient::InitializeConnection() {
 	DBusError dbus_error;
 	dbus_error_init(&dbus_error);
+	unique_ptr<DBusError, decltype(&dbus_error_free)> dbus_err_ptr {&dbus_error, dbus_error_free};
 	dbus_conn_.reset(dbus_bus_get_private(DBUS_BUS_SYSTEM, &dbus_error));
 	if (!dbus_conn_) {
 		auto err = MakeError(
 			ConnectionError,
 			string("Failed to get connection to system bus: ") + dbus_error.message + "["
 				+ dbus_error.name + "]");
-		dbus_error_free(&dbus_error);
 		return err;
 	}
```

oleorhagen (approved):
lgtm, I don't know the `DBus` stuff very well though, but code reads well imo

common/dbus/platform/asio_libdbus/dbus_client.cpp
vpodzime: I'll consider it. It didn't happen because I forgot, the reason was I thought the stack-allocated error only had static/stack-allocated data in it.

PR #1385, opened 2023-09-10
Integration test runs - client fixes

oleorhagen:
All miscellaneous changes, and nitpicks, and fixes.

### DIFF
--- a/mender-update/deployments/deployments.cpp
+++ b/mender-update/deployments/deployments.cpp
@@ -286,13 +286,16 @@ error::Error DeploymentClient::PushStatus(
 			auto resp = exp_resp.value();
 			auto content_length = resp->GetHeader("Content-Length");
 			if (!content_length) {
-				log::Error("Failed to get content length from the status API response headers");
+				log::Error(
+					"Failed to get content length from the status API response headers: "
+					+ content_length.error().String());
 				body_writer->SetUnlimited(true);
 			} else {
 				auto ex_len = common::StringToLongLong(content_length.value());
 				if (!ex_len) {
 					log::Error(
-						"Failed to convert the content length from the status API response headers to an integer");
+						"Failed to convert the content length from the status API response headers to an integer: "
+						+ ex_len.error().String());
 					body_writer->SetUnlimited(true);
 				} else {
 					received_body->resize(ex_len.value());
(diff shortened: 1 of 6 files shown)


mender-update/deployments/deployments.cpp
```diff
@@ -285,12 +285,18 @@ error::Error DeploymentClient::PushStatus(
...
-			if (!ex_len) {
+			if (!content_length) {
 				log::Error("Failed to get content length from the status API response headers");
 				body_writer->SetUnlimited(true);
 			} else {
-				received_body->resize(ex_len.value());
+				auto ex_len = common::StringToLongLong(content_length.value());
+				if (!ex_len) {
+					log::Error(
+						"Failed to convert the content length from the status API response headers to an integer");
```
lluiscampos, line 295: I would append to the log message `ex_len.error().String()`

lluiscampos: Pre-existing,

Shouldn't this (and few lines above) be a warning instead? Setting the body_writer to be unlimited makes it in effect to recover.

mender-update/deployments/deployments.cpp
```diff
@@ -449,12 +449,18 @@ error::Error DeploymentClient::PushLogs(
...
+			if (!content_length) {
+				log::Error("Failed to get content length from the status API response headers");
 				body_writer->SetUnlimited(true);
 			} else {
-				received_body->resize(ex_len.value());
+				auto ex_len = common::StringToLongLong(content_length.value());
+				if (!ex_len) {
+					log::Error(
+						"Failed to convert the content length from the status API response headers to an integer");
+					body_writer->SetUnlimited(true);
```
lluiscampos, line 460: Same suggestion(s) than before :+1:

lluiscampos (approved):
Nice fixes! I am curious to hear how many tests you managed to ~run~ pass :smiley:

oleorhagen:
> Nice fixes! I am curious to hear how many tests you managed to ~run~ pass 😃

Straight OOTB, none so far unfortunately, I've been cheating a bit to get them to run, like the certs don't work yet, but that will be todays fix, then we will see :)

Locally, I can do simple file updates with `demo` though :tada:

mender-update/deployments/deployments.cpp
oleorhagen: Not against, it certainly. I'll add a separate commit for it though

PR #1387, opened 2023-09-11
Templated DBusClient methods

vpodzime:
Ticket: none
Changelog: none

### DIFF
--- a/common/dbus.hpp
+++ b/common/dbus.hpp
@@ -21,6 +21,7 @@
 #include <memory>
 #include <string>
 #include <unordered_map>
+#include <utility>
 
 #ifdef MENDER_USE_ASIO_LIBDBUS
 #include <dbus/dbus.h>
@@ -70,6 +71,8 @@ using DBusSignalHandler = function<void(SignalValueType)>;
 // in the future.
 using SignalSpec = string;
 
+using ExpectedStringPair = expected::expected<std::pair<string, string>, error::Error>;
+
 // Note: Not a thread-safe class, create multiple instances if needed. However,
 // the implementation based on libdbus is likely to suffer from potential race
 // conditions in the library itself.
@@ -118,6 +121,7 @@ private:
 #endif // MENDER_USE_ASIO_LIBDBUS
 
 	unordered_map<SignalSpec, DBusSignalHandler<expected::ExpectedString>> signal_handlers_string_;
+	unordered_map<SignalSpec, DBusSignalHandler<ExpectedStringPair>> signal_handlers_string_pair_;
 
 	error::Error InitializeConnection();
 
--- a/common/dbus.hpp
+++ b/common/dbus.hpp
@@ -29,6 +29,7 @@
 #include <common/error.hpp>
 #include <common/expected.hpp>
 #include <common/events.hpp>
+#include <common/optional.hpp>
 
 namespace mender {
 namespace common {
@@ -37,6 +38,7 @@ namespace dbus {
 namespace error = mender::common::error;
 namespace expected = mender::common::expected;
 namespace events = mender::common::events;
+namespace optional = mender::common::optional;
 
 using namespace std;
 
@@ -57,11 +59,11 @@ extern const DBusErrorCategoryClass DBusErrorCategory;
 
 error::Error MakeError(DBusErrorCode code, const string &msg);
 
-// TODO: template <typename reply_type> or
-//       DBusCallReplyHandler<DBusString> or
-//       DBusCallStringReplyHandler?
-using DBusCallReplyHandler = function<void(expected::ExpectedString)>;
-using DBusSignalHandler = function<void(string)>;
+template <typename ReplyType>
+using DBusCallReplyHandler = function<void(ReplyType)>;
+
+template <typename SignalValueType>
+using DBusSignalHandler = function<void(SignalValueType)>;
 
 // Might need something like
 //   struct {string sender; string iface; string signal;}
@@ -78,16 +80,20 @@ public:
 
 	~DBusClient();
 
-	// TODO: template <typename reply_type> (see above)
+	template <typename ReplyType>
 	error::Error CallMethod(
 		const string &destination,
 		const string &path,
 		const string &iface,
 		const string &method,
-		DBusCallReplyHandler handler);
+		DBusCallReplyHandler<ReplyType> handler);
 
+	template <typename SignalValueType>
 	error::Error RegisterSignalHandler(
-		const string &sender, const string &iface, const string &signal, DBusSignalHandler handler);
+		const string &sender,
+		const string &iface,
+		const string &signal,
+		DBusSignalHandler<SignalValueType> handler);
 	void UnregisterSignalHandler(const string &sender, const string &iface, const string &signal);
 
 #ifdef MENDER_USE_ASIO_LIBDBUS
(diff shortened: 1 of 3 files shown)


common/dbus/platform/asio_libdbus/dbus_client.cpp
```diff
@@ -135,16 +142,34 @@ error::Error DBusClient::CallMethod(
 	// We need to create a copy here because we need to make sure the handler,
 	// which might be a lambda, even with captures, will live long enough for
 	// the finished pending call to use it.
-	DBusCallReplyHandler *handler_copy = new DBusCallReplyHandler(handler);
-	if (!dbus_pending_call_set_notify(pending, HandleReply, handler_copy, FreeHandlerCopy)) {
+	auto *handler_copy = new DBusCallReplyHandler<ReplyType>(handler);
+	if (!dbus_pending_call_set_notify(
+			pending, HandleReply<ReplyType>, handler_copy, FreeHandlerCopy<ReplyType>)) {
```
kacf, line 149: Preexisting, but do we leak `handler_copy` here if this call fails?

Even though the lifetime is short, I would consider a `unique_ptr` for `handler_copy`. If we forget to free a raw pointer, it's an invisible bug (yes, asan can catch it, but only if this block hits). OTOH, if we forget to call `unique_ptr::release()` it's a very obvious bug, which will immediately be visible.

common/dbus/platform/asio_libdbus/dbus_client.cpp
```diff
@@ -135,16 +142,34 @@ error::Error DBusClient::CallMethod(
...
 
+template error::Error DBusClient::CallMethod(
+	const string &destination,
+	const string &path,
+	const string &iface,
+	const string &method,
+	DBusCallReplyHandler<expected::ExpectedString> handler);
+
+template <>
+void DBusClient::AddSignalHandler(
```
kacf, line 164: Ended up using both template instantiation *and* template specialization! Advanced stuff! :man_scientist: :smile: 

But I agree that this combo solves the problem well.

common/dbus/platform/asio_libdbus/dbus_client.cpp
```diff
@@ -353,6 +370,11 @@ bool CheckDBusMessageSignature<expected::ExpectedString>(const string &signature
 	return signature == DBUS_TYPE_STRING_AS_STRING;
 }
 
+template <>
+bool CheckDBusMessageSignature<ExpectedStringPair>(const string &signature) {
+	return signature == string(DBUS_TYPE_STRING_AS_STRING) + DBUS_TYPE_STRING_AS_STRING;
```
kacf, line 375: Nitpick: Parentheses around the addition would help readability.

kacf (approved):
Nice solution!

common/dbus/platform/asio_libdbus/dbus_client.cpp
vpodzime: > Ended up using both template instantiation _and_ template specialization! Advanced stuff! 👨‍🔬 😄

No idea... This is just what I arrived to by trying various combinations of things. :joy:

common/dbus/platform/asio_libdbus/dbus_client.cpp
vpodzime: Yes, you're right. Actually, in this particular case, `dbus_pending_call_set_notify()` only fails if there's not enough memory for it, but that's not a good enough excuse. :)

common/dbus.hpp
```diff
@@ -21,6 +21,7 @@
 #include <memory>
 #include <string>
 #include <unordered_map>
+#include <utility>
 
 #ifdef MENDER_USE_ASIO_LIBDBUS
 #include <dbus/dbus.h>
```
oleorhagen, line 27: Nit: Pre-existing, but this is a `C` header, no?

common/dbus/platform/asio_libdbus/dbus_client.cpp
```diff
@@ -424,6 +470,16 @@ optional::optional<DBusSignalHandler<expected::ExpectedString>> DBusClient::GetS
 	}
 }
 
+template <>
+optional::optional<DBusSignalHandler<ExpectedStringPair>> DBusClient::GetSignalHandler(
+	const SignalSpec &spec) {
+	if (signal_handlers_string_pair_.find(spec) != signal_handlers_string_pair_.cend()) {
+		return signal_handlers_string_pair_[spec];
+	} else {
```
oleorhagen, line 482: nit: Unneeded else

common/dbus/platform/asio_libdbus/dbus_client.cpp
```diff
@@ -437,6 +493,13 @@ DBusHandlerResult MsgFilter(DBusConnection *connection, DBusMessage *message, vo
 	const string signature {dbus_message_get_signature(message)};
 
 	auto opt_string_handler = client->GetSignalHandler<expected::ExpectedString>(spec);
+	auto opt_string_pair_handler = client->GetSignalHandler<ExpectedStringPair>(spec);
+
+	// either no match or only one match
+	assert(
+		!(static_cast<bool>(opt_string_handler) || static_cast<bool>(opt_string_pair_handler))
```
oleorhagen, line 504: Q: Why cast to `bool`, when it's an `optional` already?

oleorhagen: So, multiple matches is a programming error?

common/dbus/platform/asio_libdbus/dbus_client.cpp
```diff
@@ -447,6 +510,16 @@ DBusHandlerResult MsgFilter(DBusConnection *connection, DBusMessage *message, vo
 		auto ex_value = ExtractValueFromDBusMessage<expected::ExpectedString>(message);
 		(*opt_string_handler)(ex_value);
 		return DBUS_HANDLER_RESULT_HANDLED;
+	} else if (opt_string_pair_handler) {
```
oleorhagen, line 517: I was nearly expecting a templated function here also :laughing:

oleorhagen (approved):
Turned our real nice imo! :tada:

common/dbus.hpp
vpodzime: It's also a 3rd party library header so not entirely sure which rule applies here.

common/dbus/platform/asio_libdbus/dbus_client.cpp
vpodzime: Because otherwise it fails to compile with `optional` not having `operator||` and `operator^`.

common/dbus/platform/asio_libdbus/dbus_client.cpp
vpodzime: Yes, but I like it.

common/dbus/platform/asio_libdbus/dbus_client.cpp
vpodzime: If you find a way to do it that way, sure! I wasn't able to. But there's definitely nothing left to lose here. :sweat_smile:

common/dbus.hpp
oleorhagen: Good point!

oleorhagen: Makes sense

common/dbus/platform/asio_libdbus/dbus_client.cpp
oleorhagen: `(opt_string_handler || opt_string_pair_handler)` compiles just fine

vpodzime: For `^` I was getting a compilation error.

common/dbus/platform/asio_libdbus/dbus_client.cpp
```diff
@@ -344,54 +454,79 @@ static void HandleReply(DBusPendingCall *pending, void *data) {
...
+
+	// either no match or only one match
+	assert(
+		!(static_cast<bool>(opt_string_handler) || static_cast<bool>(opt_string_pair_handler))
+		|| (static_cast<bool>(opt_string_handler) ^ static_cast<bool>(opt_string_pair_handler)));
+
+	if (opt_string_handler) {
+		if (!CheckDBusMessageSignature<expected::ExpectedString>(signature)) {
+			auto err = MakeError(ValueError, "Unexpected reply signature: " + signature);
+			(*opt_string_handler)(expected::unexpected(err));
```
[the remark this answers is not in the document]
vpodzime: I still consider myself a C developer :grin:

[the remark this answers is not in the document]
oleorhagen: me 2 :rofl:

PR #1389, opened 2023-09-11
Cancel() triggers "operation canceled"

kacf:
This should "fix" most `Cancel()` calls to return `operation_canceled` in their handlers, unless it is during destruction.

One notable omission is the signal handlers. Boost does produce an error even in this case, but because a signal handler is installed permanently, and does not wait for any specific event, it seems more acceptable to not call them with `operation_canceled`. And the handlers are not taking errors in their arguments currently, so it would need to change the call sites.

### DIFF
--- a/common/events.hpp
+++ b/common/events.hpp
@@ -32,7 +32,9 @@ namespace mender {
 namespace common {
 namespace events {
 
-using EventHandler = std::function<void(mender::common::error::Error err)>;
+using namespace std;
+
+using EventHandler = function<void(mender::common::error::Error err)>;
 
 #ifdef MENDER_USE_BOOST_ASIO
 namespace asio = boost::asio;
@@ -53,7 +55,7 @@ public:
 	// cancellation condition before doing its work.
 	//
 	// Thread-safe.
-	void Post(std::function<void()> func);
+	void Post(function<void()> func);
 
 private:
 #ifdef MENDER_USE_BOOST_ASIO
@@ -76,6 +78,13 @@ class Timer : public EventLoopObject {
 public:
 	Timer(EventLoop &loop);
 	~Timer() {
+		// Note: Testing pointer value here, not bool.
+		if (!destroying_) {
+			// Can happen as a consequence of destruction after move.
+			return;
+		}
+
+		*destroying_ = true;
 		Cancel();
 	}
 
@@ -91,10 +100,17 @@ public:
 	template <typename Duration>
 	void AsyncWait(Duration duration, EventHandler handler) {
 		timer_.expires_after(duration);
-		timer_.async_wait([handler](std::error_code ec) {
+		auto &destroying = destroying_;
+		timer_.async_wait([destroying, handler](error_code ec) {
+			if (*destroying) {
+				return;
+			}
+
 			if (ec) {
 				auto err = ec.default_error_condition();
-				if (err != make_error_condition(boost::system::errc::operation_canceled)) {
+				if (err == make_error_condition(boost::system::errc::operation_canceled)) {
+					handler(error::Error(make_error_condition(errc::operation_canceled), ""));
+				} else {
 					handler(error::Error(err, "Timer error"));
 				}
 			} else {
@@ -109,12 +125,13 @@ public:
 private:
 #ifdef MENDER_USE_BOOST_ASIO
 	asio::steady_timer timer_;
+	shared_ptr<bool> destroying_;
 #endif // MENDER_USE_BOOST_ASIO
 };
 
 using SignalNumber = int;
-using SignalSet = std::vector<SignalNumber>;
-using SignalHandlerFn = std::function<void(SignalNumber)>;
+using SignalSet = vector<SignalNumber>;
+using SignalHandlerFn = function<void(SignalNumber)>;
 
 class SignalHandler : public EventLoopObject, virtual public mio::Canceller {
 public:
--- a/common/http.hpp
+++ b/common/http.hpp
@@ -203,6 +203,7 @@ using OutgoingResponsePtr = shared_ptr<OutgoingResponse>;
 using ExpectedOutgoingResponsePtr = expected::expected<OutgoingResponsePtr, error::Error>;
 
 using RequestHandler = function<void(ExpectedIncomingRequestPtr)>;
+using IdentifiedRequestHandler = function<void(IncomingRequestPtr, error::Error)>;
 using ResponseHandler = function<void(ExpectedIncomingResponsePtr)>;
 
 using ReplyFinishedHandler = function<void(error::Error)>;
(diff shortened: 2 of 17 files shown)


common/http.hpp
```diff
@@ -343,16 +343,20 @@ struct ClientConfig {
...
+	HeaderHandlerCalled,
 	ReaderCreated,
-	InProgress,
+	BodyReadingInProgress,
 	ReachedEnd,
 	Done,
 };
+static inline bool AtLeast(TransactionStatus status, TransactionStatus expected_status) {
+	return static_cast<int>(status) >= static_cast<int>(expected_status);
+}
```
lluiscampos, line 356: Nice :star_struck:

common/http/platform/beast/http.cpp
```diff
@@ -607,28 +610,29 @@ void Client::ReadHeaderHandler(const error_code &ec, size_t num_read) {
...
 
 	auto cancelled = cancelled_;
+	status_ = TransactionStatus::HeaderHandlerCalled;
 	CallHandler(header_handler_);
 	if (*cancelled) {
 		return;
 	}
 
-	if (body_status_ == BodyReadingStatus::None) {
+	if (status_ == TransactionStatus::HeaderHandlerCalled) {
```
lluiscampos, line 619: At this point we have called the user header handler, where we expect that (the user) calls `MakeBodyAsyncReader` where `status_` is set to `ReaderCreated`. So which is the path when there is no body (when `response_body_length_` is 0)? Wouldn't then we always hit this `if` condition?

common/http/platform/beast/http.cpp
```diff
@@ -706,6 +710,20 @@ void Client::ReadBodyHandler(error_code ec, size_t num_read) {
...
 
 void Client::Cancel() {
+	if (client_active_) {
+		auto err =
+			error::Error(make_error_condition(errc::operation_canceled), "HTTP request cancelled");
+		if (status_ == TransactionStatus::None) {
+			CallErrorHandler(err, request_, header_handler_);
+		} else if (status_ != TransactionStatus::Done) {
+			CallErrorHandler(err, request_, body_handler_);
+		}
```
lluiscampos, line 720: Are we sure that `Cancel` can't be called in other status? What about between body parts read? Then the status would be `TransactionStatus::BodyReadingInProgress` and it is not caught here. Is that handled from somewhere else?

common/http_test.cpp
```diff
@@ -956,17 +956,12 @@ TEST(HttpTest, TestClientCancelInHeaderHandler) {
...
+			EXPECT_EQ(exp_resp.error().code, make_error_condition(errc::operation_canceled));
 		});
 
 	events::Timer timer(loop);
-	timer.AsyncWait(chrono::milliseconds(500), [&loop](error::Error err) {
-		// Should get here, without reaching the body handler first.
-
-		loop.Stop();
-	});
+	timer.AsyncWait(chrono::milliseconds(500), [&loop](error::Error err) { loop.Stop(); });
```
lluiscampos, line 964: I love when tests really help understating APIs :)

lluiscampos (commented):
Overall it looks very consistent. Nice work.

common/http/platform/beast/http.cpp
kacf: `response_body_length_` cannot be zero, because we check `response_body_read_ >= response_body_length_` some lines above this, and since the former is initialized to zero, we will return from there in that case.

common/http/platform/beast/http.cpp
kacf: Hmm, maybe you misread it, it's all statuses *except* `Done`, so they should all be caught here.

common/http/platform/beast/http.cpp
kacf: I can add a comment about it.

lluiscampos: Now I see it. Thank you.

common/http/platform/beast/http.cpp
lluiscampos: I misread :see_no_evil:

lluiscampos (approved):
Looks good from my side :+1:

mender-update/update_module/v3/platform/c++17/fs_operations.cpp
```diff
@@ -370,10 +353,12 @@ error::Error UpdateModule::DeleteStreamsFiles() {
 
 AsyncFifoOpener::AsyncFifoOpener(events::EventLoop &loop) :
 	event_loop_ {loop},
-	cancelled_ {make_shared<atomic<bool>>(true)} {
+	cancelled_ {make_shared<bool>(true)},
```
vpodzime, line 356: Removing `atomic<>` here. I guess it was never actually needed?

vpodzime (approved):
Looks good to me otherwise.

common/http/platform/beast/http.cpp
```diff
@@ -794,6 +794,22 @@ void Stream::CallErrorHandler(
 	server_.RemoveStream(shared_from_this());
 }
 
+void Stream::CallErrorHandler(
+	const error_code &ec, const IncomingRequestPtr &req, IdentifiedRequestHandler handler) {
+	CallErrorHandler(error::Error(ec.default_error_condition(), ""), req, handler);
```
oleorhagen, line 834: Q: Why don't we extract the `message`?

common/http/platform/beast/http.cpp
```diff
@@ -794,6 +794,22 @@ void Stream::CallErrorHandler(
...
+}
+
+void Stream::CallErrorHandler(
+	const error::Error &err, const IncomingRequestPtr &req, IdentifiedRequestHandler handler) {
+	stream_active_.reset();
+	handler(
+		req,
+		err.WithContext(
+			req->address_.host + ": " + MethodToString(req->method_) + " " + request_->GetPath()));
```
oleorhagen, line 844: Q: Are we _always_ guaranteed the `address` and `method` is set unconditionally, when called?

mender-update/update_module/v3/platform/c++17/fs_operations.cpp
kacf: Exactly. The intention was to use it in the the thread, but it was never actually used there.

common/events.hpp
```diff
@@ -76,6 +78,13 @@ class Timer : public EventLoopObject {
 public:
 	Timer(EventLoop &loop);
 	~Timer() {
+		// Note: Testing pointer value here, not bool.
+		if (!destroying_) {
```
oleorhagen, line 82: Not fully following, as opposed to comparing to `nullptr`?

common/events.hpp
```diff
@@ -109,12 +125,13 @@ class Timer : public EventLoopObject {
 private:
 #ifdef MENDER_USE_BOOST_ASIO
 	asio::steady_timer timer_;
+	shared_ptr<bool> destroying_;
```
oleorhagen, line 128: Q: Why is a `shared_ptr` needed? :thinking:

oleorhagen (commented):
Can't say I'm able to get into the weeds here, but on the surface looks good

common/http/platform/beast/http.cpp
kacf: We do, it's in the `ec.default_error_condition()`, which, when later using `String()`, calls `message()`.

common/http/platform/beast/http.cpp
kacf: No, but even invalid values are valid strings. `MethodToString(req->method_)` will just be `"Invalid"`, and `address_.host` will be empty. So as an error string I think it still does the job.

common/events.hpp
kacf: No, as opposed to `if (!*destroying_)`, IOW comparing the bool value. But using `nullptr` is much clearer, I'll switch to that!

common/events.hpp
kacf: Because the handler in `AsyncWait` needs to keep its own copy, and this needs to survive the destruction of the timer, since Boost may call it late.

PR #1391, opened 2023-09-13
Misc fixes 4 integration tests

oleorhagen:
Just  a collection of fixes from running the client with integration tests.

### DIFF
--- a/common/config_parser/config_parser.cpp
+++ b/common/config_parser/config_parser.cpp
@@ -13,17 +13,22 @@
 //    limitations under the License.
 
 #include <common/config_parser.hpp>
+
 #include <string>
 #include <vector>
 #include <algorithm>
+
 #include <common/json.hpp>
+#include <common/log.hpp>
 
 namespace mender {
 namespace common {
 namespace config_parser {
 
 using namespace std;
+
 namespace json = mender::common::json;
+namespace log = mender::common::log;
 
 const ConfigParserErrorCategoryClass ConfigParserErrorCategory;
 
@@ -68,7 +73,8 @@ ExpectedBool MenderConfigFromFile::ValidateServerConfig() const {
 
 	if (servers.size() == 0) {
 		if (server_url.size() == 0) {
-			// TODO - Log warning
+			log::Warning(
+				"No ServerURL found in the configuration. The client will not be able to download updates");
 		}
 	}
 	return true;
--- a/common/path/platform/c++17/path.cpp
+++ b/common/path/platform/c++17/path.cpp
@@ -14,10 +14,11 @@
 
 #include <common/path.hpp>
 
-#include <string>
-
 #include <filesystem>
+#include <string>
 #include <unordered_set>
+
+#include <common/error.hpp>
 
 namespace mender {
 namespace common {
@@ -44,6 +45,17 @@ bool IsAbsolute(const string &path) {
 
 bool FileExists(const string &path) {
 	return fs::exists(path);
+}
+
+error::Error FileDelete(const string &path) {
+	error_code ec;
+	bool deleted = fs::remove(fs::path(path), ec);
+	if (not deleted) {
+		return error::Error(
+			ec.default_error_condition(),
+			"Failed to remove the file: '" + path + "'. error: " + ec.message());
+	}
+	return error::NoError;
 }
 
 expected::ExpectedBool IsExecutable(const string &file_path, const bool warn) {
--- a/mender-update/cli/actions.cpp
+++ b/mender-update/cli/actions.cpp
@@ -14,9 +14,20 @@
 
 #include <mender-update/cli/actions.hpp>
 
-#include <common/events.hpp>
-#include <common/log.hpp>
+#include <algorithm>
+#include <iostream>
+#include <string>
 
+#include <artifact/config.hpp>
+
+#include <common/common.hpp>
+#include <common/error.hpp>
+#include <common/events.hpp>
+#include <common/key_value_database.hpp>
+#include <common/log.hpp>
+#include <common/path.hpp>
+
+#include <mender-update/cli/cli.hpp>
 #include <mender-update/daemon.hpp>
 #include <mender-update/standalone.hpp>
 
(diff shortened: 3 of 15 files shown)


common/config_parser/config_parser.cpp
```diff
@@ -68,7 +73,8 @@ ExpectedBool MenderConfigFromFile::ValidateServerConfig() const {
 
 	if (servers.size() == 0) {
 		if (server_url.size() == 0) {
-			// TODO - Log warning
```
kacf, line 71: Lol, who thought it was too much work to add this the first time around? 😂

mender-update/standalone.hpp
```diff
@@ -107,7 +107,8 @@ error::Error SaveStateData(database::KeyValueDatabase &db, const StateData &data
 
 error::Error RemoveStateData(database::KeyValueDatabase &db);
 
-ResultAndError Install(context::MenderContext &main_context, const string &src);
+ResultAndError Install(
+	context::MenderContext &main_context, const string &src, bool verify_signature = true);
```
kacf, line 111: Why not simple reuse the `enum class Signature` here?

common/path/platform/c++17/path.cpp
```diff
@@ -46,6 +46,16 @@ bool FileExists(const string &path) {
 	return fs::exists(path);
 }
 
+bool FileDelete(const string &path) {
+	error_code ec;
+	bool deleted = fs::remove(fs::path(path), ec);
+	if (not deleted) {
+		log::Error("Failed to remove the file: " + path + " error: " + ec.message());
+		return false;
```
kacf, line 54: Maybe it's better if this function returns an error. This is a quite generic function, and logging is not always what you want (it could be expected to fail, for example). The caller can log instead.

mender-update/cli/actions.cpp
```diff
@@ -28,9 +37,62 @@ namespace conf = mender::common::conf;
...
+	if (result.err != error::NoError) {
+		error::Error err =
+			db.Write(main_context.artifact_name_key, common::ByteVectorFromString("unknown"));
+		if (err != error::NoError) {
+			return err;
+		}
+		bool deleted = path::FileDelete(bootstrap_artifact_path);
+		if (!deleted) {
+			return context::MakeError(
+				context::ValueError, "Failed to delete the bootstrap Artifact");
```
kacf, line 86: `ValueError`? If you change `FileDelete` to return an error, then you can just return it as is (possibly using `WithContext`). 

So maybe:
```
return result.err.FollowedBy(err.WithContext("Failed to delete the bootstrap Artifact"));
```

mender-update/cli/actions.cpp
```diff
@@ -28,9 +37,62 @@ namespace conf = mender::common::conf;
...
+	log::Info("Installing the bootstrap Artifact");
+	const bool verify_artifact_signature = false;
+	auto result =
+		standalone::Install(main_context, bootstrap_artifact_path, verify_artifact_signature);
+
+	if (result.err != error::NoError) {
+		error::Error err =
+			db.Write(main_context.artifact_name_key, common::ByteVectorFromString("unknown"));
+		if (err != error::NoError) {
+			return err;
```
kacf, line 81: Here I would return `result.err.FollowedBy(err)`.

mender-update/cli/actions.cpp
```diff
@@ -28,9 +37,62 @@ namespace conf = mender::common::conf;
...
+		}
+		bool deleted = path::FileDelete(bootstrap_artifact_path);
+		if (!deleted) {
+			return context::MakeError(
+				context::ValueError, "Failed to delete the bootstrap Artifact");
+		}
+		return context::MakeError(context::ValueError, "Failed to install the bootstrap Artifact");
```
kacf, line 88: Maybe:
```
return result.err.WithContext("Failed to install the bootstrap Artifact");
```

mender-update/cli/actions.cpp
```diff
@@ -28,9 +37,62 @@ namespace conf = mender::common::conf;
...
+	}
+	bool deleted = path::FileDelete(bootstrap_artifact_path);
+	if (!deleted) {
+		return context::MakeError(context::ValueError, "Failed to delete the bootstrap Artifact");
```
kacf, line 92: Same here, with regards to error return.

mender-update/standalone.hpp
oleorhagen: I pondered it, but it would mean adding an `artifact` dependency to the `standalone` file. 

I'm not against it at all, actually pro, but didn't know if you'd consider it poor style.

kacf: `standalone` already "sits higher" in the hierarchy of dependencies, so I think it's ok. In the `.cpp` file I believe we are already using it.

common/path/platform/c++17/path.cpp
oleorhagen: Agreed, fixing

oleorhagen:
Pipeline passed: https://gitlab.com/Northern.tech/Mender/mender/-/pipelines/1009443380

common/config_parser/config_parser.cpp
vpodzime: Did we even have logging at that time?

mender-update/cli/actions.cpp
```diff
@@ -28,9 +39,66 @@ namespace conf = mender::common::conf;
...
+		main_context.GetConfig().paths.GetBootstrapArtifactFile()};
+	// Check if the DB is populated - then install conditionally
+	auto &db = main_context.GetMenderStoreDB();
+	auto exp_key = db.Read(main_context.artifact_name_key);
+	if (exp_key) {
+		// Key exists. Do nothing
+		return error::NoError;
+	}
+	error::Error err = exp_key.error();
+	if (err != error::NoError
```
vpodzime, line 60: `error::NoError` should never be used with `expected::unexpected()`, right? We should have an `assert()` or check there.

vpodzime (commented):
Looks good to me otherwise.

mender-update/cli/actions.cpp
```diff
@@ -14,12 +14,23 @@
...
 #include <common/events.hpp>
+#include <common/key_value_database.hpp>
 #include <common/log.hpp>
+#include <common/path.hpp>
 
+#include <mender-update/cli/cli.hpp>
 #include <mender-update/daemon.hpp>
 #include <mender-update/standalone.hpp>
 
+#include <artifact/config.hpp>
```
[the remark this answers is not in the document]
oleorhagen: I guess it could go along with the `mender-update` ones, but I thought I'd separate them. but no biggie, I'll do whatever yous think makes the most sense.

mender-update/cli/actions.cpp
```diff
@@ -28,9 +39,66 @@ namespace conf = mender::common::conf;
...
-
-namespace daemon = mender::update::daemon;
+namespace error = mender::common::error;
+namespace kv_db = mender::common::key_value_database;
+namespace path = mender::common::path;
 namespace standalone = mender::update::standalone;
+namespace daemon = mender::update::daemon;
+
+
+error::Error maybeInstallBootstrapArtifact(context::MenderContext &main_context) {
```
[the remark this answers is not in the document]
oleorhagen: All open to suggestions

mender-update/cli/actions.cpp
oleorhagen: Hmm, what do you mean?

We do this all the time? 

Remember that the case is inverted here, we check for `success` above. This is the `error` condition.

common/path/platform/c++17/path.cpp
```diff
@@ -46,6 +47,17 @@ bool FileExists(const string &path) {
 	return fs::exists(path);
 }
 
+error::Error FileDelete(const string &path) {
+	error_code ec;
+	bool deleted = fs::remove(fs::path(path), ec);
+	if (not deleted) {
```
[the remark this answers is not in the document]
oleorhagen: I personally _think_ so. But no strong feelings here

common/config_parser/config_parser.cpp
oleorhagen: Looks like it, but they were done in parallel

mender-update/cli/actions.cpp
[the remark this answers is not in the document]
vpodzime: `static error::Error DoMaybeInstallBootstrapArtifact()`?

mender-update/cli/actions.cpp
kacf: > They should never contain error::NoError, AFAICT. Hence that part of the condition is unnecessary/confusing.

And actually, the test for `KeyError` implies that `err != error::NoError`, so it's unnecessary.

mender-update/cli/actions.cpp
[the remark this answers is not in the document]
kacf: +1

kacf (approved):
👍

mender-update/cli/actions.cpp
oleorhagen: Ahh, now I see what you mean

PR #1392, opened 2023-09-14
DBus server

vpodzime:


common/CMakeLists.txt
```diff
@@ -278,7 +278,7 @@ add_dependencies(tests crypto_test)
 find_package(PkgConfig REQUIRED)
 find_package(Boost 1.35 REQUIRED)
 pkg_check_modules(dbus REQUIRED dbus-1)
-add_library(common_dbus STATIC dbus/dbus.cpp dbus/platform/asio_libdbus/dbus_client.cpp)
+add_library(common_dbus STATIC dbus/dbus.cpp dbus/platform/asio_libdbus/dbus.cpp)
```
kacf, line 281: > Also rename dbus_client.cpp to dbus.cpp because it now contains non-client code. At the same time it's not big enough for splitting it into multiple files to be worth the extra complexity.

It might be worth it, because the server and the client will not be used by the same binaries. Mender-auth will only use the server, and mender-update will only use the client, if I'm not mistaken. Of course this doesn't *require* us to split it, but might help reduce code size if we can keep them independent. Any thoughts on this?

common/dbus/platform/asio_libdbus/dbus.cpp
```diff
@@ -529,6 +529,51 @@ DBusHandlerResult MsgFilter(DBusConnection *connection, DBusMessage *message, vo
 	}
 }
 
+static inline string GetMethodSpec(
+	const string &service, const string &interface, const string &method) {
+	return service + "::" + interface + method;
```
kacf, line 538: No separator between `interface` and `method`?

common/dbus.hpp
```diff
@@ -184,6 +187,34 @@ class DBusObject {
...
+		DBusPeer(loop),
+		service_name_ {service_name} {};
+
+	~DBusServer() override;
+
+	error::Error AdvertiseObject(DBusObjectPtr obj);
+
+	// Only a convenience version for tests. Double-check that obj outlives the
+	// DBusServer!
+	error::Error AdvertiseObject(DBusObject &obj) {
```
kacf, line 204: Since it is an overload, perhaps it should have "Test" in its name to prevent accidental usage?

common/dbus_test.cpp
```diff
@@ -14,23 +14,59 @@
 
 #include <common/dbus.hpp>
 
+#include <stdlib.h> // setenv()
```
kacf, line 17: Such comments tend to go out of date, or?

common/CMakeLists.txt
```diff
@@ -283,6 +283,7 @@ target_link_libraries(common_dbus PUBLIC
   common_log
   common_error
   common_events
+  common_processes
```
kacf, line 286: Shouldn't it be the test depending on this, and not the library?

kacf (approved):
Some possibilities/suggestions, but looks good as is too!

common/dbus.hpp
vpodzime: Not a strong opinion here, but it's just a variant for use with a reference, just like we have in many of our I/O classes. I just added the comment because I got myself caught in the trap of declaring `server` before `object` in the test.

common/dbus_test.cpp
vpodzime: Generally, yes, maybe. But this one is quite special, I think. Including `stdlib.h` when there is `cstdlib` is a bit weird, but `setenv()` just doesn't exist in the latter.

common/CMakeLists.txt
vpodzime: Yup! Another good catch!

common/CMakeLists.txt
vpodzime: Lot of the code belongs to `DBusPeer`. So we would have to split things in 3 files, some functions could no longer be `static` and we would have to add some internal header file. With the client and the server being much smaller parts on top of a bigger pile shared by both. I'd suggest we keep things simple for now and we will see if code size optimization is needed and worth it here.

common/dbus.hpp
kacf: Ok, as long as there's no additional risk, it's fine.

common/dbus_test.cpp
kacf: Ah, right, I didn't even think about `cstdlib` in this instance (but I should have!). Maybe put a slightly bigger comment then, on why `cstdlib` can't be used.

common/dbus.hpp
```diff
@@ -132,6 +132,40 @@ class DBusClient : public events::EventLoopObject {
...
+using DBusMethodHandler = function<ReturnType(void)>;
+
+class DBusObject {
+public:
+	DBusObject(const string &path) :
+		path_ {path} {};
+
+	const string &GetPath() {
+		return path_;
+	};
```
oleorhagen, line 168: Nit: unneeded `;` after `}`

common/dbus.hpp
```diff
@@ -132,6 +132,40 @@ class DBusClient : public events::EventLoopObject {
 	optional::optional<DBusSignalHandler<SignalValueType>> GetSignalHandler(const SignalSpec &spec);
 };
 
+// Might need something like
```
oleorhagen, line 153: Nit: Missing `include <vector>`

common/dbus/platform/asio_libdbus/dbus.cpp
```diff
@@ -578,6 +578,178 @@ optional::optional<DBusMethodHandler<ExpectedStringPair>> DBusObject::GetMethodH
...
+	dbus_error_init(&dbus_error);
+
+	// We could also do DBUS_NAME_FLAG_ALLOW_REPLACEMENT for cases where two of
+	// processes request the same name, but it would require handling of the
+	// NameLost signal and triggering termination.
+	if (dbus_bus_request_name(
+			dbus_conn_.get(), service_name_.c_str(), DBUS_NAME_FLAG_DO_NOT_QUEUE, &dbus_error)
+		== -1) {
+		dbus_conn_.reset();
+		auto err = MakeError(
```
oleorhagen, line 597: Nit: err is already defined.

common/dbus/platform/asio_libdbus/dbus.cpp
```diff
@@ -578,6 +578,178 @@ optional::optional<DBusMethodHandler<ExpectedStringPair>> DBusObject::GetMethodH
...
 }
 
+error::Error DBusServer::InitializeConnection() {
+	auto err = DBusPeer::InitializeConnection();
+	if (err != error::NoError) {
+		return err;
+	}
+
+	DBusError dbus_error;
+	dbus_error_init(&dbus_error);
```
oleorhagen, line 588: For future safeness, I would wrap this in a smart pointer

oleorhagen (approved):
Looks pretty. 

I'd be lying if I said I understand what is going on, so just left a few superficial comments.

common/dbus/platform/asio_libdbus/dbus.cpp
vpodzime: yeah, but not it this scope and it's here only to split the code into multiple lines, really.

common/dbus/platform/asio_libdbus/dbus.cpp
vpodzime: I know you would. :wink:

oleorhagen (approved):

PR #1394, opened 2023-09-18
Patches to enable building on QNX

kacf:
Still TBD:
- [x] In release mode, Googletest includes the `optional` container for stdc++, which conflicts with our `optional-lite` library. Need to find a solution for this.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -130,6 +130,10 @@ add_compile_options(-Werror -Wall -Wsuggest-override)
 
 include(cmake/build_mode.cmake)
 
+if("${STD_FILESYSTEM_LIB_NAME}" STREQUAL "")
+  set(STD_FILESYSTEM_LIB_NAME stdc++fs)
+endif()
+
 if($CACHE{COVERAGE})
   add_custom_target(coverage_enabled COMMAND true)
 else()
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -21,6 +21,16 @@ configure_file(mender-version.h.in mender-version.h)
 
 project(mender)
 
+# The manual lists Debug, Release, RelWithDebInfo and MinSizeRel as standard build types. Let's set
+# minimum size as the default, since network and I/O are likely to be bigger bottlenecks than CPU
+# usage.
+#
+# Source:
+# https://cmake.org/cmake/help/v3.27/manual/cmake-buildsystem.7.html#default-and-custom-configurations
+if("${CMAKE_BUILD_TYPE}" STREQUAL "")
+  set(CMAKE_BUILD_TYPE "MinSizeRel")
+endif()
+
 include(cmake/asan.cmake)
 include(cmake/threadsan.cmake)
 # set(CMAKE_VERBOSE_MAKEFILE ON)
@@ -116,14 +126,9 @@ else()
 endif()
 
 # Default for all components.
-add_compile_options(-std=c++11 -Werror -Wall -Wsuggest-override)
+add_compile_options(-Werror -Wall -Wsuggest-override)
 
-if (CMAKE_BUILD_TYPE STREQUAL "NDebug")
-add_compile_definitions(NDEBUG=1)
-endif()
-
-# Use this with target_compile_options for platform specific components that need it.
-set(PLATFORM_SPECIFIC_COMPILE_OPTIONS -std=c++17)
+include(cmake/build_mode.cmake)
 
 if($CACHE{COVERAGE})
   add_custom_target(coverage_enabled COMMAND true)
@@ -192,3 +197,5 @@ add_subdirectory(artifact)
 add_subdirectory(Documentation EXCLUDE_FROM_ALL)
 
 configure_file(config.h.in config.h)
+
+message(STATUS "Build type: ${CMAKE_BUILD_TYPE}")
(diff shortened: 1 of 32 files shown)


README.md
```diff
@@ -229,6 +229,15 @@ During the current, early stage of development using a higher, cross-compilation
 system such as Yocto is advisable. Once things are sufficiently stabilized, a set of steps for
 manual cross-compilation will be added here.
 
+## QNX
+
+```
+mkdir build && cd build
```
oleorhagen, line 237: S:
```
cmake -S . -B build && QNX_TARGET=<compiler-selection> cmake --build ./build -D CMAKE_TOOLCHAIN_FILE=cmake/qnx.cmake
```

README.md
```diff
@@ -229,7 +229,9 @@ During the current, early stage of development using a higher, cross-compilation
 system such as Yocto is advisable. Once things are sufficiently stabilized, a set of steps for
 manual cross-compilation will be added here.
 
-## QNX
+### QNX
```
oleorhagen, line 232: S: `squash`

common/CMakeLists.txt
```diff
@@ -30,7 +30,11 @@ target_link_libraries(common_path
 check_cxx_source_compiles("#include <filesystem>
   int main() { std::filesystem::path p(\"/\"); }" STD_FILESYSTEM_AVAILABLE)
 if(NOT "${STD_FILESYSTEM_AVAILABLE}" STREQUAL 1)
-  target_link_libraries(common_path PUBLIC stdc++fs)
+  if ("${CMAKE_SYSTEM_NAME}" STREQUAL "QNX")
```
oleorhagen, line 33: Q: Why can't this be in the `qnx.cmake` file?

CMakeLists.txt
```diff
@@ -21,6 +21,16 @@ configure_file(mender-version.h.in mender-version.h)
...
 project(mender)
 
+# The CMake documentation is quite ambiguous on the subject of available build types, and I haven't
+# been able to find an authoritative list of the default available ones. However, the manual lists
+# Debug, Release, RelWithDebInfo and MinSizeRel as "typical" types, so let's use those.
+#
+# Source:
+# https://cmake.org/cmake/help/v3.27/variable/CMAKE_BUILD_TYPE.html#variable:CMAKE_BUILD_TYPE
+if("${CMAKE_BUILD_TYPE}" STREQUAL "")
+  set(CMAKE_BUILD_TYPE "Release")
```
oleorhagen, line 31: I was actually thinking we would set the default to `MinSizeRel` ?

CMakeLists.txt
```diff
@@ -116,15 +126,24 @@ else()
...
 
 # Default for all components.
-add_compile_options(-std=c++11 -Werror -Wall -Wsuggest-override)
-
-if (CMAKE_BUILD_TYPE STREQUAL "NDebug")
-add_compile_definitions(NDEBUG=1)
+add_compile_options(-Werror -Wall -Wsuggest-override)
+
+# See comment about CMAKE_BUILD_TYPE above.
+if ("${CMAKE_BUILD_TYPE}" MATCHES "Rel")
```
oleorhagen, line 132: S: Split this out into a separate `cmake` file

CMakeLists.txt
```diff
@@ -21,6 +21,16 @@ configure_file(mender-version.h.in mender-version.h)
 
 project(mender)
 
+# The CMake documentation is quite ambiguous on the subject of available build types, and I haven't
```
oleorhagen, line 24: https://cmake.org/cmake/help/latest/manual/cmake-buildsystem.7.html#id38

lluiscampos (approved):

README.md
kacf: On second thought I think I will leave it as is. We do the `cmake ..` variant on Linux, and if we don't do `cd build`, then `make` doesn't work.

oleorhagen: Sure fine, just a suggestion. `cmake --build build` has one benefit over `make` however, and that is that it runs your build system, whether it is `make` or not. So if your default `buildsystem` is not make, this will still work

CMakeLists.txt
kacf: Yeah, maybe that's a better default. I suppose more people will care about size than about speed. I/O and network are nearly always the bottlenecks anyway.

CMakeLists.txt
kacf: How did I not find that the first time around?? I was looking all over the place... Thanks! :open_mouth:

common/CMakeLists.txt
kacf: It can. Check out the last commit.

common/optional.hpp
```diff
@@ -15,6 +15,18 @@
 #ifndef MENDER_COMMON_OPTIONAL_HPP
 #define MENDER_COMMON_OPTIONAL_HPP
 
+#include <version>
```
vpodzime, line 18: We don't use anything from `<version>`, do we? I only see our custom macro.

oleorhagen (approved):

oleorhagen:
Looks good :smile_cat:

vpodzime (approved):
Looks good to me otherwise.

common/optional.hpp
kacf: You're right, it was a leftover. Removed.

PR #1398, opened 2023-09-20
MEN-6273: Extend Update Module API to include payload sizes

lluiscampos:


lluiscampos:
I can squash the three commits into one once reviewed. I developed in these three steps and I think it will help reviewing :+1:

mender-update/cli/cli_test.cpp
```diff
@@ -422,6 +424,74 @@ artifact_name=test
...
+
+case "$1" in
+    ProvidePayloadFileSizes)
+        echo "Yes"
+        ;;
+    Download)
+        exit 1
+        ;;
+    DownloadWithFileSizes)
+		exit 0
```
kacf, line 451: Indentation.

mender-update/update_module/v3/update_module.cpp
```diff
@@ -64,6 +65,34 @@ UpdateModule::DownloadData::DownloadData(
...
+	}
+	auto &processStdOut = exp_output.value();
+	if (processStdOut == "Yes") {
+		return true;
+	} else if (processStdOut == "No" || processStdOut == "") {
+		return false;
+	}
+	return expected::unexpected(error::Error(
+		make_error_condition(errc::protocol_error),
+		"Unexpected output from the process for ProvidePayloadFileSizes state"));
```
kacf, line 81: Include what it was maybe?

mender-update/daemon/state_test.cpp
```diff
@@ -128,6 +142,129 @@ vector<StateTransitionsTestCase> idle_and_sync_test_cases {
...
+			{
+				"downloading",
+				"installing",
+				"rebooting",
+				"installing",
+				"success",
+			},
+		.install_outcome = InstallOutcome::SuccessfulInstall,
+		.provide_file_sizes = ProvideFileSizes::Yes,
+		.generate_idle_sync_scripts = true,
```
kacf, line 183: This is not needed here, is it?

mender-update/daemon/states.cpp
```diff
@@ -366,16 +385,29 @@ void UpdateDownloadState::ParseArtifact(Context &ctx, sm::EventPoster<StateEvent
...
+	} else {
+		ctx.deployment.update_module->AsyncDownload(
+			ctx.event_loop, *ctx.deployment.artifact_payload, [&poster](error::Error err) {
+				if (err != error::NoError) {
+					log::Error(err.String());
+					poster.PostEvent(StateEvent::Failure);
+					return;
+				}
+
+				poster.PostEvent(StateEvent::Success);
```
kacf, line 408: Suggestion: Put this handler in a variable first, then you don't have to specify it twice.

kacf (approved):
Beautiful!

I think you should keep the commits as they are, since they are already self contained, despite being incremental.

vpodzime (approved):
Just a nitpick, don't see anything else obviously wrong. But I have to say I have basically no clue how the update modules work with these separate loops (why don't we do recursive loops?) etc. But this code just seems to follow the same patterns, so...

mender-update/daemon/state_test.cpp
```diff
@@ -234,7 +234,7 @@ vector<StateTransitionsTestCase> idle_and_sync_test_cases {
 				"ArtifactCommit_Enter_00",
 				"ArtifactCommit",
 				"ArtifactCommit_Leave_00",
-				"Cleanup",
+				"Cleanup", // Leave me, or clang-format is not nice
```
oleorhagen, line 377: :laughing:

oleorhagen (approved):
Nothing further to add :)

Well done!

mender-update/daemon/states.cpp
lluiscampos: Thank you!

PR #1402, opened 2023-09-21
MEN-6658: Run the authentication loop once upon bootstrap

oleorhagen:


mender-auth/cli/actions.cpp
```diff
@@ -80,6 +82,39 @@ error::Error DoBootstrap(shared_ptr<MenderKeyStore> keystore, const bool force)
 			return err;
 		}
 	}
+	return err;
+}
+
+error::Error DoAuthenticate(context::MenderContext &main_context) {
+	events::EventLoop loop;
+	auto &config = main_context.GetConfig();
+	if (config.server_url == "") {
```
lluiscampos, line 91: Related to [this](https://github.com/mendersoftware/mender/pull/1408/files#r1341056123), shouldn't you also check here for entries in the servers array?

mender-auth/cli/actions.cpp
```diff
@@ -80,6 +82,39 @@ error::Error DoBootstrap(shared_ptr<MenderKeyStore> keystore, const bool force)
...
+			} else {
+				log::Error(resp.error().String());
+			}
+			loop.Stop();
+		},
+		config.tenant_token);
+	if (err != error::NoError) {
+		return err;
+	}
+	loop.Run();
```
lluiscampos, line 117: Shouldn't we have a timer in parallel for a `loop.Stop` after a sensible timeout? Could it happen that this hangs for very long? :thinking:

mender-auth/cli/actions.cpp
```diff
@@ -80,6 +82,39 @@ error::Error DoBootstrap(shared_ptr<MenderKeyStore> keystore, const bool force)
...
+		config.server_url,
+		config.paths.GetKeyFile(),
+		config.paths.GetInventoryScriptsDir(),
+		[&loop, &config](auth_client::APIResponse resp) {
+			log::Info("Got Auth response");
+			if (resp) {
+				log::Info(
+					"Successfully authorized with the server: " + config.server_url + resp.value());
+			} else {
+				log::Error(resp.error().String());
```
lluiscampos, line 109: This will still return `error::NoError`. We should take an `error` by reference in the lambda and set it on actual errors. Or is this by design?

mender-auth/cli/actions.cpp
oleorhagen: Depends who merges first? :laughing: 

Alright, but probably safest to change it.. or not? :thinking: 

Honestly not sure. You decide

mender-auth/cli/actions.cpp
oleorhagen: As far as I understand we are only supposed to _try_ and authenticate with the server.

Acceptance criteria:
* `Report result on command line.`

Whether it does or not is not so important, as long as we log it.

mender-auth/cli/actions.cpp
oleorhagen: It is a very valid point imo! 

Tbh, I'm a little scared of this for our main loop.

In this instance, we could probably have a timeout on the loop, like we do for our test-loop, just to be sure.

In any case, this is supposed to be ran from the CLI, so if it hangs, you'd expect the user to stop it.

Alas, I agree. What is a sensible timeout you think? `5 sec` ? :thinking:

lluiscampos: As you said the user is most likely going to cancel, so for this kind of timeout I would say 30 seconds.

Question: do we print a info message like "authenticating to https:/..."? This would help for the user to cancel if it takes long.

mender-auth/cli/actions.cpp
lluiscampos: Okey.

mender-auth/cli/actions.cpp
lluiscampos: Leave it as is.

(and merge fast!)

lluiscampos (approved):

PR #1403, opened 2023-09-21
mender-auth DBus

vpodzime:
TODO:
- [ ] adapt `mender-auth/ipc/server_test.cpp`


common/dbus/platform/asio_libdbus/dbus.cpp
```diff
@@ -429,7 +436,7 @@ ExpectedStringPair ExtractValueFromDBusMessage(DBusMessage *message) {
 		dbus_error_free(&dbus_error);
 		return expected::unexpected(err);
 	}
-	return std::pair<string, string> {string(value1), string(value1)};
+	return std::pair<string, string> {string(value1), string(value2)};
```
kacf, line 439: This commit makes it sound like a new feature, but is this really a fix which adds the pieces that were missing?

mender-auth/ipc/server_test.cpp
```diff
@@ -45,10 +54,31 @@ using TestEventLoop = mender::common::testing::TestEventLoop;
...
+		setenv("DBUS_SYSTEM_BUS_ADDRESS", dbus_sock_path.c_str(), 1);
+	};
+
+	static void TearDownTestSuite() {
+		dbus_daemon_proc_->EnsureTerminated();
+		unsetenv("DBUS_SYSTEM_BUS_ADDRESS");
+	};
 
 	void SetUp() override {
+#if defined(__has_feature)
```
kacf, line 77: Omg, I didn't know one could do this! This is much better than what I'm doing here: https://github.com/mendersoftware/mender/pull/1394/commits/cf354f7b6afeb330311e320ba4f34a1898de5544

mender-auth/ipc/server_test.cpp
```diff
@@ -138,40 +167,27 @@ TEST_F(ListenClientTests, TestListenFetchJWTToken) {
...
+	dbus::DBusClient client {loop};
+	err = client.CallMethod<expected::ExpectedBool>(
+		"io.mender.AuthenticationManager",
+		"/io/mender/AuthenticationManager",
+		"io.mender.Authentication1",
+		"FetchJwtToken",
+		[&loop, &timer](expected::ExpectedBool ex_value) {
+			ASSERT_TRUE(ex_value) << ex_value.error().message;
+			EXPECT_TRUE(ex_value.value());
+			timer.AsyncWait(chrono::seconds {1}, [&loop](error::Error err) { loop.Stop(); });
```
kacf, line 186: Why do we use a timer here? Can't we wait for the signal instead?

kacf (approved):
Nice!

mender-auth/ipc/server.cpp
```diff
@@ -18,93 +18,59 @@
 #include <iostream>
 #include <string>
 
-#include <common/common.hpp>
+#include <api/auth.hpp>
 #include <common/error.hpp>
-#include <common/events_io.hpp>
-#include <common/http.hpp>
+#include <common/expected.hpp>
+#include <common/dbus.hpp>
```
oleorhagen, line 24: Nit: sort

mender-auth/ipc/server.cpp
```diff
@@ -18,93 +18,59 @@
...
+	dbus_obj->AddMethodHandler<dbus::ExpectedStringPair>(
+		"io.mender.AuthenticationManager", "io.mender.Authentication1", "GetJwtToken", [this]() {
+			return dbus::StringPair {GetJWTToken(), GetServerURL()};
+		});
+	dbus_obj->AddMethodHandler<expected::ExpectedBool>(
+		"io.mender.AuthenticationManager",
+		"io.mender.Authentication1",
+		"FetchJwtToken",
+		[this, private_key_path, identity_script_path]() {
+			auto err = auth_client::FetchJWTToken(
```
oleorhagen, line 50: Nit: this will panic if the `server_url == ""`

mender-auth/ipc/server_test.cpp
```diff
@@ -34,9 +39,13 @@ using namespace std;
 
 namespace conf = mender::common::conf;
 namespace error = mender::common::error;
+namespace events = mender::common::events;
+namespace expected = mender::common::expected;
+namespace dbus = mender::common::dbus;
```
oleorhagen, line 44: Nit: sort

mender-auth/ipc/server.cpp
```diff
@@ -18,93 +18,59 @@
...
+				client_,
+				server_url_,
+				private_key_path,
+				identity_script_path == "" ? default_identity_script_path_ : identity_script_path,
+				[this](auth_client::APIResponse resp) {
+					CacheAPIResponse(server_url_, resp);
```
oleorhagen, line 56: Q: Should we log an error in case `!resp`?

oleorhagen (approved):
Looks really nice imo!

mender-auth/ipc/server_test.cpp
vpodzime: We can. I just replaced things in the original test that used a separate thread for stopping the loop after a second. I guess I was happy enough with the improvement I made! :joy:

common/dbus/platform/asio_libdbus/dbus.cpp
```diff
@@ -420,6 +432,23 @@ ExpectedStringPair ExtractValueFromDBusMessage(DBusMessage *message) {
...
+template <>
+expected::ExpectedBool ExtractValueFromDBusMessage(DBusMessage *message) {
+	DBusError dbus_error;
+	dbus_error_init(&dbus_error);
+	bool result;
+	if (!dbus_message_get_args(
+			message, &dbus_error, DBUS_TYPE_BOOLEAN, &result, DBUS_TYPE_INVALID)) {
+		auto err = MakeError(
+			ValueError,
+			string("Failed to extract reply data from reply message: ") + dbus_error.message + " ["
```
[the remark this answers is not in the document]
vpodzime: Something needs to be casted so that the string's `+` operator kicks in. To me it makes most sense to do it at the first element of the expression and just append things to it with `+`.

mender-auth/ipc/server.cpp
vpodzime: Good catch! But this is not the right place to handle such case. In fact, we probably shouldn't allow setting `server_url_` to `""`, right?

mender-auth/ipc/server.cpp
vpodzime: Not a bad idea :grin:

common/dbus/platform/asio_libdbus/dbus.cpp
vpodzime: I'm splitting it into two commits -- a `fix` one and a `feat` one.

mender-auth/ipc/server.cpp
```diff
@@ -18,93 +18,59 @@
...
+	dbus_obj->AddMethodHandler<expected::ExpectedBool>(
+		"io.mender.AuthenticationManager",
+		"io.mender.Authentication1",
+		"FetchJwtToken",
+		[this, private_key_path, identity_script_path]() {
+			auto err = auth_client::FetchJWTToken(
+				client_,
+				server_url_,
+				private_key_path,
+				identity_script_path == "" ? default_identity_script_path_ : identity_script_path,
```
oleorhagen, line 54: One thing I just stumbled upon, which was pre-existing, but still:

We don't pass in the `tenant_token`, which we will need to do.

kacf:
Something is wrong with the rebase I think. You've gotten an extra commit from Ole which is already in master (revealed when I tried to do `git range-diff`).

mender-auth/ipc/server.cpp
vpodzime: As discussed on Slack -- running `mender-auth` with `server_url == ""` just doesn't make sense. So I'll just add a check and early error return here and leave the bigger picture to a separate PR.

mender-auth/ipc/server.cpp
vpodzime: Let's resolve the issue in a follow-up PR.

mender-auth/ipc/server.cpp
```diff
@@ -37,6 +37,9 @@ using namespace std;
 
 // Register DBus object handling auth methods and signals
 error::Error Caching::Listen(const string &private_key_path, const string &identity_script_path) {
+	// Cannot serve new tokens when not knowing where to fetch them from.
+	AssertOrReturnError(server_url_ != "");
```
oleorhagen, line 41: :100:

oleorhagen (approved):

mender-auth/ipc/server.cpp
vpodzime: Actually, we **do pass** the `tenant_token_` in,. it's just the last argument even after the handler (because it has a default value).

mender-auth/ipc/server.cpp
vpodzime: https://github.com/mendersoftware/mender/pull/1408

mender-auth/ipc/server.cpp
oleorhagen: Oh, I see it now! :eye:

PR #1404, opened 2023-09-26
feat: Implement generic TCP forwarding upon 101 Switching Protocols.

kacf:
Ticket: MEN-6665

Plus many support commits.

### DIFF
--- a/common/http/platform/beast/http.cpp
+++ b/common/http/platform/beast/http.cpp
@@ -695,6 +695,10 @@ void Client::ReadHeaderHandler(const error_code &ec, size_t num_read) {
 	response_->status_code_ = http_response_parser_->get().result_int();
 	response_->status_message_ = string {http_response_parser_->get().reason()};
 
+	logger_.Debug(
+		"Received response: " + to_string(response_->status_code_) + " "
+		+ response_->status_message_);
+
 	string debug_str;
 	for (auto header = http_response_parser_->get().cbegin();
 		 header != http_response_parser_->get().cend();
--- a/common/http/platform/beast/http.cpp
+++ b/common/http/platform/beast/http.cpp
@@ -125,11 +125,6 @@ Client::Client(
 	cancelled_ {make_shared<bool>(true)},
 	resolver_(GetAsioIoContext(event_loop)),
 	body_buffer_(HTTP_BEAST_BUFFER_SIZE) {
-	// This is equivalent to:
-	//   response_buffer_.reserve(body_buffer_.size());
-	// but compatible with Boost 1.67.
-	response_buffer_.prepare(body_buffer_.size() - response_buffer_.size());
-
 	ssl_ctx_.set_verify_mode(ssl::verify_peer);
 
 	if (client.client_cert_path != "" and client.client_cert_key_path != "") {
@@ -274,6 +269,16 @@ void Client::ResolveHandler(
 	resolver_results_ = results;
 
 	stream_ = make_shared<ssl::stream<tcp::socket>>(GetAsioIoContext(event_loop_), ssl_ctx_);
+
+	if (!response_buffer_) {
+		// We can reuse this if preexisting.
+		response_buffer_ = make_shared<beast::flat_buffer>();
+
+		// This is equivalent to:
+		//   response_buffer_.reserve(body_buffer_.size());
+		// but compatible with Boost 1.67.
+		response_buffer_->prepare(body_buffer_.size() - response_buffer_->size());
+	}
 
 	http_response_parser_ = make_shared<http::response_parser<http::buffer_body>>();
 
@@ -522,7 +527,7 @@ void Client::ReadHeader() {
 	if (is_https_) {
 		http::async_read_some(
 			*stream_,
-			response_buffer_,
+			*response_buffer_,
 			*http_response_parser_,
 			[this, cancelled](const error_code &ec, size_t num_read) {
 				if (!*cancelled) {
@@ -532,7 +537,7 @@ void Client::ReadHeader() {
 	} else {
 		http::async_read_some(
 			stream_->next_layer(),
-			response_buffer_,
+			*response_buffer_,
 			*http_response_parser_,
 			[this, cancelled](const error_code &ec, size_t num_read) {
 				if (!*cancelled) {
@@ -658,7 +663,7 @@ void Client::AsyncReadNextBodyPart(
 	if (is_https_) {
 		http::async_read_some(
 			*stream_,
-			response_buffer_,
+			*response_buffer_,
 			*http_response_parser_,
 			[this, cancelled](const error_code &ec, size_t num_read) {
 				if (!*cancelled) {
@@ -668,7 +673,7 @@ void Client::AsyncReadNextBodyPart(
 	} else {
 		http::async_read_some(
 			stream_->next_layer(),
-			response_buffer_,
+			*response_buffer_,
 			*http_response_parser_,
 			[this, cancelled](const error_code &ec, size_t num_read) {
 				if (!*cancelled) {
--- a/common/http/platform/beast/http.cpp
+++ b/common/http/platform/beast/http.cpp
@@ -118,6 +118,114 @@ private:
 	friend class Server;
 };
 
+template <typename StreamType>
+class RawSocket : virtual public io::AsyncReadWriter {
+public:
+	RawSocket(shared_ptr<StreamType> stream, shared_ptr<beast::flat_buffer> buffered) :
+		destroying_ {make_shared<bool>(false)},
+		stream_ {stream},
+		buffered_ {buffered} {
+		// If there are no buffered bytes, then we don't need it.
+		if (buffered_ && buffered_->size() == 0) {
+			buffered_.reset();
+		}
+	}
+
+	~RawSocket() {
+		*destroying_ = true;
+		Cancel();
+	}
+
+	error::Error AsyncRead(
+		vector<uint8_t>::iterator start,
+		vector<uint8_t>::iterator end,
+		io::AsyncIoHandler handler) override {
+		// If we have prebuffered bytes, which can happen if the HTTP parser read the
+		// header and parts of the body in one block, return those first.
+		if (buffered_) {
+			return DrainPrebufferedData(start, end, handler);
+		}
+
+		read_buffer_ = asio::buffer(&*start, end - start);
+		auto &destroying = destroying_;
+		stream_->async_read_some(
+			read_buffer_,
+			[destroying, handler](const boost::system::error_code &ec, size_t num_read) {
+				if (*destroying) {
+					return;
+				}
+
+				if (ec == asio::error::operation_aborted) {
+					handler(expected::unexpected(error::Error(
+						make_error_condition(errc::operation_canceled),
+						"Could not read from socket")));
+				} else if (ec) {
+					handler(expected::unexpected(
+						error::Error(ec.default_error_condition(), "Could not read from socket")));
+				} else {
+					handler(num_read);
+				}
+			});
+		return error::NoError;
+	}
+
+	error::Error AsyncWrite(
+		vector<uint8_t>::const_iterator start,
+		vector<uint8_t>::const_iterator end,
+		io::AsyncIoHandler handler) override {
+		write_buffer_ = asio::buffer(&*start, end - start);
+		auto &destroying = destroying_;
+		stream_->async_write_some(
+			write_buffer_,
+			[destroying, handler](const boost::system::error_code &ec, size_t num_written) {
+				if (*destroying) {
+					return;
+				}
+
+				if (ec == asio::error::operation_aborted) {
+					handler(expected::unexpected(error::Error(
+						make_error_condition(errc::operation_canceled),
+						"Could not write to socket")));
+				} else if (ec) {
+					handler(expected::unexpected(
+						error::Error(ec.default_error_condition(), "Could not write to socket")));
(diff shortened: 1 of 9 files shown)


common/io/io.cpp
```diff
@@ -92,50 +92,93 @@ Error Copy(Writer &dst, Reader &src, vector<uint8_t> &buffer) {
 	}
 }
 
-void AsyncCopy(Writer &dst, AsyncReader &src, function<void(Error)> finished_handler) {
+struct CopyData {
+	CopyData(size_t limit) :
+		buf(MENDER_BUFSIZE),
+		limit(limit) {
```
vpodzime, line 98: I'd use `{}` here since it's used for `copied` below, but YMMV. And especially for the vectors I still find it quite confusing, so...

common/http/platform/beast/http.cpp
```diff
@@ -118,6 +118,107 @@ class BodyAsyncReader : virtual public io::AsyncReader {
...
+	}
+
+	error::Error AsyncRead(
+		vector<uint8_t>::iterator start,
+		vector<uint8_t>::iterator end,
+		io::AsyncIoHandler handler) override {
+		// If we have prebuffered bytes, which can happen if the HTTP parser read the
+		// header and parts of the body in one block, return those first.
+		if (buffered_) {
+			size_t to_copy = min(static_cast<size_t>(end - start), buffered_->size());
```
oleorhagen, line 146: Could split it out for readability, and call it something like `emptyBuffer` or similar?

vpodzime (approved):
Don't see anything wrong, but I have to admit I've completely lost track/knowledge of what's going on in the HTTP code. And it's a really scary one, I can say that.

oleorhagen (commented):
Looks good I think.

One suggestion going forwards, is to keep these PRs a little smaller, and maybe do the bug fixes incrementally. 

The PR is pretty involved as is, and hence, takes quite a lot of effort and focus to go through, trying to understand what is going on.

This would have been easier to digest, if it was a little smaller.

common/io/io.cpp
```diff
@@ -127,6 +127,135 @@ void AsyncCopy(WriterPtr dst, AsyncReaderPtr src, function<void(Error)> finished
...
+
+		AsyncWriterPtr writer;
+		ReaderPtr reader;
+		function<void(Error)> finished_handler;
+		shared_ptr<vector<uint8_t>> buf;
+		size_t expected_written;
+	};
+
+	Functor initial {dst, src, finished_handler, buf, 0};
+	initial(0);
```
lluiscampos, line 232: Really nice :+1:

common/io/io.cpp
```diff
@@ -127,6 +127,135 @@ void AsyncCopy(WriterPtr dst, AsyncReaderPtr src, function<void(Error)> finished
...
+		finished_handler(error::NoError);
+		return;
+	}
+
+	auto err = writer->AsyncWrite(
+		buf->begin(),
+		buf->begin() + *exp_size,
+		AsyncCopyWriterFunctor {writer, reader, finished_handler, buf, *exp_size});
+	if (err != error::NoError) {
+		finished_handler(exp_size.error());
```
lluiscampos, line 279: This is wrong, it should be `finished_handler(err);`

common/io/io.cpp
```diff
@@ -127,6 +127,135 @@ void AsyncCopy(WriterPtr dst, AsyncReaderPtr src, function<void(Error)> finished
...
+	}
+	if (exp_size.value() == 0) {
+		finished_handler(error::NoError);
+		return;
+	}
+
+	auto err = writer->AsyncWrite(
+		buf->begin(),
+		buf->begin() + *exp_size,
+		AsyncCopyWriterFunctor {writer, reader, finished_handler, buf, *exp_size});
```
lluiscampos: I would use `exp_size.value()` here to be consistent with few lines above.

I've been myself being inconsistent, but from now I will try to stick to `.value()` for `expected` and de-reference `*...` for shared_ptr (instead of `.get()`).

lluiscampos (commented):
I am half way through (I have a meeting now) but this is what I found so far.

The actual switch to TCP forwarding looks good, I couldn't spot anything there :eyes:

kacf:
> One suggestion going forwards, is to keep these PRs a little smaller, and maybe do the bug fixes incrementally.
> 
> The PR is pretty involved as is, and hence, takes quite a lot of effort and focus to go through, trying to understand what is going on.
> 
> This would have been easier to digest, if it was a little smaller.

This did actually occur to me as well, but part of the problem was that I didn't know if some of the earlier commits (especially the response and socket handling) actually worked, without the final "Switching Protocols" commit. But now that you pointed it out, I probably could have been better at identifying commits that were unrelated to this, and submitted them earlier. I guess I kept on thinking that "surely this is the last issue", and it kept growing 🙂. Thanks, I'll keep it mind for later and try to improve.

common/io/io.cpp
kacf: Done!

And indeed it is confusing for vectors, but doing it like this both expresses the correct intent, and does the correct thing too:

```c++
CopyData(size_t limit) :
		buf(MENDER_BUFSIZE),
		limit {limit} {
```

* `buf` constructed using a function, since it is not the value we are assigning here, but the size.
* `limit` constructed using brace-init, since we are assigning a value.

mender-auth/http_forwarder/http_forwarder_test.cpp
```diff
@@ -528,3 +528,427 @@ TEST(HttpForwarderTests, TargetTerminatesDownload) {
...
+				server_socket = exp_socket.value();
+
+				auto reader = make_shared<BodyOfXes>();
+				io::AsyncCopy(server_socket, reader, [](error::Error err) {
+					ASSERT_EQ(err, error::NoError);
+				});
+
+				auto received = make_shared<vector<uint8_t>>();
+				auto writer = make_shared<io::ByteWriter>(received);
+				writer->SetUnlimited(true);
```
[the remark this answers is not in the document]
kacf: Keeping it like this is intentional on my part. Once we get to the embedded platforms, we may have to change all of these to have a limit, so it's a big advantage that they are searchable.

mender-auth/http_forwarder/http_forwarder_test.cpp
```diff
@@ -528,3 +528,427 @@ TEST(HttpForwarderTests, TargetTerminatesDownload) {
...
+				client_socket, reader, [](error::Error err) { ASSERT_EQ(err, error::NoError); });
+
+			auto received = make_shared<vector<uint8_t>>();
+			auto writer = make_shared<io::ByteWriter>(received);
+			writer->SetUnlimited(true);
+			io::AsyncCopy(
+				writer,
+				client_socket,
+				[received, &expected, &copies, &loop](error::Error err) {
+					ASSERT_EQ(err, error::NoError);
```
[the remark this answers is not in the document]
kacf: You're right, it can use `EXPECT`.

mender-auth/http_forwarder/http_forwarder_test.cpp
```diff
@@ -528,3 +528,427 @@ TEST(HttpForwarderTests, TargetTerminatesDownload) {
...
+			ASSERT_TRUE(exp_req) << exp_req.error().String();
+			auto req = exp_req.value();
+
+			auto exp_resp = exp_req.value()->MakeResponse();
+			ASSERT_TRUE(exp_resp) << exp_resp.error().String();
+			auto resp = exp_resp.value();
+
+			resp->SetStatusCodeAndMessage(101, "Switching Protocols");
+			auto err = resp->AsyncSwitchProtocol([&copies, &loop](
+													 io::ExpectedAsyncReadWriterPtr exp_socket) {
```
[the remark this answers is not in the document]
kacf: Yeah, we have a few of these already. I have no idea which rule in clang-format is triggering this, and I haven't seen enough of them to be bothered to fix it.

mender-auth/http_forwarder/http_forwarder_test.cpp
```diff
@@ -145,6 +145,8 @@ TEST(HttpForwarderTests, BasicRequest) {
 	err = forwarder.AsyncForward("http://127.0.0.1:0", "http://127.0.0.1:" TEST_PORT "/");
 	ASSERT_EQ(err, error::NoError);
 
+	EXPECT_GE(forwarder.GetPort(), 1024);
```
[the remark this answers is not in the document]
kacf: I would, but this should have been included in an earlier commit which is already in master, so it's not related to any of the other commits in this PR.

mender-auth/http_forwarder/http_forwarder.cpp
```diff
@@ -197,7 +197,18 @@ void Server::ResponseHeaderHandler(
...
 
 	auto exp_body_reader = resp_in->MakeBodyAsyncReader();
-	if (exp_body_reader) {
+
+	if (resp_in->GetStatusCode() == http::StatusSwitchingProtocols) {
+		if (exp_body_reader) {
+			connection->logger_.Error(
+				"Response both requested to switch protocol, and has a body, which is not supported");
+			exp_body_reader.value()->Cancel();
+			connections_.erase(req_in);
```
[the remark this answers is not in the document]
kacf: Hmm, good point. I suppose we might want to reply with 501 Not Implemented or something like that.

common/http/platform/beast/http.cpp
```diff
@@ -118,6 +118,107 @@ class BodyAsyncReader : virtual public io::AsyncReader {
...
+			copy_n(static_cast<const uint8_t *>(buffered_->cdata().data()), to_copy, start);
```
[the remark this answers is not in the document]
kacf: Apparently the Boost `flat_buffer` is an ultra flexible buffer which returns a void pointer instead of byte pointer.

common/io/io.cpp
kacf: Well spotted! :eye:

common/io/io.cpp
kacf: Yeah, I've also been inconsistent. I'm also leaning towards `value()` since it works better in cases like `exp_ptr.value()->Function()` (`*exp_ptr->Function()` does not work).

common/io/io.cpp
```diff
@@ -92,50 +92,93 @@ Error Copy(Writer &dst, Reader &src, vector<uint8_t> &buffer) {
...
+	size_t copied {0};
+	size_t limit;
+};
+
+void AsyncCopy(
+	Writer &dst, AsyncReader &src, function<void(Error)> finished_handler, size_t stop_after) {
 	AsyncCopy(
 		WriterPtr(&dst, [](Writer *) {}),
 		AsyncReaderPtr(&src, [](AsyncReader *) {}),
 		finished_handler);
```
lluiscampos, line 111: Missing to pass `stop_after` to `AsyncCopy`.

common/io/io.cpp
```diff
@@ -92,50 +92,93 @@ Error Copy(Writer &dst, Reader &src, vector<uint8_t> &buffer) {
...
+	}
 }
 
-void AsyncCopy(AsyncWriter &dst, Reader &src, function<void(Error)> finished_handler) {
+void AsyncCopy(
+	AsyncWriter &dst, Reader &src, function<void(Error)> finished_handler, size_t stop_after) {
 	AsyncCopy(
 		AsyncWriterPtr(&dst, [](AsyncWriter *) {}),
 		ReaderPtr(&src, [](Reader *) {}),
 		finished_handler);
```
lluiscampos, line 176: Same here, missing to pass `stop_after`. It is also missing in the rest of versions of `AsyncCopy` below.

lluiscampos (changes requested):
Requesting changes for the two small errors that I reported: returning the wrong err (see above review) and not passing the limit to the ptr versions of `AsyncCopy` (see below).

It looks good otherwise! Well done.

common/io/io.cpp
kacf: Embarrassed! :see_no_evil: Thanks!

mender-auth/http_forwarder/http_forwarder.cpp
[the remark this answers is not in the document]
kacf: Done, it's now doing exactly that!

kacf:
@lluiscampos: Updated!

lluiscampos (approved):

common/http/platform/beast/http.cpp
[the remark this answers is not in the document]
oleorhagen: of course it is..

mender-auth/http_forwarder/http_forwarder.cpp
```diff
@@ -196,14 +196,30 @@ void Server::ResponseHeaderHandler(
...
-				"Response both requested to switch protocol, and has a body, which is not supported");
+			string msg =
+				"Response both requested to switch protocol, and has a body, which is not supported";
+			connection->logger_.Error(msg);
 			exp_body_reader.value()->Cancel();
-			connections_.erase(req_in);
+			resp_out->SetStatusCodeAndMessage(http::StatusNotImplemented, msg);
+			// Remove body if we set it previously.
+			resp_out->SetHeader("Content-Length", "0");
+			auto err = resp_out->AsyncReply([cancelled, this, req_in](error::Error err) {
```
oleorhagen, line 212: NIt: Ignored, error, no?

kacf: Oh man, I must be tired. Thanks, fixed!

PR #1405, opened 2023-09-27
DBus fixes

vpodzime:
A big important one and a minor one.

### DIFF
--- a/common/dbus/platform/asio_libdbus/dbus.cpp
+++ b/common/dbus/platform/asio_libdbus/dbus.cpp
@@ -417,7 +417,7 @@ ExpectedStringPair ExtractValueFromDBusMessage(DBusMessage *message) {
 		dbus_error_free(&dbus_error);
 		return expected::unexpected(err);
 	}
-	return std::pair<string, string> {string(value1), string(value1)};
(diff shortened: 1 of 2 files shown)


kacf (approved):

common/dbus/platform/asio_libdbus/dbus.cpp
```diff
@@ -417,7 +417,7 @@ ExpectedStringPair ExtractValueFromDBusMessage(DBusMessage *message) {
 		dbus_error_free(&dbus_error);
 		return expected::unexpected(err);
 	}
-	return std::pair<string, string> {string(value1), string(value1)};
+	return StringPair {string(value1), string(value2)};
```
oleorhagen, line 433: Did we not comment on this in the original PR? :thinking:

oleorhagen (approved):
Nice!

common/dbus/platform/asio_libdbus/dbus.cpp
vpodzime: Yes, moved the fix into a separate commit and then moved the fix commit to this fix PR. :grin:

PR #1406, opened 2023-09-27
Refactor the configuration into one package

oleorhagen:
See commit msg

### DIFF
--- a/conf/config.go
+++ b/conf/config.go
@@ -22,12 +22,12 @@ import (
 	"github.com/pkg/errors"
 	log "github.com/sirupsen/logrus"
 
-	"github.com/mendersoftware/mender/client"
 	"github.com/mendersoftware/mender/dbus"
 )
 
 const (
 	DefaultUpdateControlMapBootExpirationTimeSeconds = 600
+	Pkcs11URIPrefix                                  = "pkcs11:"
 )
 
 type MenderConfigFromFile struct {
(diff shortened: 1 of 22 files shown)


oleorhagen:
Sorry, did not get around it :)

conf/config.go
```diff
@@ -127,6 +127,80 @@ type DualRootfsDeviceConfig struct {
...
+// MenderServer is a placeholder for a full server definition used when
+// multiple servers are given. The fields corresponds to the definitions
+// given in MenderConfig.
+type MenderServer struct {
+	ServerURL string
+	// TODO: Move all possible server specific configurations in
+	//       MenderConfig over to this struct. (e.g. TenantToken?)
+}
+
+type Config struct {
```
lluiscampos, line 197: In this package now, `conf.Config` doesn't really describe what this struct holds. When it was `client.Config` made more sense.

If it is not too much work, can we rename to something explicit like `conf.ClientConfig` or `conf.HttpConfig`?

lluiscampos (approved):
I think we could rename the struct (see below), but being this soon legacy code I won't insist in beatifications. Looks good!

lluiscampos (approved):

conf/config.go
oleorhagen: Went for `HttpConfig`

PR #1407, opened 2023-09-28
test: Drain event queue before checking number of connections.

kacf:


lluiscampos (approved):
I've seen this before :+1:

PR #1408, opened 2023-09-28
chore: Exit mender-auth with an early error if no servers specified

vpodzime:
With no servers, mender-auth just doesn't know which servers to try to authenticate with.

Ticket: none
Changelog: none

### DIFF
--- a/mender-auth/cli/actions.cpp
+++ b/mender-auth/cli/actions.cpp
@@ -96,6 +96,14 @@ ExpectedActionPtr DaemonAction::Create(
 }
 
 error::Error DaemonAction::Execute(context::MenderContext &main_context) {
+	auto &config = main_context.GetConfig();
+	if (!none_of(config.servers.cbegin(), config.servers.cend(), [](const string &it) {
+			return it != "";
+		})) {
+		log::Error("Cannot run in daemon mode with no server URL specified");
+		return error::MakeError(error::ExitWithFailureError, "");
+	}
+
 	auto err = DoBootstrap(keystore_, force_bootstrap_);
 	if (err != error::NoError) {
 		log::Error("Failed to bootstrap: " + err.String());
@@ -104,7 +112,7 @@ error::Error DaemonAction::Execute(context::MenderContext &main_context) {
 
 	events::EventLoop loop {};
 
-	ipc::Server ipc_server {loop, main_context.GetConfig()};
+	ipc::Server ipc_server {loop, config};
 
 	err = ipc_server.Listen();
 	if (err != error::NoError) {


mender-auth/cli/actions.cpp
```diff
@@ -96,6 +96,14 @@ ExpectedActionPtr DaemonAction::Create(
 }
 
 error::Error DaemonAction::Execute(context::MenderContext &main_context) {
+	auto &config = main_context.GetConfig();
+	if (!none_of(config.servers.cbegin(), config.servers.cend(), [](const string &it) {
```
lluiscampos, line 100: Is this correct? We also need to check for `server_url` field, as the `servers` array could be empty but a single server could be specified in `ServerURL`. We have a [validation method for that](https://github.com/mendersoftware/mender/blob/fb296beb6aa695ab0da29980dd540a1c6177f187/common/config_parser/config_parser.cpp#L66), why not using it?

I think @kacf has a fix to be summited soon that is is actually merging these two

lluiscampos (changes requested):

mender-auth/cli/actions.cpp
oleorhagen: If I understood it correctly, the `servers` will always be in the `servers` array going forwards, and hence the `server_url` field should be moot going forwards

kacf: It's correct, I told @vpodzime to use only `servers`, since I'm removing `server_url` as part of my conf fixes.

lluiscampos: Okey :see_no_evil:

lluiscampos: Still, why can't we use the validation method mentioned above?

kacf: I think @vpodzime's point is that it should be done during config file parsing. I don't really have a strong opinion, it could fit in both places.

lluiscampos (approved):

mender-auth/cli/actions.cpp
kacf: Since I'm working on the config I think I need to backtrack on what I said: It looks like it could be beneficial to keep the check for any available server in the mender-auth code:

1. The first reason, which is arguably more of an annoyance, is that by moving the check to the `common/conf` code, every test that parses config, even if it is to test something minor, now needs a `ServerURL` in it, or it won't validate and the test fails for reasons unrelated to what is being tested.
2. More importantly though: The code is shared between mender-auth and mender-update, but it is perfectly legal for mender-update to run without a ServerURL. Maybe you are running mender-auth with different launch options, or maybe it has been replaced with a dummy binary with a simple endpoint to facilitate no-auth in a safe, isolated network.
3. It can arguably be easier to let the production code decide what it needs via an error return, instead of keeping the check separately in the config parsing code.

What do you think, @vpodzime?

vpodzime: My way of thinking was that *mender-update* doesn't need any ServerURL and also *mender-auth* doesn't need it in all cases. So I put the check to the place in the code where it's 100% obvious it will be needed (for `Listen()` right below it). I was a bit surprised how "deep" it was, but I still don't see a better place.

PR #1409, opened 2023-09-29
First round of fixes related to config file.

kacf:


lluiscampos (approved):

common/config_parser/config_parser.cpp
```diff
@@ -426,6 +426,12 @@ ExpectedBool MenderConfigFromFile::ValidateConfig() {
 	if (!server_conf) {
 		return server_conf;
 	}
+
+	if (!dbus_enabled) {
+		log::Error("Turning DBus off is no longer supported");
+		return false;
+	}
```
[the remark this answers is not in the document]
kacf: Now that I think of it, why even keep this option? We can just remove it as well. We are committed to DBus at this point, it's extremely unlikely that this will ever be an allowed configuration option.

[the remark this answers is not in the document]
kacf: I did exactly that. Let me know if you disagree.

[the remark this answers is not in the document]
lluiscampos: I agree. Good thinking.

lluiscampos (approved):

vpodzime (approved):
Looks good to me.

common/config_parser/config_parser.cpp
```diff
@@ -327,7 +305,7 @@ ExpectedBool MenderConfigFromFile::LoadFile(const string &path) {
 		}
 	}
 
-	/* Vectors/arrays now */
```
oleorhagen, line 330: ?

common/config_parser.hpp
```diff
@@ -89,9 +89,11 @@ class MenderConfigFromFile {
 	/** Connectivity connection handling and transfer parameters */
 	ClientConnectivity connectivity;
 
-	/** Rootfs device paths */
-	string rootfs_part_A;
-	string rootfs_part_B;
+	/** Rootfs device paths. These are not parsed by the client anymore, since rootfs updates
+		are now handled by an update module. But for historical reasons, they still share config
+		files, so these options can still be in the file. */
+	// string rootfs_part_A;
```
oleorhagen, line 96: Q: Are we, if they are not set, just going to ignore it until the UM fails?

artifact/v3/scripts/executor.cpp
```diff
@@ -254,7 +254,7 @@ Error ScriptRunner::AsyncRunScripts(
 	if (!exp_scripts) {
 		// Missing directory is OK
 		if (exp_scripts.error().IsErrno(ENOENT)) {
-			log::Warning("Found no state script directory (" + script_path + "). Continuing on");
+			log::Debug("Found no state script directory (" + script_path + "). Continuing on");
```
oleorhagen, line 257: Hah, I have the same one locally :smile:

common/conf/conf.cpp
```diff
@@ -210,11 +210,15 @@ expected::ExpectedSize MenderConfig::ProcessCmdlineArgs(
...
 
-	auto ex_log_level = log::StringToLogLevel(log_level);
-	if (!ex_log_level) {
-		return expected::unexpected(ex_log_level.error());
+	SetLevel(log::kDefaultLogLevel);
+
+	if (log_level != "") {
+		auto ex_log_level = log::StringToLogLevel(log_level);
+		if (!ex_log_level) {
+			return expected::unexpected(ex_log_level.error());
```
oleorhagen, line 218: Hmm, is it better to just log a warning and go with the default?

oleorhagen (approved):
Fun one pointer :laughing:

common/config_parser/config_parser.cpp
kacf: They are not in order anymore, that's why I removed this comment and the `Deal with plain string values first` above. It's needed because I have to parse the vectors first for artifact keys and servers, and I didn't want to swap the order of all of them (huge diff, and not really that important).

common/config_parser.hpp
kacf: Yep! It's out of our area of responsibility, basically.

common/config_parser/config_parser.cpp
```diff
@@ -426,6 +412,7 @@ ExpectedBool MenderConfigFromFile::ValidateConfig() {
 	if (!server_conf) {
 		return server_conf;
 	}
+
```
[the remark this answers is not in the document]
kacf: Never mind this one. The whole function will disappear in my followup PR.

common/conf/conf.cpp
kacf: I think we should avoid silently accepting invalid options (well, silently in the sense that you have to actively check the log), unless there's a high risk of compromising robustness. I don't think that's the case here.

common/config_parser.hpp
```diff
@@ -87,27 +90,36 @@ class MenderConfigFromFile {
...
+	// int update_control_map_expiration_time_seconds = 0;
 
-	/** Expiration timeout for the control map when just booted */
-	int update_control_map_boot_expiration_time_seconds = 600; // 10 min
+	/** Expiration timeout for the control map when just booted. The Update Control feature has
+		been removed from the C++ client. */
+	// int update_control_map_boot_expiration_time_seconds = 600; // 10 min
 
 	/** Poll interval for checking for new updates */
 	int update_poll_interval_seconds = 1800;
```
lluiscampos, line 125: Few lines below, we have:
```
/** Poll interval for checking for update (check-update) */
int state_script_retry_interval_seconds = 60;
```

which looks wrong to me.

I am working on something related locally so I just noticed.

PR #1410, opened 2023-10-02
One occasional crash fix + test output prettyfication.

kacf:


kacf:
I hope this might fix some of the sporadic failures we've been having.

lluiscampos (approved):

vpodzime (commented):
1. Is the hex noise really printed for all tests or only for those that rely on `GetParam()`? I've never seen it.
2. And I believe the last commit deserves some more details in the commit message.

kacf:
> 1. Is the hex noise really printed for all tests or only for those that rely on `GetParam()`? I've never seen it.

Because we have hidden the ones that came before. I just decided to add it everywhere now so that it's spread around when copy pasting.

> 2\. And I believe the last commit deserves some more details in the commit message.

It's work in progress, just ignore it for now. I can't reproduce it locally, that's why.

kacf:
@lluiscampos, @vpodzime: Could you have a second look? The first fix wasn't enough, and I had to do a few more. The last two commits are new.

lluiscampos (approved):
It seems like you have good control of the issue. The reasoning makes sense and the code looks good.

vpodzime:
> the code looks good

Not risking ruining this nice picture by actually looking at the code. :grin:

PR #1411, opened 2023-10-03
Misc fixes from integration test runs

oleorhagen:


common/state_machine.hpp
```diff
@@ -166,6 +167,15 @@ class StateMachineRunner : virtual public EventPoster<EventType> {
 			}
 		}
 
+		auto loop_queue = event_queue_;
+		while (!loop_queue.empty()) {
+			// auto event_ = even;
```
lluiscampos, line 172: Leftovers :scissors:

support/mender-updated.service
```diff
@@ -9,7 +9,7 @@ Conflicts=mender-client.service
 Type=idle
 User=root
 Group=root
-ExecStart=/usr/bin/mender-update daemon
+ExecStart=/usr/bin/mender-update --log-level=trace daemon
```
lluiscampos, line 12: This we need to revert

mender-update/deployments/deployments.cpp
```diff
@@ -141,6 +141,12 @@ error::Error DeploymentClient::CheckNewDeployments(
 			}
 		} else if (status == http::StatusNoContent) {
 			api_handler(CheckUpdatesAPIResponse {nullopt});
+		} else {
+			log::Info(
```
lluiscampos, line 145: I think this should be a `Warning`. Do you have a good reason not to?

mender-update/deployments/deployments.cpp
```diff
@@ -141,6 +141,12 @@ error::Error DeploymentClient::CheckNewDeployments(
...
+				"DeploymentClient::CheckNewDeployments - received unhandled http response: "
+				+ to_string(status));
+			api_handler(expected::unexpected(MakeError(
+				DeploymentAbortedError, "received unhandled http response: " + to_string(status))));
```
lluiscampos, line 149: Nit: `HTTP` instead of `http` - I didn't find any other log message using lowercase `http`.

artifact/v3/header/header.cpp
```diff
@@ -87,6 +87,18 @@ ExpectedHeader Parse(io::Reader &reader, ParserConfig conf) {
 
 	tok = lexer.Next();
 	vector<ArtifactScript> state_scripts {};
+	if (tok.type == token::Type::ArtifactScripts) {
+		// Make sure the `scripts` dir exists
+		if (not path::FileExists(conf.artifact_scripts_filesystem_path)) {
+			log::Trace(
+				"Creating the Artifact script directory: " + conf.artifact_scripts_filesystem_path);
+			error::Error err = path::CreateDirectory(conf.artifact_scripts_filesystem_path);
```
lluiscampos, line 95: What about parent directories? `filesystem::create_directory` will not create them for us and hence this fix is incomplete.

mender-update/daemon/states.hpp
```diff
@@ -280,7 +280,8 @@ class StateScriptState : virtual public StateType {
 		script_executor::Action action,
 		chrono::seconds retry_interval,
 		const string &artifact_script_path,
-		const string &rootfs_script_path) :
+		const string &rootfs_script_path,
+		script_executor::OnError on_error_ = script_executor::OnError::Fail) :
```
lluiscampos, line 284: Why give it a default value? I would vote for it to be mandatory.

mender-update/daemon/states.hpp
```diff
@@ -289,7 +289,8 @@ class StateScriptState : virtual public StateType {
 			rootfs_script_path,
 		},
 		state_ {state},
-		action_ {action} {};
+		action_ {action},
+		on_error_ {on_error_} {};
```
lluiscampos, line 293: Nit: the commit message says `ignore_error` but the fix is about `on_error` parameter (which can happen to have an "ignore" value).

lluiscampos (approved):
Many nice fixes! Overall looks good, I pointed out few things but nothing to block the PR from my side.

The detail on `state_machine.cpp` looks good to me but I am not knowledgeable enough there.

mender-update/daemon/states.hpp
kacf: +1

api/auth/auth.cpp
```diff
@@ -231,48 +231,40 @@ error::Error FetchJWTToken(
 }
 
 void Authenticator::ExpireToken() {
-	unique_lock<mutex> lock {auth_lock_};
```
kacf, line 234: Nit: Would perhaps say, "The code is run in the event loop, **not in a separate thread**", just to be extra clear.

common/http/platform/beast/http.cpp
```diff
@@ -475,6 +478,9 @@ void Client::ConnectHandler(const error_code &ec, const asio::ip::tcp::endpoint
 		return;
 	}
 
+	boost::asio::socket_base::keep_alive option(true);
+	stream_->next_layer().set_option(option);
```
kacf, line 482: Remember to use [the config option](https://github.com/mendersoftware/mender/blob/feature-c%2B%2B-client/common/config_parser.hpp#L51).

kacf (changes requested):
Nice, many gems here! And by that I mean: Stupid mistakes we made that made me laugh! 😄

common/state_machine.hpp
```diff
@@ -166,6 +167,15 @@ class StateMachineRunner : virtual public EventPoster<EventType> {
...
+			auto event_shadow = loop_queue.front();
+			loop_queue.pop();
```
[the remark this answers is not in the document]
oleorhagen: I will remove this. Was not meant to be a part of the PR

mender-update/daemon/states.hpp
oleorhagen: :ok:

mender-update/deployments/deployments.cpp
```diff
@@ -285,7 +285,7 @@ error::Error DeploymentClient::PushStatus(
 			auto resp = exp_resp.value();
 			auto content_length = resp->GetHeader("Content-Length");
 			if (!content_length) {
-				log::Error(
+				log::Debug(
```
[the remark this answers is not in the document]
kacf: Right, I missed that this can happen (actually, it can *only* happen) when there is no body. But that is indeed possible here.

However, this means that the rest of the code is wrong: Since it can only happen when there is no body, we should not set `body_writer->SetUnlimited(true)` (it won't misbehave, but it masks the intention), so I would remove that in addition to the fix you did.

common/http/platform/beast/http.cpp
```diff
@@ -475,6 +481,9 @@ void Client::ConnectHandler(const error_code &ec, const asio::ip::tcp::endpoint
...
```
kacf, line 485: What about here?

common/path/platform/c++17/path.cpp
```diff
@@ -70,6 +82,12 @@ expected::ExpectedBool IsExecutable(const string &file_path, const bool warn) {
 	return true;
 }
 
+void Permissions(const string &file_path, const vector<Perms> perms) {
+	fs::perms p;
+	std::for_each(perms.cbegin(), perms.cend(), [&p](const Perms perm) { p |= perm_map.at(perm); });
+	fs::permissions(file_path, p);
```
kacf, line 88: 👍

artifact/v3/header/header.cpp
lluiscampos: > It is create directory singular

Yes, I think the helper is good

> I don't think we can run into this situation

The path can be overridden, so we can run into this situation.

I think we either log and error or fix all the way. As it is now in the scenario of a custom path it will error _creating_ the directory rather than running the scripts, which I think is more confusing than helping.

kacf:
From `git range-diff`:

```
18:  2c9dc7ac ! 18:  82edabc5 fix(executore): Also check the version file for the rootfs scripts
    @@ Metadata
     Author: Ole Petter Orhagen <ole.orhagen@northern.tech>
     
      ## Commit message ##
    -    fix(executore): Also check the version file for the rootfs scripts
    +    fix(executor): Also check the version file for the rootfs scripts
```

I liked the Italian version! 😄

common/http/platform/beast/http.cpp
oleorhagen: :see_no_evil: 


https://external-content.duckduckgo.com/iu/?u=https%3A%2F%2Fmedia1.tenor.com%2Fimages%2F6f30ece33994552dd67b30e40134771f%2Ftenor.gif%3Fitemid%3D12347446&f=1&nofb=1&ipt=8f0a7d9fb05a49d78c9cdbf68ae9ded4de4d1f6d5115318584bc55fdb648c6b4&ipo=images

artifact/v3/header/header.cpp
kacf: If I'm tie breaker, I think I will vote for creating the whole directory chain. This is a function we are likely to reuse as well, and I think this behavior is in general more useful. And it is certainly not wrong to do it in this specific case either.

oleorhagen: Alright, `mkdir --parents cpp version` coming up :rocket:

common/http.hpp
```diff
@@ -272,9 +272,10 @@ class OutgoingRequest : public Request {
 
 class Stream;
 
-class IncomingRequest : public Request,
-						virtual public io::Canceller,
-						public enable_shared_from_this<IncomingRequest> {
+class IncomingRequest :
+	public Request,
+	virtual public io::Canceller,
+	public enable_shared_from_this<IncomingRequest> {
```
kacf, line 278: From the range-diff it looks like almost this entire commit stems from the `SO_KEEPALIVE` commit, where this formatting was changed. So you could perhaps get rid of it by squashing it there.

mender-update/daemon/states.cpp
```diff
@@ -70,7 +70,9 @@ void StateScriptState::OnEnter(Context &ctx, sm::EventPoster<StateEvent> &poster
 	string state_name {script_executor::Name(this->state_, this->action_)};
 	log::Debug("Executing the  " + state_name + " State Scripts...");
 	auto err = this->script_.AsyncRunScripts(
-		this->state_, this->action_, [state_name, &poster](error::Error err) {
+		this->state_,
+		this->action_,
+		[state_name, &poster](error::Error err) {
```
kacf, line 75: The only exception is this.

Btw, do you know how to run clang-format individually on each commit? Ping me if you want me to show you.

kacf (approved):

common/http.hpp
oleorhagen: Well, yes and no. 

I touched the file in that commit. But not this part of the file...

So the formatting has changed, since the last time we touched  the file.

I therefore judged to keep it out of the commit, as it is not related to the functionality change, but feel free to correct me still.

mender-update/daemon/states.cpp
lluiscampos: @kacf ping! I'd like to learn your process.

kacf: For future reference, check out https://gitlab.com/kacf/gems.

PR #1412, opened 2023-10-03
MEN-6677: Implement state script retry-later functionality

lluiscampos:


artifact/v3/scripts/executor.cpp
```diff
@@ -152,8 +152,8 @@ ScriptRunner::ScriptRunner(
 	chrono::seconds state_script_timeout,
 	const string &artifact_script_path,
 	const string &rootfs_script_path,
-	mender::common::processes::OutputCallback stdout_callback,
-	mender::common::processes::OutputCallback stderr_callback) :
+	processes::OutputCallback stdout_callback,
```
oleorhagen, line 157: hah, thanks! 

I've been thinking this every time I've looked at it...

artifact/v3/scripts/executor.cpp
```diff
@@ -177,36 +181,88 @@ void ScriptRunner::LogErrAndExecuteNext(
...
+void ScriptRunner::MaybeSetupRetryTimeoutTimer() {
+	if (!this->retry_timeout_timer_) {
+		log::Debug("Setting retry timer for " + to_string(this->retry_timeout_.count()) + "ms");
+		// First run on this script
+		this->retry_timeout_timer_.reset(new events::Timer(this->loop_));
+		this->retry_timeout_timer_->AsyncWait(this->retry_timeout_, [this](error::Error err) {
+			if (err.code == make_error_condition(errc::operation_canceled)) {
+				// The timer did not fire up. Do nothing
+			} else {
+				this->retry_interval_timer_->Cancel();
```
oleorhagen, line 251: S: Log the error

artifact/v3/scripts/executor.cpp
```diff
@@ -177,36 +181,88 @@ void ScriptRunner::LogErrAndExecuteNext(
...
 }
 
+void ScriptRunner::HandleScriptRetry(
+	vector<string>::iterator current_script,
+	vector<string>::iterator end,
+	bool ignore_error,
+	HandlerFunction handler) {
+	log::Info(
+		"Script returned Retry Later exit code, re-retrying in "
+		+ to_string(this->retry_interval_.count() / 1000) + "s");
```
oleorhagen, line 225: Q: Could we not go via `chrono` somehow?

oleorhagen (approved):
Interesting.

I was thinking this was going to be a part of the state machine, but this is good too :)

lluiscampos:
> I was thinking this was going to be a part of the state machine, but this is good too :)

Me too. But reading on the golang code I saw that everything was done [in the `executor`](https://github.com/mendersoftware/mender/blob/719cc9990b5b35cb69783b3bb15ae0e1d5d0e589/statescript/executor.go#L313) so I tried to mimic.

kacf:
> I was thinking this was going to be a part of the state machine, but this is good too

This was my thinking too, but I guess it was never said explicitly. The Golang client isn't really the best inspiration for things related to the retrying and state machinery, since it was essentially the reason for doing a different approach in the C++ client.

That being said, I don't think it's critical for this feature, since we don't have any plans for features that would allow a different state flow while inside a retry cycle. And if we did, we'd probably re-implement Update Control instead.

So yeah, this is good as is. Since Ole is more into the executor than me, I'll trust his judgment here, and won't do my own review unless there is something specific you want me to look at.

artifact/v3/scripts/executor.cpp
lluiscampos: Not really, that I could find. On the one hand the [stream output operator](https://en.cppreference.com/w/cpp/chrono/duration/operator_ltlt) is C++20 only, and on the other hand the duration conversions are only allowed from bigger to smaller unit (seconds to milliseconds, for example, but not the other way around!).

Related,

We use `milliseconds` only for the sake of tests. On production the duration unit (coming from the config file or from defaults) is seconds. Do you have a different suggestion on how could I keep using seconds in the code and only tweak it for tests?

artifact/v3/scripts/executor.hpp
```diff
@@ -107,19 +109,34 @@ class ScriptRunner {
...
+	chrono::milliseconds retry_timeout_;
 	string artifact_script_path_;
 	string rootfs_script_path_;
 	processes::OutputCallback stdout_callback_;
 	processes::OutputCallback stderr_callback_;
 	Error error_script_error_;
 	vector<string> collected_scripts_;
 	unique_ptr<processes::Process> script_;
+	unique_ptr<events::Timer> retry_interval_timer_;
+	unique_ptr<events::Timer> retry_timeout_timer_;
```
vpodzime, line 139: My gut feeling tells me that doing allocation and deallocation to avoid a `bool` flag on the timer is a waste of resources. Not a biggie, of course, but I think adding that `bool active_` flag and `bool Timer::GetActive() { return active_ }` wouldn't be that much work.

artifact/v3/scripts/executor.cpp
```diff
@@ -222,18 +279,23 @@ Error ScriptRunner::Execute(
...
+					&& this->script_->GetExitStatus() == state_script_retry_exit_code;
+				if (is_script_retry_error) {
+					MaybeSetupRetryTimeoutTimer();
+					return HandleScriptRetry(current_script, end, ignore_error, handler);
+				} else {
+					if (ignore_error) {
+						return LogErrAndExecuteNext(
+							err, current_script, end, ignore_error, handler);
+					}
+					return HandleScriptError(err, handler);
```
vpodzime, line 293: You could do `else if ()` and then `return HandleScriptError()` outside of the `if-else if`.

artifact/v3/scripts/executor.cpp
```diff
@@ -222,18 +279,23 @@ Error ScriptRunner::Execute(
 		this->loop_,
 		[this, current_script, end, ignore_error, handler](Error err) {
 			if (err != error::NoError) {
-				if (ignore_error) {
-					return LogErrAndExecuteNext(err, current_script, end, ignore_error, handler);
+				const bool is_script_retry_error =
+					err.code == processes::MakeError(processes::NonZeroExitStatusError, "").code
+					&& this->script_->GetExitStatus() == state_script_retry_exit_code;
```
vpodzime, line 284: Nice!

vpodzime (approved):
Looks good to me otherwise.

artifact/v3/scripts/executor.cpp
oleorhagen: Alright, makes sense :)

artifact/v3/scripts/executor.hpp
```diff
@@ -71,15 +71,15 @@ class ScriptRunner {
 public:
 	ScriptRunner(
 		events::EventLoop &loop,
-		chrono::seconds state_script_timeout,
+		chrono::milliseconds script_timeout,
```
oleorhagen, line 74: Q: Would it not be better to keep it as is, in seconds?

lluiscampos: Yes, I would like to. Can you help me [here](https://github.com/mendersoftware/mender/pull/1412#discussion_r1348307973)?

oleorhagen: Shit. Maybe ur right. This was harder than I thought.

Unsure what to do here. I would kinda like the production code to be in seconds, but then again slow tests.

No hard opinions really. Do what you see fit here :)

artifact/v3/scripts/executor.cpp
kacf: Could you use `chrono::duration_cast<chrono::seconds>(this->retry_interval_).count()`? It always rounds down, but I guess that isn't a big concern here.

lluiscampos: Oh, I missed `duration_cast`, thanks!

artifact/v3/scripts/executor.hpp
lluiscampos: Being Friday, I will try the template way. I'll put it on a separate commit so that we can decide to drop.

lluiscampos: I failed.

oleorhagen: No worries, let's keep templates out of this anywho.

`milliseconds` is fine, let's keep it and move on :)

PR #1413, opened 2023-10-03
chore: Take Update Control into account when upgrading.

kacf:
Since we don't support it, fail instead.

Ticket: MEN-6647
Changelog: Update Control support has been removed from the client.

### DIFF
--- a/mender-update/daemon/state_machine/state_machine.cpp
+++ b/mender-update/daemon/state_machine/state_machine.cpp
@@ -15,6 +15,7 @@
 #include <mender-update/daemon/state_machine.hpp>
 
 #include <common/conf.hpp>
+#include <common/key_value_database.hpp>
 #include <common/log.hpp>
 
 #include <mender-update/daemon/states.hpp>
@@ -24,6 +25,7 @@ namespace update {
 namespace daemon {
 
 namespace conf = mender::common::conf;
+namespace kvdb = mender::common::key_value_database;
 namespace log = mender::common::log;
 
 StateMachine::StateMachine(Context &ctx, events::EventLoop &event_loop) :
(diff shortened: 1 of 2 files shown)


lluiscampos (commented):
The acceptance criteria says "Upon startup, the key must be removed from the database if, and only if, there is no deployment ongoing" - where is that done?

And why don't we need to also fail a deployment from Download state if we get a deployment with update control maps?

kacf:
> The acceptance criteria says "Upon startup, the key must be removed from the database if, and only if, there is no deployment ongoing" - where is that done?

Oh good catch, I missed that! Will add it.

> And why don't we need to also fail a deployment from Download state if we get a deployment with update control maps?

All of the other states except `kUpdateStateArtifactReboot` either trigger a failure, or are on the failure path already, so it's not necessary anywhere else. But this may need revisiting later if we add more states that we can recover from, such as resuming a download after a spontaneous reboot.

mender-update/daemon/state_machine/state_machine.cpp
```diff
@@ -318,8 +321,19 @@ void StateMachine::LoadStateFromDb() {
...
 	if (!exp_loaded.value()) {
 		log::Debug("No existing deployment data, starting from initial state");
+
+		auto err = store.Remove(ctx_.mender_context.update_control_maps);
+		if (err != error::NoError) {
+			log::Error(
+				"Error removing " + ctx_.mender_context.update_control_maps
+				+ " key from database: " + err.String());
+			// Nothing we can do about it.
+		}
```
lluiscampos, line 335: This will forever log an Error when booting. I think is best to check if the key exists before removing it, and only then log an error.

lluiscampos (approved):
It looks good otherwise. Take my comment as a suggestion, you probably know best :)

mender-update/daemon/state_machine/state_machine.cpp
kacf: It's smart enough to [ignore removing a non-existing key](https://github.com/mendersoftware/mender/blob/feature-c%2B%2B-client/common/key_value_database/platform/lmdb/lmdb.cpp#L73-L76).

PR #1415, opened 2023-10-04
chore: Consolidate struct members for server URL.

kacf:
Use only the list, instead of the list and the single entry.

Ticket: MEN-6662


lluiscampos (commented):
It looks good to me :+1: We'll do the final review after the other tasks are in place

lluiscampos (approved):

vpodzime (approved):
ACK

PR #1421, opened 2023-10-11
chore: Make it possible to configure repo for no-build situations.

kacf:
This still allows installing modules, generators, and Documentation, but does not need any dependencies to be installed except for the compiler.


oleorhagen (approved):
Nice!

I would consider moving it to the `cmake` dir though

kacf:
> I would consider moving it to the `cmake` dir though

Yeah, I was also thinking about it. My thinking was that since this is the main build file, and not just a sub component, I concluded it probably belongs in the root dir. Especially the sub directory inclusion becomes a bit weird if this is itself in a subdirectory (would need `add_subdirectory(../common)`).

PR #1423, opened 2023-10-12
MEN-6679: Implement help screen for all options

lluiscampos:


lluiscampos:
Test MEN-6679 :robot:

lluiscampos:
Test https://northerntech.atlassian.net/browse/MEN-6679 :robot:

mender-update/cli/cli.cpp
```diff
@@ -55,12 +55,6 @@ const cli::Command cmd_install {
 				.description =
 					"Return exit code 4 if a manual reboot is required after the Artifact installation",
 			},
-			// TODO: not implemented
-			cli::Option {
-				.long_option = "passphrase-file",
-				.description =
-					"Passphrase file for decrypting an encrypted private key. '-' loads passphrase from stdin",
-			},
```
kacf: > Changelog: Deprecate `mender install --passphrase-file` CLI option. This option was added by mistake in previous versions of the Mender client and had no use (private key is not required to install an Artifact).

I think this should also be rephrased, it makes it sound too serious. How about: "Remove `--passphrase-file` argument from command scopes where it has no use. Move it to `mender-auth daemon --passphrase-file`, which is the only appropriate location."

tests/entrypoint.sh
```diff
@@ -23,4 +23,6 @@ if [ -n "$TENANT_TOKEN" ]; then
 fi
 
 /etc/init.d/ssh start
-mender --no-syslog daemon
+mender-auth daemon &
+sleep 10
+mender-update daemon
```
kacf, line 28: Why 10 seconds, isn't it more like 1 (if even that)?

common/conf/conf_cli_help.cpp
```diff
@@ -35,7 +35,62 @@ const size_t max_width = 78;
...
+   - Supported log levels incudes: 'debug', 'info', 'warning', 'error', 'panic' and
+     'fatal'.
+
+Environment variables:
+   - MENDER_CONF_DIR - configuration (default: )"
+										 + default_paths.GetPathConfDir() + R"().
+   - MENDER_DATA_DIR - identity, inventory and update modules (default: )"
+										 + default_paths.GetPathDataDir() + R"().
+   - MENDER_DATASTORE_DIR - runtime datastore (default: )"
+										 + default_paths.GetDataStore() + R"().)";
```
kacf, line 91: > `R"().)"`

Lol, confusing! But correct!

common/conf/conf_cli_help.cpp
```diff
@@ -35,7 +35,62 @@ const size_t max_width = 78;
...
+		.description = "Mender state data DIRECTORY path",
+		.default_value = default_paths.GetPathDataDir(),
+		.parameter = "DIR"},
+	CliOption {
+		.long_option = "log-file",
+		.short_option = "L",
+		.description = "FILE to log to",
+		.parameter = "FILE"},
+	CliOption {
+		.long_option = "log-level",
```
kacf, line 65: It's unfortunate that these need to be specified separately from the options that are actually used in parsing. Isn't there a way we can unify them?

kacf (commented):
Wow! You have done a better job than most cli libs I've used! 😃

common/cli/cli.cpp
```diff
@@ -0,0 +1,171 @@
...
+	stream << endl;
+
+	stream << "GLOBAL OPTIONS:" << endl;
+	PrintOptions(cli.global_options, stream);
+}
+
+void PrintCliCommandHelp(const App &cli, const string &command_name, ostream &stream) {
+	auto match_on_name = [command_name](const Command &cmd) { return cmd.name == command_name; };
+
+	auto cmd = std::find_if(cli.commands.begin(), cli.commands.end(), match_on_name);
```
oleorhagen: Nice!

common/common.cpp
```diff
@@ -69,5 +69,28 @@ string JoinStrings(const vector<string> &str, const string &delim) {
...
+	const vector<string> &str, const string &delim, const size_t max_width) {
+	vector<string> ret;
+	auto s = str.begin();
+	if (s == str.end()) {
+		return ret;
+	}
+
+	string line = *s;
+	s++;
+	for (; s != str.end(); s++) {
```
oleorhagen, line 82: S: Why not add this extra functionality as a function to `JoinStrings`, and then give it an empty default? 

Then we don't have to duplicate this?f

common/conf.hpp
```diff
@@ -20,6 +20,7 @@
 #include <unordered_set>
 #include <common/config_parser.hpp>
 #include <common/path.hpp>
+#include <common/cli.hpp>
```
oleorhagen: nit; sort

common/conf/conf.cpp
```diff
@@ -203,6 +214,11 @@ expected::ExpectedSize MenderConfig::ProcessCmdlineArgs(
 		}
 	}
 
+	if (help_arg) {
+		cli::PrintCliHelp(app);
+		return expected::unexpected(error::MakeError(error::ExitWithSuccessError, ""));
```
oleorhagen, line 228: No definite answer obvs, but should we error on no args?

mender-auth/cli/cli.cpp
```diff
@@ -38,6 +40,109 @@ namespace setup = mender::common::setup;
...
+
+const cli::App cli_mender_auth = {
+	.name = "mender-auth",
+	.short_description = "manage and start Mender Auth",
+	.long_description =
+		R"(mender-auth integrates both the mender-auth daemon and commands for manually
+   performing tasks performed by the daemon (see list of COMMANDS below).
+
+Global flag remarks:
+   - Supported log levels incudes: 'debug', 'info', 'warning', 'error', 'panic' and
```
oleorhagen: Nit: We also have the `trace` level

mender-auth/cli/cli.cpp
```diff
@@ -38,6 +40,109 @@ namespace setup = mender::common::setup;
...
+     'fatal'.
+
+Environment variables:
+   - MENDER_CONF_DIR - configuration (default: )"
+		+ default_paths.GetPathConfDir() + R"().
```
oleorhagen: NICE! Thourough

mender-auth/cli/cli.cpp
```diff
@@ -38,6 +40,109 @@ namespace setup = mender::common::setup;
...
+				.description = "Set logging level",
+				.default_value = "info",
+			},
+			// TODO: not implemented
+			cli::Option {
+				.long_option = "trusted-certs",
+				.short_option = "E",
+				.description = "Trusted server certificates FILE path",
+				.parameter = "FILE"},
+			// TODO: not implemented
```
oleorhagen: ?

common/conf_test.cpp
```diff
@@ -337,21 +336,21 @@ TEST(ConfTests, LogLevel) {
 	{
 		vector<string> args {"--log-level", "error"};
 		conf::MenderConfig config;
-		config.ProcessCmdlineArgs(args.begin(), args.end(), cli::App {});
+		config.ProcessCmdlineArgs(args.begin(), args.end(), conf::CliApp {});
```
oleorhagen, line 339: Hmm, why not keep the `cli` namespace? :thinking:

oleorhagen (approved):
Really, really nice!

:rocket: :medal_military:

common/conf/conf_cli_help.cpp
lluiscampos: I thought about this too at the beginning.

Once I merged everything in the same namespace, I think it makes sense to do it. I'll try.

mender-auth/cli/cli.cpp
lluiscampos: On this commit I just marked everything that was not implemented and later commits implement or remove the flags.

Should I squash all in one commit?

common/conf_test.cpp
lluiscampos: I wrote it in the commit message.

Originally I wanted to keep both independent, but it turns out that `cli` depends on `conf` to get the default paths and `conf` depends on `cli` to print the help pages on explicit `--help` argument.

On the other hand it was relatively little code in `cli` and also `conf` had already functionality around cmd line parsing... so I decided to merge (it is a separate commit so can be dropped if you push for keeping them separated).

common/conf/conf.cpp
lluiscampos: I don't understand. You mea in combination with `--help` or in general? In general it will fail later with something like "an action needs to be specified".

From the options parsed here, is only `--help` and `--version` that are allowed with no command, the rest will require a command.

Also, note that `--version` does not accept other options or the command while `--help` accepts them but ignores. This is based on personal taste and I am open to make both same strict.

tests/entrypoint.sh
lluiscampos: Completely out of the blue, I haven't even run the container... :sweat_smile:

As this still depends on D-Bus being finalized I'll set it to 1 and make a mental note to test it a bit later (we use it in some tests, so we should catch errors here anyway).

common/CMakeLists.txt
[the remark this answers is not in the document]
lluiscampos: Thank you for this tip and for all the changelog text improvements :memo:

common/conf/conf.cpp
oleorhagen: No, I was thinking of the `exit code` when no arguments are provided :)

oleorhagen: > In general it will fail later with something like "an action needs to be specified".

Roger, let's keep it as it is then :)

mender-auth/cli/cli.cpp
oleorhagen: Up to you really. I think either way is fine

common/cli/cli.cpp
```diff
@@ -0,0 +1,171 @@
...
+
+const Option help_option = {
+	.long_option = "help",
+	.short_option = "h",
+	.description = "show help",
+	.default_value = "false",
+};
+
+template <typename InputIterator>
+using ColumnFormatter = function<string(typename iterator_traits<InputIterator>::value_type)>;
```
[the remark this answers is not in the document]
lluiscampos: We should thank :robot:

common/cli/cli.cpp
```diff
@@ -0,0 +1,171 @@
...
+using namespace std;
+
+namespace common = mender::common;
+
+const size_t max_width = 78;
+const string indent = "   ";   // 3 spaces
+const string separator = "  "; // 2 spaces
+
+const Option help_option = {
+	.long_option = "help",
```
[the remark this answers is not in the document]
lluiscampos: `long` is a reserved keyword in C/C++. It has to be `long_opt` or `l_option` or something like that so I'd rather keep it explicit `long_option`.

common/common.cpp
lluiscampos: I will try.

oleorhagen: True, didn't think of that...

common/conf/conf.cpp
```diff
@@ -164,8 +168,11 @@ expected::ExpectedSize MenderConfig::ProcessCmdlineArgs(
 			"-L",
 			"--log-level",
 			"-l",
+			"--trusted-certs",
+			"-E",
 		},
 		{
+			"--skipverify"
```
lluiscampos, line 175: :lady_beetle: Missing trailing `,`

mender-update/cli/cli.cpp
lluiscampos: Thanks. It is applicable to `daemon` and `bootstrap`, though.

common/conf_test.cpp
lluiscampos: I decided to keep them together, and I did yet another refactoring for `ProcessCmdlineArgs` to get the flags from the commands structs instead of having the strings duplicated.

lluiscampos:
This one is now on top of #1427 - please review that one first. Apart from that, the only relatively big change is the last commit (@kacf have a look) while the rest are minor follow-ups from review comments.

common/CMakeLists.txt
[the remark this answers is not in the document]
lluiscampos: @oleorhagen Fake news:

```
commit b89bdd6563026bd7fd6371b89f0a6304bc91532f
Author: Lluis Campos <lluis.campos@northern.tech>
Date:   Thu Oct 12 16:12:38 2023 +0200

    feat: Print app and commands help on `--help` and args errors
    
    Split the commands from the old `mender` cli into `mender-auth` and
    `mender-update`, adding all the options that we used to support.
    With one exception (see changelog) all options have been preserved.
    
    Some options are not actually implemented and will be addressed in
    upcoming commits.
    
    Changelog(chore): `--forcebootstrap` is not a global cli option anymore,
    but an option specific to `mender-auth daemon` and `mender-auth
    bootstrap` commands.
    
    Ticket: MEN-6679
    
Error: A 'Changelog' is required for the given commit
Commit b89bdd6563026bd7fd6371b89f0a6304bc91532f does not adhere to the conventional commit specification, used in the Mender project
See https://github.com/mendersoftware/mendertesting/blob/master/commitlint/grammar.md for more information
```

I'll see if I can fix the commit linting instead :thinking:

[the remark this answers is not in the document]
lluiscampos: https://github.com/mendersoftware/mendertesting/pull/298

common/conf/conf.cpp
```diff
@@ -155,29 +158,7 @@ expected::ExpectedSize MenderConfig::ProcessCmdlineArgs(
...
-			"-E",
-		},
-		{
-			"--skipverify",
-			"--version",
-			"-v",
-			"--help",
-			"-h",
-		});
+		start, end, GlobalOptsSetWithValue(), GlobalOptsSetWithoutValue());
```
kacf, line 161: Good!

But can we also do the same with the options in `ParseAuthArguments` and `ParseUpdateArguments`? I know, I'm being annoying! 😄

While you're at it, could you also add an `else` block to the end of each flag matcher in those two functions, with an `assert(false)`, just to make sure we catch the case where a flag is not handled.

lluiscampos: Good suggestions, both of them.

kacf: No, I think it's good! 👍 Now there is no duplication.

kacf (approved):
Great!

`git range-diff` is telling me that you removed the `--version` test though. Was that intentional? I didn't completely follow why the `DoMain` change was now omitted.

lluiscampos:
> Great!
> 
> `git range-diff` is telling me that you removed the `--version` test though. Was that intentional? I didn't completely follow why the `DoMain` change was now omitted.

It was merged in the preceding PR:  #1427 :+1:

kacf:
Ah of course, that fooled me.

PR #1425, opened 2023-10-16
test: Add test for missing certificate.

kacf:
Ticket: MEN-6787

### DIFF
--- a/common/http_test.cpp
+++ b/common/http_test.cpp
@@ -1824,6 +1824,58 @@ TEST(HttpsTest, WrongSelfSignedCertificateError) {
 	EXPECT_FALSE(client_hit_body);
 }
 
+TEST(HttpsTest, NoCertificateError) {
+	GTEST_SKIP() << "MEN-6787";
+
+	TestEventLoop loop;
+
+	bool client_hit_header {false};
+	bool client_hit_body {false};
+
+	mendertesting::TemporaryDirectory tmpdir;
+	string script = R"(#! /bin/sh
+	  exec openssl s_server -www )";
+	script += " -key server.localhost.key";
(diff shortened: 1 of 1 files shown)


common/http_test.cpp
```diff
@@ -1824,6 +1824,58 @@ TEST(HttpsTest, WrongSelfSignedCertificateError) {
...
+	err = client.AsyncCall(
+		req,
+		[&client_hit_header, &loop](http::ExpectedIncomingResponsePtr exp_resp) {
+			client_hit_header = true;
+			ASSERT_FALSE(exp_resp);
+			loop.Stop();
+		},
+		[&client_hit_body, &loop](http::ExpectedIncomingResponsePtr exp_resp) {
+			client_hit_body = true; // This should never happen
+			loop.Stop();
```
oleorhagen, line 1869: Nit: Unnecessary loop.Stop

oleorhagen (approved):

common/http_test.cpp
kacf: Correct, but I believe I've hit some of these while implementing things, and it terminates faster then. I don't consider it wrong, so I think I'll leave it.

oleorhagen: It's just fint, however, this got me curious. How does it terminate faster?

kacf: If it gets here, it means that `ASSERT_FALSE` caused it to return before the first `loop.Stop()` would run. So if we don't stop it here it has to wait for the timer.

Of course, there are a lot of ways this could be done differently. We could use `EXPECT_FALSE` instead. Or call `loop.Stop()` before `ASSERT_FALSE`. I didn't spend a lot of time optimizing this, to be honest! 😄

PR #1426, opened 2023-10-16
test: Add test for verifying hostname when doing TLS handshake.

kacf:
Ticket: MEN-6788

### DIFF
--- a/common/http_test.cpp
+++ b/common/http_test.cpp
@@ -1824,6 +1824,58 @@ TEST(HttpsTest, WrongSelfSignedCertificateError) {
 	EXPECT_FALSE(client_hit_body);
 }
 
+TEST(HttpsTest, CertificationWithWrongHostName) {
+	GTEST_SKIP() << "MEN-6788";
+
+	TestEventLoop loop;
+
+	bool client_hit_header {false};
+	bool client_hit_body {false};
+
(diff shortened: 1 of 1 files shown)


common/http_test.cpp
```diff
@@ -1824,6 +1824,58 @@ TEST(HttpsTest, WrongSelfSignedCertificateError) {
...
+	err = client.AsyncCall(
+		req,
+		[&client_hit_header, &loop](http::ExpectedIncomingResponsePtr exp_resp) {
+			client_hit_header = true;
+			ASSERT_FALSE(exp_resp);
+			loop.Stop();
+		},
+		[&client_hit_body, &loop](http::ExpectedIncomingResponsePtr exp_resp) {
+			client_hit_body = true; // This should never happen
+			loop.Stop();
```
oleorhagen, line 1869: Nit: I know this is copy and paste, but `loop.Stop()` should not be necessary here right?

oleorhagen (approved):

common/http_test.cpp
kacf: Correct, but I believe I've hit some of these while implementing things, and it terminates faster then. I don't consider it wrong, so I think I'll leave it.

PR #1427, opened 2023-10-16
Align cli namespaces' APIs and add test for `--version`

lluiscampos:


lluiscampos:
This is the remaining diff between the `main.cpp` files:

```
$ diff -u mender-update/main.cpp mender-auth/main.cpp 
--- mender-update/main.cpp	2023-10-16 14:46:20.161511103 +0200
+++ mender-auth/main.cpp	2023-10-16 14:54:21.547930984 +0200
@@ -15,18 +15,15 @@
 #include <string>
 #include <vector>
 
-#include <common/setup.hpp>
-#include <mender-update/cli/cli.hpp>
+#include <mender-auth/cli/cli.hpp>
 
 using namespace std;
 
 int main(int argc, char *argv[]) {
-	mender::common::setup::GlobalSetup();
-
 	vector<string> args;
 	if (argc > 1) {
 		args = vector<string>(argv + 1, argv + argc);
 	}
 
-	return mender::update::cli::Main(args);
+	return mender::auth::cli::Main(args);
 }
```

@kacf  @oleorhagen, do we need to do `GlobalSetup()` from `mender-auth`?

kacf:
> do we need to do `GlobalSetup()` from `mender-auth`?

We don't have to, but I think it's a good idea. In mender-update it was introduced in order to prevent termination if a process crashes while reading the Update Module Download pipe, which would send SIGPIPE. Mender-auth doesn't do that, but it could still crash if the terminal it's printing to suddenly closes, for example.

mender-auth/cli/cli.cpp
```diff
@@ -130,6 +130,21 @@ error::Error DoMain(
 	return action.value()->Execute(context);
 }
 
+int Main(const vector<string> &args, function<void(context::MenderContext &ctx)> test_hook) {
```
kacf, line 133: Commit message is a bit funky. Did you mean to write `mender::client::one` as a namespace?

kacf (approved):

mender-auth/cli/cli.cpp
lluiscampos: Hahaha. I meant to write ... `mender:update:cli` with `mender:auth`'s one (as in its `:cli`).

I'll rewrite.

lluiscampos:
> > do we need to do `GlobalSetup()` from `mender-auth`?
> 
> We don't have to, but I think it's a good idea. In mender-update it was introduced in order to prevent termination if a process crashes while reading the Update Module Download pipe, which would send SIGPIPE. Mender-auth doesn't do that, but it could still crash if the terminal it's printing to suddenly closes, for example.

All right. I seems harmless so I'll add it :+1:

PR #1429, opened 2023-10-17
Hardware crypto support

oleorhagen:
Implementation of the HSM support for the cpp client

### DIFF
--- a/src/api/auth.hpp
+++ b/src/api/auth.hpp
@@ -20,10 +20,11 @@
 #include <vector>
 
 #include <common/conf.hpp>
+#include <common/crypto.hpp>
 #include <common/dbus.hpp>
 #include <common/error.hpp>
-#include <common/expected.hpp>
 #include <common/events.hpp>
+#include <common/expected.hpp>
 #include <common/http.hpp>
 #include <common/optional.hpp>
 
@@ -33,11 +34,11 @@ namespace auth {
 
 using namespace std;
 
+namespace crypto = mender::common::crypto;
 namespace dbus = mender::common::dbus;
 namespace error = mender::common::error;
-namespace expected = mender::common::expected;
 namespace events = mender::common::events;
-
+namespace expected = mender::common::expected;
 
 enum AuthClientErrorCode {
 	NoError = 0,
@@ -72,7 +73,7 @@ using AuthenticatedAction = function<void(ExpectedAuthData)>;
 error::Error FetchJWTToken(
 	mender::http::Client &client,
 	const vector<string> &servers,
-	const string &private_key_path,
+	const crypto::Args &args,
 	const string &device_identity_script_path,
 	APIResponseHandler api_handler,
 	const string &tenant_token = "");
--- a/src/api/auth/auth.cpp
+++ b/src/api/auth/auth.cpp
@@ -106,7 +106,7 @@ static void TryAuthenticate(
 error::Error FetchJWTToken(
 	mender::http::Client &client,
 	const vector<string> &servers,
-	const string &private_key_path,
+	const crypto::Args &crypto_args,
 	const string &device_identity_script_path,
 	APIResponseHandler api_handler,
 	const string &tenant_token) {
@@ -133,7 +133,7 @@ error::Error FetchJWTToken(
 		request_body_map.insert({"tenant_token", tenant_token});
 	}
 
-	auto expected_public_key = crypto::ExtractPublicKey(private_key_path);
+	auto expected_public_key = crypto::ExtractPublicKey(crypto_args);
 	if (!expected_public_key) {
 		return expected_public_key.error();
 	}
@@ -147,7 +147,7 @@ error::Error FetchJWTToken(
 
 	// Sign the body
 	auto expected_signature =
-		crypto::SignRawData(private_key_path, common::ByteVectorFromString(request_body));
+		crypto::SignRawData(crypto_args, common::ByteVectorFromString(request_body));
 	if (!expected_signature) {
 		return expected_signature.error();
 	}
@@ -265,12 +265,12 @@ static void TryAuthenticate(
 error::Error FetchJWTToken(
 	mender::http::Client &client,
 	const vector<string> &servers,
-	const string &private_key_path,
+	const crypto::Args &crypto_args,
 	const string &device_identity_script_path,
 	APIResponseHandler api_handler,
 	const string &tenant_token) {
 	return http::FetchJWTToken(
-		client, servers, private_key_path, device_identity_script_path, api_handler, tenant_token);
+		client, servers, crypto_args, device_identity_script_path, api_handler, tenant_token);
 }
 
 void Authenticator::ExpireToken() {
--- a/src/api/auth_test.cpp
+++ b/src/api/auth_test.cpp
@@ -123,7 +123,11 @@ TEST_F(AuthTests, FetchJWTTokenBasicTest) {
 			loop.Stop();
 		};
 	auto err = auth::FetchJWTToken(
-		client, servers, private_key_path, test_device_identity_script, handle_jwt_token_callback);
+		client,
+		servers,
+		{private_key_path},
+		test_device_identity_script,
+		handle_jwt_token_callback);
 
 	loop.Run();
 
@@ -197,7 +201,11 @@ TEST_F(AuthTests, FetchJWTTokenFailoverTest) {
 			loop.Stop();
 		};
 	auto err = auth::FetchJWTToken(
-		client, servers, private_key_path, test_device_identity_script, handle_jwt_token_callback);
+		client,
+		servers,
+		{private_key_path},
+		test_device_identity_script,
+		handle_jwt_token_callback);
 
 	loop.Run();
 
@@ -246,7 +254,11 @@ TEST_F(AuthTests, FetchJWTTokenFailTest) {
 		EXPECT_THAT(resp.error().String(), ::testing::HasSubstr("No more servers"));
 	};
 	auto err = auth::FetchJWTToken(
-		client, servers, private_key_path, test_device_identity_script, handle_jwt_token_callback);
+		client,
+		servers,
+		{private_key_path},
+		test_device_identity_script,
+		handle_jwt_token_callback);
 
 	loop.Run();
 
--- a/src/common/CMakeLists.txt
+++ b/src/common/CMakeLists.txt
@@ -134,6 +134,7 @@ target_compile_options(common_http PRIVATE ${PLATFORM_SPECIFIC_COMPILE_OPTIONS})
 target_link_libraries(common_http PUBLIC
   Boost::beast
   common
+  common_crypto
   common_events
   common_error
   common_log
--- a/src/common/crypto.hpp
+++ b/src/common/crypto.hpp
@@ -40,10 +40,12 @@ enum CryptoErrorCode {
 };
 
 class PrivateKey;
-using ExpectedPrivateKey = expected::expected<unique_ptr<PrivateKey>, error::Error>;
+using ExpectedPrivateKey = expected::expected<PrivateKey, error::Error>;
 
 class PrivateKey {
 public:
+	PrivateKey() {};
+
 	static ExpectedPrivateKey LoadFromPEM(const string &private_key_path, const string &passphrase);
 	static ExpectedPrivateKey LoadFromPEM(const string &private_key_path);
 	static ExpectedPrivateKey Generate(const unsigned int bits, const unsigned int exponent);
@@ -52,7 +54,11 @@ public:
 	};
 	error::Error SaveToPEM(const string &private_key_path);
 #ifdef MENDER_CRYPTO_OPENSSL
-	unique_ptr<EVP_PKEY, void (*)(EVP_PKEY *)> key;
+	unique_ptr<EVP_PKEY, void (*)(EVP_PKEY *)> key {nullptr, [](EVP_PKEY *) { return; }};
+
+	EVP_PKEY *Get() {
+		return key.get();
+	}
 
 private:
 	PrivateKey(unique_ptr<EVP_PKEY, void (*)(EVP_PKEY *)> &&private_key) :
--- a/src/common/crypto.hpp
+++ b/src/common/crypto.hpp
@@ -39,34 +39,57 @@ enum CryptoErrorCode {
 	VerificationError,
 };
 
+struct Args {
+	const string private_key_path;
+	const string private_key_passphrase;
+	const string ssl_engine;
+};
+
 class PrivateKey;
 using ExpectedPrivateKey = expected::expected<PrivateKey, error::Error>;
+
+#ifdef MENDER_CRYPTO_OPENSSL
+class OpenSSLResourceHandle;
+using ResourceHandlePtr = unique_ptr<OpenSSLResourceHandle, void (*)(OpenSSLResourceHandle *)>;
+using PkeyPtr = unique_ptr<EVP_PKEY, void (*)(EVP_PKEY *)>;
+#endif // MENDER_CRYPTO_OPENSSL
 
 class PrivateKey {
 public:
 	PrivateKey() {};
 
-	static ExpectedPrivateKey Load(const string &private_key_path) {
-		return Load(private_key_path, "");
+	static ExpectedPrivateKey Load(const Args &args);
+	static ExpectedPrivateKey LoadFromPEM(const string &private_key_path) {
+		return LoadFromPEM(private_key_path, "");
 	}
-	static ExpectedPrivateKey Load(const string &private_key_path, const string &passphrase);
 	static ExpectedPrivateKey LoadFromPEM(const string &private_key_path, const string &passphrase);
-	static ExpectedPrivateKey LoadFromPEM(const string &private_key_path);
+	static ExpectedPrivateKey LoadFromHSM(const Args &args);
 	static ExpectedPrivateKey Generate(const unsigned int bits, const unsigned int exponent);
 	static ExpectedPrivateKey Generate(const unsigned int bits) {
 		return PrivateKey::Generate(bits, MENDER_DEFAULT_RSA_EXPONENT);
 	};
 	error::Error SaveToPEM(const string &private_key_path);
+
 #ifdef MENDER_CRYPTO_OPENSSL
-	unique_ptr<EVP_PKEY, void (*)(EVP_PKEY *)> key {nullptr, [](EVP_PKEY *) { return; }};
+	PkeyPtr key {nullptr, [](EVP_PKEY *) { return; }};
+
+	ResourceHandlePtr resource_handle_ {nullptr, [](OpenSSLResourceHandle *) { return; }};
 
 	EVP_PKEY *Get() {
 		return key.get();
 	}
 
-private:
-	PrivateKey(unique_ptr<EVP_PKEY, void (*)(EVP_PKEY *)> &&private_key) :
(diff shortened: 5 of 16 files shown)


common/crypto.hpp
```diff
@@ -46,6 +46,10 @@ class PrivateKey {
 public:
 	PrivateKey() {};
 
+	static ExpectedPrivateKey Load(const string &private_key_path) {
+		return Load(private_key_path, "");
+	}
```
lluiscampos: In the `.cpp` file, we have a similar wrapper:
```
ExpectedPrivateKey PrivateKey::LoadFromPEM(const string &private_key_path) {
	return PrivateKey::LoadFromPEM(private_key_path, "");
}
```

Consider moving the definition of `Load` to the cpp or to bring `LoadFromPEM` to the header :pray:

common/crypto.hpp
```diff
@@ -15,6 +15,8 @@
 #ifndef MENDER_COMMON_CRYPTO_HPP
 #define MENDER_COMMON_CRYPTO_HPP
 
+#include <common/crypto/platform/openssl/openssl_config.h>
```
lluiscampos, line 18: ~This cannot be right: the platform agnostic header shouldn't know about openssl.~

Edit: too bad that we need this include for the definitions below. At least it should be guarded with `ifdef MENDER_CRYPTO_OPENSSL`

common/crypto.hpp
```diff
@@ -39,34 +41,67 @@ enum CryptoErrorCode {
 	VerificationError,
 };
 
+struct Args {
+	const string private_key_path;
+	const string private_key_passphrase;
+	const string ssl_engine;
+};
```
lluiscampos, line 48: Good thinking!

common/crypto.hpp
```diff
@@ -39,34 +41,67 @@ enum CryptoErrorCode {
...
+		key(std::move(private_key)),
+		engine {std::move(engine)} {};
+#else
+	PrivateKey(
+		unique_ptr<EVP_PKEY, void (*)(EVP_PKEY *)> &&private_key,
+		unique_ptr<OSSL_PROVIDER, int (*)(OSSL_PROVIDER *)> &&default_provider,
+		unique_ptr<OSSL_PROVIDER, int (*)(OSSL_PROVIDER *)> &&hsm_provider) :
+		key(std::move(private_key)),
+		default_provider {std::move(default_provider)},
+		hsm_provider {std::move(hsm_provider)} {};
```
lluiscampos, line 103: Can we have some alias around?
```
using PkeyPtr = unique_ptr<EVP_PKEY, void (*)(EVP_PKEY *)>;
using ProviderPtr = unique_ptr<OSSL_PROVIDER, int (*)(OSSL_PROVIDER *)>;
...
```

For being a header file, it became quite hard to read :eye:

common/crypto/platform/openssl/crypto.cpp
```diff
@@ -148,8 +160,135 @@ ExpectedPrivateKey PrivateKey::LoadFromPEM(
...
-	return PrivateKey::LoadFromPEM(private_key_path, "");
+#ifdef MENDER_CRYPTO_OPENSSL_LEGACY
+ExpectedPrivateKey PrivateKey::LoadFromHSM(const Args &args) {
+	log::Trace("Loading the private key from HSM");
+
+	ENGINE_load_builtin_engines();
+	auto engine = unique_ptr<ENGINE, void (*)(ENGINE *)>(
+		ENGINE_by_id(args.ssl_engine.c_str()), engine_free_func);
+
+	if (engine == NULL) {
```
lluiscampos, line 171: Elsewhere we are using `nullptr`

common/crypto/platform/openssl/crypto.cpp
```diff
@@ -148,8 +160,135 @@ ExpectedPrivateKey PrivateKey::LoadFromPEM(
...
+				"Unhandled OpenSSL type: expected PrivateKey, got: " + info_type_string));
+			break;
+		}
+	}
+
+	return expected::unexpected(MakeError(
+		SetupError,
+		"Failed to load the private key from the hardware security module: "
+			+ GetOpenSSLErrorMessage()));
+}
```
lluiscampos, line 283: It is really hard to follow. I am glad we have a new OpenSSL expert in the team :sweat_smile: 

I browsed the man pages and your code makes sense, but I am not able to provide any better feedback than that.

mender-auth/cli/actions.cpp
```diff
@@ -45,21 +45,12 @@ shared_ptr<MenderKeyStore> KeystoreFromConfig(
...
 		pem_file = config.security.auth_private_key;
 		ssl_engine = config.security.ssl_engine;
 		static_key = cli::StaticKey::Yes;
-	}
-	if (config.https_client.key == "" && config.security.auth_private_key == "") {
+	} else {
 		pem_file = config.paths.GetKeyFile();
-		ssl_engine = config.https_client.ssl_engine;
 		static_key = cli::StaticKey::No;
 	}
```
lluiscampos, line 55: I think it is not right to associate `StaticKey::No` to everything but `config.security.auth_private_key != ""`. In other words, this sets no static keys always for the https configuration, potentially working incorrectly with hardware crypto for example (it will work fine as-is, because we don't have code to attempt regenerating the https key anyway, but still the design is wrong I think).

Sorry for opening now this melon, but I think this should be a filed exposed in the configuration. It is just too much of a guess to figure this out for the user.

lluiscampos: By the way, thank you for addressing this! :muscle:

lluiscampos (commented):
Brave work!

common/crypto.hpp
oleorhagen: I am also pondering typing some of these, so that it will be harder to mess up the argument order, but might be I'm going overboard here...

common/crypto.hpp
oleorhagen: Yeh, issue is the `PrivateKey` needs to keep the `HSM` resources alive, so the `ENGINE` or the `PROVIDER`.

I have been thinking about how to abstract this away, but I can't think of any good ways atm.

All open to suggestions here ofc

common/crypto/platform/openssl/crypto.cpp
oleorhagen: It's a shame, I have been trying to make this as readable as possible. I could perhaps comment it out more?

mender-auth/cli/actions.cpp
oleorhagen: Discussed f2f, keeping it as it is.

common/crypto.hpp
kacf: One thing which occurred to me after our talk: Do the types need to be qualified? They are just pointers, so can't they be forward declared with `class ENGINE;` and that should be enough?

mender-auth/cli/keystore.cpp
```diff
@@ -70,11 +70,11 @@ error::Error MenderKeyStore::Load() {
 }
 
 error::Error MenderKeyStore::Save() {
-	if (!key_) {
+	if (!key_.Get()) {
```
kacf, line 74: Not wrong, but maybe it's more idiomatic to provide

```
operator bool() {
    return Get() != nullptr;
}
```

in PrivateKey?

common/crypto.hpp
kacf: +1!

This should also help on reading the code where you use the constructors below.

common/crypto/platform/openssl/crypto.cpp
```diff
@@ -74,6 +80,11 @@ auto bn_free = [](BIGNUM *bn) {
 		BN_free(bn);
 	}
 };
+auto engine_free_func = [](ENGINE *e) {
```
kacf, line 83: Nitpick: Regular function syntax here, maybe?

common/crypto/platform/openssl/crypto.cpp
```diff
@@ -148,8 +160,135 @@ ExpectedPrivateKey PrivateKey::LoadFromPEM(
...
+	}
+
+	// Go through all objects in the context till we find the first private key
+	while (not OSSL_STORE_eof(ctx.get())) {
+		OSSL_STORE_INFO *info = OSSL_STORE_load(ctx.get());
+
+		if (info == NULL) {
+			return expected::unexpected(MakeError(
+				SetupError,
+				"Failed to read the private key information the hardware security module: "
```
kacf, line 249: "...**from** the hardware security module"

common/crypto/platform/openssl/crypto.cpp
```diff
@@ -148,8 +160,135 @@ ExpectedPrivateKey PrivateKey::LoadFromPEM(
...
+			"OSSL_STORE_OPEN: Failed to load the private key from the hardware security module: "
+				+ GetOpenSSLErrorMessage()));
+	}
+
+	// Go through all objects in the context till we find the first private key
+	while (not OSSL_STORE_eof(ctx.get())) {
+		OSSL_STORE_INFO *info = OSSL_STORE_load(ctx.get());
+
+		if (info == NULL) {
+			return expected::unexpected(MakeError(
```
kacf, line 247: Or should we try the next key here with `continue`?

This goes for several `returns` below as well.

common/http/platform/beast/http.cpp
```diff
@@ -270,11 +272,19 @@ error::Error Client::Initialize() {
...
+				return exp_key.error().WithContext(
+					"Error loading private key from " + client_config_.client_cert_key_path);
+			}
+
+			const int ret =
+				SSL_CTX_use_PrivateKey(ssl_ctx_[i].native_handle(), exp_key.value().Get());
+			if (ret != 1) {
+				log::Error(
+					"Failed to add the PrivateKey: " + client_config_.client_cert_key_path
+					+ " to the SSL CTX. The HTTP client will not be functioning normally");
```
kacf, line 287: I think we should return error here as well, we have opted for returns on all other error paths.

mender-auth/cli/actions.cpp
```diff
@@ -97,7 +88,12 @@ error::Error DoAuthenticate(context::MenderContext &main_context) {
 	auto err = auth_client::FetchJWTToken(
 		client,
 		config.servers,
-		config.paths.GetKeyFile(),
+		{
+			config.security.auth_private_key == "" ? config.paths.GetKeyFile()
+												   : config.security.auth_private_key,
+			"",
```
kacf, line 94: What about passphrase?

mender-auth/cli/actions.cpp
```diff
@@ -151,8 +147,10 @@ error::Error DaemonAction::Execute(context::MenderContext &main_context) {
 	ipc::Server ipc_server {loop, config};
 
 	err = ipc_server.Listen(
-		config.security.auth_private_key == "" ? config.paths.GetKeyFile()
-											   : config.security.auth_private_key,
+		{config.security.auth_private_key == "" ? config.paths.GetKeyFile()
+												: config.security.auth_private_key,
+		 "",
```
kacf, line 152: What about passphrase?

mender-auth/cli/keystore.cpp
oleorhagen: Good suggestion

common/crypto.hpp
oleorhagen: I was pondering that all evening last night!

I'm not sure if it's a good idea though, is it considered kosher?

mender-auth/cli/actions.cpp
oleorhagen: Yes, glad you pointed it out.

Not sure what to do about it atm. Do we currently have any support for it? I can't find it.

This could be a good opportunity for adding it though

common/crypto.hpp
kacf: Yes, this is quite common, actually. It doesn't apply to us, but in public library development there is an important function to doing so. If your private struct is exposed in the header, there is a chance it could, maybe unintentionally, be used directly in the source files of whoever is using your library. This is bad, because it binds the memory layout of that struct to the binary that was built, instead of to the library. This means you can no longer update the struct, because you risk breaking compatibility with already released binaries. OTOH, if you only forward declare it, then any such unintentional use would be flagged by the compiler as illegal, and the users would be forced to use the factory functions you provide to get a pointer to the struct.

mender-auth/cli/actions.cpp
kacf: It's being set [here](https://github.com/mendersoftware/mender/pull/1429/files#diff-3540d3121deba3ec9ef7262bf0093173c7717d1bfc172970cfe1a527644ac9e4R75) by command line options.

oleorhagen: lol, of course!

common/crypto/platform/openssl/crypto.cpp
oleorhagen: I _think_ that it will not hurt, but in general I would expect people to be so specific in the `URL` that they choose the correct index, but maybe not. 

Continuing

common/crypto/platform/openssl/crypto.cpp
```diff
@@ -50,6 +55,30 @@ using namespace std;
...
+#else
+class OpenSSLResourceHandle {
+public:
+	ProviderPtr default_provider_;
+	ProviderPtr hsm_provider_;
+};
+#endif // MENDER_CRYPTO_OPENSSL_LEGACY
+auto resource_handle_free_func = [](OpenSSLResourceHandle *h) {
+	if (h) {
+		h->~OpenSSLResourceHandle();
```
kacf, line 78: Need to use `delete h` here, to pair it with the `new` below, otherwise the memory of `OpenSSLResourceHandle` itself won't be freed.

But calling `delete` is the default, so actually, you don't need a custom deleter at all.

common/crypto/platform/openssl/crypto.cpp
```diff
@@ -147,11 +182,135 @@ ExpectedPrivateKey PrivateKey::LoadFromPEM(
...
+				+ GetOpenSSLErrorMessage()));
+	}
+	log::Debug("Successfully initialised the HSM engine");
+
+	auto private_key = unique_ptr<EVP_PKEY, void (*)(EVP_PKEY *)>(
+		ENGINE_load_private_key(
+			engine.get(),
+			args.private_key_path.c_str(),
+			(UI_METHOD *) NULL,
+			NULL /*callback_data */),
```
kacf, line 217: Should use `nullptr`.

common/crypto/platform/openssl/crypto.cpp
```diff
@@ -147,11 +182,135 @@ ExpectedPrivateKey PrivateKey::LoadFromPEM(
...
+
+	// Go through all objects in the context till we find the first private key
+	while (not OSSL_STORE_eof(ctx.get())) {
+		OSSL_STORE_INFO *info = OSSL_STORE_load(ctx.get());
+
+		if (info == NULL) {
+			log::Error(
+				"Failed to load the store info from the hardware security module: trying the next object in the context: "
+				+ GetOpenSSLErrorMessage());
+		}
```
kacf, line 272: Missing `continue`? I see there is one below, but that's after `info` has already been used, so is it correct?

common/crypto/platform/openssl/crypto.cpp
```diff
@@ -147,11 +182,135 @@ ExpectedPrivateKey PrivateKey::LoadFromPEM(
...
+			SetupError,
+			"OSSL_STORE_OPEN: Failed to load the private key from the hardware security module: "
+				+ GetOpenSSLErrorMessage()));
+	}
+
+	// Go through all objects in the context till we find the first private key
+	while (not OSSL_STORE_eof(ctx.get())) {
+		OSSL_STORE_INFO *info = OSSL_STORE_load(ctx.get());
+
+		if (info == NULL) {
```
kacf, line 268: Use `nullptr`.

mender-auth/cli/keystore.hpp
```diff
@@ -68,12 +68,22 @@ class MenderKeyStore {
 	error::Error Save();
 	error::Error Generate();
 
+	string KeyName() {
+		return key_name_;
+	}
+	string SSLEngine() {
+		return ssl_engine_;
+	};
+	string PassPrhase() {
```
kacf, line 77: Spelling: `PassPhrase`

common/crypto/platform/openssl/crypto.cpp
oleorhagen: Oh.. true.

But I do need a custom deleter, if I am to use this way of hiding the resource handle definition in the `cpp` file, for otherwise the unique_ptr requires knowledge of the `OpenSSLResourceHandle` struct.

So I guess question is. Is it enough to call delete on `h` ? Or do i also need to explictly call the destructor?

common/crypto/platform/openssl/crypto.cpp
oleorhagen: I've seen that in other work we have used `NULL` when interacting with `C`, so I stuck to that here. 
I can change it, but just so we are clear

common/crypto/platform/openssl/crypto.cpp
oleorhagen: The other continue is for the case where we don't get a `private-key` from the `info` struct, but some other type.

common/crypto/platform/openssl/crypto.cpp
oleorhagen: Same as above

common/crypto/platform/openssl/crypto.cpp
kacf: Right. It's probably because we copied it from C examples or whatever. I'd use `nullptr` since it has an important type safety role.

common/crypto/platform/openssl/crypto.cpp
kacf: Yes, `delete` is enough, it both calls the destructor, and frees the memory.

However, I suspect you will run into the same problem, and I have a hunch why: Since this is a simple object, the destructor is (implicitly) defined in the header, where the full type is not available. Try to declare `~OpenSSLResourceHandle();` in the header, and put an empty one in the cpp file, which I think should be enough to move its definition there, avoiding the need for a full type.

I think...

common/crypto/platform/openssl/crypto.cpp
kacf: Ok, so we do need it here as well?

common/crypto/platform/openssl/crypto.cpp
oleorhagen: Seems `delete` does both :raised_hands:  :)

common/crypto/platform/openssl/crypto.cpp
oleorhagen: Yes :) thank you!

common/crypto/platform/openssl/crypto.cpp
oleorhagen: > However, I suspect you will run into the same problem, and I have a hunch why: Since this is a simple object, the destructor is (implicitly) defined in the header, where the full type is not available. Try to declare ~OpenSSLResourceHandle(); in the header, and put an empty one in the cpp file, which I think should be enough to move its definition there, avoiding the need for a full type.

> I think...

The compiler kept asking for `sizeof(OpenSSLResourceHandle)`, so I don't think we are as lucky :/

lluiscampos (approved):
It looks good from my side. Nothing more to add :+1:

kacf (approved):
Thanks!

PR #1430, opened 2023-10-17
CI job for compatibility build + amend code for Boost 1.67

lluiscampos:


oleorhagen (approved):

.gitlab-ci.yml
```diff
@@ -42,6 +42,30 @@ test:
     paths:
       - coverage.lcov
 
+backwards-compat:
+  stage: test
+  image: debian:10
```
kacf, line 47: Should put a comment here that this should always be oldest LTS.

oleorhagen (approved):

.gitlab-ci.yml
```diff
@@ -42,6 +42,30 @@ test:
...
+  before_script:
+    - apt update && apt install -yyq g++ git make lcov pkg-config liblmdb++-dev libboost-dev libboost-log-dev libboost-regex-dev libssl-dev libarchive-dev libdbus-1-dev curl dbus
```
kacf, line 49: Can you add `stunnel4 tinyproxy-bin` right away? Then we won't get a mismerge with #1433.

.gitlab-ci.yml
```diff
@@ -42,6 +42,30 @@ test:
     paths:
       - coverage.lcov
 
+backwards-compat:
+  stage: test
```
kacf, line 46: This is build only, right? Should it be named as such? Or moved to a different stage?

kacf (approved):

.gitlab-ci.yml
lluiscampos: This job is meant to only build, not to run tests, so it shouldn't be necessary.

kacf: Doh! Of course. As pointed out by me also ten seconds later! 😄

.gitlab-ci.yml
lluiscampos: I think it belongs to "test" stage in the sense that it tests backwards compatibility. We only want to make sure that it compiles (if the tests cannot run in old LTS we can live with that).

I will name it `test:backward-compat` for consistency with the test stage.

PR #1432, opened 2023-10-17
fix(auth): Attempt to authenticate to all servers from config

vpodzime:
TODO:
- [x] extra unit tests for some failing
- [x] extra unit tests for all failing

Ticket: MEN-6468
Changelog: none


vpodzime:
@kacf with these changes I think it's time to get rid of `MenderConfigFromFile.server_url`. Do I recall correctly that you already have some commit for that?

kacf (approved):
Nice work! I've got no comments!

> with these changes I think it's time to get rid of `MenderConfigFromFile.server_url`. Do I recall correctly that you already have some commit for that?

Yes, it's in https://github.com/mendersoftware/mender/pull/1415.

PR #1433, opened 2023-10-18
Http proxy support

kacf:


common/http/http.cpp
```diff
@@ -112,6 +112,12 @@ error::Error BreakDownUrl(const string &url, BrokenDownUrl &address) {
 		address.path = tmp.substr(split_index);
 	}
 
+	if (address.host.find("@") != string::npos) {
+		return error::Error(
+			make_error_condition(errc::not_supported),
+			"URL Username and password is not supported");
+	}
```
vpodzime, line 121: I know it shouldn't really happen, but this makes me think about `address` being half- or even all-set when this error is returned. And thus can then easily be used for requests. While I have some doubts that all the code that calls this (or `SetAddress()`) actually checks the returned value (error) and clears the fields in case of error.

common/conf.hpp
```diff
@@ -233,6 +235,8 @@ class MenderConfig : public cfg_parser::MenderConfigFromFile {
 public:
 	Paths paths {};
 
+	http::ClientConfig http_client_config;
```
vpodzime, line 238: Wouldn't it be better to add `http::ClientConfig GetClientConfig()`? What if something changes some value the config is based on **after** `ProcessCmdlineArgs()`?

vpodzime (commented):
Suggestions for the first 3 commits. I'll get to the actual meat here later.

common/http/http.cpp
kacf: Good point. I'll add this to all the error returns.

common/conf.hpp
kacf: I suppose that's good future proofing, yes. I'll switch to that.

common/conf.hpp
```diff
@@ -273,8 +275,14 @@ class MenderConfig : public cfg_parser::MenderConfigFromFile {
 		vector<string>::const_iterator end,
 		const CliApp &app);
 
+	const http::ClientConfig &GetHttpClientConfig() const {
```
oleorhagen, line 278: Q: why `const http::ClientConfig` ? Did we not have a conversation about this not making a difference before? :thinking:

oleorhagen: Also, good idea!

common/http/platform/beast/http.cpp
```diff
@@ -235,28 +235,34 @@ Client::Client(
...
-		ssl_ctx_.use_certificate_file(client.client_cert_path, boost::asio::ssl::context_base::pem);
-		ssl_ctx_.use_private_key_file(
-			client.client_cert_key_path, boost::asio::ssl::context_base::pem);
-	}
+	for (auto i = 0; i < MENDER_BOOST_BEAST_SSL_CTX_COUNT; i++) {
+		ssl_ctx_[i].set_verify_mode(client.skip_verify ? ssl::verify_none : ssl::verify_peer);
+
+		if (client.client_cert_path != "" and client.client_cert_key_path != "") {
+			ssl_ctx_[i].set_options(boost::asio::ssl::context::default_workarounds);
+			ssl_ctx_[i].use_certificate_file(
```
oleorhagen, line 250: Note that this will throw an exception in the case of a missing file. 

Same below

common/http/platform/beast/http.cpp
```diff
@@ -353,14 +438,21 @@ io::ExpectedAsyncReadWriterPtr Client::SwitchProtocol(IncomingResponsePtr req) {
...
 			stream, response_data_.response_buffer_);
-	} else {
+	case SocketMode::Tls:
+		return make_shared<RawSocket<ssl::stream<tcp::socket>>>(
+			make_shared<ssl::stream<tcp::socket>>(std::move(stream->next_layer())),
+			response_data_.response_buffer_);
+	case SocketMode::Plain:
 		return make_shared<RawSocket<tcp::socket>>(
-			make_shared<tcp::socket>(std::move(stream->next_layer())),
+			make_shared<tcp::socket>(std::move(stream->next_layer().next_layer())),
```
oleorhagen, line 451: Q: `get_lowest_layer()` ?

oleorhagen (approved):
Phew, our `http` code is something else at this point :see_no_evil:

common/conf.hpp
kacf: Yes, but it's a reference, not a value, and then it makes a difference.

Using `const &` is actually a micro-optimization, and not really critical. My intention here was to return plain `http::ClientConfig`, no const, no reference. However, because it happens to be a member of the class, and we don't need to construct a new one, we can return a const reference as an alternative. As long as the caller assigns it to a new struct, this makes no practical difference, it just saves a few CPU cycles.

The only case where it would make a difference is if the caller assigns it to a reference. If we later changed it to return plain `http::ClientConfig`, which we would have to if the function needs to construct the data, then this wouldn't work anymore.

So yeah, I could change to plain `http::ClientConfig` since it is, in an API sense, more correct. I'm fine with either.

common/http/platform/beast/http.cpp
kacf: I tried, but this doesn't work, because there is a hidden fourth layer inside `tcp::socket`, `boost::asio::basic_stream_socket`.

common/http/platform/beast/http.cpp
kacf: Pre-existing, but indeed, good catch! I'll fix both of them.

common/http_proxy_test.cpp
```diff
@@ -0,0 +1,913 @@
...
+	EXPECT_NE(error::NoError, err);
+	EXPECT_EQ(err.code, http::MakeError(http::InvalidUrlError, "").code) << err.String();
+
+	err = test("https", http::ClientConfig {.https_proxy = "bogus"});
+	EXPECT_NE(error::NoError, err);
+	EXPECT_EQ(err.code, http::MakeError(http::InvalidUrlError, "").code) << err.String();
+
+	err = test("https", http::ClientConfig {.https_proxy = "http://localhost/a-path"});
+	EXPECT_NE(error::NoError, err);
+	EXPECT_EQ(err.code, http::MakeError(http::InvalidUrlError, "").code) << err.String();
```
lluiscampos, line 342: We don't have any test setting `.no_proxy`, can we add one?

lluiscampos (approved):
It looks good!

I haven't read the unit tests in detail, though :see_no_evil:

common/http/http.cpp
```diff
@@ -423,5 +425,39 @@ ExponentialBackoff::ExpectedInterval ExponentialBackoff::NextInterval() {
 	return current_interval;
 }
 
+bool HostNameMatchesNoProxy(const string &host, const string &no_proxy) {
```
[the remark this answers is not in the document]
kacf: It's *much* cleaner, actually! I went looking for `std::string::split`, but I forgot to check our own `common.hpp`.

common/http/platform/beast/http.cpp
```diff
@@ -321,6 +329,83 @@ error::Error Client::AsyncCall(
...
+			request_->SetMethod(Method::CONNECT);
+			BrokenDownUrl proxy_address;
+			auto err = BreakDownUrl(https_proxy_, proxy_address);
+			if (err != error::NoError) {
+				return err.WithContext("HTTPS proxy URL is invalid");
+			}
+			if (proxy_address.path != "" && proxy_address.path != "/") {
+				return MakeError(
+					InvalidUrlError, "A URL with a path is not legal for a proxy address");
+			}
```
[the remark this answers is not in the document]
kacf: shared above...? Am I missing something?

common/http_proxy_test.cpp
kacf: No, because any value is valid there, at least the way we have implemented it. I guess you could say that `..myhost` is invalid, but without a proper standard it's hard to decide what to do with it other than treat it as a wildcard suffix with value `.myhost`.

common/http/http.cpp
```diff
@@ -425,6 +425,45 @@ ExponentialBackoff::ExpectedInterval ExponentialBackoff::NextInterval() {
...
+		return getenv(primary.c_str());
+	} else if (secondary_set) {
+		return getenv(secondary.c_str());
+	} else {
+		return "";
+	}
+}
+
+// The proxy variables aren't standardized, but this page was useful for the common patterns:
+// https://superuser.com/questions/944958/are-http-proxy-https-proxy-and-no-proxy-environment-variables-standard
```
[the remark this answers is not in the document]
kacf: Sneaky! I'll add detection of this too, even though it is unlikely to affect us.

common/http/platform/beast/http.cpp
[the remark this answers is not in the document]
lluiscampos: I mean that this 9 lines are the exact same as above in https://github.com/mendersoftware/mender/pull/1433/files#diff-7100548573dbc4eefb8ff288d50eb33b3c038a36e5573de55fa670b1f0221c46R380-R388 -> we could refactor in a tiny helper function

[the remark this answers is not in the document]
kacf: It's not that easy. Some parts are the same, but there are subtle differences sprinkled all over it: Several references to "https" instead of "http", the lower path switches out the `request_` member, and the path construction is different. It wouldn't be impossible to break some of it out, but all ways that I've found significantly hurt readability.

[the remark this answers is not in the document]
lluiscampos: You are right :see_no_evil:

common/http/platform/beast/http.cpp
oleorhagen: :see_no_evil:

common/conf.hpp
oleorhagen: Alright, makes sense. Also fine with either

common/http/platform/beast/http.cpp
```diff
@@ -235,28 +235,34 @@ Client::Client(
 	const ClientConfig &client, events::EventLoop &event_loop, const string &logger_name) :
 	event_loop_ {event_loop},
 	logger_name_ {logger_name},
+	http_proxy_ {client.http_proxy},
+	https_proxy_ {client.https_proxy},
+	no_proxy_ {client.no_proxy},
```
vpodzime, line 241: `no_proxy_` is not used anywhere, it's only set here.

common/http/http.cpp
vpodzime: Some description of what this function does and the expected inputs look would be really nice.

vpodzime (commented):
Late to the party, but at least the comment regarding `no_proxy_` could be important.

common/http/http.cpp
kacf: I simply copied the comment from about 15 lines up. It's meant to cover all four of these functions, but I guess it's not so clear since this function has a different name.

common/http/platform/beast/http.cpp
kacf: Very good catch! This is indeed a bug. And a good example of why unit tests != integration tests.

PR #1435, opened 2023-10-18
test: Fix suboptimal job count when building.

kacf:
`--load-average` causes it to oscillate between full capacity (which rapidly raises load average) and far below capacity (which lowers load average again). This causes the overall execution time to be much slower than just using all the CPUs constantly.

Probably this was a typo, and using all CPUs is what was intended all along.

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -32,7 +32,7 @@ test:
     - chmod +x /usr/local/bin/mender-artifact
   script:
     - cmake -D COVERAGE=ON .
-    - make -j --load-average=$(nproc --all) --keep-going coverage
+    - make --jobs=$(nproc --all) --keep-going coverage
   tags:
     - mender-qa-worker-generic
   artifacts:
@@ -63,7 +63,7 @@ test:backward-compat:
     - apt update & apt install -yyq --allow-unauthenticated cmake
   script:
     - cmake .
-    - make -j --load-average=$(nproc --all) --keep-going
+    - make --jobs=$(nproc --all) --keep-going
   tags:
     - mender-qa-worker-generic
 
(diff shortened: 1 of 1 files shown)


vpodzime (commented):
This wasn't a typo, actually. On my system, for example, `--load-average` performs better because it sometimes starts even more jobs. And we had a discussion about this @oleorhagen

vpodzime:
Oh, and another aspect is that in container environments, it actually respects the load on the system. That's why it's easily slower in our CI.

oleorhagen:
Did we time it in CI at some point @kacf ?

kacf:
No, but I don't really see how `load-average` could be faster. With `--job` we use 100% of the capacity, all the time. Launching more jobs than that isn't going to improve the situation (it could for I/O intensive work, but I don't think it applies here).

It was *significantly* slower in the container on my system, by a factor of 2 at least.

vpodzime:
> No, but I don't really see how `load-average` could be faster.

It can be faster if N jobs are not enough for your N CPUs. :wink: And that can easily happen. Just like it can happen that a system is already under a heavy load when building. There's no silver bullet.

kacf:
"Can happen", maybe, but this doesn't agree with current evidence. And it's not what I would expect either, since CPU is the main resource being used.

> Just like it can happen that a system is already under a heavy load when building

But why does that matter? You just commit to 100% anyway, and the kernel might give you less, but the performance will not be significantly affected (a bit more context switching, but I doubt this will make any practical difference).

In any case, I think the debate over whether `load-average` is a better measure or not, is not the main issue here. The main issue is that `--load-average` is buggy, because it oscillates between extreme commitment, and low commitment, and every time the latter happens, the CPUs are mostly idle for the next 10-15 seconds or so. This keeps repeating in cycles.

vpodzime:
> In any case, I think the debate over whether `load-average` is a better measure or not, is not the main issue here. The main issue is that `--load-average` is buggy, because it oscillates between extreme commitment, and low commitment, and every time the latter happens, the CPUs are mostly idle for the next 10-15 seconds or so. This keeps repeating in cycles.

On my system (with 4 CPU cores), `-j -l4` results in slightly faster builds than `-j4`. Not sure if you have a buggy `make` or what's going on, but I don't see the symptoms you describe.

PR #1437, opened 2023-10-19
chore: Proxy HTTP requests from clients through mender-auth.

kacf:
Also, in the tests, `TEST_PORT_2` seems to not be necessary (never used together with `TEST_PORT`), so just use the latter.

Ticket: MEN-6664

### DIFF
--- a/mender-auth/ipc/server.cpp
+++ b/mender-auth/ipc/server.cpp
@@ -59,19 +59,7 @@ error::Error Caching::Listen(const string &private_key_path, const string &ident
 				servers_,
 				private_key_path,
 				identity_script_path == "" ? default_identity_script_path_ : identity_script_path,
-				[this](auth_client::APIResponse resp) {
-					auth_in_progress_ = false;
-					CacheAPIResponse(resp);
-					if (resp) {
-						dbus_server_.EmitSignal<dbus::StringPair>(
-							"/io/mender/AuthenticationManager",
-							"io.mender.Authentication1",
-							"JwtTokenStateChange",
-							dbus::StringPair {resp.value().token, resp.value().server_url});
-					} else {
-						log::Error("Failed to fetch new token: " + resp.error().String());
-					}
-				},
+				[this](auth_client::APIResponse resp) { FetchJwtTokenHandler(resp); },
 				tenant_token_);
 			if (err != error::NoError) {
 				log::Error("Failed to trigger token fetching: " + err.String());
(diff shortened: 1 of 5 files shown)


mender-auth/ipc/server.cpp
```diff
@@ -84,6 +72,35 @@ error::Error Caching::Listen(const string &private_key_path, const string &ident
...
+void Caching::FetchJwtTokenHandler(auth_client::APIResponse &resp) {
+	auth_in_progress_ = false;
+
+	forwarder_.Cancel();
+
+	if (resp) {
+		// ":0" port number means pick random port in user range.
+		auto err = forwarder_.AsyncForward("http://127.0.0.1:0", resp.value().server_url);
+		if (err == error::NoError) {
+			Cache(resp.value().token, forwarder_.GetUrl());
```
kacf, line 84: @vpodzime: The cache still contains `cached_server_url_`, but now it's the URL of the internal forwarder. It makes sense programmatically, but I'm a bit undecided if I'm muddying up the semantics, and making it unclear this way. What do you think?

oleorhagen (approved):

mender-auth/ipc/server.hpp
```diff
@@ -29,21 +29,26 @@
...
 namespace events = mender::common::events;
 namespace log = mender::common::log;
 namespace path = mender::common::path;
 
+namespace auth_client = mender::api::auth;
+
+namespace http_forwarder = mender::auth::http_forwarder;
+
 class Caching {
 public:
```
vpodzime, line 53: I guess now it's all too obvious this class should no longer be called `Caching`. We refer to it as `Server` anyway (see `using Server...` below) and it's much more than just a caching layer for auth data.

mender-auth/ipc/server.cpp
vpodzime: I don't know the story behind this all. But intuitively, I'd keep the old behavior in a class and then made a subclass with a proxy forwarder and serving its URL as the server URL for clients. But if we want to have everything in one class, it makes most sense to just cache the local URL here like your code does.

vpodzime (commented):
Looks good to me otherwise.

mender-auth/ipc/server.cpp
kacf: I think using two classes here is a bit overkill, since we only need the functionality of one of them. In that case I'll leave it as it is.

mender-auth/ipc/server.hpp
kacf: True, I went for `AuthenticatingForwarder`.

vpodzime (approved):
ACK

PR #1441, opened 2023-10-19
Fix Artifact EC verification

oleorhagen:


common/crypto/platform/openssl/crypto.cpp
```diff
@@ -430,6 +439,97 @@ expected::ExpectedString SignRawData(
...
+	if (s == NULL) {
+		return expected::unexpected(MakeError(
+			SetupError,
+			"Failed to extract the s(ignature) part from the ECDSA signature in the binary representation"
+				+ GetOpenSSLErrorMessage()));
+	}
+
+	// Set the r&s values in the SIG struct
+	// r & s now owned by ecSig
+	int ret {ECDSA_SIG_set0(ecSig.get(), r.release(), s.release())};
```
kacf, line 486: Possible subtle gotcha here: If `ECDSA_SIG_set0` fails, I assume that it will not take ownership. So you should actually use `.get()` here, and then do `.release()` after the check below instead. Thanks to @lluiscampos for making me aware of this in an earlier PR.

common/crypto/platform/openssl/crypto.cpp
```diff
@@ -430,6 +439,97 @@ expected::ExpectedString SignRawData(
...
+	if (ret != OPENSSL_SUCCESS) {
+		return expected::unexpected(MakeError(
+			SetupError,
+			"Failed to set the signature parts in the ECDSA structure" + GetOpenSSLErrorMessage()));
+	}
+
+	/* Allocate some array guaranteed to hold the DER-encoded structure */
+	vector<uint8_t> arr(256);
```
kacf, line 494: Could we give a name to the value?

common/crypto/platform/openssl/crypto.cpp
```diff
@@ -430,6 +439,97 @@ expected::ExpectedString SignRawData(
...
 }
 
+const size_t ecdsa256curveBits = 256;
+const size_t ecdsa256keySize = 32;
+
+// Try and decode the keys from pure binary, assuming that the points on the
+// curve (r,s), have been concatenated together (r || s), and simply dumped to
+// binary. Which is what we did in the `mender-artifact` tool.
+// (See MEN-1740) for some insight into previous issues, and the chosen fix.
+template <BIGNUM *DecoderFn(const unsigned char *, int, BIGNUM *)>
```
kacf, line 449: Note that since this is a function pointer and not a type, you can save binary size by passing it as an argument instead.

common/crypto/platform/openssl/crypto.cpp
```diff
@@ -472,11 +571,17 @@ expected::ExpectedBool VerifySignData(
 	ret = EVP_PKEY_verify(
 		pkey_signer_ctx.get(), signature.data(), signature.size(), shasum.data(), shasum.size());
 	if (ret < 0) {
+		const string openssl_error_msg {GetOpenSSLErrorMessage()};
+		if (openssl_error_msg.find("asn1 encoding") != string::npos) {
```
kacf, line 575: Oh man, is this the best way? They don't provide an error code for it?

I'm wondering if this might be unsafe. Imagine the scenario if openssl is upgraded to a version where this string has subtly changed. Then suddenly the client will not successfully validate any artifact anymore. And you don't know this until you try to do the *next* update. Maybe it's better if we always try?

oleorhagen: First thing I did was copy-paste the error code.

Guess what happened next :see_no_evil: :crying_cat_face: :sob: 

But I agree, maybe they have some error type or something

common/crypto/platform/openssl/crypto.cpp
oleorhagen: `a`? :laughing:

common/crypto/platform/openssl/crypto.cpp
oleorhagen: Uhhh, very good point! :grin:

common/crypto/platform/openssl/crypto.cpp
oleorhagen: haha, shite. It's true, and I was so pleased with my template :roll_eyes:

common/crypto/platform/openssl/crypto.cpp
```diff
@@ -430,6 +439,97 @@ expected::ExpectedString SignRawData(
...
+	auto ecSig = unique_ptr<ECDSA_SIG, void (*)(ECDSA_SIG *)>(ECDSA_SIG_new(), ECDSA_SIG_free);
+	if (ecSig == nullptr) {
+		return expected::unexpected(MakeError(
+			SetupError,
+			"Failed to allocate the structure for the ECDSA signature" + GetOpenSSLErrorMessage()));
+	}
+
+	auto r = unique_ptr<BIGNUM, void (*)(BIGNUM *)>(
+		DecoderFn(signature.data(), ecdsa256keySize, NULL), BN_free);
+	if (r == NULL) {
```
lluiscampos, line 469: Inconsistent use of `nullptr` and `NULL`.

common/crypto/platform/openssl/crypto.cpp
lluiscampos: Indeed, the template use here ~is~ was very neat

common/crypto/platform/openssl/crypto.cpp
```diff
@@ -472,11 +571,17 @@ expected::ExpectedBool VerifySignData(
 	ret = EVP_PKEY_verify(
 		pkey_signer_ctx.get(), signature.data(), signature.size(), shasum.data(), shasum.size());
 	if (ret < 0) {
```
lluiscampos, line 573: Pre-existing, and basically everywhere in the file, but shouldn't these be:
```suggestion
	if (ret < OPENSSL_SUCCESS) {
```
?

common/crypto/platform/openssl/crypto.cpp
```diff
@@ -472,11 +571,17 @@ expected::ExpectedBool VerifySignData(
...
+		const string openssl_error_msg {GetOpenSSLErrorMessage()};
+		if (openssl_error_msg.find("asn1 encoding") != string::npos) {
+			log::Debug(
+				"Failed to verify the signature with the supported OpenSSL binary formats. Falling back to the custom Mender encoded binary format for ECDSA signatures");
+			return VerifyECDSASignData(public_key_path, shasum, signature);
+		}
 		return expected::unexpected(MakeError(
 			VerificationError,
-			"Failed to verify signature. OpenSSL PKEY verify failed: " + GetOpenSSLErrorMessage()));
+			"Failed to verify the new der encoded signature. OpenSSL PKEY verify failed: "
```
lluiscampos, line 582: A bit difficult to read, what does "new" mean in this context? The signature did not validate and we didn't try the fallback method neither. At least use `DER` instead:
```suggestion
			"Failed to verify the DER encoded signature. OpenSSL PKEY verify failed: "
```

But also, it should be `DER/PEM` ?

lluiscampos:
Ni commit message, by the way :memo:

common/crypto/platform/openssl/crypto.cpp
oleorhagen: After a long weekend, I've settled on `der_encoded_byte_array`

common/crypto/platform/openssl/crypto.cpp
oleorhagen: No, `OPENSSL_SUCCESS` is `1`.

ret is an error if `< 0`. Verification failed if `0`, and succeeded if `1`

But this should obviously be communicated better.

common/crypto/platform/openssl/crypto.cpp
oleorhagen: In OpenSSL 3.x we get no error codes during verification, so I will fall back to ECDSA in case of all errors here.

common/crypto/platform/openssl/crypto.cpp
vpodzime: > I assume that it will not take ownership.

I was trying to check this in the man page, but it doesn't mention what the exact behavior is. Not sure if a potential leak here is not better than a potential double-free.

common/crypto/platform/openssl/crypto.cpp
```diff
@@ -430,6 +439,97 @@ expected::ExpectedString SignRawData(
...
+	const vector<uint8_t> &signature);
+
+static expected::ExpectedBool VerifyECDSASignData(
+	const string &public_key_path,
+	const mender::sha::SHA &shasum,
+	const vector<uint8_t> &signature) {
+	expected::ExpectedBytes exp_der_encoded_signature =
+		TryASN1EncodeMenderCustomBinaryECFormat<BN_bin2bn>(signature, shasum)
+			.or_else([&signature, &shasum](error::Error) {
+				return TryASN1EncodeMenderCustomBinaryECFormat<BN_lebin2bn>(signature, shasum);
```
vpodzime, line 521: I guess we should log the error? And maybe even with various log levels based on the actual error? (might be easier to do inside the `TryANS1...()` function)

common/crypto/platform/openssl/crypto.cpp
```diff
@@ -21,11 +21,20 @@
 #include <vector>
 #include <memory>
 
+
+#ifdef MENDER_CRYPTO_OPENSSL_LEGACY
+#include <openssl/engine.h>
+#else
+#include <openssl/provider.h>
+#include <openssl/store.h>
+#endif // MENDER_CRYPTO_OPENSSL_LEGACY
```
vpodzime, line 30: These (the above) don't seem to be required for the code below. At least not for the new one.

vpodzime (commented):
My two cents. Good job overall!

common/crypto/platform/openssl/crypto.cpp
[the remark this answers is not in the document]
oleorhagen: I changed it to take a function argument instead of a template now, and then it's still `DecoderFn`. But maybe we should name it something else?

common/crypto/platform/openssl/crypto.cpp
oleorhagen: The `manpage` does this in the example:
```
     if (!ECDSA_SIG_set0(obj, rbn, sbn))
            /* error */
        /* Set these to NULL since they are now owned by obj */
        rbn = sbn = NULL;
```

oleorhagen: So I guess releasing after is the way to go

common/crypto/platform/openssl/crypto.cpp
oleorhagen: How about we add the error to the last `TryASN1` call, and then the receiver logs it as it sees fit? :thinking:

common/crypto/platform/openssl/crypto.cpp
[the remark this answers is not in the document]
vpodzime: It's probably enough to give the arguments names (and potentially explain what the function is supposed to do).

common/crypto/platform/openssl/crypto.cpp
vpodzime: If there are multiple errors, they should all end up somewhere, IMHO.

common/crypto/platform/openssl/crypto.cpp
kacf: Looks good!

kacf (approved):

common/crypto/platform/openssl/crypto.cpp
oleorhagen: Yes, that's what I think too

common/crypto/platform/openssl/crypto.cpp
oleorhagen: ahh, hahaha.

common/crypto/platform/openssl/crypto.cpp
```diff
@@ -430,6 +432,122 @@ expected::ExpectedString SignRawData(
...
+
+static expected::ExpectedBool VerifyECDSASignData(
+	const string &public_key_path,
+	const mender::sha::SHA &shasum,
+	const vector<uint8_t> &signature) {
+	expected::ExpectedBytes exp_der_encoded_signature =
+		TryASN1EncodeMenderCustomBinaryECFormat(signature, shasum, BN_bin2bn)
+			.or_else([&signature, &shasum](error::Error big_endian_error) {
+				log::Error(
+					"Failed to decode the signature binary blob from our custom binary format assuming the big-endian encoding, error: "
```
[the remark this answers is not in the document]
oleorhagen: Yeh, was my initial thinking as well.

PR #1444, opened 2023-10-23
Several fixes for Debian 10 buster.

kacf:


common/CMakeLists.txt
```diff
@@ -10,6 +10,9 @@ add_library(common_error STATIC
   error.cpp
   error/platform/c++11/error.cpp
 )
+# compile_options should be inherited by all our code, but common_error is used "virtually
+# everywhere", so just let it propagate from here.
```
lluiscampos, line 16: I am not convinced. My first grep for `target_link_libraries` got me [here](https://github.com/mendersoftware/mender/blob/01144d3d2a368c64a0f24b60a05c84f9ffac3b6a/api/CMakeLists.txt#L4), which does not use `common_error`.

lluiscampos (approved):

common/CMakeLists.txt
kacf: It does, but transitively, through the other dependencies.

But I see now that there are a few that don't, such as `common_setup` and `common`.

Update: I just discovered that `add_compile_options` works on "current directory and below`. This would be much better.

oleorhagen (approved):

cmake/boost.cmake
```diff
@@ -0,0 +1,66 @@
...
+if(MENDER_DOWNLOAD_BOOST AND NOT ${Boost_FOUND})
+  FetchContent_Declare(
+    Boost
+    # SYSTEM is only supported in CMake 3.25 and later, but is necessary in order to exclude Boost
+    # from our restrictive warnings-as-errors, in particular -Wsuggest-override. See workaround
+    # below.
+    SYSTEM
+    DOWNLOAD_EXTRACT_TIMESTAMP ON
+    URL https://github.com/boostorg/boost/releases/download/boost-1.83.0/boost-1.83.0.tar.gz
+    URL_MD5 58db882403e0c16b334760f3c3b76ff8
```
vpodzime, line 22: This hurts. The network layers used here have integrity checks so we don't need to duplicate that. And this has nothing to do with a security check. So please either drop it or replace it with a better checksum algorithm.

cmake/boost.cmake
```diff
@@ -0,0 +1,66 @@
...
+  #
+  # This is taken from this page:
+  # https://stackoverflow.com/questions/37434946/how-do-i-iterate-over-all-cmake-targets-programmatically
+  function(get_all_targets var)
+    set(targets)
+    get_all_targets_recursive(targets ${CMAKE_CURRENT_SOURCE_DIR})
+    set(${var} ${targets} PARENT_SCOPE)
+  endfunction()
+
+  macro(get_all_targets_recursive targets dir)
```
vpodzime, line 38: Would make more sense to define the macro before it's used. But not a biggie.

cmake/boost.cmake
```diff
@@ -4,6 +4,12 @@ find_package(Boost 1.74 COMPONENTS log)
 
 option(MENDER_DOWNLOAD_BOOST "Download Boost if it is not found (Default: OFF)" OFF)
 
+if(NOT MENDER_DOWNLOAD_BOOST AND NOT ${Boost_FOUND})
+  message(FATAL_ERROR
+    "Boost not found. Either make sure a recent enough Boost development package is installed (libboost-dev), or use `-D MENDER_DOWNLOAD_BOOST=ON`."
```
vpodzime, line 9: Should this also mention the required minimum version?

vpodzime (commented):
Looks good to me otherwise.

cmake/boost.cmake
kacf: The minimum version is printed on the previous line because of the `find_package` above, so this should be easy to deduce from the log.

cmake/boost.cmake
kacf: Fixed.

common/CMakeLists.txt
kacf: @lluiscampos: This is completely changed now. The source was moved into the `src` folder, which allows us to use directory-scoped flags (like `add_compile_options`) again. So we don't need this hack anymore.

It was more logical to do the rename first, so the new commit is the first commit in the series, somewhat counter-intuitively. Then the rest of the changes are based on top of that, but are not significantly changed other than to take the `add_compile_options` opportunity into account.

lluiscampos: It looks much cleaner now :+1:

lluiscampos (approved):

PR #1447, opened 2023-10-23
fix(https): Enable server host name verification in TLS handshake

vpodzime:
It's not done automatically and we need to use a callback for
this. However, it's a common thing so Boost has the callback
(functor object) ready for us.

Ticket: MEN-6788
Changelog: none

### DIFF
--- a/common/http/platform/beast/http.cpp
+++ b/common/http/platform/beast/http.cpp
@@ -16,9 +16,10 @@
 
 #include <algorithm>
 
-#include <boost/asio/ip/tcp.hpp>
-#include <boost/asio/ssl/verify_mode.hpp>
 #include <boost/asio.hpp>
+#include <boost/asio/ip/tcp.hpp>
+#include <boost/asio/ssl/host_name_verification.hpp>
+#include <boost/asio/ssl/verify_mode.hpp>
 
 #include <common/common.hpp>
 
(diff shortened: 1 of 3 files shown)


vpodzime:
Only the last commit is relevant for this PR, the rest is from #1438

common/http_test.cpp
```diff
@@ -1825,8 +1825,6 @@ TEST(HttpsTest, WrongSelfSignedCertificateError) {
 }
 
 TEST(HttpsTest, CertificationWithWrongHostName) {
-	GTEST_SKIP() << "MEN-6788";
```
kacf, line 1828: There are more like this in the proxy tests (grep for it).

kacf (approved):
👍

In a security library, I find it weird that this is opt-in instead of opt-out. But it's a good Boost abstraction, so looks pretty nice!

common/http/platform/beast/http.cpp
```diff
@@ -18,6 +18,7 @@
 
 #include <boost/asio/ip/tcp.hpp>
 #include <boost/asio/ssl/verify_mode.hpp>
+#include <boost/asio/ssl/host_name_verification.hpp>
```
oleorhagen, line 21: Nit: include order

oleorhagen (approved):

common/http_test.cpp
vpodzime: Oh, in a different file! :+1:

PR #1454, opened 2023-10-26
fix: Omit service/destination from method handling spec

vpodzime:
If the method call arrives to the particular `DBusServer` instance, it must have the matching destination set.

Ticket: MEN-6808
Changelog: none

### DIFF
--- a/api/auth_test.cpp
+++ b/api/auth_test.cpp
@@ -263,10 +263,7 @@ TEST_F(AuthDBusTests, AuthenticatorBasicTest) {
 	dbus::DBusServer dbus_server {loop, "io.mender.AuthenticationManager"};
 	auto dbus_obj = make_shared<dbus::DBusObject>("/io/mender/AuthenticationManager");
 	dbus_obj->AddMethodHandler<dbus::ExpectedStringPair>(
-		"io.mender.AuthenticationManager",
-		"io.mender.Authentication1",
-		"GetJwtToken",
-		[JWT_TOKEN, SERVER_URL]() {
+		"io.mender.Authentication1", "GetJwtToken", [JWT_TOKEN, SERVER_URL]() {
 			return dbus::StringPair {JWT_TOKEN, SERVER_URL};
 		});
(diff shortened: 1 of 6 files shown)


kacf (approved):
Is this because the service gets turned into `1:0` or something like that, so it can't be matched anymore? I remember reading about it some time ago, but unsure if I'm understanding it correctly.

vpodzime:
> Is this because the service gets turned into `1:0` or something like that, so it can't be matched anymore? I remember reading about it some time ago, but unsure if I'm understanding it correctly.

It's based on the difference between how `mender-update` makes DBus calls:
```
method call time=1698319153.077783 sender=:1.54 -> destination=io.mender.AuthenticationManager serial=2 path=/io/mender/AuthenticationManager; interface=io.mender.Authentication1; member=GetJwtToken
```
and how `mender-connect` works:
```
method call time=1698319172.376246 sender=:1.55 -> destination=:1.10 serial=6 path=/io/mender/AuthenticationManager; interface=io.mender.Authentication1; member=GetJwtToken
```

I'm not quite sure this will be enough to eliminate this failure:
```
WARN[2023-10-26T11:11:40Z] call to GetJWTToken on the Mender D-Bus API failed: GDBus.Error:org.freedesktop.DBus.Error.UnknownMethod: Method "GetJwtToken" with signature "" on interface "io.mender.Authentication1" doesn't exist
```
but it seems to be a requirement. (note that we don't advertise the methods in any way so it's interesting that it claims that the method doesn't exist while it works for `mender-update`).

kacf:
Hmm ok. I think I understand it. So the destination has already been resolved by the time it gets to the process (otherwise it wouldn't get there). Why they're calling it in different ways is another question, but it's not really relevant for the discussion.

The last point I don't know...

lluiscampos:
I am running integration tests with this fix and I will report soon :+1:

lluiscampos:
@vpodzime I updated your PR from latest feature branch for compatibility with `meta-mender` :+1:

lluiscampos (approved):
It works! :rocket:

PR #1457, opened 2023-10-30
chore: Add info log line on successful authentication

lluiscampos:
On regular operation, `mender-auth` will log errors until accepted by the server and then it will just keep silent. This log should not be very noisy and will give the user a confirmation that the device is has been accepted.

### DIFF
--- a/mender-auth/ipc/server.cpp
+++ b/mender-auth/ipc/server.cpp
@@ -90,6 +90,7 @@ void AuthenticatingForwarder::FetchJwtTokenHandler(auth_client::APIResponse &res
 			Cache(resp.value().token, resp.value().server_url);
 		}
 
+		log::Info("Successfully received new authorization data");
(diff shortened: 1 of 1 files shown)


lluiscampos:
The main motivation, really, is to have something easy to grep for in the tests.
* Current tests: https://github.com/mendersoftware/integration/blob/0ef4269f2d3dc608c729fc18efb7739bf8d8c014/tests/helpers.py#L75
* Tentative workaround: https://github.com/mendersoftware/integration/pull/2466/commits/d02ff59d333685ea75a8c00d49acf967ccd3a9e2#diff-7fb11d01efb2e6788562533f26902ed0563b68bfbd51ccecf879ec3ad594f869R77 --> but I rather keep it to grep on `mender-authd` and have an explicit entry there instead.

oleorhagen (approved):

kacf (approved):

PR #1459, opened 2023-10-31
Several fixes to make sure second updates work

kacf:
Ticket: MEN-6807

### DIFF
--- a/src/mender-update/http_resumer.hpp
+++ b/src/mender-update/http_resumer.hpp
@@ -89,6 +89,7 @@ private:
 	// The header handler needs to manipulate inner_reader_ in order to replace it in
 	// subsequent requests.
 	friend class HeaderHandlerFunctor;
+	friend class BodyHandlerFunctor;
 };
 
 // Main class to download the Artifact, which will react to server
--- a/src/mender-update/http_resumer.hpp
+++ b/src/mender-update/http_resumer.hpp
@@ -135,7 +135,7 @@ private:
 	void DoCancel();
 
 	shared_ptr<DownloadResumerClientState> resumer_state_;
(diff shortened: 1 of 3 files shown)


src/mender-update/http_resumer.hpp
```diff
@@ -89,6 +89,7 @@ class DownloadResumerAsyncReader : virtual public io::AsyncReader {
 	// The header handler needs to manipulate inner_reader_ in order to replace it in
 	// subsequent requests.
```
lluiscampos, line 90: Nit: please update the comment

src/mender-update/http_resumer/http_resumer.cpp
```diff
@@ -374,7 +381,7 @@ error::Error DownloadResumerClient::AsyncCall(
 
 	if (!*cancelled_) {
 		return error::Error(
-			make_error_condition(errc::operation_in_progress), "HTTP call already ongoing");
+			make_error_condition(errc::operation_in_progress), "HTTP resumer call already ongoing");
```
lluiscampos, line 384: Thank you

lluiscampos (approved):
The fixes look good.

What happened with the destructor? Why is not needed now to call `Cancel()`

kacf:
> What happened with the destructor? Why is not needed now to call `Cancel()`

I discovered that it already calls `Cancel()`, just indirectly, through the inner reader which also gets destroyed. So the issue was that it wasn't destroyed to begin with.

kacf:
Oh but hold on! That won't work if there is no inner reader, for example in between waits! I'll make a test case for that too.

kacf:
Agreed to merge early to unblock, but I will follow up on the issue I posted above.

PR #1460, opened 2023-10-31
MEN-6813: Serialize identity keys as single strings when possible

lluiscampos:
Use the same logic than for inventory keys: if the vector has one
element serialize as string, else submit as a list of strings.

### DIFF
--- a/src/common/identity_parser.hpp
+++ b/src/common/identity_parser.hpp
@@ -26,6 +26,8 @@ namespace kvp = mender::common::key_value_parser;
 
 kvp::ExpectedKeyValuesMap GetIdentityData(const string &identity_data_generator);
 
+string DumpIdentityData(const kvp::KeyValuesMap &identity_data);
+
 } // namespace identity_parser
 } // namespace common
 } // namespace mender
--- a/src/common/identity_parser/identity_parser.cpp
+++ b/src/common/identity_parser/identity_parser.cpp
@@ -14,6 +14,8 @@
 
 #include <common/identity_parser.hpp>
 
+#include <common/common.hpp>
+#include <common/json.hpp>
 #include <common/key_value_parser.hpp>
 #include <common/processes.hpp>
 
@@ -35,6 +37,38 @@ kvp::ExpectedKeyValuesMap GetIdentityData(const string &identity_data_generator)
 
 	auto ex_key_values = kvp::ParseKeyValues(ex_line_data.value());
 	return ex_key_values;
+}
+
+string DumpIdentityData(const kvp::KeyValuesMap &identity_data) {
+	stringstream top_ss;
+	top_ss << "{";
+	auto key_vector = common::GetMapKeyVector(identity_data);
+	std::sort(key_vector.begin(), key_vector.end());
+	for (const auto &key : key_vector) {
+		top_ss << R"(")" << json::EscapeString(key) << R"(":)";
+		if (identity_data.at(key).size() == 1) {
+			top_ss << R"(")" + json::EscapeString(identity_data.at(key)[0]) + R"(")";
+		} else {
+			stringstream items_ss;
+			items_ss << "[";
+			for (const auto &str : identity_data.at(key)) {
+				items_ss << R"(")" + json::EscapeString(str) + R"(",)";
+			}
+			auto items_str = items_ss.str();
+			// replace the trailing comma with the closing square bracket
+			items_str[items_str.size() - 1] = ']';
+			top_ss << items_str;
+		}
+		top_ss << ",";
+	}
+	auto str = top_ss.str();
+	if (str[str.size() - 1] == ',') {
+		// replace the trailing comma with the closing curly bracket
+		str.pop_back();
+	}
+	str.push_back('}');
+
+	return str;
 }
 
 } // namespace identity_parser
--- a/src/api/auth/auth.cpp
+++ b/src/api/auth/auth.cpp
@@ -115,13 +115,8 @@ error::Error FetchJWTToken(
 	if (!expected_identity_data) {
 		return expected_identity_data.error();
 	}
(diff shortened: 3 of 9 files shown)


lluiscampos:
This does not work as expected. Please wait for the review.

src/common/identity_parser.hpp
```diff
@@ -26,6 +26,8 @@ namespace kvp = mender::common::key_value_parser;
 
 kvp::ExpectedKeyValuesMap GetIdentityData(const string &identity_data_generator);
 
+string DumpIdentityData(kvp::KeyValuesMap identity_data);
```
vpodzime, line 29: `const kvp::KeyValuesMap &identity_data`

src/common/identity_parser/identity_parser.cpp
```diff
@@ -37,6 +39,39 @@ kvp::ExpectedKeyValuesMap GetIdentityData(const string &identity_data_generator)
...
 }
 
+string DumpIdentityData(kvp::KeyValuesMap identity_data) {
+	stringstream top_ss;
+	top_ss << "{";
+	auto key_vector = common::GetMapKeyVector(identity_data);
+	std::sort(key_vector.begin(), key_vector.end());
+	for (const auto &key : key_vector) {
+		top_ss << json::EscapeString(key);
+		top_ss << R"(:)";
```
vpodzime, line 49: The use of the raw string on this line is a big confusing,it's unnecessary, right?

src/common/identity_parser/identity_parser.cpp
```diff
@@ -37,6 +39,39 @@ kvp::ExpectedKeyValuesMap GetIdentityData(const string &identity_data_generator)
...
+			auto items_str = items_ss.str();
+			// replace the trailing comma with the closing square bracket
+			items_str[items_str.size() - 1] = ']';
+			top_ss << items_str;
+		}
+		top_ss << R"(,)";
+	}
+	auto str = top_ss.str();
+	if (str[str.size() - 1] == ',') {
+		// replace the trailing comma with the closing square bracket
```
vpodzime, line 67: A curly bracket is being added below in this case.

vpodzime (approved):
Looks good to me otherwise.

src/common/identity_parser/identity_parser.cpp
```diff
@@ -37,6 +39,39 @@ kvp::ExpectedKeyValuesMap GetIdentityData(const string &identity_data_generator)
...
+			top_ss << "\"" + json::EscapeString(identity_data[key][0]) + "\"";
+		} else {
+			stringstream items_ss;
+			items_ss << "[";
+			for (const auto &str : identity_data[key]) {
+				items_ss << "\"" + json::EscapeString(str) + "\",";
+			}
+			auto items_str = items_ss.str();
+			// replace the trailing comma with the closing square bracket
+			items_str[items_str.size() - 1] = ']';
```
kacf, line 60: I think this will break if `identity_data` is empty. The question is, can it be? You should probably use the same check as below.

src/common/identity_parser/identity_parser.cpp
lluiscampos: It is not necessary.

Ironically, I thought it would be _less_ confusing if I used raw strings consistently instead of combining the two typings :thinking:

src/common/identity_parser.hpp
lluiscampos: One issue if I take the const reference is that I cannot do operator '[] `

```
/home/lluis/northern.work/mender/src/common/identity_parser/identity_parser.cpp:49:38: error: passing ‘const KeyValuesMap’ {aka ‘const std::unordered_map<std::__cxx11::basic_string<char>, std::vector<std::__cxx11::basic_string<char> > >’} as ‘this’ argument discards qualifiers [-fpermissive]
   49 |                 if (identity_data[key].size() == 1) {
      |                                      ^
```

So I'll use non-const reference :+1:

src/common/identity_parser/identity_parser.cpp
lluiscampos: On empty `identity_data` the code won't get into this for loop. Then, it could be a problem on a map entry with a key and a `vector<string>` of 0 elements.

For the later, I added some unit tests for this and figured that it cannot really happen because even on blank values (say, an script echoing `key="`) the parser inserts an empty string in the vector.

I think it is okey to leave as is (with the tests, coming up in the next push).

For reference, if we fix here we must also follow-up [here](https://github.com/mendersoftware/mender/blob/2ad0d486f5aba4f760a3af85b00321ae85f5c446/src/mender-update/inventory.cpp#L104).

src/common/identity_parser.hpp
oleorhagen: Why not use `.at` in any case?

lluiscampos: Thank you :see_no_evil:

src/common/identity_parser/identity_parser.cpp
```diff
@@ -37,6 +39,39 @@ kvp::ExpectedKeyValuesMap GetIdentityData(const string &identity_data_generator)
 	return ex_key_values;
 }
 
+string DumpIdentityData(kvp::KeyValuesMap identity_data) {
```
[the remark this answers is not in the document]
lluiscampos: And remove the default overload for `unordered_map<string, vector<string>> std_map` from [here](https://github.com/mendersoftware/mender/blob/2ad0d486f5aba4f760a3af85b00321ae85f5c446/src/common/json/platform/nlohmann/nlohmann_json.cpp#L233)?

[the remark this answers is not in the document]
oleorhagen: Sorry, I thought `kvp::KeyValuesMap` was a separate type

kacf (approved):

PR #1462, opened 2023-11-01
fix: Don't cache server URL and token locally in process.

kacf:
```
    fix: Don't cache server URL and token locally in process.
    
    The problem if we do is that is that we may then connect to a
    forwarder address which doesn't exist anymore. This produces
    "connection refused", not "Not authorized", and therefore it does not
    trigger an authentication request by itself. So always ask.
    
    This can be easily reproduced by bringing up both mender-auth and
    mender-update, and then restarting mender-auth after the first auth
    cycle.
    
    This makes the whole `AuthenticatorExternalTokenUpdateTest` test
    irrelevant, since we no longer need to test the caching capability of
    the client side authenticator. So just remove it.
```


src/common/http/platform/beast/http.cpp
```diff
@@ -1887,7 +1887,9 @@ void Server::PrepareNewStream() {
 void Server::AsyncAccept(StreamPtr stream) {
 	acceptor_.async_accept(stream->socket_, [this, stream](const error_code &ec) {
 		if (ec) {
-			log::Error("Could not accept connection: " + ec.message());
+			if (ec != errc::operation_canceled) {
```
oleorhagen, line 1890: Q: should we add a lower level log for all instances?

vpodzime (commented):
I believe this makes many more tests irrelevant.

oleorhagen (approved):

src/common/http/platform/beast/http.cpp
kacf: Originally I did not expect to get any errors here at all. Then I realized that this particular one makes sense, but I still don't expect to get other errors, so I think it's best to keep it.

kacf:
> I believe this makes many more tests irrelevant.

Care to elaborate? 🙂 Everything else passed so..

lluiscampos (approved):

kacf:
> > I believe this makes many more tests irrelevant.
> 
> Care to elaborate? 🙂 Everything else passed so..

I can't really spot any. I'll merge for now, but let me know if there is some test that doesn't fit anymore.

vpodzime:
> I can't really spot any. I'll merge for now, but let me know if there is some test that doesn't fit anymore.

I'm surprised [this](https://github.com/mendersoftware/mender/blob/feature-c%2B%2B-client/src/api/client_test.cpp#L224) doesn't fail, for example. I went through the tests and it looks like they are not strict enough to start failing due to more calls of the `GetJwtToken()` method.

kacf:
@vpodzime: Thanks, I attempted to fix this [here](https://github.com/mendersoftware/mender/pull/1504).

PR #1463, opened 2023-11-01
feat(watchdog): Add application watchdog support for the systemd service

oleorhagen:
This adds support for an application watchdog through the systemd's supported
`WatchdogSec=` feature. In which the application will have to `ping` systemd, or
the service will restart.

Ticket: MEN-6763
Changelog: Add application watchdog support for the systemd service by default.
The default value is set to 24-hours, and if `mender-update` has not given any
sign of life within this interval, the service will be restarted by `systemd`.
The timeout value can be configured through `WatchdogSec=` in the
`mender-updated` service file, removed, or the feature can be disabled through
`MENDER_USE_SYSTEMD` when building.


oleorhagen:
Should probably add a simple integration test for this

kacf (changes requested):
I suggest one change in order to avoid that setting the poll interval to be longer than 24h starts triggering the watchdog unnecessarily. Not many customers are using this, but we know of a few, mainly those that are extremely bandwidth sensitive.

Make a dummy state, similar to the check-update state, which is on a hard-coded schedule, let's say 23h, for example. It does nothing, but simply spins the event loop, which kicks the watchdog. What do you think?

src/common/CMakeLists.txt
```diff
@@ -299,3 +299,17 @@ target_link_libraries(dbus_test PUBLIC
...
 gtest_discover_tests(dbus_test ${MENDER_TEST_FLAGS} NO_PRETTY_VALUES)
 add_dependencies(tests dbus_test)
+
+add_library(common_watchdog STATIC)
+if (MENDER_USE_SYSTEMD)
+  find_package(PkgConfig REQUIRED)
+  pkg_check_modules(SYSTEMD REQUIRED IMPORTED_TARGET libsystemd)
+  target_sources(common_watchdog PRIVATE watchdog/systemd/watchdog.cpp)
+  target_compile_options(common_crypto PRIVATE ${PLATFORM_SPECIFIC_COMPILE_OPTIONS})
+  target_link_libraries(common_crypto PUBLIC
```
vpodzime, line 309: `common_crypto`?

src/common/watchdog.hpp
```diff
@@ -0,0 +1,30 @@
...
+#ifndef MENDER_COMMON_WATCHDOG_HPP
+#define MENDER_COMMON_WATCHDOG_HPP
+
+namespace mender {
+namespace common {
+namespace watchdog {
+
+using namespace std;
+
+void Kick();
```
vpodzime, line 24: Love this peaceful name of the ping function! :heart: :grin: :+1:

src/common/watchdog/dud/watchdog.cpp
```diff
@@ -0,0 +1,27 @@
...
+
+#include <common/watchdog.hpp>
+
+namespace mender {
+namespace common {
+namespace watchdog {
+
+void Kick() {
+	return;
+}
```
vpodzime, line 23: I'd call this `noop` instead of `dud`, but not a biggie.

src/common/watchdog/systemd/watchdog.cpp
```diff
@@ -0,0 +1,46 @@
...
+#include <common/log.hpp>
+
+namespace mender {
+namespace common {
+namespace watchdog {
+
+using namespace std;
+
+void Kick() {
+  log::Trace("Kicking the application watchdog");
```
vpodzime, line 28: weird indentation

src/common/watchdog/systemd/watchdog.cpp
```diff
@@ -0,0 +1,46 @@
...
+	uint64_t usec;
+	if (sd_watchdog_enabled(0, &usec) > 1) {
+		int ret {sd_notify(0 /* Do not unset NOTIFY_SOCKET env var */, "WATCHDOG=1")};
+		if (ret < 0) {
+			log::Error(
+				"Failed to kick the systemd service watchdog, received error code: "
+				+ to_string(ret));
```
vpodzime, line 35: > On failure, these calls return a negative errno-style error code. 

Indicates that `strerror(-ret)` could give some useful details.

src/common/watchdog/systemd/watchdog.cpp
```diff
@@ -0,0 +1,46 @@
...
+		}
+		if (ret == 0) {
+			log::Error(
+				"The NOTIFY_SOCKET environment variable is not set, so it is not possible to kick the watchdog");
```
vpodzime, line 39: Can this even happen when `sd_watchdog_enabled()` returns a positive value?

vpodzime (commented):
A nice and smart thing to do! :clap: I'm looking forward to seeing a nice and generic watchdog implementation in *libntech* that will be used on non-systemd platforms :wink:

src/common/watchdog.hpp
oleorhagen: :laughing:

src/common/watchdog/dud/watchdog.cpp
oleorhagen: Yeh - much better :smile:

src/common/watchdog/systemd/watchdog.cpp
oleorhagen: Alright, good to know! I read it, but didn't fully comprehend :)

src/common/watchdog/systemd/watchdog.cpp
oleorhagen: Probably not, unless something weird happens in the meantime.

Defense, or clean code? :thinking:

oleorhagen:
> I suggest one change in order to avoid that setting the poll interval to be longer than 24h starts triggering the watchdog unnecessarily. Not many customers are using this, but we know of a few, mainly those that are extremely bandwidth sensitive.
> 
> Make a dummy state, similar to the check-update state, which is on a hard-coded schedule, let's say 23h, for example. It does nothing, but simply spins the event loop, which kicks the watchdog. What do you think?

I see where you're coming from - but I'm not quite sure this does not defeat the purpose. In this case, the rest of the state-machine could freeze up, but the watchdog is still kicked on the regular, no?

In this case, I think they should adjust the interval in their service config file instead. Which begs the question - maybe we should add this to our Yocto-build setup then? Instead of hard-coding it here? :thinking:

oleorhagen:
Acceptance test: https://github.com/mendersoftware/meta-mender/pull/2000

kacf:
> I see where you're coming from - but I'm not quite sure this does not defeat the purpose. In this case, the rest of the state-machine could freeze up, but the watchdog is still kicked on the regular, no?
> 
> In this case, I think they should adjust the interval in their service config file instead. Which begs the question - maybe we should add this to our Yocto-build setup then? Instead of hard-coding it here? 🤔

Well, I was thinking it would only be kicked in the idle state, where there is the least chance of state machine freeze up. I agree it does increase the chance of freezing *slightly*, but the idle state is not really where we see the problem popping up in the first place, is it?

src/common/watchdog/systemd/watchdog.cpp
oleorhagen: haha, alright. Sold!

vpodzime (approved):
ACK

oleorhagen:
> > I see where you're coming from - but I'm not quite sure this does not defeat the purpose. In this case, the rest of the state-machine could freeze up, but the watchdog is still kicked on the regular, no?
> > In this case, I think they should adjust the interval in their service config file instead. Which begs the question - maybe we should add this to our Yocto-build setup then? Instead of hard-coding it here? 🤔
> 
> Well, I was thinking it would only be kicked in the idle state, where there is the least chance of state machine freeze up. I agree it does increase the chance of freezing _slightly_, but the idle state is not really where we see the problem popping up in the first place, is it?

What do we gain by this?

kacf:
We gain that customers who have longer than 24h schedules do not start getting client restarts in their logs every 24h. I don't think it has any ill effects, but it will look like something is wrong when reading the log.

The most ideal would be to have the `WatchDogSec = max(24h, POLLING_INTERVAL + 1)`, but I suppose there's no way to set it programmatically.

oleorhagen:
Meta-mender support: https://github.com/mendersoftware/meta-mender/pull/2060

kacf (approved):
> Meta-mender support: [mendersoftware/meta-mender#2060](https://github.com/mendersoftware/meta-mender/pull/2060)

👍

oleorhagen:
@mender-test-bot start pipeline --pr meta-mender/2060

oleorhagen:
@mender-test-bot start pipeline --pr meta-mender/2060 --pr integration/2561

PR #1464, opened 2023-11-01
chore(state-machine): Panic in the case of non-handled non-deferred e…

oleorhagen:


src/common/state_machine.hpp
```diff
@@ -223,6 +223,21 @@ class StateMachineRunner : virtual public EventPoster<EventType> {
 			// actionable until at least one state machine reaches a different state.
 			PostToEventLoop();
 		}
+
+		// Check if there are any non-deferred events in the queue - then fail if
+		queue<EventType> queue_copy = event_queue_;
+		while (not queue_copy.empty()) {
```
kacf, line 229: It's expected to have non-deferred events here, because this is after the handlers have run. I think the check needs to be between the `if (!to_run.empty()) {` block and the one above it.

There is also a much easier way to check this. When we are doing `event_queue_.push(event)` above, also increment a `deferred_count` counter. If `deferred_count < event_queue_.size()`, then the check is positive.

kacf (changes requested):
Good safeguard!

src/common/state_machine.hpp
oleorhagen: Oh! Very good idea!

I had a suspicion you would come up with something smarter than me here :laughing:

oleorhagen: Alright, new try. I am still using my original approach, but moved it to before we run the machine loop.

Ur approach, although clever, did not work when initializing (since we add the start events to the to_run queue, and I didn't feel like special casing it.

src/common/dbus/platform/asio_libdbus/dbus.cpp
```diff
@@ -67,15 +67,14 @@ void HandleReply(DBusPendingCall *pending, void *data);
 DBusHandlerResult MsgFilter(DBusConnection *connection, DBusMessage *message, void *data);
 
 error::Error DBusPeer::InitializeConnection() {
-	DBusError dbus_error;
-	dbus_error_init(&dbus_error);
-	dbus_conn_.reset(dbus_bus_get_private(DBUS_BUS_SYSTEM, &dbus_error));
+	std::unique_ptr<DBusError, void (*)(DBusError *)> dbus_error {new DBusError, dbus_error_free};
```
kacf, line 70: I support the incentive, but I suspect you have in fact have introduced a memory leak here. There are two levels of resources here, the `DBusError` struct itself, and the content it contains. `dbus_error_free` only frees the latter.

I can see two ways to solve it:

1. Use a lambda with both `dbus_error_free` and `delete dbus_error` in it.
2. Don't use `new DBusError`, but keep the variable on the stack instead, as before, and use `&dbus_error` as the argument when constructing the smart pointer (with a different name), together with `dbus_error_free`. The `unique_ptr` now doesn't own the pointer itself, but it owns the contents.

Same for the other commit.

src/common/state_machine.hpp
kacf: This can be easily fixed too. Just keep a counter for the `run_queue` as well, but don't increment it during the initialization step, only when adding events in the "normal" flow. Then the check becomes `deferred_count < added_to_run_queue`. I think it's better than copying the whole queue on every state transition.

kacf (changes requested):

src/common/state_machine.hpp
oleorhagen: True, I did dabble with the idea. But decided it's not really better. This implementation reads pretty straight forward, and does not have to touch the other pieces of code, and handle special cases.

src/common/dbus/platform/asio_libdbus/dbus.cpp
oleorhagen: Oh shite. True. Rookie mistake!

src/common/dbus/platform/asio_libdbus/dbus.cpp
```diff
@@ -67,31 +74,36 @@ void HandleReply(DBusPendingCall *pending, void *data);
...
+	dbus_error_init(dbus_error.get());
+	dbus_conn_.reset(dbus_bus_get_private(DBUS_BUS_SYSTEM, dbus_error.get()));
 	if (!dbus_conn_) {
 		auto err = MakeError(
 			ConnectionError,
-			string("Failed to get connection to system bus: ") + dbus_error.message + "["
-				+ dbus_error.name + "]");
-		dbus_error_free(&dbus_error);
+			string("Failed to get connection to system bus: ") + dbus_error.get()->message + "["
+				+ dbus_error.get()->name + "]");
```
kacf, line 85: `.get()` should only be used when it's required to, here it can be omitted.

kacf (approved):

PR #1466, opened 2023-11-01
chore: Alias mender-authd.service -> mender-client.service

oleorhagen:
This aliases the systemd.service for `mender-authd` to `mender-client`, so that
other existing third-party services relying on it will still work transparently.

Ticket: MEN-6812

-------

### DIFF
--- a/support/mender-authd.service
+++ b/support/mender-authd.service
@@ -14,3 +14,4 @@ Restart=always
 
 [Install]
 WantedBy=multi-user.target
+Alias=mender-client.service


lluiscampos (approved):
This is what I was hoping for :rocket: 

By reading the doc this seems to do what we want. I also think it is _more_ correct to put it in `mender-authd` than in `mender-updated` :+1:

oleorhagen:
> This is what I was hoping for 🚀
> 
> By reading the doc this seems to do what we want. I also think it is _more_ correct to put it in `mender-authd` than in `mender-updated` 👍

I think it _has_ to be in `authd` considering the dependencies

oleorhagen:
lazy 1 - world 1

lluiscampos:
Not an issue, but this will also make it work for tests or uses doing `systemctl restart mender-client`. For the tests at least it would be best to catch and fix all of these.

oleorhagen:
Testing it now

lluiscampos:
It works fine on my end :+1:

PR #1468, opened 2023-11-02
test: Verify that identity is marshaled the same way as in Go

oleorhagen:
This adds a test to verify that identity is marshaled in the same way as it is
in the Golang client code.

See the test here:
`https://github.com/mendersoftware/mender/blob/mendersoftware/device/identity_data_test.go#L130`

Note that we still might have a UTF-8 issue here. But this is a separate topic.

Ticket: MEN-6825

-------

### DIFF
--- a/src/common/identity_parser_test.cpp
+++ b/src/common/identity_parser_test.cpp
@@ -117,3 +117,94 @@ TEST_F(IdentityParserTests, DumpIdentityEmptyIdentity) {
 
 	ASSERT_EQ(R"({})", json_str);
 }
+
+
+TEST_F(IdentityParserTests, VerifyIdentityKeyOrder) {
+	string script = R"(#!/bin/sh
+echo "foo=bar"
+echo "key=value=23"
+echo "some value=bar"
+echo "mac=de:ad:be:ef:00:01"
+exit 0
+)";
+	auto ret = PrepareTestScript(script);
+	ASSERT_TRUE(ret);
+
+	kv_p::ExpectedKeyValuesMap ex_data = id_p::GetIdentityData(test_script_fname);
+	ASSERT_TRUE(ex_data);
+
+	kv_p::KeyValuesMap key_values_map = ex_data.value();
+	auto json_str = id_p::DumpIdentityData(key_values_map);
+
+	ASSERT_EQ(
+		R"({"foo":"bar","key":"value=23","mac":"de:ad:be:ef:00:01","some value":"bar"})", json_str);
+}
+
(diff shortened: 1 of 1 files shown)


kacf:
I assume these are passing because all the operations involved in this particular test case are in fact using auto-conversion. So which ones aren't using it? Database? HTTP requests maybe?

oleorhagen:
> I assume these are passing because all the operations involved in this particular test case are in fact using auto-conversion. So which ones aren't using it? Database? HTTP requests maybe?

I assumed they we're passing, at least that is the case for me locally, because my `locale` is `UTF-8`

kacf:
But that should also break, in theory, because the database is not doing conversion to UTF-8, AFAIK. OTOH, it possible that nothing will break as long as the locale on the system doesn't change, since even if you introduce inconsistencies, they don't become visible until you change the locale. I think we need to hunt specifically for a failing case.

lluiscampos:
Nit: the link in the commit message (https://github.com/mendersoftware/mender/blob/mendersoftware/device/identity_data_test.go#L130) will disappear after we merge the feature branch. Use a link for an specific master commit instead

oleorhagen:
> But that should also break, in theory, because the database is not doing conversion to UTF-8, AFAIK. OTOH, it possible that nothing will break as long as the locale on the system doesn't change, since even if you introduce inconsistencies, they don't become visible until you change the locale. I think we need to hunt specifically for a failing case.

Yes, but I feel we are discussing something else than this PR now. I maybe should/could/can keep the UTF-8 out of this. It was originally just meant to verify that we marshal the JSON in the same way (disregarding the UTF-8/string issues [if there are any])

kacf:
Hmm, maybe you're right. To be honest it's hard to wrap my head around this.

But do you agree that we should try to deliberately find a failing scenario, just to verify that this is in fact a problem? It might involve setting your own locale to something weird, but it's for testing purposes.

oleorhagen:
> Hmm, maybe you're right. To be honest it's hard to wrap my head around this.
> 
> But do you agree that we should try to deliberately find a failing scenario, just to verify that this is in fact a problem? It might involve setting your own locale to something weird, but it's for testing purposes.

Yes, absolutely!

I will link to this conversation in the string representation research ticket

kacf (approved):
Good! Then I request that you leave the Unicode tests out of this PR, because I'm still not convinced that they would reveal the problem the way we want, so it's better not to submit them while we are still unsure.

oleorhagen:
> Good! Then I request that you leave the Unicode tests out of this PR, because I'm still not convinced that they would reveal the problem the way we want, so it's better not to submit them while we are still unsure.

:apple: :orange_circle:

PR #1472, opened 2023-11-03
fix: Fix several fallback config parsing bugs.

kacf:
First of all, reverse the order of loading. If the fallback is to have least precedence, then it must be loaded first for the other one to override it.

Then, to match the behavior of the Golang client, we need to change two things:

* We need to allow `ArtifactVerifyKey` and `ServerURL` to appear in both files. This was not previously possible because they were automatically put into the list variant and would trigger an error that you can't use both at the same time.

* The list variants, `ArtifactVerifyKeys` and `Servers` need to override the specified lists in the fallback config file, not add to them.

Changelog: None
Ticket: None

### DIFF
--- a/src/common/conf_test.cpp
+++ b/src/common/conf_test.cpp
@@ -752,3 +752,27 @@ TEST(ConfTests, ProxyEnvironmentVariables) {
 		EXPECT_FALSE(result);
 	}
 }
+
+TEST(ConfTests, FallbackConfig) {
+	mtesting::TemporaryDirectory tmpdir;
+
+	string conf_file = path::Join(tmpdir.Path(), "mender.conf");
+	{
+		ofstream f(conf_file);
+		f << R"({"ServerURL": "https://right-server.com"})";
+		ASSERT_TRUE(f.good());
+	}
+
+	string fallback_conf_file = path::Join(tmpdir.Path(), "fallback-mender.conf");
+	{
+		ofstream f(fallback_conf_file);
+		f << R"({"ServerURL": "https://wrong-server.com"})";
+		ASSERT_TRUE(f.good());
+	}
+
+	vector<string> args {"--config", conf_file, "--fallback-config", fallback_conf_file};
+	conf::MenderConfig config;
+	config.ProcessCmdlineArgs(args.begin(), args.end(), conf::CliApp {});
+	ASSERT_EQ(config.servers.size(), 1);
+	EXPECT_EQ(config.servers[0], "https://right-server.com");
+}
--- a/src/common/config_parser/config_parser.cpp
+++ b/src/common/config_parser/config_parser.cpp
@@ -206,6 +206,7 @@ ExpectedBool MenderConfigFromFile::LoadFile(const string &path) {
 
 	e_cfg_value = cfg_json.Get("ArtifactVerifyKeys");
 	if (e_cfg_value) {
+		this->artifact_verify_keys.clear();
 		const json::Json value_array = e_cfg_value.value();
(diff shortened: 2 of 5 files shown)


src/common/conf_test.cpp
```diff
@@ -752,3 +752,27 @@ TEST(ConfTests, ProxyEnvironmentVariables) {
...
 	}
 }
+
+TEST(ConfTests, FallbackConfig) {
+	mtesting::TemporaryDirectory tmpdir;
+
+	string conf_file = path::Join(tmpdir.Path(), "mender.conf");
+	{
+		ofstream f(conf_file);
+		f << R"({"ServerURL": "https://right-server.com"})";
```
oleorhagen, line 762: Example.com? :laughing:

src/common/config_parser/config_parser.cpp
```diff
@@ -227,26 +228,30 @@ ExpectedBool MenderConfigFromFile::LoadFile(const string &path) {
...
 		const json::ExpectedString e_cfg_string = value_json.GetString();
 		if (e_cfg_string) {
-			if (artifact_verify_keys.size() != 0) {
+			if (artifact_verify_keys_field_used_) {
 				auto err = MakeError(
 					ConfigParserErrorCode::ValidationError,
 					"Both 'ArtifactVerifyKey' and 'ArtifactVerifyKeys' are set");
 				return expected::unexpected(err);
 			}
+			assert(artifact_verify_keys.size() <= 1);
```
oleorhagen, line 245: Q: Why?

kacf: Because if `artifact_verify_keys_field_used_` is false, then it should not be possible to have more than one key in the list, since only the `ArtifactVerifyKey` singleton form has been used. We're just making sure of that logic here.

src/common/conf_test.cpp
kacf: I guess that's the go-to choice! But we don't connect to it anyway, it's just parsing.

lluiscampos (approved):

src/common/config_parser/config_parser.cpp
oleorhagen: But why did you remove the test for the multiple elements case? :thinking:

kacf: To be accurate it was a test for *any* element, not multiple elements. And we cannot use that test anymore because we need to be able to replace an element with a new element (the override case, first bullet point in the commit message). So there can be an element already, or no element. We clear the list, so it makes no difference in the lines below.

There should never be more than one element, because then `artifact_verify_keys_field_used_` will be true (the only way to get multiple elements into the list).

oleorhagen: Roger :green_circle:

kacf:
@lluiscampos, @oleorhagen: How does [this result](https://gitlab.com/Northern.tech/Mender/mender-qa/-/jobs/5477550262) look to you?

oleorhagen:
> @lluiscampos, @oleorhagen: How does [this result](https://gitlab.com/Northern.tech/Mender/mender-qa/-/jobs/5477550262) look to you?

`test_rootfs_image_rejected` passed last nightly :/

lluiscampos:
@kacf Your branch is not up to date. At least one of the failures is due to missing https://github.com/mendersoftware/mender/pull/1471

Also, please set `BUILD_CPP_CLIENT` to "true" for your non-bot builds. This time it worked by mistake :sweat_smile:

lluiscampos:
> This time it worked by mistake 😅

Fix here:
* https://github.com/mendersoftware/mender-qa/pull/707

PR #1474, opened 2023-11-03
Fix key regeneration issue when keys are set in the config

oleorhagen:
Simple fix for keys set in our configuration file


lluiscampos (approved):
Good fixes. Thanks

PR #1480, opened 2023-11-07
chore: Temporary add the version to an inventory script

lluiscampos:
To be reverted in MEN-6840

### DIFF
--- a/support/mender-inventory-provides
+++ b/support/mender-inventory-provides
@@ -7,6 +7,10 @@ set -e
 
 /usr/bin/mender-update show-provides
 
+# Submit client version
+# TODO: MEN-6840: Remove when developing the built-in version
+echo mender_client_version=`/usr/bin/mender-update --version`
+
 # Also submit device_type, which is in a separate file, not in the database. We don't know if the
 # file contains a newline, so it's important that this is done last, to avoid corrupting other
 # entries.


oleorhagen (approved):

support/mender-inventory-provides
```diff
@@ -12,4 +12,8 @@ set -e
 # entries.
 grep '^device_type=' /var/lib/mender/device_type
 
+# Submit client version
+# TODO: MEN-6840: Remove when developing the built-in version
+echo mender_client_version=`/usr/bin/mender-update --version`
```
oleorhagen, line 17: Q: Should we skip `/usr/bin` and rely on the PATH?

kacf: Beware of the comment in the previous block! Put this one above it.

kacf (approved):
Pre-approving, but check my comment!

support/mender-inventory-provides
lluiscampos: > Q: Should we skip `/usr/bin` and rely on the PATH?

Few lines above we are already using `/usr/bin/`. I am not comfortable changing it, really. (Also for this specific case it will be reverted, but your point stands valid for the pre-existing code)

lluiscampos: > Beware of the comment in the previous block! Put this one above it.

Thank you!

PR #1481, opened 2023-11-07
fix: Refuse to install an unsigned Artifact when keys are available

lluiscampos:


oleorhagen (approved):
Glad we found this one!

:medal_military:

PR #1485, opened 2023-11-08
MEN-6671: Check for both config files

lluiscampos:


support/modules/rootfs-image
```diff
@@ -152,7 +173,7 @@ EOF
         ;;
 
     ArtifactRollback)
```
lluiscampos, line 175: @kacf I have an issue here because the module should be able to rollback even when the config file is missing. As the logic is set, the config file is required for setting `upgrade_available`, `mender_boot_part`, etc.

How can we handle this?

kacf: Hmm, yes, tricky. I *think* we can actually ignore that error here, because we know that the bootloader will roll back for us, even if we do nothing here. Does that make sense?

The `VerifyRollbackReboot` should still catch errors if something goes wrong in that logic.

kacf: In other words, do it opportunistically if we have the info and the tools, but accept that we sometimes can't do it, and return success anyway.

support/modules/rootfs-image
```diff
@@ -44,20 +25,55 @@ else
...
-upgrade_available="$(${PRINTENV} upgrade_available)"
-upgrade_available="${upgrade_available#upgrade_available=}"
+parse_conf_file() {
+    found_conf_file=
+    # Try first the fallback config file, which has least precedence
+    if grep -q RootfsPartA "${MENDER_DATASTORE_DIR:-/var/lib/mender}/mender.conf" 2> /dev/null; then
+        CONF_FILE="${MENDER_DATASTORE_DIR:-/var/lib/mender}/mender.conf"
+        found_conf_file=yes
+    fi
+    if grep -q RootfsPartA "${MENDER_CONF_DIR:-/etc/mender}/mender.conf" 2> /dev/null; then
```
kacf, line 35: This is ok, but wouldn't it be both easier and more robust to slap a loop around the parsing below, once over `/data`, once over `/etc`? You wouldn't even need to check a condition then, you can just use something like:

```
tmp="$(jq -r .RootfsPartA < "$CONF_FILE")"
MENDER_ROOTFS_PART_A="${tmp:-${MENDER_ROOTFS_PART_A}}"
```

An advantage then is that if `jq` is installed, it's guaranteed to be parsed correctly, whereas this block can still misparse `RootfsPartA` if it happens to be in a comment for example. Not that I expect it, but you know. 🙂

Just make sure that the `jq` and `sed` calls use ` || true` to avoid `set -e` kicking in.

support/modules/rootfs-image
```diff
@@ -77,6 +93,12 @@ ensure_correct_root_mounted() {
     exit 1
 }
 
+check_requirements() {
+    parse_conf_file || exit 1
```
kacf, line 97: `set -e` is active, so the `exit` is not necessary.

support/modules/rootfs-image
lluiscampos: `json` does not allow comments. But in any case what you say makes sense. You could have a `RootfsPartAA` key or `RootfsPartA` as part of a tenant token (I guess it is technically _possible_ :laughing: ).

I'll clean it up, thanks.

kacf (approved):

lluiscampos:
Tested in https://github.com/mendersoftware/integration/pull/2487 before the rework. Tested manually after the rework :+1:

PR #1486, opened 2023-11-08
MEN-6671: Resume downloads when the first header handler errors

lluiscampos:
This will cover the case for example when the host cannot temporary be resolved or the handshake fails.

It is explicitly tested on an integration test.

### DIFF
--- a/src/mender-update/http_resumer/http_resumer.cpp
+++ b/src/mender-update/http_resumer/http_resumer.cpp
@@ -108,6 +108,17 @@ private:
 void HeaderHandlerFunctor::operator()(http::ExpectedIncomingResponsePtr exp_resp) {
 	auto resumer_client = resumer_client_.lock();
 	if (resumer_client) {
+		// If an error has already occurred, schedule the next AsyncCall directly
+		if (!exp_resp) {
+			resumer_client->logger_.Warning(exp_resp.error().String());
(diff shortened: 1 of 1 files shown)


src/mender-update/http_resumer/http_resumer.cpp
```diff
@@ -108,6 +108,19 @@ class BodyHandlerFunctor {
 void HeaderHandlerFunctor::operator()(http::ExpectedIncomingResponsePtr exp_resp) {
 	auto resumer_client = resumer_client_.lock();
 	if (resumer_client) {
+		// If an error has already occurred, schedule the next AsyncCall directly
+		if (!exp_resp) {
+			resumer_client->logger_.Warning(exp_resp.error().String());
+			// resumer_client->CallUserHandler(exp_resp);
+			// return;
```
kacf, line 115: Remove?

kacf (approved):

src/mender-update/http_resumer/http_resumer.cpp
lluiscampos: Yes, of course

lluiscampos:
Tested in https://github.com/mendersoftware/integration/pull/2487 before the rework. Not running again the pipeline.

PR #1488, opened 2023-11-09
chore: Fix inability to capture log output.

kacf:
The `clog` stream has some peculiar properties, and can't be captured properly, neither using stdout nor stderr. We don't really need it, so just switch to `cout` instead.

### DIFF
--- a/src/artifact/lexer.hpp
+++ b/src/artifact/lexer.hpp
@@ -52,8 +52,12 @@ public:
 	Token &Next() {
 		auto entry = tar_reader_->Next();
 		if (!entry) {
-			log::Trace("Error reading the next tar entry: " + entry.error().message);
-			this->current = Token {Type::EOFToken};
+			if (entry.error().code == tar::MakeError(tar::TarEOFError, "").code) {
+				this->current = Token {Type::EOFToken};
+			} else {
+				log::Error("Error reading the next tar entry: " + entry.error().String());
+				this->current = Token {Type::Unrecognized};
+			}
 			return this->current;
 		}
 		log::Trace("Entry name: " + entry.value().Name());
--- a/src/artifact/parser.cpp
+++ b/src/artifact/parser.cpp
@@ -114,7 +114,7 @@ ExpectedArtifact Parse(io::Reader &reader, config::ParserConfig config) {
 	if (!expected_manifest) {
 		return expected::unexpected(parser_error::MakeError(
 			parser_error::Code::ParseError,
-			"Failed to parse the manifest: " + expected_manifest.error().message));
+			"Failed to parse the manifest: " + expected_manifest.error().String()));
 	}
 	auto manifest = expected_manifest.value();
 
--- a/src/artifact/parser.cpp
+++ b/src/artifact/parser.cpp
@@ -56,26 +56,25 @@ ExpectedArtifact VerifyEmptyPayloadArtifact(
 	}
 	// TODO - When augmented sections are added - check for these also
 	log::Trace("Empty payload Artifact: Verifying empty payload");
-	auto tok = lexer.Next();
-	if (tok.type == token::Type::Payload) {
+	auto expected_payload = artifact.Next();
+	if (expected_payload) {
 		// If a payload is present, verify that it is empty
-		auto expected_payload = artifact.Next();
-		if (!expected_payload) {
-			auto err = expected_payload.error();
-			if (err.code.value() != parser_error::Code::NoMorePayloadFilesError) {
-				return expected::unexpected(parser_error::MakeError(
-					parser_error::Code::ParseError,
-					"This should never happen, we have a payload token / programmer error"));
-			}
-			return artifact;
-		}
 		auto payload = expected_payload.value();
 		auto expected_payload_file = payload.Next();
 		if (expected_payload_file) {
 			return expected::unexpected(parser_error::MakeError(
 				parser_error::Code::ParseError, "Empty Payload Artifacts cannot have a payload"));
-		}
-	}
+		} else if (
+			expected_payload_file.error().code
+			!= parser_error::MakeError(parser_error::Code::NoMorePayloadFilesError, "").code) {
+			return expected::unexpected(
+				expected_payload.error().WithContext("While verifying empty payload"));
+		} // else fall through
+	} else if (
+		expected_payload.error().code != parser_error::MakeError(parser_error::EOFError, "").code) {
+		return expected::unexpected(
+			expected_payload.error().WithContext("While verifying empty payload"));
+	} // else fall through
 	return artifact;
 }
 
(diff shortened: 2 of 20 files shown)


oleorhagen (approved):

kacf:
@oleorhagen: Please re-review. The original one-liner turned into a much bigger PR due to all the issues it exposed.

src/common/events/events_io.cpp
```diff
@@ -105,16 +105,20 @@ ReaderFromAsyncReader::ReaderFromAsyncReader(EventLoop &event_loop, mio::AsyncRe
 mio::ExpectedSize ReaderFromAsyncReader::Read(
 	vector<uint8_t>::iterator start, vector<uint8_t>::iterator end) {
 	mio::ExpectedSize read;
-	error::Error err;
 	bool finished = false;
-	err = reader_->AsyncRead(start, end, [this, &finished, &read](mio::ExpectedSize num_read) {
-		read = num_read;
-		finished = true;
-		event_loop_.Stop();
+	event_loop_.Post([start, end, this, &finished, &read]() {
```
oleorhagen, line 109: Wonder if we have more of these lying around.

src/mender-update/http_resumer/http_resumer.cpp
```diff
@@ -309,6 +309,11 @@ void DownloadResumerAsyncReader::Cancel() {
 
 error::Error DownloadResumerAsyncReader::AsyncRead(
 	vector<uint8_t>::iterator start, vector<uint8_t>::iterator end, io::AsyncIoHandler handler) {
+	if (eof_) {
```
oleorhagen, line 312: I think I might have come across this in one of my integration tests actually! Thx!

src/artifact/parser_test.cpp
```diff
@@ -161,7 +161,7 @@ TEST(ParserTest, TestParseMumboJumbo) {
 	auto artifact = mender::artifact::parser::Parse(sr);
 
 	ASSERT_FALSE(artifact) << artifact.error().message << std::endl;
-	ASSERT_EQ(artifact.error().message, "Got unexpected token : 'EOF' expected 'version'");
+	ASSERT_EQ(artifact.error().message, "Got unexpected token : 'Unrecognized' expected 'version'");
```
oleorhagen, line 164: Nice fix!

src/artifact/lexer.hpp
```diff
@@ -52,8 +52,12 @@ class Lexer {
 	Token &Next() {
 		auto entry = tar_reader_->Next();
 		if (!entry) {
-			log::Trace("Error reading the next tar entry: " + entry.error().message);
-			this->current = Token {Type::EOFToken};
+			if (entry.error().code == tar::MakeError(tar::TarEOFError, "").code) {
```
oleorhagen, line 55: Indeed :see_no_evil:  thanks!

src/artifact/parser.cpp
```diff
@@ -56,25 +56,25 @@ ExpectedArtifact VerifyEmptyPayloadArtifact(
...
-		if (!expected_payload) {
-			auto err = expected_payload.error();
-			if (err.code.value() != parser_error::Code::NoMorePayloadFilesError) {
-				return expected::unexpected(parser_error::MakeError(
-					parser_error::Code::ParseError,
-					"This should never happen, we have a payload token / programmer error"));
-			}
-			return artifact;
-		}
 		auto payload = expected_payload.value();
```
oleorhagen, line 67: Q: What if we get `EOFError`?

src/artifact/parser.cpp
```diff
@@ -186,26 +186,35 @@ ExpectedArtifact Parse(io::Reader &reader, config::ParserConfig config) {
...
 	return artifact;
 };
 
 
 ExpectedPayload Artifact::Next() {
-	// Currently only one payload supported
-	if (payload_index_ != 0) {
-		return expected::unexpected(parser_error::MakeError(
-			parser_error::Code::EOFError, "Reached the end of the Artifact"));
+	token::Token tok = lexer_.Next();
```
oleorhagen, line 194: Nice fixes!

I would suggest something like this though:

```
ExpectedPayload Artifact::Next() {
	token::Token tok = lexer_.Next();
	if (payload_index_ != 0) {
		// Currently only one payload supported
		switch (tok.type) {
		case token::Type::EOFToken:
			return expected::unexpected(parser_error::MakeError(
				parser_error::Code::EOFError, "Reached the end of the Artifact"));
		case token::Type::Payload:
			return expected::unexpected(error::Error(
				make_error_condition(errc::not_supported), "Only one artifact payload supported"));
		default:
			return expected::unexpected(parser_error::MakeError(
				parser_error::Code::ParseError, "Unexpected token: " + tok.TypeToString()));
		};
	}
	if (tok.type != token::Type::Payload) {
		return expected::unexpected(parser_error::MakeError(
			parser_error::Code::ParseError,
			"Got unexpected token " + tok.TypeToString() + " expected 'data/0000.tar"));
	}
	log::Trace("Parsing the payload");
	payload_index_++;
	return payload::Payload(*(this->lexer_.current.value), manifest);
}
```

Same thing, but structured a little differently

src/mender-update/daemon/states.cpp
```diff
@@ -391,13 +391,24 @@ void UpdateDownloadState::DoDownload(Context &ctx, sm::EventPoster<StateEvent> &
...
 		if (err != error::NoError) {
 			log::Error(err.String());
 			poster.PostEvent(StateEvent::Failure);
 			return;
 		}
 
+		auto exp_payload = ctx.deployment.artifact_parser->Next();
+		if (exp_payload) {
+			log::Error("Multiple payloads are not yet supported in daemon mode.");
+			poster.PostEvent(StateEvent::Failure);
```
oleorhagen, line 404: Q: Does it not have to return? This will post a `Failure` and then a `Success` right after

src/mender-update/daemon/states.cpp
```diff
@@ -391,13 +391,24 @@ void UpdateDownloadState::DoDownload(Context &ctx, sm::EventPoster<StateEvent> &
...
+		} else if (
+			exp_payload.error().code
+			!= artifact::parser_error::MakeError(artifact::parser_error::EOFError, "").code) {
+			log::Error(exp_payload.error().String());
+			poster.PostEvent(StateEvent::Failure);
```
oleorhagen, line 409: Same

src/mender-update/cli/cli_test.cpp
```diff
@@ -41,6 +41,23 @@ namespace processes = mender::common::processes;
 
 using namespace std;
 
+bool VerifyOnlyMessages(const string &output, const vector<string> &messages) {
+	auto lines = common::SplitString(output, "\n");
+	for (auto &line : lines) {
```
oleorhagen, line 46: Nit: `const auto`

src/mender-update/cli/cli_test.cpp
```diff
@@ -238,8 +255,11 @@ TEST(CliTest, ShowProvidesErrors) {
 void SetTestDir(const string &dir, context::MenderContext &ctx) {
 	ctx.GetConfig().paths.SetModulesPath(dir);
 	ctx.GetConfig().paths.SetModulesWorkPath(dir);
-	ctx.GetConfig().paths.SetArtScriptsPath(dir);
-	ctx.GetConfig().paths.SetRootfsScriptsPath(dir);
+
+	string scripts_dir = path::Join(dir, "scripts");
+	filesystem::create_directories(scripts_dir);
```
oleorhagen, line 260: Nit: We have similar functionality in `common`.

oleorhagen:
Some very good finds!

src/common/events/events_io.cpp
kacf: Yeah me too. I know we call the handler directly in some places, but it seems to be working fine. This is just a particular place where it wasn't working, but the event loop trickery is a bit of a special case.

src/artifact/parser.cpp
kacf: Good catch, I'm rearranging the `if` to deal with this.

src/artifact/parser.cpp
kacf: Agreed, this is clearer.

src/mender-update/daemon/states.cpp
kacf: Ooh, good catch! And rookie mistake by me! 😆

kacf:
All handled, and PR is updated!

oleorhagen:
All lights :green_circle:  

:smile:

kacf:
Already passed [here](https://gitlab.com/Northern.tech/Mender/mender-qa/-/pipelines/1073245344), the last update was just to fix formatting. Merging.

PR #1489, opened 2023-11-09
MEN-6671: Emit D-Bus signal both on authentication and unauthentication

lluiscampos:
The old client behaved this way and it seems like `mender-monitor` relies on this behavior to detect being online or offline. Other 3rd party software could also be relying on it.

This fixes the related integration tests.

### DIFF
--- a/src/mender-auth/ipc/server.cpp
+++ b/src/mender-auth/ipc/server.cpp
@@ -92,15 +92,16 @@ void AuthenticatingForwarder::FetchJwtTokenHandler(auth_client::APIResponse &res
 		}
 
 		log::Info("Successfully received new authorization data");
-		dbus_server_.EmitSignal<dbus::StringPair>(
-			"/io/mender/AuthenticationManager",
-			"io.mender.Authentication1",
-			"JwtTokenStateChange",
-			dbus::StringPair {cached_jwt_token_, cached_server_url_});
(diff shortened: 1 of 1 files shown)


src/mender-auth/ipc/server.cpp
```diff
@@ -92,15 +92,16 @@ void AuthenticatingForwarder::FetchJwtTokenHandler(auth_client::APIResponse &res
...
 	} else {
 		ClearCache();
 		log::Error("Failed to fetch new token: " + resp.error().String());
 	}
+	// Emit signal either with valid token and server url or with empty strings
+	dbus_server_.EmitSignal<dbus::StringPair>(
+		"/io/mender/AuthenticationManager",
+		"io.mender.Authentication1",
+		"JwtTokenStateChange",
+		dbus::StringPair {cached_jwt_token_, cached_server_url_});
```
oleorhagen, line 104: Q: Meaning `cached_jwt_token` and `cached_server_url` are empty?

oleorhagen (approved):

src/mender-auth/ipc/server.cpp
lluiscampos: Yes, they are blanked in `ClearCache();` above

kacf (approved):

PR #1493, opened 2023-11-14
docs: mention the deprecation of control maps

TheMeaningfulEngineer:
Ticket: MEN-6648
Changelog: Title

### DIFF
--- a/Documentation/io.mender.Update1.xml
+++ b/Documentation/io.mender.Update1.xml
@@ -6,6 +6,8 @@
     io.mender.Update1:
     @short_description: Mender Update Management API v1
 
+    !! This feature will be removed in Mender client 4.0. For server-side support duration, please refer to our [blog post](https://mender.io/blog/mender-3-6-auto-generation-of-delta-updates#:~:text=Deprecation%3A%20Synchronized%20updates).
+
     This interface lets applications interact with the update flow. It is exposed at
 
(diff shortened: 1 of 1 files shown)


tranchitella (approved):

Documentation/io.mender.Update1.xml
```diff
@@ -6,6 +6,8 @@
     io.mender.Update1:
     @short_description: Mender Update Management API v1
 
+    !! This feature will be removed in Mender client 4.0. For server-side support duration, please refer to our [blog post](https://mender.io/blog/mender-3-6-auto-generation-of-delta-updates#:~:text=Deprecation%3A%20Synchronized%20updates).
```
kacf, line 9: @TheMeaningfulEngineer: This file has been deleted in the `feature-c++-client` branch, which means that this will be deleted too as soon as we merge. Was that perhaps a mistake..?

TheMeaningfulEngineer: I think this should remain in the c++ branch for the sake of documenting the deprecation.
Would remove it only once the go client isn't part of any currently supported release bundle any more.

kacf: Ok, I will restore the file in our branch, which should then merge cleanly with this change.

In the meantime, merging here.

kacf (approved):

PR #1494, opened 2023-11-14
chore: Get rid of direct comparisons with `Error::code`.

kacf:
Such comparisons are dangerous, because if you don't also compare the category, you can get false positives.


oleorhagen (approved):

kacf:
Ignoring spurious failure (it happens in mender-monitor and mender-auth, but this code only changes mender-update).

PR #1497, opened 2023-11-15
chore: Make sure bootstrap artifact is installed before returning info.

kacf:
In a fresh provisioning situation, we need to do so or the info will be empty.

We introduce a new options struct to be able to reuse the existing installation code without printing information, since the bootstrap artifact is a implementation detail and the user might be confused otherwise. Error logs are still printed, so no real info should be lost.


oleorhagen (approved):

oleorhagen:
Looks good :)

PR #1498, opened 2023-11-15
build: Fix broken CMake install directives for dbus interface files.

kacf:


kacf:
@lluiscampos, @oleorhagen: Merging this without review, since it fixes the pipeline, and you are both gone. But I would appreciate a post-merge review anyway!

lluiscampos (commented):
:heavy_check_mark: Looks good

PR #1501, opened 2023-11-17
Implement HTTP chunked Transfer-Encoding support

kacf:


src/common/http/platform/beast/http.cpp
```diff
@@ -255,6 +255,37 @@ size_t GetContentLength(const PARSER &parser) {
...
+	const expected::ExpectedString &content_length,
+	const expected::ExpectedString &transfer_encoding) {
+	bool chunked;
+	if (transfer_encoding) {
+		if (transfer_encoding.value() != "chunked") {
+			return expected::unexpected(error::Error(
+				make_error_condition(errc::not_supported),
+				"Unsupported Transfer-Encoding: " + transfer_encoding.value()));
+		}
+		chunked = true;
```
oleorhagen, line 268: Q: Why not simply return here?

src/common/http_test.cpp
```diff
@@ -672,6 +740,80 @@ TEST(HttpTest, TestResponseBody) {
...
+		},
+		[&received_body, &loop, &count](http::ExpectedIncomingResponsePtr exp_resp) {
+			ASSERT_TRUE(exp_resp) << exp_resp.error().String();
+
+			vector<uint8_t> expected_body;
+			io::ByteWriter expected_writer(expected_body);
+			expected_writer.SetUnlimited(true);
+			io::Copy(expected_writer, *make_shared<BodyOfXes>());
+
+			ASSERT_EQ(received_body.size(), expected_body.size());
```
oleorhagen, line 801: Nit: `EXPECT`

src/common/http/platform/beast/http.cpp
kacf: You're right, I guess my logic is a bit overkill. I'll do that.

src/common/http/platform/beast/http.cpp
```diff
@@ -255,6 +255,31 @@ size_t GetContentLength(const PARSER &parser) {
...
+
+	if (content_length) {
+		auto length = common::StringToLongLong(content_length.value());
+		if (!length || length.value() < 0) {
+			return expected::unexpected(error::Error(
+				length.error().code,
+				"Content-Length contains invalid number: " + content_length.value()));
+		}
+		return length.value() > 0;
+	}
```
lluiscampos, line 278: We can use here also `GetContentLength`, right?

lluiscampos (approved):
It looks good!

As discussed privately, as long as this code does not add regressions on the current tests, we can merge this one and I'll investigate the `mender-gateway` tests failures independently.

src/common/http/platform/beast/http.cpp
```diff
@@ -476,7 +486,7 @@ io::ExpectedAsyncReaderPtr Client::MakeBodyAsyncReader(IncomingResponsePtr resp)
 			"MakeBodyAsyncReader called while reading is in progress"));
 	}
 
-	if (response_body_length_ == 0) {
+	if (GetContentLength(*response_data_.http_response_parser_) == 0) {
```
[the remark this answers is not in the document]
kacf: Yes.... Actually, I'm wondering if that whole block is unnecessary. The `MakeBodyAsyncReader` call will return error if there is no body, so the decision can be entirely based on that.

src/common/http/platform/beast/http.cpp
kacf: Not here, unfortunately. `GetContentLength` operates on `optional<size_t>`, whereas this operates on `ExpectedString`. The reason for this difference is that the former comes from data that Beast has interpreted, whereas the latter comes from raw header data.

lluiscampos: Thank you for clarifying.

src/common/http/platform/beast/http.cpp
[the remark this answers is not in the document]
kacf: @lluiscampos: I looked into this a little bit. It's possible to remove the need for using Content-Length in the resumer code, but then we need to change the Range request to not include an end range. The question is whether it's worth it. The resumer is only used for artifacts, where the length typically is known.

Is it needed for the gateway tests?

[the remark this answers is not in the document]
lluiscampos: > @lluiscampos: I looked into this a little bit. It's possible to remove the need for using Content-Length in the resumer code, but then we need to change the Range request to not include an end range. The question is whether it's worth it. The resumer is only used for artifacts, where the length typically is known.

Makes sense. Thank you for looking into it.

> Is it needed for the gateway tests?

I don't think wo.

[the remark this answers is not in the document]
kacf: Ok, merging this for now then. We can revisit if you run into problems in that area.

PR #1502, opened 2023-11-17
chore(rootfs-image): Add bootloader canary check.

kacf:
Ticket: MEN-6833

### DIFF
--- a/support/modules/rootfs-image
+++ b/support/modules/rootfs-image
@@ -81,6 +81,27 @@ set_upgrade_vars() {
     upgrade_available="${upgrade_available#upgrade_available=}"
 }
 
+check_environment_canary() {
+    mender_check_saveenv_canary="$(${PRINTENV} mender_check_saveenv_canary)"
+    if [ "$mender_check_saveenv_canary" = "mender_check_saveenv_canary=1" ]; then
+	# If the check canary exists (added during build), we need to check the real canary to make
+	# sure that the boot loader was successful in adding it during boot.
+	mender_saveenv_canary="$(${PRINTENV} mender_saveenv_canary)"
+	if [ "$mender_saveenv_canary" != "mender_saveenv_canary=1" ]; then
+	    cat 1>&2 <<'EOF'
+`mender_check_saveenv_canary` was set in the boot environment, but
+`mender_saveenv_canary` was not. This is an indication that the bootloader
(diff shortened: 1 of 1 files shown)


support/modules/rootfs-image
```diff
@@ -81,6 +81,27 @@ set_upgrade_vars() {
...
 
+check_environment_canary() {
+    mender_check_saveenv_canary="$(${PRINTENV} mender_check_saveenv_canary)"
+    if [ "$mender_check_saveenv_canary" = "mender_check_saveenv_canary=1" ]; then
+	# If the check canary exists (added during build), we need to check the real canary to make
+	# sure that the boot loader was successful in adding it during boot.
+	mender_saveenv_canary="$(${PRINTENV} mender_saveenv_canary)"
+	if [ "$mender_saveenv_canary" != "mender_saveenv_canary=1" ]; then
+	    cat 1>&2 <<'EOF'
+`mender_check_saveenv_canary` was set in the boot environment, but
```
oleorhagen, line 92: Careful on the backquotes

support/modules/rootfs-image
```diff
@@ -101,6 +122,7 @@ ensure_correct_root_mounted() {
 
 check_requirements() {
     parse_conf_file
+    check_environment_canary
```
oleorhagen, line 125: We should only warn, no hard exits?

support/modules/rootfs-image
kacf: I already got burned! 😄

That's why `'EOF'` is single-quoted, it disables all expansions.

support/modules/rootfs-image
kacf: `set -e` is set, so this is a hard exit.

support/modules/rootfs-image
oleorhagen: haha, I figured. 

I would still suggest single-quoting this instead though, and remove the verbatim expansion. but YMMV

kacf: I don't think I want to introduce backslashes if we can avoid it, and we don't need expansions here, so I think I'll leave it as is.

oleorhagen (approved):

lluiscampos (commented):
:heavy_check_mark: Looks good

PR #1503, opened 2023-11-17
chore(inventory): Make sure `mender_client_version` is always submitted.

kacf:
Ticket: MEN-6840

### DIFF
--- a/src/mender-update/inventory_test.cpp
+++ b/src/mender-update/inventory_test.cpp
@@ -22,6 +22,7 @@
 
 #include <api/client.hpp>
 #include <common/common.hpp>
+#include <common/conf.hpp>
 #include <common/error.hpp>
 #include <common/events.hpp>
 #include <common/http.hpp>
@@ -34,6 +35,7 @@ using namespace std;
 
 namespace api = mender::api;
 namespace common = mender::common;
+namespace conf = mender::common::conf;
 namespace error = mender::common::error;
 namespace events = mender::common::events;
 namespace http = mender::http;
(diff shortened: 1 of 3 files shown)


lluiscampos (changes requested):
Remember to remove it from [here](https://github.com/mendersoftware/mender/blob/7a793e54e955dfc68cfb466c32526b9c3e3fb630/support/mender-inventory-provides#L11).

Otherwise it looks good! Please run the integration tests again after removing the mentioned lines in the script to make sure the relevant tests still pass :+1:

src/mender-update/inventory_test.cpp
```diff
@@ -96,7 +99,8 @@ exit 0
 	NoAuthHTTPClient client {client_config, loop};
 
 	const string expected_request_data =
-		R"([{"name":"key1","value":["value1","value11"]},{"name":"key2","value":"value2"},{"name":"key3","value":"value3"}])";
+		R"([{"name":"key1","value":["value1","value11"]},{"name":"key2","value":"value2"},{"name":"key3","value":"value3"},{"name":"mender_client_version","value":["additional_version",")"
```
oleorhagen, line 102: Why are we submitting multiple? :thinking:

kacf: Honestly I don't really have a usecase for this. But it was the only way I could think of which doesn't silently hide the existing value if it has also been provided by a script. Most likely if you end up in this situation, you've done something wrong, but this way you will at least be able to see that.

kacf: It's still redundant. I don't think it should be doing it at all.

If two different scripts provided the value, then we'd also make an array, so this is consistent with that.

oleorhagen: :+1:  Alright. Let's hope we don't mess something up :pinched_fingers:

lluiscampos (approved):

PR #1504, opened 2023-11-21
test: Take into account that token is not internally cached anymore.

kacf:
The test was actually passing already, but this was because both
clients would initiate their requests before the round trip calling
`GetJwtToken` would finish, thereby putting the second client in the
`pending_actions_` queue of the first. This caused the number of
tokens requested to still be one, contrary to what one would expect
after disabling caching. And typically not what would happen in
production.

Fix this by forcing the two clients to be called in sequence by stopping the event loop between them. Now `GetJwtToken` is called twice, as expected.


vpodzime (approved):
ACK

vpodzime:
> a race condition

I wouldn't call this a race condition, but not a biggie.

PR #1505, opened 2023-11-21
chore: Fix assert when device is unauthorized.

kacf:
Ticket: MEN-6856

### DIFF
--- a/src/api/client.cpp
+++ b/src/api/client.cpp
@@ -28,7 +28,15 @@ namespace http = mender::http;
 namespace log = mender::common::log;
 
 http::ExpectedOutgoingRequestPtr APIRequest::WithAuthData(const auth::AuthData &auth_data) {
-	AssertOrReturnUnexpected(auth_data.server_url != "");
+	if (auth_data.server_url == "") {
+		if (auth_data.token != "") {
+			return expected::unexpected(
+				auth::MakeError(auth::APIError, "Token did not come with a server URL"));
+		} else {
+			return expected::unexpected(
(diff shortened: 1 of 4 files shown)


lluiscampos (approved):

src/api/client.cpp
```diff
@@ -28,7 +28,15 @@ namespace http = mender::http;
 namespace log = mender::common::log;
 
 http::ExpectedOutgoingRequestPtr APIRequest::WithAuthData(const auth::AuthData &auth_data) {
-	AssertOrReturnUnexpected(auth_data.server_url != "");
+	if (auth_data.server_url == "") {
+		if (auth_data.token != "") {
+			return expected::unexpected(
+				auth::MakeError(auth::APIError, "Token did not come with a server URL"));
+		} else {
+			return expected::unexpected(
```
oleorhagen, line 36: Q: Should we at least log that it is due to a missing `server_url` say debug?

kacf: It's not due to a missing `server_url`, it's due to not being authorized, in which case mender-authd returns an empty server URL, even if it has one available to it.

oleorhagen: Yes, semantically. But here, in this function, that is what is happening, no?

kacf: Sure, but do we need to print it? This is more of an implementation detail IMHO, and it's easy to find where this error came from, since it's unique. It's also because this code path is so common (all unauthorized clients end up here, printing it over and over), that I'm a bit reluctant to print more than we need.

PR #1506, opened 2023-11-21
minor fixes

oleorhagen:
Just some minor things I found during testing


src/common/path/platform/c++17/path.cpp
```diff
@@ -97,7 +97,7 @@ error::Error Permissions(const string &file_path, const vector<Perms> perms) {
 	if (perms.size() == 0) {
 		return error::NoError;
 	}
-	fs::perms p;
+	fs::perms p {};
```
kacf, line 100: Is there a difference?

src/common/crypto/platform/openssl/crypto.cpp
```diff
@@ -478,21 +478,48 @@ expected::ExpectedString ExtractPublicKey(const Args &args) {
 	if (ret != OPENSSL_SUCCESS) {
 		return expected::unexpected(MakeError(
 			SetupError,
-			"Failed to extract the public key from: (" + args.private_key_path
+			"Failed to extract the public key from bio: (" + args.private_key_path
 				+ "): OpenSSL BIO write failed: " + GetOpenSSLErrorMessage()));
 	}
 
-	int pending = BIO_ctrl_pending(bio_public_key.get());
+	EVP_PKEY *evp_public_key = PEM_read_bio_PUBKEY(bio_public_key.get(), NULL, NULL, NULL);
```
kacf, line 485: I assume this does not need freeing, since you're grabbing it from `bio_public_key`, but can it be made const then? Makes the intention a bit clearer.

That is, unless `PEM_write_bio_PUBKEY` below requires it to be non-const.

src/common/crypto/platform/openssl/crypto.cpp
```diff
@@ -478,21 +478,48 @@ expected::ExpectedString ExtractPublicKey(const Args &args) {
...
+		return expected::unexpected(MakeError(
+			SetupError,
+			"Failed to extract the public key from the bio key " + args.private_key_path
+				+ "):" + GetOpenSSLErrorMessage()));
+	}
+
+	auto bio_public_key_new =
+		unique_ptr<BIO, void (*)(BIO *)>(BIO_new(BIO_s_mem()), bio_free_all_func);
+
+	if (!bio_public_key_new.get()) {
```
kacf, line 497: Suggestion: `if (!bio_public_key_new)`.

src/common/crypto/platform/openssl/crypto.cpp
```diff
@@ -478,21 +478,48 @@ expected::ExpectedString ExtractPublicKey(const Args &args) {
...
 			SetupError,
-			"Failed to extract the public key from: (" + args.private_key_path
+			"Failed to extract the public key from bio: (" + args.private_key_path
 				+ "): OpenSSL BIO write failed: " + GetOpenSSLErrorMessage()));
 	}
 
-	int pending = BIO_ctrl_pending(bio_public_key.get());
+	EVP_PKEY *evp_public_key = PEM_read_bio_PUBKEY(bio_public_key.get(), NULL, NULL, NULL);
+
+	if (evp_public_key == NULL) {
```
oleorhagen, line 487: Like you probably can tell. This is not the proper way of doing this. However - it works - and it allows us to move on.

I will suggest to create a ticket, which digs deeper into this later, and does use the default provider for writing the public key to PKCS#8. 

Then move on with this for now

kacf (approved):

src/common/crypto/platform/openssl/crypto.cpp
kacf, line 487: Oh, and `nullptr`.

src/common/path/platform/c++17/path.cpp
oleorhagen: Was just CI complaining about it. This fixed it...

So I'm guessing a false positive? :shrug:

kacf: Hmm, probably. I think theoretically there could be a difference if the class has a `initializer_list` constructor, in which case it might be called with an empty one, instead of the constructor with no arguments (not 100% sure). Any well behaved class should not distinguish between those two cases though.

src/common/crypto/platform/openssl/crypto.cpp
oleorhagen: It does need freeing, I was too quick. Thx!

src/common/crypto/platform/openssl/crypto.cpp
oleorhagen: Discussed f2f. This is fine.

Will create a task, which maybe an intern can have a look at at some point.

oleorhagen:
@kacf I pushed another commit for the `ed25519` fix. Please have a :eyes:

src/common/crypto/platform/openssl/crypto.cpp
```diff
@@ -544,16 +544,54 @@ expected::ExpectedString ExtractPublicKey(const Args &args) {
...
+static expected::ExpectedBytes SignED25519(EVP_PKEY *pkey, const vector<uint8_t> &digest) {
+	size_t sig_len;
+
+	auto md_ctx = unique_ptr<EVP_MD_CTX, void (*)(EVP_MD_CTX *)>(EVP_MD_CTX_new(), EVP_MD_CTX_free);
+	if (md_ctx == nullptr) {
+		return expected::unexpected(MakeError(
+			SetupError, "Failed to initialize the OpenSSL md_ctx: " + GetOpenSSLErrorMessage()));
+	}
+
+	int ret {EVP_DigestSignInit(md_ctx.get(), NULL, NULL, NULL, pkey)};
```
kacf, line 556: `nullptr`. Maybe we should just do a mass replace in this file.

src/common/crypto/platform/openssl/crypto.cpp
```diff
@@ -600,15 +656,7 @@ expected::ExpectedString Sign(const Args &args, const mender::sha::SHA &shasum)
 }
 
 expected::ExpectedString SignRawData(const Args &args, const vector<uint8_t> &raw_data) {
```
kacf, line 658: This function seems redundant now, it just calls the inner function with the same arguments. Can it be removed?

src/common/crypto/platform/openssl/crypto.cpp
```diff
@@ -544,16 +544,54 @@ expected::ExpectedString ExtractPublicKey(const Args &args) {
 	return string(key_vector.begin(), key_vector.end());
 }
 
-expected::ExpectedBytes SignData(const Args &args, const vector<uint8_t> &digest) {
-	auto exp_private_key = PrivateKey::Load(args);
-	if (!exp_private_key) {
-		return expected::unexpected(exp_private_key.error());
+static expected::ExpectedBytes SignED25519(EVP_PKEY *pkey, const vector<uint8_t> &digest) {
```
kacf, line 547: I'm confused, both `SignED25519` and `SignGeneric` take the same input, so why is this called `digest` (`raw_data` is passed in)? `SignGeneric` does perform a digest step internally, but this seems unrelated to what its input argument is.

Maybe a better way is to use the same argument names, but to call the functions `SignMessage` and `SignDigest`, respectively.

src/common/crypto/platform/openssl/crypto.cpp
oleorhagen: Yes it can. 

I seem to recall we had a discussion about this naming with @lluiscampos when it was first introduced.

Which name do you prefer?

kacf: I'd call it just `Sign`, since we sign both raw data and digest data using this function, depending on which key type it is.

kacf (approved):

oleorhagen:
@mender-test-bot start pipeline --pr integration/2493 --pr meta-mender/2002

src/common/crypto/platform/openssl/crypto.cpp
```diff
@@ -478,21 +478,61 @@ expected::ExpectedString ExtractPublicKey(const Args &args) {
 	if (ret != OPENSSL_SUCCESS) {
 		return expected::unexpected(MakeError(
 			SetupError,
-			"Failed to extract the public key from: (" + args.private_key_path
+			"Failed to extract the public key from bio: (" + args.private_key_path
```
lluiscampos, line 481: To be consistent with the message below, shouldn't this be "... from _the_ bio _key_:" ? If it is intended to be different, we should explicitly say something else.

src/common/crypto/platform/openssl/crypto.cpp
```diff
@@ -478,21 +478,61 @@ expected::ExpectedString ExtractPublicKey(const Args &args) {
...
+				+ "):" + GetOpenSSLErrorMessage()));
+	}
+
+	auto bio_public_key_new =
+		unique_ptr<BIO, void (*)(BIO *)>(BIO_new(BIO_s_mem()), bio_free_all_func);
+
+	if (bio_public_key_new == nullptr) {
+		return expected::unexpected(MakeError(
+			SetupError,
+			"Failed to extract the public key from the public key " + args.private_key_path
```
lluiscampos, line 513: ".. from the public key" +` args.private_key_path`?

src/common/crypto/platform/openssl/crypto.cpp
lluiscampos: Removing it makes sense to me :+1:

lluiscampos (approved):
Not that I have a good proposal, but I think that the error messages could be improved a bit. At the high level I would like to clearly differentiate between errors initializing things and errors on operations.

src/common/crypto/platform/openssl/crypto.cpp
oleorhagen: All open to good suggestions here

oleorhagen:
@mender-test-bot start pipeline --pr integration/2493  --pr meta-mender/2002

oleorhagen:
@mender-test-bot start pipeline  --pr integration/2493   --pr meta-mender/2002

oleorhagen:
New try with kirkstone-next: https://gitlab.com/Northern.tech/Mender/mender-qa/-/pipelines/1083754382

src/common/crypto/platform/openssl/crypto.cpp
oleorhagen: I changed them all to `from private key `. 

Then the `GetOpenSSLError` will hopefully return the useful parts.

oleorhagen:
@mender-test-bot start pipeline --pr integration/2493

src/common/crypto.hpp
```diff
@@ -108,9 +108,9 @@ expected::ExpectedString EncodeBase64(vector<uint8_t> to_encode);
 
 expected::ExpectedBytes DecodeBase64(string to_decode);
 
-expected::ExpectedString Sign(const Args &args, const sha::SHA &shasum);
+expected::ExpectedString Sign(const Args &args, const vector<uint8_t> &raw_data);
 
-expected::ExpectedString SignRawData(const Args &args, const vector<uint8_t> &raw_data);
+expected::ExpectedString Sign(const Args &args, const vector<uint8_t> &raw_data);
```
lluiscampos, line 113: This declaration now duplicated, we can just remove.

src/common/crypto/platform/openssl/crypto.cpp
```diff
@@ -478,21 +478,61 @@ expected::ExpectedString ExtractPublicKey(const Args &args) {
...
```
lluiscampos, line 513: This line still says "from the public key" + args.private_key

lluiscampos (approved):

src/common/crypto/platform/openssl/crypto.cpp
oleorhagen: Yes, I unified them to this. I couldn't really come up with something better. The error shows up in the error message, so it felt most right this way. But open to suggestions

PR #1507, opened 2023-11-22
MEN-6671: Pass the actual keepalive user setting to the HTTP configuration

lluiscampos:


src/common/conf/conf.cpp
```diff
@@ -261,6 +261,7 @@ expected::ExpectedSize MenderConfig::ProcessCmdlineArgs(
 	http_client_config_.client_cert_path = https_client.certificate;
 	http_client_config_.client_cert_key_path = https_client.key;
 	http_client_config_.skip_verify = skip_verify;
+	http_client_config_.disable_keep_alive = connectivity.disable_keep_alive;
```
kacf, line 264: Argh! I was so meticulous when I went through the config. Did I really miss it?? :sob:

kacf (approved):

src/common/conf/conf.cpp
lluiscampos: This is one out of many, your catch ratio was still very high!

oleorhagen: There was also a `ssl_engine` setting missing. Just to throw some more coal on the fire :p

oleorhagen (approved):

src/common/conf/conf.cpp
kacf: Haha, you couldn't pass up that chance, could you! 😃

lluiscampos:
As discussed in person, the user configuration `DisableKeepAlive` refers to HTTP Keep-Alive, not TCP keepalive. This "fix" is therefore unnecessary. I'll follow-up in other PRs

src/common/conf/conf.cpp
oleorhagen: hahaha, no :laughing:

PR #1508, opened 2023-11-22
feat: recover from a corrupted db and save a working version.

merlin-northern:
Changelog: Title
Ticket: MEN-6848

### DIFF
--- a/store/dbstore.go
+++ b/store/dbstore.go
@@ -26,7 +26,8 @@ import (
 )
 
 const (
-	DBStoreName = "mender-store"
+	DBStoreName        = "mender-store"
+	dbFileBrokenPrefix = "-broken"
 )
 
 var (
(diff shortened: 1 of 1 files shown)


merlin-northern:
@tranchitella give me early review please, there are no tests (yet).

store/dbstore.go
```diff
@@ -26,7 +26,9 @@ import (
 )
 
 const (
-	DBStoreName = "mender-store"
+	DBStoreName         = "mender-store"
+	dbFileBrokenPrefix  = "-broken"
+	dbFileWorkingPrefix = "-working"
```
tranchitella, line 31: As mentioned on Slack, I'd try to minimize the diff and keep the change concise. We don't need to backup the working database, so I'd remove all this part. All we need to do is to rename `mender-store` into something else (e.g. `mender-store-broken`) and create a new empty database if opening the LMDB fails. That's it, nothing else.

tranchitella (changes requested):

store/dbstore.go
```diff
@@ -61,23 +63,95 @@ func NewDBStore(dirpath string) *DBStore {
...
+				log.Debugf("copying %s->%s: %+v", currentDbPath, workingDbPath, err)
+				if err != nil {
+					os.RemoveAll(workingDbPath)
+				} else {
+					log.Infof("Saved a working DB backup in %s", workingDbPath)
+				}
+			}
+		}
+	}
+}
```
tranchitella, line 153: I don't think we need this whole func.

merlin-northern: removed.

store/dbstore.go
merlin-northern: fixed.

store/dbstore.go
```diff
@@ -61,16 +62,50 @@ func NewDBStore(dirpath string) *DBStore {
...
+			}
+			if openErr := newEnv.Open(
+				currentDbPath,
+				lmdb.NoSubdir|noSyncFlag,
+				0600,
+			); openErr != nil {
+				log.Errorf("Failed to open empty DB environment: %v", err)
+				return nil
+			}
+			log.Info("Started with a fresh DB due to a corruption of the previous one.")
```
tranchitella, line 102:
```suggestion
			log.Info("Started with an empty DB because the original one is corrupted.")
```

store/dbstore.go
```diff
@@ -61,16 +62,50 @@ func NewDBStore(dirpath string) *DBStore {
...
+			log.Infof("Moving %s to %s", currentDbPath, brokenDbPath)
+			_ = os.RemoveAll(brokenDbPath)
+			renameErr := os.Rename(currentDbPath, brokenDbPath)
+			if renameErr != nil {
+				log.Errorf("Failed to move %s to %s: %v", currentDbPath, brokenDbPath, err)
+				return nil
+			}
+			newEnv, err := lmdb.NewEnv()
+			if err != nil {
+				log.Errorf("Failed to create DB environment: %v", err)
```
tranchitella, line 91:
```suggestion
				log.Errorf("Failed to create the DB environment: %v", err)
```

store/dbstore.go
```diff
@@ -61,16 +62,50 @@ func NewDBStore(dirpath string) *DBStore {
...
+				return nil
+			}
+			if openErr := newEnv.Open(
+				currentDbPath,
+				lmdb.NoSubdir|noSyncFlag,
+				0600,
+			); openErr != nil {
+				log.Errorf("Failed to open empty DB environment: %v", err)
```
tranchitella, line 99:
```suggestion
				log.Errorf("Failed to open an empty DB environment: %v", err)
```

tranchitella (approved):

store/dbstore.go
merlin-northern: fixed, but we do not refer to it as "the" anywhere in the logs.

store/dbstore.go
merlin-northern: fixed.

merlin-northern:

PR #1509, opened 2023-11-22
MEN-6671: Close TCP socket when HTTP request/response is completed

lluiscampos:
By  calling `DoCancel()` before calling the user handler in all cases
except with a successful 101 Switching Protocols response.

As the user handler will use the `response_`, remove the reset of if
(and of the `request_`) from the `DoCancel()`. This has no side effects
as these classes have noop as destructors and the objects will anyway
later be set to the new request/response in subsequent calls.

### DIFF
--- a/src/common/http/platform/beast/http.cpp
+++ b/src/common/http/platform/beast/http.cpp
@@ -569,10 +569,8 @@ void Client::CallErrorHandler(
 
 void Client::CallErrorHandler(
 	const error::Error &err, const OutgoingRequestPtr &req, ResponseHandler handler) {
-	*cancelled_ = true;
-	cancelled_ = make_shared<bool>(true);
-	stream_.reset();
 	status_ = TransactionStatus::Done;
+	DoCancel();
 	handler(expected::unexpected(
 		err.WithContext(MethodToString(req->method_) + " " + req->orig_address_)));
 }
@@ -967,13 +965,12 @@ void Client::ReadHeaderHandler(const error_code &ec, size_t num_read) {
 		CallHandler(header_handler_);
 		if (!*cancelled) {
 			status_ = TransactionStatus::Done;
+			if (response_->status_code_ != StatusCode::StatusSwitchingProtocols) {
+				// Make an exception for 101 Switching Protocols response, where the TCP connection
+				// is meant to be reused.
+				DoCancel();
+			}
 			CallHandler(body_handler_);
-
-			// After body handler has run, set the request to cancelled. The body
-			// handler may have made a new request, so this is not necessarily the same
-			// request as is currently active (note use of shared_ptr copy, not
-			// `cancelled_`).
-			*cancelled = true;
 		}
 		return;
 	}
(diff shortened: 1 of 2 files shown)


lluiscampos:
@kacf This is my overkill approach. It should be possible to do it on the reader level, but I couldn't make it work.

src/api/auth/auth.cpp
```diff
@@ -103,6 +103,13 @@ static void TryAuthenticate(
 	const string signature,
 	APIResponseHandler api_handler);
 
+static void callAPiHandler(
```
kacf, line 106: Leading CamelCase.

src/api/auth/auth.cpp
```diff
@@ -103,6 +103,13 @@ static void TryAuthenticate(
...
+	mender::http::Client &client, APIResponseHandler api_handler, ExpectedAuthData exp) {
+	// Explicitly cancel the HTTP client to ensure that the socket gets closed
+	api_handler(exp);
+	client.Cancel();
```
kacf, line 110: A couple of things:

* I would also comment that we are not interested in the body.
* It might better to cancel before calling the handler, if possible, because the handler may start new requests.
* Keep in mind that cancelling may trigger cancellation messages elsewhere in the code, so make sure you try this in practice a bit.

kacf (approved):

src/api/auth/auth.cpp
lluiscampos: > A couple of things:
> 
>     * I would also comment that we are not interested in the body.

We do read the body. We were miss-lead yesterday when we looked at the code together but the whole body is read (and the json response from the server is logged) before calling the user.

>     * It might better to cancel before calling the handler, if possible, because the handler may start new requests.
> 

I tried this first, but the tests would segfault.

The segfault and the fact that the reader(s) are not properly destroyed made me suspect that the issue is the resistivity of this code, where a failed auth triggers a new fetch (with the potential next server url) passing around the handlers. In other words I would expect that we don't need to do `client.Cancel()` here but I couldn't find a different way to fix this problem.

kacf: > We do read the body. We were miss-lead yesterday when we looked at the code together but the whole body is read (and the json response from the server is logged) before calling the user.

We do? But then this is not the appropriate fix. It means there is a bug somewhere which prevents closing the socket on a completed request. We should find where that is so it can be fixed centrally.

vpodzime: Yeah, to me having to call `client.Cancel()` here explicitly looks like a bug.

lluiscampos: The bug was indeed somewhere else. Rework coming up :eyes:

lluiscampos: Thank you both for being skeptical about this "fix". A different approach is coming up in the next force-push :eyes:

kacf (approved):

kacf:
Much better fix!

vpodzime (commented):
Just one "high-level" comment -- the word `cancel` is not a good fit (anymore).

PR #1510, opened 2023-11-22
MEN-6862: Deprecate `Connectivity` settings on the client     

lluiscampos:


lluiscampos:
@TheMeaningfulEngineer @tranchitella Please review the changelog entry :eyes:

src/common/config_parser.hpp
```diff
@@ -45,14 +45,6 @@ struct ClientSecurity {
 	string ssl_engine;
 };
 
-/** Connectivity instructs the client how we want to treat the keep alive
-	connections and when a connection is considered idle and therefore closed */
-struct ClientConnectivity {
-	bool disable_keep_alive = false;
-	// Removed in Mender v4.0.0, because we don't have a connection cache there.
-	// int idle_conn_timeout_seconds = 0;
-};
```
kacf, line 54: Should we also keep the commented out struct? No strong opinion, I realize it can be a bit messy too, but maybe good for historical reasons?

kacf (approved):

tranchitella (approved):

src/common/config_parser.hpp
lluiscampos: I kept it commented out now.

PR #1512, opened 2023-11-23
Various test fixes

kacf:


src/common/path/platform/c++17/path.cpp
```diff
@@ -75,6 +75,17 @@ error::Error FileDelete(const string &path) {
 	return error::NoError;
 }
 
+error::Error DeleteRecursively(const string &path) {
+	assert(string(path.begin(), path.begin() + 4) == "/tmp");
```
oleorhagen, line 79: ?

oleorhagen: Only supported for `tmp`?

src/mender-update/daemon/state_test.cpp
```diff
@@ -3109,7 +3112,16 @@ class StateDeathTest : public testing::TestWithParam<StateTransitionsTestCase> {
 				"artifact-name",
 				"--output-path",
 				ArtifactPath(),
-			});
+			};
+
+			auto scripts = MakeTestArtifactScripts(
+				GetParam(), tmpdir.Path(), path::Join(tmpdir.Path(), "state.log"));
+			for (auto &script : scripts) {
```
oleorhagen, line 3119: nit: const

src/mender-update/daemon/states.cpp
```diff
@@ -292,9 +292,18 @@ void UpdateDownloadState::OnEnter(Context &ctx, sm::EventPoster<StateEvent> &pos
 }
 
 void UpdateDownloadState::ParseArtifact(Context &ctx, sm::EventPoster<StateEvent> &poster) {
+	string art_scripts_path = ctx.mender_context.GetConfig().paths.GetArtScriptsPath();
+
+	// Clear the artifact scripts directory so we don't risk old scripts lingering.
+	auto err = path::DeleteRecursively(art_scripts_path);
```
oleorhagen, line 298: Good thing we figured this out at the end :sweat:

oleorhagen (approved):

src/common/path/platform/c++17/path.cpp
kacf: Haha, this was my safeguard!

src/mender-update/daemon/state_test.cpp
kacf: Nah, not for local variables.

kacf:
Original pipeline [here](https://gitlab.com/Northern.tech/Mender/mender-qa/-/pipelines/1082939839), only unit test fixes in the latest push.

src/mender-update/daemon/state_test.cpp
oleorhagen: I was thinking about the loop

kacf: Yes, but it's also a local variable. It doesn't have any effect outside of this function, so I think it's overkill to put it everywhere.

oleorhagen: I've always thought of it as a readability thing, but I know you're a pragmatist here :p

kacf: Gah, you make an annoyingly good argument! I admit I had not thought of it from that angle, and I guess it does express intent. Ok then, you convinced me, `const` it is!

oleorhagen: hahaha :tada:

PR #1514, opened 2023-11-28
feat: recover from a corrupted db and save a working version.

tranchitella:
Changelog: Title
Ticket: MEN-6848
(cherry picked from commit edeb192efa3fa613924fd18b3fef577e5bc33f1e)

### DIFF
--- a/cli/setup.go
+++ b/cli/setup.go
@@ -922,7 +922,7 @@ func (opts *setupOptionsType) maybeAddHostLookup() {
 	}
 
 	// Seek to last character
-	_, err = f.Seek(-1, os.SEEK_END)
+	_, err = f.Seek(-1, io.SeekEnd)
 	if err != nil {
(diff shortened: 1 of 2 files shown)


cli/setup.go
```diff
@@ -922,7 +922,7 @@ func (opts *setupOptionsType) maybeAddHostLookup() {
 	}
 
 	// Seek to last character
-	_, err = f.Seek(-1, os.SEEK_END)
+	_, err = f.Seek(-1, io.SeekEnd)
```
merlin-northern, line 925: and were did that come from? we also want to cp it?

merlin-northern (approved):
one comment; one change I do not recognize, otherwise looks good to me.

cli/setup.go
tranchitella: PIpeline fails without this change because newer golang deprecafted io.SEEK_END

PR #1518, opened 2023-11-29
chore: Fix missing deployment logs in certain cases.

kacf:
I haven't been able to reproduce this in our CI, only locally, so it's possible that the behavior has changed between libc versions. Checking for existence should be safe in either case though.

Our integration tests check for the submission of deployment logs for failed deployments already.

Ticket: MEN-6867

### DIFF
--- a/src/mender-update/deployments/platform/boost_log/deployments.cpp
+++ b/src/mender-update/deployments/platform/boost_log/deployments.cpp
@@ -81,10 +81,14 @@ error::Error DeploymentLog::PrepareLogDirectory() {
 
 error::Error DeploymentLog::DoPrepareLogDirectory() {
 	fs::path dir_path(data_store_dir_);
-	bool created = fs::create_directories(dir_path);
-	if (created) {
-		// should basically never happen, but if we happened to create the
-		// directory, it's empty and thus well-prepared
+
+	// `create_directories` may fail on some platforms with an error if the destination exists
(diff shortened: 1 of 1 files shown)


src/mender-update/deployments/platform/boost_log/deployments.cpp
```diff
@@ -81,10 +81,14 @@ error::Error DeploymentLog::PrepareLogDirectory() {
...
 	fs::path dir_path(data_store_dir_);
-	bool created = fs::create_directories(dir_path);
-	if (created) {
-		// should basically never happen, but if we happened to create the
-		// directory, it's empty and thus well-prepared
+
+	// `create_directories` may fail on some platforms with an error if the destination exists
+	// as a symlink, even if the target is a valid directory. So let's check for existence
+	// first.
+	if (not fs::exists(dir_path)) {
```
oleorhagen, line 88: I'm so proud I'm in tears :smiling_face_with_tear:

oleorhagen (approved):

src/mender-update/deployments/platform/boost_log/deployments.cpp
kacf: Hehe, the `not` is nice!

lluiscampos: I have to confess that I started using it also (as per yesterday on my local development). It is amazing how the `not` `not` nice nitpicks spread.

PR #1519, opened 2023-11-29
fix: Make partition checking more robust in every state.

kacf:
```
commit fde5b0469a7f2bd2c81638c1ded553d0bd233512
Date:   Tue Nov 28 12:45:02 2023

    chore: Add a message when deployment has finished.
    
commit ab31b5626362b91ee3bbc0eaf8f81caad5d56f61
Date:   Tue Nov 28 09:40:04 2023

    fix: Make partition checking more robust in every state.
    
    In particular, this prevents the client from erronously committing a
    rootfs-image update if it's paused in the ArtifactReboot state, and
    then restarted through `systemctl restart` instead of through a
    reboot. Now it checks that the root partition is what we expect.
    
    This has been a bug in the Golang client too, so we want a changelog
    entry for this.
    
    Changelog: The client no longer erronously commits a rootfs-image
    artifact after being restarted using `systemctl restart` in the
    `ArtifactReboot` state.
    
    Ticket: MEN-6633
    
```

### DIFF
--- a/support/modules/rootfs-image
+++ b/support/modules/rootfs-image
@@ -107,7 +107,7 @@ EOF
     return 0
 }
 
-device_matches_root() {
+check_device_matches_root() {
     case "$1" in
         /dev/ubi*)
             # UBI is a bit peculiar. Trying to take the device number of the root device does not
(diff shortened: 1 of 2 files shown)


support/modules/rootfs-image
```diff
@@ -107,7 +107,7 @@ EOF
     return 0
 }
 
-device_matches_root() {
+check_device_matches_root() {
```
oleorhagen, line 110: Thanks! Much nicer imo!

oleorhagen (approved):

PR #1520, opened 2023-11-29
chore: Remove in-memory database.

kacf:
It was used heavily in the Golang client for testing, but we ended up never using it in the C++ client.


kacf:
This database was never used in production, so I think I'll skip the mender-qa pipeline for this one.

oleorhagen (approved):

PR #1521, opened 2023-11-29
chore: Back up and recreate database in case of database corruption.

kacf:
Ticket: MEN-6860

The second commit is the heart of the fix.

### DIFF
--- a/src/common/key_value_database/platform/lmdb/lmdb.cpp
+++ b/src/common/key_value_database/platform/lmdb/lmdb.cpp
@@ -79,8 +79,7 @@ error::Error LmdbTransaction::Remove(const string &key) {
 	}
 }
 
-KeyValueDatabaseLmdb::KeyValueDatabaseLmdb() :
-	env_ {make_unique<lmdb::env>(lmdb::env::create())} {
+KeyValueDatabaseLmdb::KeyValueDatabaseLmdb() {
 }
 
 KeyValueDatabaseLmdb::~KeyValueDatabaseLmdb() {
@@ -90,23 +89,20 @@ KeyValueDatabaseLmdb::~KeyValueDatabaseLmdb() {
 error::Error KeyValueDatabaseLmdb::Open(const string &path) {
 	Close();
 
+	env_ = make_unique<lmdb::env>(lmdb::env::create());
+
 	try {
 		env_->open(path.c_str(), MDB_NOSUBDIR, 0600);
 	} catch (std::runtime_error &e) {
+		env_.reset();
 		return MakeError(LmdbError, e.what());
 	}
 
-	successfully_opened_ = true;
 	return error::NoError;
 }
 
 void KeyValueDatabaseLmdb::Close() {
-	if (!successfully_opened_) {
-		return;
-	}
-
-	env_->close();
-	successfully_opened_ = false;
+	env_.reset();
 }
 
 expected::ExpectedBytes KeyValueDatabaseLmdb::Read(const string &key) {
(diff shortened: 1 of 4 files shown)


src/common/key_value_database/platform/lmdb/lmdb.cpp
```diff
@@ -90,23 +89,23 @@ KeyValueDatabaseLmdb::~KeyValueDatabaseLmdb() {
...
 
-	successfully_opened_ = true;
 	return error::NoError;
 }
 
 void KeyValueDatabaseLmdb::Close() {
-	if (!successfully_opened_) {
-		return;
+	if (env_) {
+		env_->close();
```
oleorhagen, line 151: Q: Is the env not closed when it goes out of scope?

oleorhagen (approved):

src/common/key_value_database/platform/lmdb/lmdb.cpp
kacf: Oh, you're right, this is redundant now.

kacf:
Original mender-qa pipeline [here](https://gitlab.com/Northern.tech/Mender/mender-qa/-/pipelines/1088813812). I'm not going to rerun for the tiny cleanups.

src/common/key_value_database/platform/lmdb/lmdb.cpp
```diff
@@ -79,34 +89,65 @@ error::Error LmdbTransaction::Remove(const string &key) {
...
+
+			auto err2 = OpenInternal(path, false);
+			if (err2 != error::NoError) {
+				return err.FollowedBy(err2);
+			}
+			return error::NoError;
+		} catch (fs::filesystem_error &e) {
+			env_.reset();
+			return err.FollowedBy(error::Error(e.code().default_error_condition(), e.what())
+									  .WithContext("Opening LMDB database failed"));
```
lluiscampos, line 138: Shouldn't this be also `LmdbError`?
```suggestion
			return err.FollowedBy(MakeError(LmdbError, e.what())
									  .WithContext("Opening LMDB database failed"));
```

oleorhagen (approved):

lluiscampos (approved):

src/common/key_value_database/platform/lmdb/lmdb.cpp
```diff
@@ -12,15 +12,25 @@
 //    See the License for the specific language governing permissions and
 //    limitations under the License.
 
-#include <common/common.hpp>
 #include <common/key_value_database_lmdb.hpp>
 
+#include <filesystem>
```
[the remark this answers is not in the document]
kacf: You mean `lmdb++.h`? It is actually a C++ header, they just didn't adopt the `hpp` naming.

src/common/key_value_database/platform/lmdb/lmdb.cpp
kacf: The error doesn't come from LMDB though, it comes from our file manipulation, outside of the database, so I thought it's better to be explicit about that.

Strictly speaking the exception category below (which I think cannot happen, but I didn't want to risk not having it), is not an `LmdbError` either, but I couldn't find a better one to give it, since `runtime_error` does not contain `code`. But maybe it's better to use `error::GenericError` there, since we have no idea where it comes from then. I'll do that, unless you object.

lluiscampos: Thanks. All things considered, it is fine as it is :+1:

PR #1522, opened 2023-11-29
chore: Add changelog for HSM change

oleorhagen:
Changelog(feat): The client no longer compiles with deprecated OpenSSL APIs.
This means that the OpenSSL HSM Engine API is deprecated for clients compiled with
OpenSSLv3, and the end-user will have to provide a v3 PKCS11 provider for the
interaction with external HSMs. For OpenSSL 1, the Engine API is still supported.


oleorhagen:
Seeing the discussions we've had around the Engine API deprecation in the integration PR, I think it'd be wise to add a proper `Changelog` for this, seeing the old client compiles with the deprecated APIs just fine.

lluiscampos (changes requested):
The changelog is not rendering as you expect, see the comment from the bot. I think it is a bug in the generator is related to setting the changelog type in the changelog entry itself rather than in the commit tittle.

Otherwise, the message is really good!

PR #1523, opened 2023-11-29
MEN-6834: Make all JSON parsing case insensitive

lluiscampos:
Although the JSON standard defines the keys as case sensitive, the
    golang standard library implements it as case insensitive. Therefore to
    avoid regressions we need to make all our JSON parsing case insensitive.
    
    To implement it, we define a custom `nlohmann::basic_json` with a custom
    `std::map` that compares the keys as lowercase. See:
    * https://json.nlohmann.me/api/basic_json/basic_json/
    * https://json.nlohmann.me/api/basic_json/object_t/


src/common/json.hpp
```diff
@@ -46,7 +48,37 @@ enum JsonErrorCode {
...
+public:
+	bool operator()(const string &lhs, const string &rhs) const {
+		return common::StringToLower(lhs) < common::StringToLower(rhs);
+	}
+};
+
+template <class Key, class T, class IgnoredLess, class Allocator = allocator<pair<const Key, T>>>
+class CaseInsensitiveMap : public map<const Key, T, CaseInsensitiveLess, Allocator> {
+public:
+	using CaseInsensitiveMapType = map<const Key, T, CaseInsensitiveLess, Allocator>;
```
lluiscampos, line 61: FYI, I tried (the hard way) with `unordered_map` as I think it would be a better fit (plus we could reuse a tiny piece of code from the HTTP) but although the documentation seems to indicate that it should work, it does not work. See [here](https://github.com/nlohmann/json/issues/1385) or [here](https://github.com/nlohmann/json/issues/164).

lluiscampos:

src/common/config_parser_test.cpp
```diff
@@ -581,3 +581,26 @@ TEST_F(ConfigParserTests, ValidateServerConfig) {
 	EXPECT_THAT(ret.error().String(), testing::HasSubstr("ServerURL"));
 	EXPECT_THAT(ret.error().String(), testing::HasSubstr("Servers"));
 }
+
+TEST_F(ConfigParserTests, CaseInsensitive) {
+	ofstream os(test_config_fname);
+	os << R"({
+  "artifactverifykey": "ArtifactVerifyKey_value",
+  "deviceTypeFile": "DeviceTypeFile_value",
+  "SERVERURL": "ServerURL_value"
```
kacf, line 590: Maybe we should have a separate test where you have the same key with different casing? I'm a little bit interested to see how it handles it, especially when compared to having two completely identical keys.

kacf (approved):
Nice!

src/common/config_parser_test.cpp
lluiscampos: It behaves the same with two identical keys than with same key with different casing: the entry in the map gets overridden and only the last one stands.

I'll add a test, thanks for the suggestion.

lluiscampos:
The last force-push only added a unit-test, not running again the full pipeline. See the last pipeline [here](https://gitlab.com/Northern.tech/Mender/mender-qa/-/pipelines/1089545281).

PR #1524, opened 2023-11-30
feat: Add a termination signal handler the mender-auth

oleorhagen:
This adds a signal handler for SIG{TERM,INT,QUIT} to mender-auth, so that the
daemon exits gracefully in these circumstances by stopping the event loop.

Ticket: MEN-6733
Changelog: None

### DIFF
--- a/src/mender-auth/cli/actions.cpp
+++ b/src/mender-auth/cli/actions.cpp
@@ -143,6 +143,17 @@ error::Error DaemonAction::Execute(context::MenderContext &main_context) {
 
 	events::EventLoop loop {};
 
+	events::SignalHandler signal_handler {loop};
+
+	err = signal_handler.RegisterHandler(
+		{SIGTERM, SIGINT, SIGQUIT}, [&loop](events::SignalNumber signum) {
+			log::Info("Termination signal received, shutting down gracefully");
+			loop.Stop();
+		});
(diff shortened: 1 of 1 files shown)


src/mender-auth/cli/actions.cpp
```diff
@@ -143,6 +143,14 @@ error::Error DaemonAction::Execute(context::MenderContext &main_context) {
 
 	events::EventLoop loop {};
 
+	events::SignalHandler signal_handler {loop};
+
+	signal_handler.RegisterHandler(
+		{SIGTERM, SIGINT, SIGQUIT}, [&loop](events::SignalNumber signum) {
+			log::Info("Termination signal received, shutting down gracefully");
+			loop.Stop();
+		});
```
lluiscampos, line 152: We should check the return value
```suggestion
	err = signal_handler.RegisterHandler(
		{SIGTERM, SIGINT, SIGQUIT}, [&loop](events::SignalNumber signum) {
			log::Info("Termination signal received, shutting down gracefully");
			loop.Stop();
		});
	if (err != error::NoError) {
		return err;
	}
```

lluiscampos (approved):
So simple :sweat_smile: 

See my comment below, but pre-aproving :+1:

oleorhagen:

PR #1531, opened 2023-12-07
MEN-6836: Do not call the user's callback when returning an error

lluiscampos:
The design is that the entry level function `WithToken` should either
return error synchronously or eventually call the handler
asynchronously.

This fixes the issue where `mender-update` will abort with an unhandled
event. What was happening before is that an error while connecting to
the bus would post two events and later the state machine will handle
the second one from the `Idle` state.

### DIFF
--- a/src/api/auth/auth.cpp
+++ b/src/api/auth/auth.cpp
@@ -392,7 +392,6 @@ error::Error Authenticator::WithToken(AuthenticatedAction action) {
 		if (err != error::NoError) {
 			// Should never fail. We rely on the signal heavily so let's fail
 			// hard if it does.
-			action(expected::unexpected(err));
 			return err;
 		}
 	}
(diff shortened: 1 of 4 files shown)


src/api/auth/auth.cpp
```diff
@@ -392,7 +392,6 @@ error::Error Authenticator::WithToken(AuthenticatedAction action) {
 		if (err != error::NoError) {
 			// Should never fail. We rely on the signal heavily so let's fail
 			// hard if it does.
-			action(expected::unexpected(err));
```
lluiscampos, line 395: The test that I wrote verifies this fix.

The other fix below is harder to trigger as we would need somehow a hook in `WithToken` to make the DBus inaccessible _after_ we pass this first part. One hack could be to make a test friend class that could manipulate `watching_token_signal_`, but I regarded that as too ugly to put in production code.

I think it is not worth the effort.

lluiscampos:

src/api/auth/auth.cpp
kacf: Agreed!

kacf (approved):

oleorhagen (commented):
Nice find :)

PR #1541, opened 2023-12-18
Add build time optimizations to minimize for size

oleorhagen:
This adds a few commits to optimize the mender binaries to be optimized for size.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -2,6 +2,13 @@ cmake_minimum_required(VERSION 3.15)
 
 project(mender)
 
+# Set the default build type to MinSizeRel. We care more about size than speed.
+if(NOT CMAKE_BUILD_TYPE)
+  set(CMAKE_BUILD_TYPE MinSizeRel CACHE STRING
+    "Choose the type of build, options are: None Debug Release RelWithDebInfo MinSizeRel."
+    FORCE)
+endif()
+
 add_subdirectory(support)
 add_subdirectory(Documentation)
 
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -9,6 +9,8 @@ if(NOT CMAKE_BUILD_TYPE)
     FORCE)
 endif()
 
+message(STATUS "Building with build type: ${CMAKE_BUILD_TYPE}")
+
 add_subdirectory(support)
 add_subdirectory(Documentation)
 


oleorhagen:
Seems we might have to skip the `-Oz` optimization.

CMakeLists.txt
```diff
@@ -2,6 +2,13 @@ cmake_minimum_required(VERSION 3.15)
 
 project(mender)
 
+# Set the default build type to MinSizeRel. We care more about size than speed.
+if(NOT CMAKE_BUILD_TYPE)
+  set(CMAKE_BUILD_TYPE MinSizeRel CACHE STRING
+    "Choose the type of build, options are: None Debug Release RelWithDebInfo MinSizeRel."
```
vpodzime, line 14: Shouldn't `Release` be changed to yield size-optimized binaries instead?

CMakeLists.txt
```diff
@@ -5,6 +5,9 @@ project(mender)
 # Optimize aggressively for size when building MinSizeRel
 string(APPEND CMAKE_CXX_FLAGS_MINSIZEREL " -Oz")
 
+# Enable Link-Time-Optimizations when building MinSizeRel
+string(APPEND CMAKE_CXX_FLAGS_MINSIZEREL " -flto")
```
vpodzime, line 9: Not sure how portable this is. AFAIK, only recent versions of GCC and Clang can do this (reliably).

vpodzime (commented):
I cannot comment on what is important here. The fact that these are long-running daemons doesn't mean they need to be size-optimized sacrificing performance. One could argue that performance is the key thing for something that always runs on a device. To me at least `-Oz` is too aggressive.

And like I said, I believe `Release` should be the build type for releases and thus what should yield binaries for releasing.

CMakeLists.txt
oleorhagen: Why? `MinSizeRel` is literally a minimum sized release. Which is exactly what we want.

CMakeLists.txt
oleorhagen: Yes, I too am a little unsure about it. I'll check if we can do a check with CMake, and then add it conditionally.

oleorhagen:
> I cannot comment on what is important here. The fact that these are long-running daemons doesn't mean they need to be size-optimized sacrificing performance. One could argue that performance is the key thing for something that always runs on a device. To me at least `-Oz` is too aggressive.
> 
> And like I said, I believe `Release` should be the build type for releases and thus what should yield binaries for releasing.

My impression is that people care about the binary size, at least we are getting quite a few questions around it. Which is why it seems sensible to optimize for it by default.

CMakeLists.txt
oleorhagen: And when I say release, I mean `Release`

CMakeLists.txt
oleorhagen: Updated

vpodzime: Nice!

CMakeLists.txt
lluiscampos: I agree in that optimizing `MinSizeRel` is more correct. A user can always set `CMAKE_BUILD_TYPE` to Release and then I think it is expected to not tamper with these optimizations.

lluiscampos (approved):
I also think that `-Oz` is too much. It may have impact in slower CPUs on constrain devices. Should we drop it? We are always on time to add it later.

Great work - the commit by commit review was a pleasure :star_struck:  Thanks @vpodzime for your input too :+1:

CMakeLists.txt
vpodzime: To me, release binaries and packages should be built with the `Release` build type. And its parameters should determine how the released binaries/packages look like. This also gives more flexibility, take the `-Oz` flag for example -- it makes total sense for `MinSizeRel`, but `Release` may not use it because it's too aggressive. I guess this gets us to the broader discussion below, though.

vpodzime:
> My impression is that people care about the binary size, at least we are getting quite a few questions around it. Which is why it seems sensible to optimize for it by default.

I can't tell, I have no data. However, there's definitely a difference between people worrying about size given their experience with the Go client's binaries and people worrying about every 100K in binary size. I'd say that if people worry about the <100K difference made by `-Oz` sacrificing a bit of performance, they would *way more* prefer a build with `mender-update` being able to authenticate and thus not needing `mender-auth`, DBus, half of the HTTP code, etc. But that's yet another story.

oleorhagen:
> > My impression is that people care about the binary size, at least we are getting quite a few questions around it. Which is why it seems sensible to optimize for it by default.
> 
> I can't tell, I have no data. However, there's definitely a difference between people worrying about size given their experience with the Go client's binaries and people worrying about every 100K in binary size. I'd say that if people worry about the <100K difference made by `-Oz` sacrificing a bit of performance, they would _way more_ prefer a build with `mender-update` being able to authenticate and thus not needing `mender-auth`, DBus, half of the HTTP code, etc. But that's yet another story.

haha, yes. That is another discussion :)

oleorhagen:

oleorhagen:
Merged without the `-lto` optimization added. It made the unit tests hang actually. So something funny going on there...

Didn't bother looking into it rn. merging

PR #1547, opened 2024-01-02
fix(standalone): Make sure to verify the depends and provides

oleorhagen:
Simple fix, making sure that we verify the depends and provides in standalone mode as well. Which was simply forgotten.

Ticket: MEN-6929
Changelog: None


lluiscampos (approved):
We should have a test for this, either here or in the acceptance tests.

Caching it in an integration test of an independent repository is too late IMO.

Otherwise, nicely solved!

kacf (approved):
Nice tests!

src/mender-update/cli/cli_test.cpp
```diff
@@ -463,6 +475,92 @@ artifact_name=test
...
+
+		mtesting::RedirectStreamOutputs output;
+		int exit_status = cli::Main(
+			args, [&tmpdir](context::MenderContext &ctx) { SetTestDir(tmpdir.Path(), ctx); });
+		EXPECT_EQ(exit_status, 1) << exit_status;
+
+		EXPECT_EQ(output.GetCout(), R"(Installing artifact...
+Installation failed. System not modified.
+)");
+		EXPECT_EQ(output.GetCerr(), "");
```
kacf, line 560: No logging at all? Isn't that a bit suspicious?

oleorhagen: Yes. But it is the same for all the tests.

I did a closer look. And the only logging on the path, which is not showing is from the artifact parser. And the reason it is not showing is that the log level is lower than `Info`. So this looks right to me :)

PR #1551, opened 2024-01-05
chore: clean up platform logic in main.cmake

aduskett:
Remove all of the extra if (${PLATFORM} STREQUAL linux_x86) lines in main.cmake and unify them into a single if statement. Logic for other platforms can be added to the else() statement at a later date.


kacf (approved):
👍 Thanks!

PR #1552, opened 2024-01-05
Build: add ccache support

aduskett:
On a system with an AMD 6900HS, ccache dramatically speeds up recompiling from
source.
    
First run: 5m25.182s
Second run: 0m2.373s    
A speedup of 162.5x compared to the first run!


aduskett:
Another thing to look into: If the Mender maintainers were to set up a docker volume with a ccache, build and test times would be cut dramatically. Possibly by more than 50%!

main.cmake
```diff
@@ -44,6 +45,12 @@ option(MENDER_ARTIFACT_GZIP_COMPRESSION "Enable GZIP compression support when do
 option(MENDER_ARTIFACT_LZMA_COMPRESSION "Enable LZMA compression support when downloading and extracting Artifacts (Default: ON)" ON)
 option(MENDER_ARTIFACT_ZSTD_COMPRESSION "Enable Zstd compression support when downloading and extracting Artifacts (Default: ON)" ON)
 
+find_program(CCACHE ccache)
+if(CCACHE)
+  set_property(GLOBAL PROPERTY RULE_LAUNCH_COMPILE "${CCACHE}")
```
kacf, line 50: The cmake docs specify that `RULE_LAUNCH_COMPILE` [should not be used](https://cmake.org/cmake/help/latest/prop_tgt/RULE_LAUNCH_COMPILE.html#rule-launch-compile), and `CMAKE_CXX_COMPILER_LAUNCHER` should be used instead.

main.cmake
```diff
@@ -44,6 +45,12 @@ option(MENDER_ARTIFACT_GZIP_COMPRESSION "Enable GZIP compression support when do
 option(MENDER_ARTIFACT_LZMA_COMPRESSION "Enable LZMA compression support when downloading and extracting Artifacts (Default: ON)" ON)
 option(MENDER_ARTIFACT_ZSTD_COMPRESSION "Enable Zstd compression support when downloading and extracting Artifacts (Default: ON)" ON)
 
+find_program(CCACHE ccache)
+if(CCACHE)
```
kacf, line 49: I'm a little bit uneasy about adding it as the default. I've been using it for a long time already (by specifying `-D CMAKE_CXX_COMPILER_LAUNCHER=ccache`), and most of the time it's been fine. But I have had problems in few weird instances, particularly when using code coverage, where blowing out the entire cache was the only thing that worked. This is rather unexpected if it's the default and you don't know that ccache is being used.

.gitlab-ci.yml
```diff
@@ -80,7 +80,7 @@ test:backward-compat:
   stage: test
   image: ubuntu:22.04
   before_script:
-    - apt update && apt install -yyq clang cmake git make pkg-config liblmdb++-dev libboost-dev libboost-log-dev libssl-dev libarchive-dev libdbus-1-dev curl dbus stunnel4 tinyproxy-bin netcat
+    - apt update && apt install -yyq ccache clang cmake git make pkg-config liblmdb++-dev libboost-dev libboost-log-dev libssl-dev libarchive-dev libdbus-1-dev curl dbus stunnel4 tinyproxy-bin netcat
```
kacf, line 83: This is going to require more changes in our CI to work with a persistent cache. We don't use the shared runners, we have private ones that are provisioned on every run, therefore sharing a cache is a bit complicated. I'm not sure if this is worth the effort, since the runners are quite beefy, and the client is about as big as it's going to get now.

main.cmake
aduskett: Agreed. I can wrap this in if(BUILD_TESTS), but it may be worth looking into if tests are failing because ccache is enabled.

.gitlab-ci.yml
aduskett: Fair enough. It's up to you if you want me to remove ccache from the apt install!

main.cmake
oleorhagen: I too am not a fan of adding this as the default. I've run into some issues with `ccache` myself.

main.cmake
```diff
@@ -44,6 +45,17 @@ option(MENDER_ARTIFACT_GZIP_COMPRESSION "Enable GZIP compression support when do
...
 
+# Coverage may fail when building tests.
+if(NOT BUILD_TESTS)
+  find_program(CCACHE ccache)
+  if(CCACHE)
+    message(STATUS "Found ccache at ${CCACHE}")
+    set(CMAKE_C_COMPILER_LAUNCHER ${CCACHE})
+    set(CMAKE_CXX_COMPILER_LAUNCHER ${CCACHE})
+  endif()
+endif()
```
kacf, line 56: I don't think we should enable it by default at all. May I suggest a convenience option instead? Something like:

```
option(CCACHE "Enable ccache" OFF)
```

And then use that to decide.

Btw, I think `CMAKE_LINKER_COMPILER_LAUNCHER` should also be added above.

.gitlab-ci.yml
kacf: I guess we could leave it here just to make sure it works. We need to make sure we enable it then.

main.cmake
oleorhagen: Mostly tests, but I'm still skeptical of it. I see that `Gentoo` recommends against it: https://wiki.gentoo.org/wiki/Handbook:AMD64/Working/Features#Caching_compilation_objects. 

But it's really nice to have the option to enable it though I think.

main.cmake
aduskett: > I don't think we should enable it by default at all. May I suggest a convenience option instead? Something like:
> 
> ```
> option(CCACHE "Enable ccache" OFF)
> ```
> 
> And then use that to decide.
> 
> Btw, I think `CMAKE_LINKER_COMPILER_LAUNCHER` should also be added above.

Good idea. However, that option would override the CCACHE variable from find_program, which seems to make cmake quite upset. So I changed the option to "ENABLE_CCACHE" instead.

main.cmake
```diff
@@ -8,6 +8,7 @@ if (POLICY CMP0135)
   cmake_policy(SET CMP0135 NEW)
 endif (POLICY CMP0135)
 
+
```
oleorhagen, line 11: Nit: Random extra whitespace

main.cmake
```diff
@@ -44,6 +46,18 @@ option(MENDER_ARTIFACT_GZIP_COMPRESSION "Enable GZIP compression support when do
 option(MENDER_ARTIFACT_LZMA_COMPRESSION "Enable LZMA compression support when downloading and extracting Artifacts (Default: ON)" ON)
 option(MENDER_ARTIFACT_ZSTD_COMPRESSION "Enable Zstd compression support when downloading and extracting Artifacts (Default: ON)" ON)
 
+# Coverage may fail when building tests.
```
oleorhagen, line 49: Q: Should this be a `message` perhaps? :thinking:

main.cmake
```diff
@@ -44,6 +46,18 @@ option(MENDER_ARTIFACT_GZIP_COMPRESSION "Enable GZIP compression support when do
...
+if(ENABLE_CCACHE)
+  find_program(CCACHE ccache)
+  if(CCACHE)
+    message(STATUS "Found ccache at ${CCACHE}")
+    set(CMAKE_C_COMPILER_LAUNCHER ${CCACHE})
+    set(CMAKE_CXX_COMPILER_LAUNCHER ${CCACHE})
+    set(CMAKE_LINKER_COMPILER_LAUNCHER ${CCACHE})
+  endif()
```
oleorhagen, line 57: Q: Should we log a warning if `ccache` is enabled and not found?

oleorhagen (commented):
Looks good I think. It's delightfully small :)

I think I would change the commit type from `feat` to `build` though. As this really does not add any new features to the client.

oleorhagen:
@aduskett is this still WIP?

main.cmake
aduskett: Probably not. It's not a requirement. I could add a message at the bottom if CCache is being used.

main.cmake
aduskett: Maybe a warning?

main.cmake
aduskett: Fixed

aduskett:
> @aduskett is this still WIP?

Sorry, I've been busy with a lot of things in life. I pushed the suggested changes. Sorry about that!

main.cmake
oleorhagen: Yes, maybe even :)

oleorhagen:
> > @aduskett is this still WIP?
> 
> Sorry, I've been busy with a lot of things in life. I pushed the suggested changes. Sorry about that!

No worries, we all know the feeling. I always feel bad when I have to leave reviews waiting for a few days ;)

oleorhagen (approved):
Looks good to me :)

PR #1553, opened 2024-01-09
chore(context): Log provides/depends mismatches at error level

oleorhagen:
Just make sure that they show up in the logs, as the mismatch can be pretty confusing otherwise.


kacf (approved):
Great, thanks!

(I suspect some unit tests need adjusting, but we'll see)

oleorhagen:
> Great, thanks!
> 
> (I suspect some unit tests need adjusting, but we'll see)

haha, me too. Just too lazy to run it locally :see_no_evil:

PR #1556, opened 2024-01-16
build: Modify syntax for compatibility with CMake 3.18

lluiscampos:
Older CMake versions do not support `--component=...` syntax, so use an space instead.


kacf (approved):

oleorhagen:
Don't we have to support `3.16` for the `buster` builds? :thinking:

lluiscampos:
> Don't we have to support `3.16` for the `buster` builds? 🤔

Yes, that is the intention. These was breaking 3.18 and older, but I didn't investigate how far back.

PR #1559, opened 2024-01-24
Revert "build: Set MinSizeRel as the default build type"

kacf:
It's redundant, since it already exists in 84c13e9fe.

This reverts commit 0b923f656ea8621678b43f849ac71daba32737f3.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -2,13 +2,6 @@ cmake_minimum_required(VERSION 3.15)
 
 project(mender)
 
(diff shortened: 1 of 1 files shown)


kacf:
@oleorhagen: I think this is not needed, or did something in the old logic fail for you?

oleorhagen:
> @oleorhagen: I think this is not needed, or did something in the old logic fail for you?

Curious. Rebase miss mby?

oleorhagen (approved):

kacf:
Not running the full pipeline on this one, it should have no effect on the build.

PR #1564, opened 2024-01-31
Support commits for reusing common modules in other projects.

kacf:


oleorhagen (commented):
Last commit smells a bit to me.

Should we not now package it as an independent library, and use it. Instead of manipulating paths and adding new config files?

kacf:
If we wanted to go the libntech route, then yes. But I don't think this is quite at that level yet, it has not been proven in any other project, and we are not aiming for binary compatibility or anything like that, at least not now. This is a pure vendor-style static dependency, so my plan was to use it exactly the same way that we're using it in the client.

Does it make sense? (And did I understand you correctly?)

main.cmake
```diff
@@ -1,3 +1,6 @@
+set(MENDER_SOURCE_DIR "${CMAKE_CURRENT_SOURCE_DIR}")
+set(MENDER_SRC_DIR "${CMAKE_CURRENT_SOURCE_DIR}/src")
```
lluiscampos, line 2: `MENDER_SOURCE_DIR` and `MENDER_SRC_DIR` are very similar. Could we call the former one `MENDER_TOPLEVEL_DIR`, `MENDER_ROOT_DIR` or something similar?

lluiscampos (approved):

kacf:
For the record, I misunderstood @oleorhagen, I thought he wanted a shared library, but he was talking about separate projects on the cmake level, which can then be imported.

@oleorhagen: Check it now. The split has been completely rewritten, and I think this turned out much nicer!

main.cmake
kacf: With the changes after oleorhagen's comments, the variable was removed, so this is irrelevant now.

kacf:
Only the two last commits are new, btw.

src/common/CMakeLists.txt
```diff
@@ -156,3 +156,7 @@ if (MENDER_USE_SYSTEMD)
 else()
   target_sources(common_watchdog PRIVATE watchdog/noop/watchdog.cpp)
 endif()
+
+# Header-only.
+add_library(common_state_machine INTERFACE)
```
oleorhagen, line 213: Clever!

oleorhagen (approved):
Very nice!!

PR #1574, opened 2024-02-07
chore: Add missing dependency to `README` instructions

lluiscampos:


oleorhagen (approved):
Thx. At this point we have this replicated in quite a few places now. Should we think about unifying this somehow perchance?

PR #1575, opened 2024-02-08
Some build fixes

kacf:


lluiscampos (approved):

oleorhagen (approved):
```
Apparently this does not show up in the regular client by mere
coincidence, but it does show up in repos using its common module.
```

This is like straight outta the style-guide :sweat_smile:

PR #1578, opened 2024-02-12
chore: Make sure the event loop is kicked even without a starting event.

kacf:
Otherwise the first state will not be entered unless there is an event to trigger it.


lluiscampos (approved):
I assume the motivation is to make work an state machine without initial events [like this](https://github.com/mendersoftware/mender/blob/ce941e5564340422891579683be867b2e867edb3/src/mender-update/daemon/state_machine/state_machine.cpp#L434-L436).

kacf:
Exactly.

Keeping the ones you mention is still preferred though, because those two are recurring events, and the ones posted there are just the first ones. OTOH the place where I need this fix, the state will be entered exactly once at the beginning, and then never again. Therefore having an event is overkill.

PR #1579, opened 2024-02-13
feat: Replace `--data` flag with `--datastore` flag.

kacf:
Note in particular that we leave one instance of the old flag behind in our tests. The flag was never tested explicitly, just implicitly through a lot of other tests. So I just picked one at random where I kept the old flag.

Changelog: Replace `--data` flag with `--datastore` flag. This aligns better with the environment variables with the same names. The old `--data` flag is still accepted for backwards compatibility, but note that it does not have the same meaning as the `MENDER_DATA_DIR` environment variable (hence the rename).

Ticket: None


kacf:
As discussed in [MEN-6985](https://northerntech.atlassian.net/browse/MEN-6985?focusedCommentId=104181).

lluiscampos (approved):

kacf:
This already [passed mender-qa](https://gitlab.com/Northern.tech/Mender/mender-qa/-/pipelines/1174568035), the last change only fixed formatting.

PR #1580, opened 2024-02-14
fix: Submit inventory as soon as the device is accepted.

kacf:
Note that this fix only works "once". IOW if you reject a device and then re-admit it later, then it won't submit inventory again, but will follow the regular schedule.

Changelog: Title
Ticket: None


tranchitella (approved):

lluiscampos (commented):
:heavy_check_mark: It looks good to me too!

PR #1581, opened 2024-02-14
fix: doing so return 1 in Download_Enter state scripts aborts instead of retrying in an endless loop

SebOpsahl:
changelog: title
ticket: MEN-6319

### DIFF
--- a/app/state.go
+++ b/app/state.go
@@ -731,6 +731,16 @@ func (uf *updateFetchState) Update() *datastore.UpdateInfo {
 	return &uf.update
 }
 
+func (uf *updateFetchState) HandleError(
+	ctx *StateContext,
+	c Controller,
+	err menderError,
+) (State, bool) {
+	log.Error(err.Error())
+
+	return NewUpdateErrorState(err, uf.Update()), false
+}
+
 func (uf *updateFetchState) PermitLooping() bool {
 	return false
 }
--- a/app/state_test.go
+++ b/app/state_test.go
@@ -1331,6 +1331,32 @@ var stateTransitionsWithUpdateModulesTestCases []stateTransitionsWithUpdateModul
 	},
 
 	{
+		caseName: "Error in Download state, with rollback",
+		stateChain: []State{
+			&updateFetchState{},
+			&updateStoreState{},
+			&updateCleanupState{},
+			&updateStatusReportState{},
+			&idleState{},
+		},
+		artifactStateChain: []string{
+			"Download_Enter_00",
+			"Download",
+			"Download_Error_00",
+			"Cleanup",
+		},
+		reportsLog: []string{
+			"downloading",
+			"failure",
+		},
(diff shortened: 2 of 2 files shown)


app/state.go
```diff
@@ -731,6 +731,22 @@ func (uf *updateFetchState) Update() *datastore.UpdateInfo {
...
+) (State, bool) {
+	log.Error(err.Error())
+
+	// Default for most update states. Some states will override this.
+	if uf.Update().SupportsRollback == datastore.RollbackSupported {
+		return NewUpdateRollbackState(uf.Update()), false
+	} else {
+		setBrokenArtifactFlag(ctx.Store, uf.Update().ArtifactName())
+		return NewUpdateErrorState(err, uf.Update()), false
+	}
```
kacf, line 747: The `NewUpdateRollbackState` is not necessary, because as I mentioned earlier, when we are in the download state (or earlier), we have not changed the system yet, so no rollback is necessary. This means that `setBrokenArtifactFlag` is not necessary either, because we didn't change the system, so we know that the already installed artifact is still functioning. It's a rollback-by-doing-nothing, essentially. 🙂

So in fact I believe you can keep only the `NewUpdateErrorState` return, and that should be enough to solve the problem AFAICS!

I think this should also fix the unit test problem you are having.

app/state_test.go
```diff
@@ -1330,6 +1330,32 @@ var stateTransitionsWithUpdateModulesTestCases []stateTransitionsWithUpdateModul
 		installOutcome: tests.SuccessfulRollback,
 	},
 
+	{
+		caseName: "Error in Download state, with rollback",
```
kacf, line 1334: I think this is a duplicate of an existing test case, isn't it?

app/state_test.go
```diff
@@ -2583,14 +2609,16 @@ var stateTransitionsWithUpdateModulesTestCases []stateTransitionsWithUpdateModul
...
+			&updateCleanupState{},
+			&updateStatusReportState{},
 			&idleState{},
 		},
 		artifactStateChain: []string{
 			"Download_Enter_00",
 			"Download_Error_00",
 		},
-		reportsLog: []string{""},
+		reportsLog: []string{"failure"},
```
kacf, line 2621: This change looks exactly right to me! 👍

kacf (changes requested):

app/state_test.go
SebOpsahl: Well there is not any test for Download state with rollback, only with no rollback. If I understood one of the corrections from the meeting it was to add the test case with rollback enabled. But I could have understood wrong. It is not a duplicate since the rollback is different, should I then keep it or remove it?

kacf: Ah, you're right, my bad. Please keep it! 👍

kacf (approved):
Beautiful!

SebOpsahl:

PR #1583, opened 2024-02-15
MEN-6995: Add `io::BufferedReader` class

lluiscampos:
Implement an Reader with an internal buffer so that data can be read twice. It is meant to be used in an special case where the Artifact headers are first parsed (aka `Read`) but then we still want to stream the Artifact from the start.


src/common/io.hpp
```diff
@@ -271,6 +271,47 @@ class StringReader : virtual public Reader {
...
+	BufferedReader(io::Reader &reader) :
+		wrapped_reader_ {reader},
+		buffer_reader_ {buffer_} {};
+
+	expected::ExpectedSize Read(
+		vector<uint8_t>::iterator start, vector<uint8_t>::iterator end) override;
+
+	error::Error Rewind();
+
+	void StopBuffering();
```
lluiscampos, line 311: @kacf This is my naive approach: save all the data and rewind.

However as it is designed now if the user does not call `StopBuffering()` we will copy the whole Artifact in memory :boom:  - I need to rethink a bit how to best limit it but please give me your preliminary feedback :pray:

src/common/io/io.cpp
```diff
@@ -573,6 +584,53 @@ error::Error FileReader::Rewind() {
 	return error::NoError;
 }
 
+ExpectedSize BufferedReader::Read(vector<uint8_t>::iterator start, vector<uint8_t>::iterator end) {
+	Vsize bytes_read_buffer {0};
+	Vsize bytes_read_wrapper {0};
+
+	if (!rewind_done_) {
```
kacf, line 591: I agree with the general approach here, but to me the logic seems backwards, and I have some suggestions:

1. Instead of checking for `!rewind_done_`, check for `rewind_done_`, and then do the buffered reading in there. Don't bother concatenating the two sources, this is IMHO not necessary. All consumers support short reads.
2. As soon as it returns `0`, set `rewind_done_ = false` and fall through . From there you follow up with the `wrapped_reader`.
3. (Bonus): If you also implement `ByteReader::Rewind() { bytes_read_ = 0; }`, then you can call it from `BufferedReader::Rewind()` which will allow `Rewind()` to be used multiple times, as long as `StopBuffering()` hasn't been called. I know, this isn't necessary, but I would argue it improves code quality as well.
4. And finally, if `!rewind_done_ && stop_done_`, then you should call `buffer_.clear()` to save memory.

src/common/io/io.cpp
```diff
@@ -573,6 +584,53 @@ error::Error FileReader::Rewind() {
...
+			return ex_bytes_read;
+		}
+		bytes_read_wrapper = ex_bytes_read.value();
+	}
+
+	return bytes_read_buffer + bytes_read_wrapper;
+}
+
+error::Error BufferedReader::Rewind() {
+	if (rewind_done_) {
```
kacf, line 623: If you follow my suggestion above then you don't need this check, but there should be a check for `stop_done_` instead.

kacf (approved):
Good solution, I think prefer this over my original idea with two outputs. I have some suggestions, but treat them as exactly that. This is good as-is, also.

lluiscampos:
@kacf @oleorhagen You are welcome to have a look at this beast. There are still some TODOs that I want to address, and I'd like to simplify also the tests (I wrote them in TDD style), but I would appreciate early feedback on the design of the classes.

I hope that it is understandable... Enjoy :popcorn:

src/common/io/io.cpp
```diff
@@ -573,6 +588,46 @@ error::Error FileReader::Rewind() {
...
+
+	// Read from the wrapped reader and save copy into the buffer
+	auto bytes_read = wrapped_reader_.Read(start, end);
+	if (!bytes_read) {
+		return bytes_read;
+	}
+	if (!stop_done_) {
+		buffer_.insert(buffer_.end(), start, end);
+	}
+	if (rewind_consumed_ && stop_done_) {
```
kacf, line 615: Micro-optimization/readability: You can replace with `} else if (rewind_consumed_) {` here.

src/common/events/events_io.cpp
```diff
@@ -137,6 +137,76 @@ mio::ExpectedSize ReaderFromAsyncReader::Read(
...
+	leaf_readers_.insert({reader, TeeReaderRequestedRead {}});
+	return reader;
+}
+
+error::Error TeeReader::ReadyToAsyncRead(
+	TeeReader::TeeReaderLeafPtr leaf_reader,
+	vector<uint8_t>::iterator start,
+	vector<uint8_t>::iterator end,
+	mio::AsyncIoHandler handler) {
+	// TODO: should this be an assert() instead?
```
kacf, line 153: I suggest [`AssertOrReturnError`](https://github.com/mendersoftware/mender/blob/master/src/common/error.hpp#L24).

src/common/events/events_io.cpp
```diff
@@ -137,6 +137,76 @@ mio::ExpectedSize ReaderFromAsyncReader::Read(
...
+	}
+
+	return error::NoError;
+}
+
+void TeeReader::DoAsyncRead() {
+	auto handler = [this](mio::ExpectedSize result) {
+		if (!result) {
+			for (auto const &it : leaf_readers_) {
+				it.second.handler(result);
```
kacf: We need to be very careful when calling handlers which can themselves modify `leaf_readers_`. It may be safer to make a copy of the list you're going to execute before you execute it.

src/common/events/events_io.cpp
```diff
@@ -137,6 +137,76 @@ mio::ExpectedSize ReaderFromAsyncReader::Read(
...
+			return;
+		};
+
+		for (auto &it : leaf_readers_) {
+			std::copy_n(buffer_.begin(), result.value(), it.second.start);
+		}
+		buffer_.clear();
+
+		for (auto const &it : leaf_readers_) {
+			it.second.handler(result);
```
kacf: Same here.

src/common/events/events_io.cpp
```diff
@@ -137,6 +137,76 @@ mio::ExpectedSize ReaderFromAsyncReader::Read(
...
+	};
+
+	auto min_read = std::min_element(
+		leaf_readers_.cbegin(),
+		leaf_readers_.cend(),
+		[](const std::pair<TeeReaderLeafPtr, TeeReaderRequestedRead> r1,
+		   std::pair<TeeReaderLeafPtr, TeeReaderRequestedRead> r2) {
+			return (r1.second.end - r1.second.start) < (r2.second.end - r2.second.start);
+		});
+	buffer_.resize(min_read->second.end - min_read->second.start);
```
kacf: Why do we even need a buffer here? Can't you just read directly into the iterator range of the first reader, and then copy to the other afterwards?

kacf (commented):
Nice implementation of TeeReader!

My review queue is unexpectedly long today, so I've only reviewed _chore: Add first version of TeeReader_ yet. I hope that doesn't block you in any way.

kacf (commented):
Nice and tidy solution. I think it's not even that complicated, once we actually figured out what to solve!

I've left most TODOs alone, I assume there will be another round.

src/common/events/events_io.cpp
lluiscampos: Thank you

src/common/events/events_io.cpp
lluiscampos: Thank you. I actually thought about this with some "not staring to pixels" time :)

src/common/events/events_io.cpp
lluiscampos: Right, thanks.

src/common/events/events_io.cpp
```diff
@@ -139,7 +139,9 @@ mio::ExpectedSize ReaderFromAsyncReader::Read(
 
 TeeReader::TeeReaderLeafPtr TeeReader::AttachSecondaryReader() {
 	auto reader = make_shared<TeeReaderLeaf>(shared_from_this());
-	leaf_readers_.insert({reader, TeeReaderRequestedRead {}});
+	leaf_readers_.insert({reader, TeeReaderLeafContext {}});
+	auto bytes_missing = source_reader_->Rewind();
```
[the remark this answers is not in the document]
lluiscampos: It is a good idea. I am following up on a separate commit.

src/common/events/events_io.cpp
```diff
@@ -139,7 +139,9 @@ mio::ExpectedSize ReaderFromAsyncReader::Read(
...
+	leaf_readers_[reader].buffer_bytes_missing = bytes_missing;
```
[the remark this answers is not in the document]
lluiscampos: I will also do this

src/common/events/events_io.cpp
```diff
@@ -137,6 +137,161 @@ mio::ExpectedSize ReaderFromAsyncReader::Read(
...
+	auto handler = [this](mio::ExpectedSize result) {
+		if (!result) {
+			CallAllHandlers(result);
+			return;
+		};
+
+		auto start_iterator = leaf_readers_.begin()->second.pending_read.start;
+		auto read_bytes = result.value();
+		for_each(
+			++leaf_readers_.begin(),
```
kacf, line 225: This is to start in the second position, not the first, right? Could you use `begin() + 1` instead? I don't think `++` has any ill effects here, but it's confusing to read because it also modifies its argument, which is a temporary (I'm mildly surprised it's even allowed here).

kacf (approved):
Nice binary diffstat: +1111,-0 🙂

src/common/io/io.cpp
```diff
@@ -573,6 +588,127 @@ error::Error FileReader::Rewind() {
...
+	buffer_reader_.Rewind();
+	rewind_done_ = true;
+	rewind_consumed_ = false;
+	return buffer_.size();
+}
+
+ExpectedSize AsyncBufferedReader::StopBufferingAndRewind() {
+	auto result = Rewind();
+	stop_done_ = true;
+	return result;
```
oleorhagen, line 698: Q: Why not return Rewind directly?

src/common/events/events_io.cpp
kacf: Right, because you can't jump an arbitrary number of steps, just one at a time. But `std::next` is even more descriptive, so 👍.

src/common/events/events_io.cpp
```diff
@@ -137,6 +137,175 @@ mio::ExpectedSize ReaderFromAsyncReader::Read(
...
+	return error::NoError;
+}
+
+void TeeReader::CallAllHandlers(mio::ExpectedSize result) {
+	// Makes a copy of the handlers and then calls them sequentially
+	vector<mio::AsyncIoHandler> handlers;
+	for (auto const &it : leaf_readers_) {
+		handlers.push_back(it.second.pending_read.handler);
+	}
+	for (auto const &h : handlers) {
```
oleorhagen, line 211: Q: why auto const?

oleorhagen (approved):
Beautiful!

src/common/events/events_io.cpp
lluiscampos: Swapping them back

PR #1584, opened 2024-02-19
chore: Make it more convenient to use *our* `Expected`.

kacf:
Which is always used with `error::Error`.

### DIFF
--- a/src/common/expected.hpp
+++ b/src/common/expected.hpp
@@ -51,24 +51,26 @@ tl::unexpected<V> unexpected(V &v) {
 }
 #endif
 
-using ExpectedString = expected<string, error::Error>;
-using ExpectedBytes = expected<vector<uint8_t>, error::Error>;
-using ExpectedInt = expected<int, error::Error>;
-using ExpectedInt64 = expected<int64_t, error::Error>;
-using ExpectedDouble = expected<double, error::Error>;
-using ExpectedBool = expected<bool, error::Error>;
-using ExpectedSize = expected<size_t, error::Error>;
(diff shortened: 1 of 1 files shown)


kacf:
@oleorhagen, @lluiscampos: What do you think? I'm thinking "why didn't we do this from the get-go?" It makes it almost unnecessary to define types like `ExpectedBool`, because `Expected<bool>` is always just two characters longer, and doesn't need a defined type. Especially with single-use `Expected` types, this is convenient.

src/common/expected.hpp
```diff
@@ -51,24 +51,26 @@ tl::unexpected<V> unexpected(V &v) {
...
+using ExpectedBool = Expected<bool>;
+using ExpectedSize = Expected<size_t>;
+using ExpectedBool = Expected<bool>;
+using ExpectedLong = Expected<long>;
+using ExpectedLongLong = Expected<long long>;
+using ExpectedStringVector = Expected<vector<string>>;
 
 template <typename T>
-using ExpectedVector = expected<vector<T>, error::Error>;
+using ExpectedVector = Expected<vector<T>>;
```
oleorhagen, line 70: I think I would just leave these tbh

oleorhagen (approved):
Hah, I've been having this on my TODO for ages

src/common/expected.hpp
kacf: I thought they could be "inspirational", but I'm fine either way. I certainly will not fix all instances.

lluiscampos (approved):
Nice clean-up

PR #1587, opened 2024-02-21
fix: Ensure enough body writer capacity if no Content-Length sent by deployments API

vpodzime:
Discovered in review of a copy-pasted version of this code into mendersoftware/mender-update-orchestrator#30.

Ticket: none
Changelog: none


vpodzime:

kacf (approved):
Good call!

PR #1588, opened 2024-02-21
fix: Fix error checking for Update Module's working directory

lluiscampos:
Changelog: Fix a bug where a failure in checking the working directory will not exit cleanly from `Cleanup` state.


lluiscampos:
@kacf Does this make sense? Otherwise I don't understand the logic inside that `if` block. Please review carefully the changelog too.

kacf (approved):
I believe that's a mistake, yes. Good catch!

lluiscampos:

PR #1592, opened 2024-02-27
fix: so the Download_error state script gets executed for failures due to signature

SebOpsahl:
changelog: title
ticket: MEN-6402

Changed the code so when encountering failures due to signature it goes into the update-error state, which will allow it to run Download_Error.

### DIFF
--- a/app/state.go
+++ b/app/state.go
@@ -808,7 +808,8 @@ func (u *updateStoreState) Handle(ctx *StateContext, c Controller) (State, bool)
 	installer, err := c.ReadArtifactHeaders(u.imagein)
 	if err != nil {
 		log.Errorf("Fetching Artifact headers failed: %s", err)
-		return NewUpdateStatusReportState(&u.update, client.StatusFailure), false
+
+		return NewUpdateCleanupState(&u.update, client.StatusFailure), false
 	}
 
 	installers := c.GetInstallers()
--- a/app/state_test.go
+++ b/app/state_test.go
@@ -48,27 +48,28 @@ import (
 
 type stateTestController struct {
 	FakeDevice
-	updater                fakeUpdater
-	artifactName           string
-	updatePollIntvl        time.Duration
-	inventPollIntvl        time.Duration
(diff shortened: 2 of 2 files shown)


app/state.go
```diff
@@ -808,7 +808,9 @@ func (u *updateStoreState) Handle(ctx *StateContext, c Controller) (State, bool)
 	installer, err := c.ReadArtifactHeaders(u.imagein)
 	if err != nil {
 		log.Errorf("Fetching Artifact headers failed: %s", err)
-		return NewUpdateStatusReportState(&u.update, client.StatusFailure), false
+		merr := NewTransientError(fmt.Errorf("Error reading artifact headers: %s", err.Error()))
+
+		return NewUpdateErrorState(merr, &u.update), false
```
kacf, line 813: Hmm, I think the `UpdateErrorState` leads to `ArtifactFailure` being executed. However, we cannot execute any `Artifact` states unless `Download` has completed to successfully (for several reasons, but most importantly because it's insecure).

So I think this should be `NewUpdateCleanupState` instead. I think it should still execute error scripts as long as [the correct status](https://github.com/mendersoftware/mender/blob/3.5.x/app/state.go#L1388-L1392) is passed into it.

Other than that I agree with the fix!

kacf (changes requested):
If you look at what the mender-test-bot rendered as the changelog, the sentence is a bit weird because the `fix:` prefix is not included as part of the message. Maybe rephrase this a little.

app/state.go
SebOpsahl: Okay, I see, my thinking was that UpdateErrorState{} should trigger the execution of the error script in transitionState(), but I now see that going into NewUpdateCleanUpdateState{} is a better and more secure solution.Thanks :)

SebOpsahl:

SebOpsahl:
All good, it's nice to learn how to do it precise and correct :)

kacf:
Only backend failed, not relevant. Merging.

PR #1596, opened 2024-03-01
chore: Add HeaderView::GetDepends() and GetProvides()

vpodzime:
Two utility functions to get `provides` and `depends` as defined by the artifact or more specifically, one of the artifact's subheaders (represented by HeaderView).

Ticket: MEN-6960
Changelog: none


src/mender-update/context/context.cpp
```diff
@@ -387,66 +387,65 @@ expected::ExpectedBool MenderContext::MatchesArtifactDepends(const artifact::Hea
...
+	if (!common::VectorContainsString(hdr_depends["device_type"], device_type)) {
 		log::Error("Artifact device type doesn't match");
 		return false;
 	}
+	hdr_depends.erase("device_type");
 
-	if (hdr_depends.artifact_name) {
-		AssertOrReturnUnexpected(hdr_depends.artifact_name->size() > 0);
+	if (MapContainsStringKey(hdr_depends, "artifact_name")) {
+		AssertOrReturnUnexpected(hdr_depends["artifact_name"].size() > 0);
```
kacf, line 412: I'm probably missing something, but now that you are matching `provides` against `hdr_depends` everywhere, why do we need these special cases at all?

kacf (approved):

vpodzime:
@mender-test-bot start pipeline --fast

src/common/common.hpp
```diff
@@ -104,8 +104,19 @@ static inline bool VectorContainsString(const vector<string> &vec, const string
 	return std::find(vec.begin(), vec.end(), str) != vec.end();
 }
 
+static inline string StringVectorToString(const vector<string> &vec, const string delim = ",") {
+	string ret = "{";
+	auto sz = vec.size();
+	for (decltype(sz) i = 0; i < (sz - 1); i++) {
+		ret += "\"" + vec[i] + "\"" + delim;
+	}
+	ret += "\"" + vec[sz - 1] + "\"" + "}";
```
lluiscampos, line 113: This is almost the same as `JoinStrings` above.

vpodzime: Yes, but `JoinStrings()` doesn't add the quotes and curly brackets. I could use it here and add the curly brackets, but I think the quotes are also nice.

vpodzime:

PR #1597, opened 2024-03-01
fix: Fix segfault in LibArchive wrapper

lluiscampos:
According to the man page,
```
(...) On error, the read callback should invoke
archive_set_error() to register an error code and message and return -1.  
```

Not setting the error correctly was leading to a segfault in the
`string` constructor as `archive_error_string` was returning null.

### DIFF
--- a/src/artifact/tar/platform/libarchive/wrapper.cpp
+++ b/src/artifact/tar/platform/libarchive/wrapper.cpp
@@ -38,14 +38,14 @@ using ExpectedSize = expected::ExpectedSize;
 // the size of the data blocks returned.
 //
 // - On EOF return 0.
-// - On error return -1.
+// - On error, call archive_set_error and return -1.
 ssize_t reader_callback(archive *archive, void *in_reader_container, const void **buff) {
 	ReaderContainer *p_reader_container = static_cast<ReaderContainer *>(in_reader_container);
 
 	auto ret = p_reader_container->reader_.Read(
 		p_reader_container->buff_.begin(), p_reader_container->buff_.end());
 	if (!ret) {
-		log::Error("Failed to read from the archive stream: Error: " + ret.error().message);
+		archive_set_error(archive, ret.error().code.value(), "%s", ret.error().message.c_str());
 		return -1;
 	}
 
@@ -83,7 +83,7 @@ Error Handle::Init() {
 	if (r != ARCHIVE_OK) {
 		return MakeError(
 			error::GenericError,
-			"Failed to initalize the 'libarchive' C bindings. LibArchive error message: '"
+			"Failed to read from the archive stream: '"
 				+ string(archive_error_string(archive_.get()))
 				+ "' error code: " + std::to_string(archive_errno(archive_.get())));
 	}


SebOpsahl (approved):

lluiscampos:

src/artifact/tar/platform/libarchive/wrapper.cpp
```diff
@@ -81,11 +81,12 @@ Error Handle::Init() {
 	}
 	r = archive_read_open(archive_.get(), &reader_container_, nullptr, reader_callback, nullptr);
 	if (r != ARCHIVE_OK) {
-		return MakeError(
-			error::GenericError,
-			"Failed to initalize the 'libarchive' C bindings. LibArchive error message: '"
-				+ string(archive_error_string(archive_.get()))
-				+ "' error code: " + std::to_string(archive_errno(archive_.get())));
+		string err_msg = "Failed to initialize the 'libarchive' C bindings. LibArchive error code: "
+						 + std::to_string(archive_errno(archive_.get()));
```
vpodzime, line 85: According to the docs:
>  On error, the `read` callback should invoke archive_set_error() to register an error code and message

IOW, we should be setting the error. Also, the log message is misleading, this is not an initialization of the C bindings, it's actually opening the archive.

vpodzime (approved):
Good as a segfault hot-fix, but please create a ticket for properly fixing this in the future.

lluiscampos:
> Good as a segfault hot-fix, but please create a ticket for properly fixing this in the future.

Thank you for your thorough review. I had to properly fix because the original patch was incomplete - it would segfault later [here](https://github.com/mendersoftware/mender/blob/084df276b365e99a9e6f12ec3cb4ab31079e4796/src/artifact/tar/platform/libarchive/tar.cpp#L75-L80).

A related problem is that `Handle::Init()` is called in the constructor, where the error is logged but otherwise ignored. Then it will segfault the same way when reading contents.

Anyway, "proper" fix coming up :innocent:

lluiscampos:
> Also, the log message is misleading, this is not an initialization of the C bindings, it's actually opening the archive

I fixed this too by not having a double logging and moving the valuable part ("Failed to read from the archive stream") to the caller logging.

src/artifact/tar/platform/libarchive/wrapper.cpp
```diff
@@ -38,15 +38,15 @@ using ExpectedSize = expected::ExpectedSize;
...
 ssize_t reader_callback(archive *archive, void *in_reader_container, const void **buff) {
 	ReaderContainer *p_reader_container = static_cast<ReaderContainer *>(in_reader_container);
 
 	auto ret = p_reader_container->reader_.Read(
 		p_reader_container->buff_.begin(), p_reader_container->buff_.end());
 	if (!ret) {
-		log::Error("Failed to read from the archive stream: Error: " + ret.error().message);
-		return -1;
+		archive_set_error(archive, ret.error().code.value(), "%s", ret.error().message.c_str());
+		return ARCHIVE_FATAL;
```
vpodzime, line 49: The `read` callback should actually return `-1` in case of error. It's the **`open`** callback that should return `ARCHIVE_FATAL`, but we don't use `open` callback.

vpodzime (approved):
Just a nitpick. Thanks for fixing this properly!

src/artifact/tar/platform/libarchive/wrapper.cpp
lluiscampos: Thank you.

lluiscampos:

PR #1598, opened 2024-03-02
chore: Move http_resumer to common

SebOpsahl:
Due to other projects needing this change

Changelog: None

Ticket: MEN-6979


src/common/http.hpp
```diff
@@ -40,13 +40,13 @@
...
-namespace update {
+namespace common {
 namespace http_resumer {
 class DownloadResumerClient;
 class HeaderHandlerFunctor;
 class BodyHandlerFunctor;
 } // namespace http_resumer
-} // namespace update
+} // namespace common
 } // namespace mender
```
lluiscampos, line 50: Too bad we still need to keep this around.

Once we are moving code, what do you think of moving `mender::common::http_resumer` namespace to `mender::common::http::resumer`? It looks a bit more tidy and it follows for example the design of `events` and `events::io`.

It however will still require that we declare here in `http.hpp` the classes of the resumer, but we can at least remove these lines and move the declarations below inside `mender::common::http` as just:
```
namespace mender {
namespace common {
namespace http {

namespace resumer {
class DownloadResumerClient;
class HeaderHandlerFunctor;
class BodyHandlerFunctor;
} // namespace resumer

using namespace std;
...
```

At this point I am only thinking out loud, it is not a requirement

src/mender-update/CMakeLists.txt
```diff
@@ -56,7 +56,7 @@ target_sources(update_module PRIVATE
 target_compile_options(update_module PRIVATE ${PLATFORM_SPECIFIC_COMPILE_OPTIONS})
 
 add_library(mender_http_resumer STATIC
-  http_resumer/http_resumer.cpp
+  ../common/http_resumer/http_resumer.cpp
 )
 target_link_libraries(mender_http_resumer PUBLIC
   common_http
```
lluiscampos, line 62: This library definition now needs to be moved to `src/common/CMakeLists.txt`.

Generally speaking we build CMake top-down approach, which means that child directories should not require files from parent directories.

tests/src/common/CMakeLists.txt
```diff
@@ -134,3 +134,5 @@ target_link_libraries(dbus_test PUBLIC
 )
 gtest_discover_tests(dbus_test ${MENDER_TEST_FLAGS} NO_PRETTY_VALUES)
 add_dependencies(tests dbus_test)
+
+add_subdirectory(http_resumer)
```
lluiscampos, line 138: Nitpick: missing newline

tests/src/common/http_resumer/http_resumer_test.cpp
```diff
@@ -104,11 +104,12 @@ class RangeBodyOfXes : virtual public io::Reader {
...
-		// number to the power of 1.1 and round to the nearest integer. If it's odd, return
-		// 'X', if it's even, return 'x'. Due to the exponent, this pattern will change
-		// slightly throughout the sequence, giving us a chance to catch offset errors.
-		// (Note: Try printing it, the pattern is mesmerizing to watch!)
+		// Fill in a specific pattern to try to catch offset/block errors: Raise the
+		// input number to the power of 1.1 and round to the nearest integer. If
+		// it's odd, return 'X', if it's even, return 'x'. Due to the exponent, this
+		// pattern will change slightly throughout the sequence, giving us a chance
+		// to catch offset errors. (Note: Try printing it, the pattern is
+		// mesmerizing to watch!)
```
lluiscampos, line 112: This file has a lot of unrelated changes. Please remove them from this PR. (If you really want to change the style of the comments, at lest do so on a separate commit).

Tip :bulb: 

I assume you got into this because your editor modifies the line with of the code and later you did `git add <file>` or even `git commit -a`. I recommend using `git add -i` (for interactive) so that you can state patch by patch which changes you want to add to your next commit :+1:

lluiscampos (changes requested):
See my comments below, one slightly big suggestion for rework and some minor things to fix. Looking good otherwise!

lluiscampos (changes requested):
One more thing with regards to the commit formalities:
* This change does not honor a changelog entry, so either remove it or write `Changelog: None`
    * To figure out when a change needs a changelog you need to think if the change has implications for the user. Is it a new feature or a fix: definitely has an impact on the user. Is it a rename of  a variable, moving some code, fixing a test, a typo or some other internal chore: no impact to the user hence no changelog entry
* You don't need "chore" twice in the title
* It could be shorter and more descriptive with something like: "Move `http_resumer` to common". Then in the body of the commit you can write as much as you want, giving details of the motivation (other projects can use now http_resumer bla bla bla)

:+1:

tests/src/common/http_resumer/http_resumer_test.cpp
SebOpsahl: Oh sorry, this was because the formatting tests complained about this files formatting, but must have formatted it wrongfully, will fix this. Thanks for the tip :)

tests/src/common/CMakeLists.txt
SebOpsahl: Oh, will fix it now :)

tests/src/common/http_resumer/http_resumer_test.cpp
```diff
@@ -1271,4 +1271,4 @@ TEST_F(DownloadResumerTest, CallerDestroysReaderInTheMiddleOfAWait) {
 	// Reuse the client to check state clean-up
 	err = client->AsyncCall(req, user_header_handler, user_body_handler);
 	EXPECT_EQ(err, error::NoError) << "Unexpected error: " << err.message;
-}
+}
```
lluiscampos, line 1274: Unrelated change

src/mender-update/daemon/context.cpp
```diff
@@ -679,4 +679,4 @@ void Context::FinishDeploymentLogging() {
 
 } // namespace daemon
 } // namespace update
-} // namespace mender
+} // namespace mender
```
lluiscampos, line 682: Unrelated change

src/common/http.hpp
```diff
@@ -40,13 +40,15 @@
...
 
 namespace mender {
-namespace update {
-namespace http_resumer {
+namespace common {
+namespace http {
+namespace resumer {
 class DownloadResumerClient;
 class HeaderHandlerFunctor;
 class BodyHandlerFunctor;
```
lluiscampos, line 48: These can be moved few lines below to inside of `mender::common::http` as I tried to show you on my previous comment [here](https://github.com/mendersoftware/mender/pull/1598#discussion_r1514005108).

lluiscampos (approved):
Thank you for the rework. Still few nitpicks but it looks good so pre-approving :heavy_check_mark:

SebOpsahl:

SebOpsahl:

PR #1601, opened 2024-03-13
chore: updating CLI help screen

SebOpsahl:
Updating CLI help screen to show arguments for commands, and updating the help screen to show more options and commands

Changelog: None

Ticket: MEN-6986

### DIFF
--- a/src/client_shared/conf/conf_cli_help.cpp
+++ b/src/client_shared/conf/conf_cli_help.cpp
@@ -147,6 +147,19 @@ void PrintInTwoColumns(
 	}
 }
 
+string PrintArgument(optional<CliArgument> argument, ostream &stream) {
+	string cliarg;
+
+	if (argument.has_value()) {
+		if (argument.value().mandatory) {
+			cliarg = " " + argument.value().name;
+		} else {
+			cliarg = " [" + argument.value().name + "]";
+		}
+	}
+	return cliarg;
+}
+
 void PrintOptions(const vector<CliOption> &options, ostream &stream) {
 	PrintInTwoColumns(
 		options.begin(),
@@ -183,6 +196,10 @@ void PrintCommandHelp(const string &cli_name, const CliCommand &command, ostream
 		stream << " - " << command.description;
 	}
 	stream << endl << endl;
+
+	stream << "USAGE:" << endl;
+	stream << indent << cli_name << " [global options] " << command.name + " [command options]";
+	stream << PrintArgument(command.argument, stream) << endl << endl;
 
 	// Append --help option at the command level
 	vector<CliOption> options_with_help = command.options;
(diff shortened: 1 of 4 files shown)


lluiscampos (changes requested):

SebOpsahl:
Oh okey, fixed it all now in the latest commit :+1:

src/client_shared/conf/conf_cli_help.cpp
```diff
@@ -147,6 +147,19 @@ void PrintInTwoColumns(
 	}
 }
 
+string PrintArgument(CliArgument argument, ostream &stream) {
+	string cliarg;
+
+	if (!argument.name.empty()) {
```
lluiscampos, line 153: This works just fine.

A semantically better approach could be to use `optional` template, where in `CliCommand` we would instead have a `optional <CliArgument> argument;` attribute and in this line we would check with `argument.has_value()`. You can read more about this template [in here](https://github.com/mendersoftware/mender/blob/e8337ce91d9f6afec81fe7e62aa8a698c55726fa/src/common/optional.hpp) and search in our codebase for examples of usage.

src/mender-update/cli/cli.cpp
```diff
@@ -49,6 +49,11 @@ const conf::CliCommand cmd_daemon {
 const conf::CliCommand cmd_install {
 	.name = "install",
 	.description = "Mender Artifact to install - local file or a URL",
+	.argument =
+		{
+			.name = "manifest",
+			.mandatory = false,
```
lluiscampos, line 55: As I indirectly wrote [here](https://github.com/mendersoftware/mender/pull/1601#discussion_r1535637076) and directly wrote [here](https://github.com/mendersoftware/mender/pull/1601#discussion_r1535638579), for `mender-update` the argument is "artifact" and it is mandatory.

lluiscampos (changes requested):
Well done. The design is now fine, and I have a suggestion below for how to make it even better.

Requesting changes for my second comment (artifact instead of manifest), the first comment is a suggestion.

lluiscampos:
> Changelog: updating CLI help screen to show arguments for commands, and updating the help screen to show more options and commands
> 
> Ticket: MEN-6986

This is not a feature and have no effects for the user. Please mark it as `chore`

src/client_shared/conf/conf_cli_help.cpp
SebOpsahl: This sounds good, changed it to use this approach for both PR’s now 🙂

src/mender-update/cli/cli.cpp
SebOpsahl: Oh sorry, fixed it now

lluiscampos (approved):

SebOpsahl:

PR #1605, opened 2024-03-19
chore: changed parsing of some passthrough args in generators

danielskinstad:
Changelog: Returns an error when passing type, or when passing metadata to docker-artifact gen, and overrides output path
and name when passed as passthrough argument.

Ticket: MEN-7110

### DIFF
--- a/support/modules-artifact-gen/directory-artifact-gen
+++ b/support/modules-artifact-gen/directory-artifact-gen
@@ -161,9 +161,34 @@ else
   exit 1
 fi
 
-
 # Create dest_dir file in plain text
 echo "$dest_dir" > $dest_dir_file
+
+# Check the the passthrough_args and potentially modify them
+# to avoid conflicts or to let them override the already args
+# provided to mender-artifact
+# Runs in a subshell to allow overriding some parameters passed
+# to mender-artifact
+passthrough_args_modified=" "
+echo -n $passthrough_args | xargs -n 2 printf "%s %s\n" | (while read -r flag arg; do
+  if [ -n "$flag" ] && [ -n "$arg" ]; then
+    case $flag in
+      -T | --type)
+        echo "Error: Conflicting flag '$flag'. Already specified by the script."
+        exit 1
+        ;;
+      -o | --output-path)
+        output_path=$arg
+        ;;
+      -n | --name)
+        artifact_name=$arg
+        ;;
+      *)
+        passthrough_args_modified="$passthrough_args_modified $flag $arg"
+        ;;
+    esac
+  fi
+done
 
 mender-artifact write module-image \
   -T directory \
@@ -172,12 +197,13 @@ mender-artifact write module-image \
   -n $artifact_name \
   -f $update_files_tar \
   -f $dest_dir_file \
-  $passthrough_args
+  $passthrough_args_modified
 
-rm $update_files_tar
-rm $dest_dir_file
 
 echo "Artifact $output_path generated successfully:"
 mender-artifact read $output_path
-
+# End of subshell
+)
+rm $update_files_tar
+rm $dest_dir_file
 exit 0
--- a/support/modules-artifact-gen/directory-artifact-gen
+++ b/support/modules-artifact-gen/directory-artifact-gen
@@ -1,4 +1,4 @@
-#!/bin/bash
+#!/bin/sh
 
 set -e
 
@@ -53,7 +53,7 @@ dest_dir_file="dest_dir"
 file_tree=""
 passthrough_args=""
 
-while (( "$#" )); do
+while [ -n "$1" ]; do
   case "$1" in
     --device-type | -t)
       if [ -z "$2" ]; then
@@ -147,7 +147,7 @@ esac
 # Create tarball, accepts directory.
 if [ -e "${file_tree}" ]; then
   if [ -d "${file_tree}" ]; then
-    if [ "$dest_dir" == "/" ]; then
+    if [ "$dest_dir" = "/" ]; then
       echo "Error: this Update Module does not support file tree deployment at /"
       exit 1
     fi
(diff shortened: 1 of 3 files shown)


support/modules-artifact-gen/single-file-artifact-gen
```diff
@@ -1,4 +1,4 @@
-#!/bin/sh
+#!/bin/bash
```
lluiscampos, line 1: (I know this is pre-existing, as the rest of the scripts were already using `bash`)

When possible, we should limit ourselves to `sh` syntax which is more portable. See below for an alternative command for `<<<`

support/modules-artifact-gen/single-file-artifact-gen
```diff
@@ -178,6 +178,30 @@ else
   stat -c %a "${file}" > $permissions_file
 fi
 
+# Check the the passthrough_args and potentially modify them
+# to avoid conflicts or to let them override the already args
+# provided to mender-artifact
+passthrough_args_modified=" "
+parse=$(xargs -n 2 printf "%s %s\n" <<< $passthrough_args)
```
lluiscampos, line 185: An `sh` compatible alternative is
```suggestion
parse=$(echo $passthrough_args | xargs -n 2 printf "%s %s\n")
```

And similar for the next line.

Or, both combined:
```suggestion
echo $passthrough_args | xargs -n 2 printf "%s %s\n" | while read -r flag arg; do
```

support/modules-artifact-gen/single-file-artifact-gen
```diff
@@ -178,6 +178,30 @@ else
...
+while read -r flag arg; do
+  case $flag in
+    -T | --type)
+      echo "Error: Conflicting flag '$flag'. Already specified by the script."
+      exit 1
+      ;;
+    -o | --output-path)
+        output_path=$arg
+      ;;
```
lluiscampos, line 194: Indentation here is off

support/modules-artifact-gen/single-file-artifact-gen
```diff
@@ -178,6 +178,30 @@ else
...
 
+# Check the the passthrough_args and potentially modify them
+# to avoid conflicts or to let them override the already args
+# provided to mender-artifact
+passthrough_args_modified=" "
+parse=$(xargs -n 2 printf "%s %s\n" <<< $passthrough_args)
+while read -r flag arg; do
+  case $flag in
+    -T | --type)
+      echo "Error: Conflicting flag '$flag'. Already specified by the script."
```
lluiscampos, line 189: The point is not that it is already specified (after others like `-o` are also specified), the point is that it is reserved for the script to use (iow it is the script who _decides_ the payload type)

lluiscampos (changes requested):
Nicely done! See comments below.

In addition, may I ask you (in a separate commit) to modify all the generators to be compatible with `sh`? :pray:  This makes sense as the generators are meant to be used for example from CI, where in many thin containers you don't have bash.

Also, this is not a feature in my opinion. Either a fix or a chore. I vote for fix.

support/modules-artifact-gen/directory-artifact-gen
```diff
@@ -1,4 +1,4 @@
-#!/bin/bash
+#!/bin/sh
```
lluiscampos, line 1: Another thing to pay attention when migrating from `bash` to `sh` is the if statements: `==` is bash syntax, `=` is sh syntax. I am getting the following error:
```
$ ./directory-artifact-gen -n hola -t hola --dest-dir /hola /home/lluis/Documents/ -- -T something
./directory-artifact-gen: 150: [: /hola: unexpected operator
Error: Conflicting flag '-T'. Already specified by the script.
```

support/modules-artifact-gen/directory-artifact-gen
```diff
@@ -161,23 +161,46 @@ else
...
-  $passthrough_args
+  $passthrough_args_modified
 
 rm $update_files_tar
 rm $dest_dir_file
 
 echo "Artifact $output_path generated successfully:"
 mender-artifact read $output_path
-
+)
```
lluiscampos, line 205: Good catch with the subshell. I didn't see this coming! :exploding_head: 

To make it more obvious I recommend 1) adding a good comment on why we need the subshell (to be able to override the variables then passed to `mender-artifact` commands) and 2) moving the `rm` commands out of the subshell, as these variables are not being overridden.

lluiscampos (changes requested):
Looking good!

:point_right:  One fix missing wrt migration to `sh`
:point_right:  One suggestion for improving

And,

> Also, this is not a feature in my opinion. Either a fix or a chore. I vote for fix.

I didn't mean to remove the changelog entries, only the category of the commit. I think that both commits are good changes to be ported to release branches and hence they should be either fix or chore and with a user friendly changelog.

lluiscampos (approved):

danielskinstad:

support/modules-artifact-gen/directory-artifact-gen
```diff
@@ -161,23 +161,49 @@ else
...
 # Create dest_dir file in plain text
 echo "$dest_dir" > $dest_dir_file
 
+# Check the the passthrough_args and potentially modify them
+# to avoid conflicts or to let them override the already args
+# provided to mender-artifact
+# Runs in a subshell to allow overriding some parameters passed
+# to mender-artifact
+passthrough_args_modified=" "
+echo -n $passthrough_args | xargs -n 2 printf "%s %s\n" | (while read -r flag arg; do
```
MichaelSteurer, line 173: Not sure if this is the right place to put, @danielskinstad: On MacOS this fails on the command line because it outputs the '-n' and hence breaks the parser in the loop:

`$ /bin/sh -c "echo -n 123"`
prints `-n 123\n` instead of `123`

See [this thread on SO](https://stackoverflow.com/questions/64771044/why-does-echo-n-command-not-work-in-a-sh-script).

PR #1613, opened 2024-04-12
fix: Resolve PARTUUID device paths in rootfs-image

ColinFinck:
PARTUUID device paths (e.g. /dev/disk/by-partuuid/b3c4f349-1180-45e1-9a3d-0a6697f4960e) are symlinks to the regular device paths (e.g. /dev/sda2). They need to be resolved first for the rest of the code to succeed.

Changelog: Fix committing mender-update artifacts when using mender-partuuid
Ticket: None

I have been using a mender-partuuid setup under Yocto Dunfell for the past 4 years. This needed fixes from me in the past such as https://github.com/mendersoftware/meta-mender/pull/1685.
I'm now in the process of upgrading to Yocto Scarthgap (using https://github.com/enlyze/meta-mender/tree/feature/scarthgap-support-for-master) and noticed this regression.

### DIFF
--- a/support/modules/rootfs-image
+++ b/support/modules/rootfs-image
@@ -25,6 +25,22 @@ else
     SETENV=fw_setenv
 fi
 
+resolve_rootfs() {
+    case "$1" in
+        /dev/root|/dev/disk/by-partlabel/*|/dev/disk/by-partuuid/*)
+            # This is a symlink that points to the regular device
+            # (e.g. /dev/disk/by-partuuid/b3c4f349-1180-45e1-9a3d-0a6697f4960e --> /dev/sda2)
+            echo "$(readlink -f $1)"
+            ;;
+
+        *)
+            # Keep any other path as-is
+            # (cf. https://github.com/mendersoftware/mender/pull/1613#discussion_r1584353642 for the reasoning)
+            echo "$1"
+            ;;
+    esac
+}
+
 parse_conf_file() {
     MENDER_ROOTFS_PART_A=""
     MENDER_ROOTFS_PART_B=""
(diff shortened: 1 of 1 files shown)


ColinFinck:
CC @kacf

support/modules/rootfs-image
```diff
@@ -63,6 +63,13 @@ parse_conf_file() {
     MENDER_ROOTFS_PART_A="$(echo $MENDER_ROOTFS_PART_A | sed -e 's,^ubi,/dev/ubi,')"
     MENDER_ROOTFS_PART_B="$(echo $MENDER_ROOTFS_PART_B | sed -e 's,^ubi,/dev/ubi,')"
 
+    # Translate any PARTUUID device path (e.g. /dev/disk/by-partuuid/b3c4f349-1180-45e1-9a3d-0a6697f4960e)
+    # to the regular device path (e.g. /dev/sda2).
+    # This is a no-op if the device path is already a regular one.
+    MENDER_ROOTFS_PART_A="$(readlink -f $MENDER_ROOTFS_PART_A)"
+    MENDER_ROOTFS_PART_B="$(readlink -f $MENDER_ROOTFS_PART_B)"
```
kacf, line 70: IIRC we need to do an explicit check for matching `/dev/disk/by-partuuid/*` first. The reason is when using LVM, resolving the link may point to something like `/dev/dm-5` and the numbering of these devices is not always stable. Therefore we need to keep the original device reference *unless* it is a partuuid reference.

This also implies that using LVM and partuuid together might not work at all, but that is a problem for another day.

ColinFinck: I could add a check for `/dev/disk/by-partuuid`, but I wasn't aware of LVM support in Mender at all.
Even more, the same fix was accepted 4 years ago and still continues to live in meta-mender until today: https://github.com/mendersoftware/meta-mender/pull/1685/files
If LVM needs to be considered here, you may need to consider it in "mender-resize-data-part.sh.in" too.

Let me know if I shall just add a `/dev/disk/by-partuuid` check here.
In any case, I don't have a Mender LVM setup for testing.

kacf: We've never officially supported it, but some people are apparently using it. I'm talking specifically about [this section](https://github.com/mendersoftware/mender/blob/3.5.x/installer/partitions.go#L270-L271) in the Golang client, and its [associated ticket](https://northerntech.atlassian.net/browse/MEN-2302). Looks like `/dev/root` and `/dev/disk/by-partlabel` should be covered too.

So yes, if you could add all of those (including `/dev/root` and `/dev/disk/by-partlabel`) it would be great! I don't have an LVM setup either, but we can at least make the precautions we can.

Let's leave the PR you linked to out of it for now. That can be fixed separately by whoever feels motivated.

ColinFinck: @kacf Fixed in 542be8f22fca3e7bda927e9dd0c28503d5f6dd69

lluiscampos:
Hi @ColinFinck. It looks good to me. There are two issues with CI:
* Your second commit is not signed nor it follows the expected formatting. I think it is best to just squash both of your commits.
* There is an unrelated CI failure that is fixed in master. Please rebase your work on top of latest master.

Then I'll run the full pipeline for you and wait for final blessing from @kacf .

ColinFinck:
@lluiscampos Thanks for your reply! I did as you suggested in 80ac1f8c1ccac16efa28ed1075eee2d789258af1

support/modules/rootfs-image
kacf: Thanks @ColinFinck!

kacf (approved):

PR #1615, opened 2024-04-23
fix: Improve error messages when executable can't be found.

kacf:
The message already existed in the non-redirect case, but not when collecting output.

Changelog: Title
Ticket: None


lluiscampos (approved):

kacf:
I'll risk omitting the QA pipeline for this one.

PR #1618, opened 2024-04-29
chore: function for checking if files are identical

danielskinstad:
See https://github.com/mendersoftware/mender-update-orchestrator/pull/90

Ticket: MEN-7171

### DIFF
--- a/src/common/CMakeLists.txt
+++ b/src/common/CMakeLists.txt
@@ -71,6 +71,7 @@ target_link_libraries(common_io PUBLIC common)
 
 add_library(common_path STATIC)
 target_sources(common_path PRIVATE
+  path/path.cpp
   path/platform/c++17/path.cpp
   path/platform/posix/path.cpp
 )
--- a/src/common/path.hpp
+++ b/src/common/path.hpp
@@ -71,6 +71,8 @@ error::Error DeleteRecursively(const string &path);
 
 expected::ExpectedBool IsExecutable(const string &path, const bool warn = true);
 
+expected::ExpectedBool AreFilesIdentical(const string &file_one, const string &file_two);
+
 expected::ExpectedUnorderedSet<string> ListFiles(const string &in, function<bool(string)> matcher);
 
 error::Error CreateDirectory(const string &path);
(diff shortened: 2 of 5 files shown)


src/common/path/platform/c++17/path.cpp
```diff
@@ -112,6 +116,42 @@ expected::ExpectedBool IsExecutable(const string &file_path, const bool warn) {
...
+	string file_two_contents {
+		istreambuf_iterator<char>(file_two_stream), istreambuf_iterator<char>()};
+
+	size_t file_one_hash = std::hash<string> {}(file_one_contents);
+	size_t file_two_hash = std::hash<string> {}(file_two_contents);
+
+	if (file_one_hash == file_two_hash) {
+		return true;
+	}
+	return false;
```
lluiscampos, line 152:
```suggestion
	return file_one_hash == file_two_hash;
```

lluiscampos (commented):
Looks good but we are missing the test :sweat_smile:

src/common/path/platform/c++17/path.cpp
```diff
@@ -112,6 +116,40 @@ expected::ExpectedBool IsExecutable(const string &file_path, const bool warn) {
 	}
 }
 
+expected::ExpectedBool AreFilesIdentical(const string &file_one, const string &file_two) {
```
lluiscampos, line 119: After closer inspection, I realized that this function does not use any of the `c++17` features (`fs` namespace`) so it could very well be made generic and moved to `src/common/path/path.cpp`, then adding it to the same library with:
```
diff --git a/src/common/CMakeLists.txt b/src/common/CMakeLists.txt
index ada11934..b361e43d 100644
--- a/src/common/CMakeLists.txt
+++ b/src/common/CMakeLists.txt
@@ -71,6 +71,7 @@ target_link_libraries(common_io PUBLIC common)
 
 add_library(common_path STATIC)
 target_sources(common_path PRIVATE
+  path/path.cpp
   path/platform/c++17/path.cpp
   path/platform/posix/path.cpp
 )
```

lluiscampos (approved):
It looks good. I recommend you test it together with your orchestrator PR before merging, but otherwise you can merge and then update there.

I made just one more nit pick observation below (sorry to came back late with feedback yet again about where to place this code... :sweat_smile: )

lluiscampos (approved):

danielskinstad:

PR #1620, opened 2024-04-30
fix: avoid leak of tenant token on the logs

MuchoLucho:
Changelog: None
Ticket: None

the form below:

```

<OPTIONAL LONGER DESCRIPTION>

```

### Description

Thank you!

### DIFF
--- a/app/auth.go
+++ b/app/auth.go
@@ -676,7 +676,14 @@ func (m *menderAuthManagerService) MakeAuthRequest() (*client.AuthRequest, error
 
 	tentok := strings.TrimSpace(string(m.tenantToken))
 
-	log.Debugf("Tenant token: %s", tentok)
+	tentok4Log := tentok
+
+	// Truncates the tenant token for avoid any leak
+	if len(tentok) > 4 {
+		tentok4Log = tentok[0:2] + "..." + tentok[len(tentok)-2:]
+	}
+
+	log.Debugf("Tenant token: %s", tentok4Log)
 
 	// fill tenant token
 	authd.TenantToken = string(tentok)


MuchoLucho:
I am not a Go guy, so I am sure it can be made more elegant. Open to suggestions

buentead:
@MuchoLucho : Thanks for the change. From a security point of view a must have. It will be easier to share the debug logs. Very much appreciated!

app/auth.go
```diff
@@ -676,7 +676,14 @@ func (m *menderAuthManagerService) MakeAuthRequest() (*client.AuthRequest, error
...
 
-	log.Debugf("Tenant token: %s", tentok)
+	tentok4Log := tentok
+
+	// Truncates the tenant token for avoid any leak
+	if len(tentok) > 4 {
+		tentok4Log = tentok[0:2] + "..." + tentok[len(tentok)-2:]
+	}
+
+	log.Debugf("Tenant token: %s", tentok4Log)
```
lluiscampos, line 686: With regards to the general logic. It won't happen in practice, but look at the following:
* do a copy of tenok to tenok4log
* `if` something something, then redact tenok4log for sercure logging
* log unconditionally

See the issue? If the `if` is false then you log the original secret tenok.

There are many ways to solve this. One way:
```suggestion
	tentok4Log := ""

	// Truncates the tenant token for avoid any leak
	if len(tentok) > 4 {
		tentok4Log = tentok[0:2] + "..." + tentok[len(tentok)-2:]
	}

	log.Debugf("Tenant token: %s", tentok4Log)
```

and a simpler one:

```suggestion
	// Truncates the tenant token for avoid any leak
	if len(tentok) > 4 {
		log.Debugf("Tenant token: %s..%s", tentok[0:2], tentok[len(tentok)-2:])
	}
```

app/auth.go
```diff
@@ -676,7 +676,14 @@ func (m *menderAuthManagerService) MakeAuthRequest() (*client.AuthRequest, error
...
 
 	// fill tenant token
 	authd.TenantToken = string(tentok)
```
lluiscampos, line 689: Two lines below we have `log.Debugf("Authorization data: %v", authd)` which also will leak the token, right?

lluiscampos (changes requested):
Great first start in golang, @MuchoLucho ! See my comments below I think we can brush this up and merge it.

Also please add a good changelog entry, this is clearly user facing and we will release it.

lluiscampos:
3.5.x is EOL, closing this PR.

Thank you @MuchoLucho regardless for your efforts :smiley:

PR #1622, opened 2024-05-02
ci: Fix coverage reporting for branches

lluiscampos:
The code that accepts `CI_BRANCH` as a generic environment variable is actually not released upstream, and the latest pypi avilable version relies on `TRAVIS_*` type of variable.

See: https://github.com/eddyxu/cpp-coveralls/compare/v0.4.2...47f638727512126bc16b1365560854a86db70e7e#diff-b1eda260daddc41aa91c216b3b054933cc85304b6d5e3235c652f0b48c59fcf9R46

As we were missing the branch name, in the eyes of coveralls the "master" branch is frozen on golang code.


kacf (approved):
Nice find!

PR #1623, opened 2024-05-03
4.0.x: feat: Add `SYSTEMD_UNIT_DIR` CMake variable.

kacf:
Use it to customize the location of the systemd unit files. Usually it is set to `/lib/systemd/system`.

Changelog: Commit
Ticket: None

(cherry picked from commit 943b8198bff3328229f0f3055989807836c9f7d6)

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -57,10 +57,6 @@ test:backward-compat:
   image: debian:10
   before_script:
     - apt update && apt install -yyq g++ git make lcov pkg-config liblmdb++-dev libboost-dev libboost-log-dev libboost-regex-dev libssl-dev libarchive-dev libdbus-1-dev curl dbus
-    # mender-artifact install
-    - curl https://downloads.mender.io/mender-artifact/master/linux/mender-artifact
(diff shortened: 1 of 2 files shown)


kacf:
@lluiscampos: This is marked `feat`, but it's a build-time-only feature, so I think it's ok to cherry-pick to 4.0.x. And this will help getting https://github.com/mendersoftware/meta-mender/pull/2093 fixed without requiring different fixes for 4.0.x and master.

lluiscampos (approved):

lluiscampos:
> @lluiscampos: This is marked `feat`, but it's a build-time-only feature, so I think it's ok to cherry-pick to 4.0.x. And this will help getting [mendersoftware/meta-mender#2093](https://github.com/mendersoftware/meta-mender/pull/2093) fixed without requiring different fixes for 4.0.x and master.

Yes, it is arguably not a *software* feature but a build system feature :+1:

kacf:
The only error is the `test_watchdog_kick` test. The test should not have run, it was me who set the versions incorrectly when I started the job. IOW, not important, so I'm merging.

PR #1625, opened 2024-05-08
Tee reader problems

kacf:


src/common/io.hpp
```diff
@@ -297,9 +297,9 @@ class BufferedReader : virtual public Reader {
 	bool rewind_consumed_ {false};
 	bool stop_done_ {false};
 
-protected:
```
lluiscampos, line 300: This was meant for testing - see the CI failure

lluiscampos (approved):
Hats off for finding this issue. The bug is pretty clear in hindsight.

src/common/io.hpp
kacf: Right.

kacf:
TeeReader is not used by the client, only by downstream projects, so merging without a full run.

PR #1626, opened 2024-05-14
Improvements

kacf:
Just various improvements I made while investigating a bug. In the end I didn't actually need any of these fixes, but they still seem like worthwhile improvements.

### DIFF
--- a/src/common/events/events_io.cpp
+++ b/src/common/events/events_io.cpp
@@ -292,7 +292,6 @@ error::Error TeeReader::RemoveReader(TeeReader::TeeReaderLeafPtr leaf_reader) {
 
 TeeReader::~TeeReader() {
 	leaf_readers_.clear();
-	buffered_reader_->Cancel();
 }
 
 error::Error TeeReader::TeeReaderLeaf::AsyncRead(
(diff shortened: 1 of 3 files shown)


src/common/events/events_io.cpp
```diff
@@ -306,8 +314,21 @@ error::Error TeeReader::TeeReaderLeaf::AsyncRead(
...
-	p->RemoveReader(shared_from_this());
+	if (!p) {
+		// Already disconnected from the tee reader. Nothing to do.
+		return;
+	}
+
+	auto leaf_reader = p->leaf_readers_.find(shared_from_this());
+	common::io::AsyncIoHandler handler;
+	if (leaf_reader != p->leaf_readers_.end()) {
+		handler = p->leaf_readers_.find(shared_from_this())->second.pending_read.handler;
```
lluiscampos, line 325: Why do we need to copy it here? It will be copied also inside `CancelReader()` before erasing the leaf, isn't that one enough?

kacf: Hmm, maybe we don't. I think I may have used it below, but then changed the approach later, and I forgot to modify this part. Let me experiment with it.

kacf: It's not needed. Removed!

PR #1632, opened 2024-05-31
fix: Invalid Range request when content_length is 0 (unknown)

fellerts:
If an HTTP download is resumed without the client ever having learned the `content_length`, it will request the update with the following invalid header:
```
Range: bytes=0--1
```
Changelog: None
Ticket: #8880

### DIFF
--- a/src/common/http/http_resumer.cpp
+++ b/src/common/http/http_resumer.cpp
@@ -413,10 +413,12 @@ io::ExpectedAsyncReaderPtr DownloadResumerClient::MakeBodyAsyncReader(
 
 http::OutgoingRequestPtr DownloadResumerClient::RemainingRangeRequest() const {
 	auto range_req = make_shared<http::OutgoingRequest>(*user_request_);
-	range_req->SetHeader(
-		"Range",
-		"bytes=" + to_string(resumer_state_->offset) + "-"
-			+ to_string(resumer_state_->content_length - 1));
+	if (resumer_state_->content_length > 0) {
+		range_req->SetHeader(
+			"Range",
+			"bytes=" + to_string(resumer_state_->offset) + "-"
+				+ to_string(resumer_state_->content_length - 1));
+	}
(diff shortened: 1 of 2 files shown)


lluiscampos (changes requested):
Hi @fellerts and thank you very much for your contribution.

The fix looks good :rocket: I only have comments for the commit formalities and testing:

Being a bug fix, we need the `Changelog` to have something else than None so that we cherry-pick and release it into stable branch. In this case your title is good, so `Changelog: Title` should be good.

With regards to the `Ticket`. I don't know what you are referring to in your commit but our commit checked only expects Mender JIRA tickets, so better to leave to `Ticket: None`.

Finally, would you write a unit test for this? It is a good thing to test for future regressions. If you don't know how to do it I will amend your commit with the test myself and we co-author the fix :+1:

lluiscampos:
> [...] Can you point me in the right direction?

After following [these steps](https://github.com/mendersoftware/mender?tab=readme-ov-file#steps) you should have a working environment. Then you can run all the tests with `make check`.

We develop on Linux, so although CMake is cross-platform we haven't really tested it elsewhere. If you get into trouble just let me know and I'll do the test :smiley:

fellerts:
@lluiscampos thanks! I've added a test for this and fixed the commit message like you asked. I run Fedora 38 and had no issues running your tests as described.

lluiscampos (approved):

lluiscampos:
> @lluiscampos thanks! I've added a test for this and fixed the commit message like you asked. I run Fedora 38 and had no issues running your tests as described.

Excellent work. Thank you!

I hope that you don't mind that I amended the formatting and slightly modified the commit (literal `Changelog: Title` instead of copying the title) and force-pushed to your fork.

I will run it through our pipeline and merge soon.

lluiscampos:

lluiscampos:
* 4.0.x

PR #1635, opened 2024-06-05
fix: parse artifact ID before validating payload size

fellerts:
If a deployment fails before the artifact info is parsed, e.g. when a deployment is started but S3 is blocked by a firewall, the JSON state data describes an artifact with an empty `payload-types` list. If mender-update is subsequently restarted, the JSON state data unmarshaler would exit early and return invalid state data (empty deployment ID) triggering an assert in `DeploymentClient::PushStatus`. The client would never recover from this.

Changelog: Title
Ticket: None

---

I reported this issue to support@northern.tech a couple of weeks ago, so there might be an internal Mender ticket for this (edit: MEN-7327). The ticket contains steps to reproduce this problem. [Here](https://github.com/user-attachments/files/15586541/mender-update-trace.log) is a trace-level log.

I don't know where to write a test for this unfortunately, so I would appreciate some help.

### DIFF
--- a/src/mender-update/daemon/context.cpp
+++ b/src/mender-update/daemon/context.cpp
@@ -384,14 +384,19 @@ static error::Error UnmarshalJsonStateDataVersion1(const json::Json &json, State
 	const auto &exp_json_update_info = json.Get("UpdateInfo");
 	SetOrReturnIfError(const auto &json_update_info, exp_json_update_info);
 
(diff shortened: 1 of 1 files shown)


MuchoLucho:
Hello @kacf and @lluiscampos I can confirm this one is linked to https://northerntech.atlassian.net/browse/MEN-7327

lluiscampos (approved):

PR #1636, opened 2024-06-06
chore: Rewrote FilterProvides to not use regex

danielskinstad:
Replaced the regex in FilterProvides by matching a substring with the whole provides-string. This is done to remove ExceptionToErrorOrAbort. Added unit tests and a testing-function to context.

Ticket MEN-7323

### DIFF
--- a/src/mender-update/context/context.cpp
+++ b/src/mender-update/context/context.cpp
@@ -17,7 +17,6 @@
 #include <cctype>
 
 #include <algorithm>
-#include <regex>
 #include <set>
 
 #include <artifact/artifact.hpp>
@@ -234,39 +233,60 @@ expected::ExpectedString MenderContext::GetDeviceType() {
 	return expected::ExpectedString(ret);
 }
 
-static error::Error FilterProvides(
+bool CheckClearsMatch(const string &to_match, const string &clears_string) {
+	if (clears_string.empty()) {
+		return to_match.empty();
+	}
+
+	vector<std::string> sub_strings;
+	string escaped;
+	for (const auto chr : clears_string) {
+		if (chr == '*') {
+			sub_strings.push_back(escaped);
+			escaped.clear();
+		} else {
+			escaped.push_back(chr);
+		}
+	}
+	sub_strings.push_back(escaped);
+
+	// Make sure that that front of vector starts at index 0
+	if (sub_strings.front() != ""
+		&& to_match.compare(0, sub_strings.front().size(), sub_strings.front()) != 0) {
+		return false;
+	}
+	// Checks if no trailing wildcard
+	if (sub_strings.back() != ""
+		&& to_match.compare(
+			   to_match.size() - sub_strings.back().size(), to_match.size(), sub_strings.back())
+			   != 0) {
+		return false;
+	}
+
+	// Iterate over substrings, set boundary if found to avoid
+	// matching same substring twice
+	size_t boundary = 0;
+	for (const auto &str : sub_strings) {
+		if (!str.empty()) {
+			size_t find = to_match.find(str, boundary);
+			if (find == string::npos) {
+				return false;
+			}
+			boundary = find + str.size();
+		}
+	}
+	return true;
+}
+
+error::Error FilterProvides(
 	const ProvidesData &new_provides,
 	const ClearsProvidesData &clears_provides,
 	ProvidesData &to_modify) {
 	// Use clears_provides to filter out unwanted provides.
 	for (auto to_clear : clears_provides) {
-		string escaped;
-		// Potential to escape every character, though unlikely.
-		escaped.reserve(to_clear.size() * 2);
-		// Notable exception: '*', since it has special handling as a glob character.
-		string meta_characters {".^$+()[]{}|?"};
-		for (const auto chr : to_clear) {
-			if (chr == '*') {
-				// Turn every '*' glob wildcard into '.*' regex wildcard.
-				escaped.push_back('.');
-			} else if (any_of(meta_characters.begin(), meta_characters.end(), [chr](char c) {
-						   return chr == c;
-					   })) {
-				// Escape every regex special character except '*'.
-				escaped.push_back('\\');
-			}
-			escaped.push_back(chr);
-		}
-
-		regex compiled;
-		auto err = error::ExceptionToErrorOrAbort(
-			[&compiled, &escaped]() { compiled.assign(escaped, regex_constants::basic); });
-		// Should not be possible, since the whole regex is escaped.
-		AssertOrReturnError(err == error::NoError);
-
 		set<string> keys;
 		for (auto provide : to_modify) {
-			if (regex_match(provide.first.begin(), provide.first.end(), compiled)) {
+			if (CheckClearsMatch(provide.first, to_clear)) {
 				keys.insert(provide.first);
 			}
 		}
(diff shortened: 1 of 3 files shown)


danielskinstad:

lluiscampos:
@kacf could you please take this one? :pray:

src/mender-update/context/context.cpp
```diff
@@ -243,30 +242,24 @@ static error::Error FilterProvides(
...
-		// Should not be possible, since the whole regex is escaped.
-		AssertOrReturnError(err == error::NoError);
-
 		set<string> keys;
 		for (auto provide : to_modify) {
-			if (regex_match(provide.first.begin(), provide.first.end(), compiled)) {
+			if (escaped == provide.first) {
+				keys.insert(provide.first);
+				// Match escaped 'wildcard substring'
+			} else if (provide.first.rfind(escaped, 0) == 0 && wildcard) {
```
kacf, line 262: I think this could give false negatives. Imagine the string `test_test_key`. If the search string has `test_*`, then it will not match because `rfind` finds the second occurrence first, which is not at the zero index.

A quick fix is to use `find` instead. On the other hand, `find` is overkill in this case, because we know that the string will always be at the start, so perhaps it's better to use `provide.first.compare(0, escaped.size(), escaped) == 0` (see [`compare`](https://en.cppreference.com/w/cpp/string/basic_string/compare)).

src/mender-update/context/context.cpp
```diff
@@ -283,6 +276,29 @@ static error::Error FilterProvides(
...
+	for (const auto &clear : clears_provides) {
+		clears.insert(clear);
+	}
+
+	for (const auto &provide : provides) {
+		if (clears.find(provide.first) != clears.end()) {
+			return false;
+		}
+	}
+	return true;
```
kacf, line 299: I don't quite follow why this function is here. Why not use the original directly?

Also, this function returns both `false` and `unexpected` in certain cases, but none of these conditions are being tested. At least the function should live in the test file then, if the false/error conditions are only meant to catch mistakes.

src/mender-update/context/context.cpp
```diff
@@ -243,30 +242,24 @@ static error::Error FilterProvides(
...
-				escaped.push_back('.');
-			} else if (any_of(meta_characters.begin(), meta_characters.end(), [chr](char c) {
-						   return chr == c;
-					   })) {
-				// Escape every regex special character except '*'.
-				escaped.push_back('\\');
+				// Acts as a regex wildcard, break so `*` isn't included,
+				// this allows us to use the substring to match like a regex
+				wildcard = true;
+				break;
```
kacf, line 252: Lluis and I talked about this, and we think it's ok if we don't support a `*` in the middle of a string, for example `test_*_stuff`. It's difficult to do this kind of matching, and there are few use cases for it. And that is indeed what is done here. However, there are some important details to fix:

1. It needs to test that nothing comes after the wildcard. It's ok to simply return an error for this, explaining that it is not supported (you can use the `errc::not_supported` error code for this, just grep in the code). Make sure there is a test for this below as well.
2. It's a behavior change that you can no longer have a wildcard in the middle of the string, and the changelog should mention this.

@lluiscampos: Although technically this is a breaking change, I think we are ok with treating it as a small change and not triggering a new major release, right? Given its likely very low impact.

kacf (changes requested):

src/mender-update/context/context.cpp
kacf: This was bugging me over the weekend. I find that I am actually a bit uneasy that we are breaking compatibility by only allowing one wildcard at the end. Even if it is an uncommon use case to use anything else. The functionality has been out there for 5+ years already.

But maybe it's not so hard to fix. We really only need to split the string on each wildcard, and search for them one by one. For example in the case of `abc*def*ghi`, we start by searching `abc`, then search for `def` in the rest of the string, then `ghi` in the rest of that string again. There are some extra details such as making sure that the start and end are anchored to the start and end of string, but this should not be hard.

src/mender-update/context/context.cpp
danielskinstad: Well, `FilterProvides` is a static function, and I implemented the tests based on the test for `FilterProvides` in the orchestrator. I now realize, however, that the function doesn't work as intended anyways. So either I have to change the function to make it actually work or make `FilterProvides` accessible for the tests. What do you think?

src/mender-update/context/context.cpp
danielskinstad: This won't happen, because I specified `pos` (the second argument) as 0. This sets the position of the last character to be considered the beginning of a match to index 0, but this just seems like find with extra steps, so `compare` seems better and I will use that instead.

src/mender-update/context/context.cpp
kacf: It's fine to make the function non-static and add it to the header file. This is quite common to do in test situations such as this one, and is usually preferred over adding "special" test functions. The closer to the real thing you are, the better! 🙂

tests/src/mender-update/context_test.cpp
```diff
@@ -695,6 +695,106 @@ TEST(ContextArtifactTests, ArtifactMatchesContextTest) {
...
+		{"abc", "*abc", true},
+		{"abc", "a*xyz", false},
+		{"test_test_key", "***", true}};
+
+
+	for (const auto &test : test_wildcard) {
+		bool result = context::CheckClearsMatch(test.to_match, test.pattern);
+		ASSERT_EQ(result, test.expected) << "Expected '" << test.to_match << "', '" << test.pattern
+										 << "' to return " << test.expected;
+	}
```
lluiscampos, line 736: Very good test!

src/mender-update/context/context.cpp
```diff
@@ -234,39 +233,52 @@ expected::ExpectedString MenderContext::GetDeviceType() {
...
+			sub_strings.push_back(escaped);
+			escaped.clear();
+		} else {
+			escaped.push_back(chr);
+		}
+	}
+	sub_strings.push_back(escaped);
+
+	// Make sure that that front of vector starts at index 0
+	if (sub_strings.front() != "" && to_match.find(sub_strings.front()) != 0) {
```
kacf, line 254: Here you could use `compare`, but `find` also works.

src/mender-update/context/context.cpp
```diff
@@ -234,39 +233,52 @@ expected::ExpectedString MenderContext::GetDeviceType() {
...
+	// Checks if no trailing wildcard
+	if (sub_strings.back() != ""
+		&& to_match.compare(
+			   to_match.size() - sub_strings.back().size(), to_match.size(), sub_strings.back())
+			   != 0) {
+		return false;
+	}
+
+	for (const auto &str : sub_strings) {
+		if (to_match.find(str) == std::string::npos) {
```
kacf, line 266: I think you need to take into account where the previous sub-match ended, don't you? Otherwise `test_*test_*` would match `test_string`, but it shouldn't.

tests/src/mender-update/context_test.cpp
kacf: Yes, very!

src/mender-update/context/context.cpp
danielskinstad: Yes, you're right :)

kacf (approved):
Perfect! Thanks for sticking with it!

danielskinstad:

PR #1637, opened 2024-06-06
feat: Implement support for compiling without DBus.

kacf:
@lluiscampos: This looks big, but a lot of it is just moving things around and making it work under two different configurations. The heart of the feature, and hopefully the easiest way to understand what's going on here, is looking at the two files `dbus/auth.cpp` and `embedded_auth.cpp`, which are the two different implementations: One using DBus, one using direct calling.


src/api/auth/auth.cpp
```diff
@@ -63,105 +51,65 @@ error::Error MakeError(AuthenticatorErrorCode code, const string &msg) {
...
-					"Failed to request new token fetching (see mender-auth logs for details)");
-				ExpectedAuthData ex_auth_data = expected::unexpected(MakeError(
-					AuthenticationError, "Failed to request new token fetching from mender-auth"));
-				PostPendingActions(ex_auth_data);
-			}
-		});
+error::Error Authenticator::RequestNewToken() {
+	auto err = FetchJwtToken();
 	if (err != error::NoError) {
 		// A sync DBus error.
```
lluiscampos, line 98: This comment is wrong now.

src/api/auth/auth.cpp
```diff
@@ -172,73 +120,38 @@ error::Error Authenticator::RequestNewToken(optional<AuthenticatedAction> opt_ac
...
-			// Should never fail. We rely on the signal heavily so let's fail
-			// hard if it does.
-			return err;
+void Authenticator::HandleReceivedToken(
+	common::ExpectedStringPair ex_auth_dbus_data, NoTokenAction no_token) {
+	auth_timeout_timer_.Cancel();
+	ExpectedAuthData ex_auth_data;
+	if (!ex_auth_dbus_data) {
+		mlog::Error(
+			"Error from the JwtTokenStateChange DBus signal: "
```
lluiscampos, line 136: "Error receiving the token"?

src/api/auth/platform/dbus/auth.cpp
```diff
@@ -0,0 +1,93 @@
...
+//
+//    Unless required by applicable law or agreed to in writing, software
+//    distributed under the License is distributed on an "AS IS" BASIS,
+//    WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+//    See the License for the specific language governing permissions and
+//    limitations under the License.
+
+#include <api/auth.hpp>
+
+#include <mutex>
```
lluiscampos, line 17: `mutex` is not really in use, is it?

src/api/auth/platform/dbus/auth.cpp
```diff
@@ -0,0 +1,93 @@
...
+				ExpectedAuthData ex_auth_data = expected::unexpected(MakeError(
+					AuthenticationError, "Failed to request new token fetching from mender-auth"));
+				PostPendingActions(ex_auth_data);
+			}
+		});
+}
+
+} // namespace auth
+} // namespace api
+} // namespace mender
```
lluiscampos, line 93: This implementation reads so much better now :tada:

src/mender-auth/api/auth/embedded_auth.cpp
```diff
@@ -0,0 +1,68 @@
...
+
+error::Error AuthenticatorHttp::FetchJwtToken() {
+	return FetchJWTToken(
+		client_,
+		config_.servers,
+		crypto_args_,
+		config_.paths.GetIdentityScript(),
+		[this](APIResponse resp) { FetchJwtTokenHandler(resp); },
+		config_.tenant_token);
+}
```
lluiscampos, line 63: As promised: very straight forward!

src/mender-auth/api/auth/embedded_auth.cpp
```diff
@@ -0,0 +1,68 @@
...
+	});
+	return error::NoError;
+}
+
+void AuthenticatorHttp::FetchJwtTokenHandler(APIResponse resp) {
+	if (resp) {
+		token_ = resp.value().token;
+		server_url_ = resp.value().server_url;
+
+		log::Info("Successfully received new authorization data");
```
lluiscampos, line 44: I like that this message is the exact same as in `mender::auth::ipc::FetchJwtTokenHandler` - so the whole replacement is transparent to the user (and potential tests!)

src/mender-auth/api/auth/embedded_auth.cpp
```diff
@@ -0,0 +1,68 @@
...
+	} else {
+		token_.clear();
+		server_url_.clear();
+
+		log::Error("Failed to fetch new token: " + resp.error().String());
+	}
+
+	PostPendingActions(mender::api::auth::AuthData {token_, server_url_});
```
lluiscampos, line 52: Reasoning out loud and for me to understand,

Here there is a key difference between D-Bus and embedded: in D-Bus a fetch does `PostPendingActions` on error or return `false` from D-Bus, while the success case will eventually be called via the signal callback. Here we have one less level of async :+1:

src/mender-update/daemon/context.cpp
```diff
@@ -151,7 +151,11 @@ Context::Context(
 	mender::update::context::MenderContext &mender_context, events::EventLoop &event_loop) :
 	mender_context(mender_context),
 	event_loop(event_loop),
+#ifdef MENDER_USE_DBUS
 	authenticator(event_loop),
+#elif defined(MENDER_EMBED_MENDER_AUTH)
+	authenticator(event_loop, mender_context.GetConfig()),
+#endif
```
lluiscampos, line 158: These kind of ifdefs for constructors I am expecting to see more and more as we do the ports to different platforms. I don't really love it but I understand that implementing a pattern like Abstract Factories (or Factory Method, I always confuse them) is not justified.

Do you have any thoughts on this moving forward? How many variants you think is acceptable for an ifdef until the code gets unmaintainable in your opinion?

src/mender-auth/cli/actions.hpp
```diff
@@ -47,6 +47,10 @@ class Action {
 using ActionPtr = shared_ptr<Action>;
 using ExpectedActionPtr = expected::expected<ActionPtr, error::Error>;
 
+shared_ptr<MenderKeyStore> KeystoreFromConfig(
+	const conf::MenderConfig &config, const string &passphrase);
```
lluiscampos, line 51: This helper we should rather move to `mender::client_shared::conf` so that it can be used from `mender-update` without requiring this coupling. This way we are also free to revert 39ae079890f109c7f06234576e1d9dbc2d74f12 which I don't think it has value for production, only for prototypoing.

lluiscampos (commented):
Thank you very much for addressing this. I got exactly what I wanted: a PoC to put peace on my mind :lotus_position: - It actually does more than PoC as you said, because besides tests and some cosmetics, it really looks good.

See comments below for some small details that we could improve.

Other than that, what should we do with this PR for now? I suggest we merge into a branch like `poc/embedded-dbus` and create a follow-up task to fix tests and potentially revert [39ae079](https://github.com/mendersoftware/mender/pull/1637/commits/e39ae079890f109c7f06234576e1d9dbc2d74f12), and do a final production review

kacf:
> Other than that, what should we do with this PR for now? I suggest we merge into a branch like `poc/embedded-dbus` and create a follow-up task to fix tests and potentially revert [39ae079](https://github.com/mendersoftware/mender/pull/1637/commits/e39ae079890f109c7f06234576e1d9dbc2d74f12), and do a final production review

I was actually thinking we could merge it. It does not harm the original code, and I would argue that it's structured better now than before. And we know we will need it (unless we use the MCU client, but it's not looking likely; we'll know soon).

I was debating whether to make a compile job for this in the CI. I decided not to, but mostly to avoid scope creep, not because I think it's a bad idea. This would keep it from bit rotting until it is used in Zephyr, and the built in mender-auth client makes it easier to test this on Linux, without having to go through Zephyr to do it.

Or do you think it's "too far" from the production target and should stay in a feature branch?

src/mender-auth/api/auth/embedded_auth.cpp
kacf: Exactly. Even the asking is async in the DBus case, while the actual reply (via signal) is separate, so we have two levels of async calls. Here we have only one, because we do the HTTP call directly.

src/mender-update/daemon/context.cpp
kacf: I don't have a definitive answer to that. A possible solution to the case here is to require the same arguments in both cases, even when it's not needed.

But more generally I would agree that factory functions should be preferred. Hopefully we can avoid many ifdefs with that.

However, one type of ifdef which is unavoidable with the current design, is ifdef when either declaring the type, or when using `make_unique`/`make_shared`. See these examples:

```c++
#ifdef LINUX
KeyValueDatabaseLmdb database_;
#else
KeyValueDatabaseZephyr database_;
#endif

- or -

#ifdef LINUX
auto database = make_shared<KeyValueDatabaseLmdb>();
#else
auto database = make_shared<KeyValueDatabaseZephyr>();
#endif
```

This is because each platform have their own type.

Back when Ole, Vratislav and I were discussing this, we were talking about the alternative where `KeyValueDatabase` would not have different names, but would instead be defined once in each platform section, so you would have multiple `KeyValueDatabase::KeyValueDatabase` constructors for example, once for each platform. The platform name would not be embedded in the name. This would eliminate many ifdefs in the production code, which might be an advantage. I know this is how Qt does it.

In the end we decided against it in favor of using qualified names. I think there was no strong reason in favor of either approach at the time, and it was more a case of not being able to plan everything in detail that far ahead. But I don't think there is a strong reason to stay with this approach, if the alternative makes the code cleaner.

src/api/auth/platform/dbus/auth.cpp
kacf: It does! 😃

src/mender-auth/api/auth/embedded_auth.cpp
kacf: I'm a man of my word! 😄

src/mender-auth/api/auth/embedded_auth.cpp
kacf: Agreed!

src/mender-auth/cli/actions.hpp
kacf: I thought about that, but it doesn't work without moving more code from mender-auth. The problem is that this function depends on the `MenderKeyStore` type, and you can't include that file from `mender::client_shared::conf` because it would be a circular dependency. Then we would also need to move that type and the associated functions also from mender-auth to client_shared, and I figured that all the related code was too specific to mender-auth to move into the shared section. I could be convinced otherwise I suppose, but that was my reasoning.

lluiscampos:
> > Other than that, what should we do with this PR for now? I suggest we merge into a branch like `poc/embedded-dbus` and create a follow-up task to fix tests and potentially revert [39ae079](https://github.com/mendersoftware/mender/pull/1637/commits/e39ae079890f109c7f06234576e1d9dbc2d74f12), and do a final production review
> 
> I was actually thinking we could merge it. It does not harm the original code, (...)

Specifically that commit, in my opinion it harms readability for not good enough reason (the use case is internal testing, and it is an experimental feature)

The rest of the changes I agree that make the code better structured :+1: 

> 
> I was debating whether to make a compile job for this in the CI. I decided not to, but mostly to avoid scope creep, not because I think it's a bad idea. This would keep it from bit rotting until it is used in Zephyr, and the built in mender-auth client makes it easier to test this on Linux, without having to go through Zephyr to do it.

I also think it is a good idea. Once we merge it opens for this use case of having a single binary for Linux, which users might be interested. Hence we should build it.

> 
> Or do you think it's "too far" from the production target and should stay in a feature branch?

To merge it we need to fix the tests, mainly. If we fix the tests I won't oppose to merging as-is (with the embedded cli too)

kacf:
> Specifically that commit, in my opinion it harms readability for not good enough reason (the use case is internal testing, and it is an experimental feature)

The thing is that the next commit depends on it. I could cut out the command line part, leaving only the cmake flag, but then it is not testable on Linux anymore (and yes, I did have bugs in the first version).

I think I'll go for fixing the tests then! Then I will also add a CI job.

lluiscampos (approved):
Awesome :medal_sports: Thanks!

PR #1611, opened 2024-04-04
feat: Showcase the configured intervals as inventory items

MuchoLucho:
It lets the user to check how soon the next poll would be and make decisions if that doesn't happen.

Changelog: None
Ticket: None

# External Contributor Checklist

🚨 Please review the [guidelines for contributing](https://github.com/mendersoftware/mender/blob/master/CONTRIBUTING.md) to this repository.

- [x] Make sure that all commits follow the conventional commit [specification](https://www.github.com/mendersoftware/mendertesting/commitlint/grammar.md) for the Mender project.

The majority of our contributions are fixes, which means your commit should have
the form below:

```
fix: <SHORT DESCRIPTION OF FIX>

<OPTIONAL LONGER DESCRIPTION>

Changelog: <USER-FRIENDLY-CHANGE-DESCRIPTION> or <None>
Ticket: <TICKET NUMBER> or <None>
```

- [x] Make sure that all commits are signed with [`git --signoff`](https://git-scm.com/book/en/v2/Git-Tools-Signing-Your-Work). Also note that the signoff author must match the author of the commit.

### Description

This suggestion came from an Ent customer and I found it cheap to implement myself

Thank you!

PR #1640, opened 2024-06-12
feat: Showcase the configured intervals as inventory items

MuchoLucho:
It lets the user to check how soon the next poll would be and make decisions if that doesn't happen.

Changelog: None
Ticket: None

the form below:

```

<OPTIONAL LONGER DESCRIPTION>

```

### Description

Thank you!

### DIFF
--- a/support/mender-inventory-intervals
+++ b/support/mender-inventory-intervals
@@ -12,7 +12,7 @@ to_kebab_case() {
     # Remove the word "interval"
     string=$(echo "$string" | sed 's/Interval//g')
     # Replace capital letters with hyphen-lowercase
-    string=$(echo "$string" | sed -E 's/\([A-Z]\)/-\1/g')
+    string=$(echo "$string" | sed 's/\([A-Z]\)/-\1/g')
     # Convert to lowercase
     string=$(echo "$string" | tr '[:upper:]' '[:lower:]')
     # Preppend the "interval" word
--- a/support/mender-inventory-intervals
+++ b/support/mender-inventory-intervals
@@ -1,0 +1,46 @@
+#!/bin/sh
+
+json_file="/etc/mender/mender.conf"
+
+if [ ! -f "$json_file" ]; then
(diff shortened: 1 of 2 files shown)


MuchoLucho:
This is a follow up on the PR #1611 that I closed in the past

support/mender-inventory-intervals
```diff
@@ -0,0 +1,46 @@
...
+to_kebab_case() {
+    string="$1"
+    # Remove the word "interval"
+    string=$(echo "$string" | sed 's/Interval//g')
+    # Replace capital letters with hyphen-lowercase
+    string=$(echo "$string" | sed -E 's/\([A-Z]\)/-\1/g')
+    # Convert to lowercase
+    string=$(echo "$string" | tr '[:upper:]' '[:lower:]')
+    # Preppend the "interval" word
+    string=$(echo "$string" | sed -E 's/^/interval/g')
```
kacf, line 19: Should there be a hyphen after `interval`?

kacf (commented):
I suggest changing to `Changelog: Title` so users can see that this has been added! 🙂

Otherwise looks good, just one question below.

support/mender-inventory-intervals
MuchoLucho: Yes, it should look something like this:

```
$ string=RetryPollIntervalSeconds
$ string=$(echo "$string" | sed 's/Interval//g')
$ echo $string

RetryPollSeconds

$ string=$(echo "$string" | sed 's/\([A-Z]\)/-\1/g')
$ echo $string

-Retry-Poll-Seconds

$ string=$(echo "$string" | tr '[:upper:]' '[:lower:]')
$ string=$(echo "$string" | sed -E 's/^/interval/g')
$ echo "$string"

interval-retry-poll-seconds
```

kacf: Ah, because there is dangling hyphen earlier. You're right, I missed that. Looks good then!

kacf (approved):

PR #1642, opened 2024-06-18
refac: Add --stop-after flag and make state machine for standalone. 

kacf:
```
Changelog: Add `--stop-after` flag which can be used with the `install`
standalone command to stop after certain states. Use `resume` to
continue. The latter also supports the `--stop-after` flag, so that
you can wait in further states.

Ticket: MEN-7115

```


src/mender-update/standalone/states.cpp
```diff
@@ -0,0 +1,507 @@
...
+	}
+
+	if (not rollback_support.value()) {
+		bool already_failed = ResultContains(ctx.result_and_error.result, Result::Failed);
+		UpdateResult(ctx.result_and_error, {Result::Failed | Result::NoRollback, error::NoError});
+		if (already_failed) {
+			poster.PostEvent(StateEvent::NothingToDo);
+		} else {
+			// If it hadn't failed already, it's because the user asked for the rollback
+			// explicitly. In this case bail out instead of continuing.
```
lluiscampos, line 406: I don't understand this comment. Are you referring to a user call `mender-update rollback` where the update module has no support for actual rollback?

tests/src/mender-update/cli/cli_test.cpp
```diff
@@ -2028,6 +2031,199 @@ artifact_name=test
...
+SupportsRollback
+ArtifactCommit
+Cleanup
+)"));
+
+	EXPECT_TRUE(VerifyProvides(tmpdir.Path(), R"(rootfs-image.version=test
+rootfs-image.checksum=f2ca1bb6c7e907d06dafe4687e579fce76b37e4e93b7605022da52e6ccc26fd2
+artifact_name=test
+)"));
+}
```
lluiscampos, line 2142: Good test

tests/src/mender-update/cli/cli_test.cpp
```diff
@@ -2364,7 +2558,6 @@ SupportsRollback
 ArtifactCommit_Enter_01
 ArtifactCommit
 ArtifactCommit_Leave_01
-ArtifactCommit_Error_01
```
lluiscampos, line 2367: Can you explain me this behavior change? This will be different from the daemon mode AFAICS

src/mender-update/cli/actions.cpp
```diff
@@ -150,83 +152,75 @@ error::Error ShowProvidesAction::Execute(context::MenderContext &main_context) {
...
-			<< "Installation failed, and rollback also failed. System may be in an inconsistent state."
-			<< endl;
-		break;
+	} else if (ResultContains(result.result, Result::Committed)) {
+		cout << "Committed." << endl;
+	} else {
+		stringstream ss;
+		ss << std::hex << static_cast<int>(result.result);
+		log::Error("Unexpected result value: 0x" + ss.str());
 	}
```
lluiscampos, line 217: TBH, this whole if/else if cascade was reading much better with the `switch` statement.

I couldn't spot anything wrong, though. It seems to be doing the same as before

src/mender-update/cli/cli.cpp
```diff
@@ -57,6 +57,18 @@ const conf::CliCommand cmd_daemon {
...
 };
 
+const conf::CliOption opt_stop_after {
+	conf::CliOption {
+		.long_option = "stop-after",
+		.description =
+			"Stop after the given state has completed. "
+			"Choices are `Download`, `ArtifactInstall` and `ArtifactCommit`. "
+			"You can later resume the installation by using the `resume` command. "
+			"Note that the client always stops after `ArtifactInstall` if the update module supports rollback.",
```
lluiscampos, line 67: and/or if the update module requested a reboot, right?

src/mender-update/standalone/standalone.cpp
```diff
@@ -201,618 +222,435 @@ error::Error RemoveStateData(database::KeyValueDatabase &db) {
...
-		auto data = StateDataFromPayloadHeaderView(header);
-		return DoEmptyPayloadArtifact(main_context, data, options);
-	}
+error::Error StateMachine::SetStopAfterState(const string &state) {
+	using tf = common::state_machine::TransitionFlag;
+	using se = StateEvent;
+	auto &s = state_machine_;
 
-	update_module::UpdateModule update_module(main_context, header.header.payload_type);
+	// Replace transition in state machine in order to exit at given point.
```
lluiscampos, line 429: This is nice.

src/mender-update/standalone/standalone.cpp
```diff
@@ -201,618 +222,435 @@ error::Error RemoveStateData(database::KeyValueDatabase &db) {
...
-	auto &in_progress = exp_in_progress.value();
+error::Error PrepareContext(Context &ctx) {
+	const auto &default_paths {ctx.main_context.GetConfig().paths};
+	ctx.script_runner.reset(new executor::ScriptRunner(
+		ctx.loop,
+		chrono::seconds {ctx.main_context.GetConfig().state_script_timeout_seconds},
+		chrono::seconds {ctx.main_context.GetConfig().state_script_retry_interval_seconds},
+		chrono::seconds {ctx.main_context.GetConfig().state_script_retry_timeout_seconds},
+		default_paths.GetArtScriptsPath(),
+		default_paths.GetRootfsScriptsPath()));
```
lluiscampos, line 458: Shouldn't these be user defined from the configuration?

src/mender-update/standalone/standalone.cpp
```diff
@@ -201,618 +222,435 @@ error::Error RemoveStateData(database::KeyValueDatabase &db) {
...
-	if (err != error::NoError) {
-		err = err.FollowedBy(
-			script_runner.RunScripts(executor::State::Download, executor::Action::Error));
-		err = err.FollowedBy(update_module.Cleanup());
-		err = err.FollowedBy(RemoveStateData(main_context.GetMenderStoreDB()));
-		return {Result::FailedNothingDone, err};
+	if (!in_progress) {
+		return {
+			Result::NoUpdateInProgress,
+			context::MakeError(context::NoUpdateInProgressError, "Cannot resume")};
```
lluiscampos, line 537: Suggestion: elaborate the user message a bit more - "Cannot resume. No update in progress found."

src/mender-update/standalone/standalone.cpp
```diff
@@ -201,618 +222,435 @@ error::Error RemoveStateData(database::KeyValueDatabase &db) {
...
-	if (result.result == Result::Committed) {
-		if (reboot.value() != update_module::RebootAction::No) {
-			result.result = Result::InstalledAndCommittedRebootRequired;
-		} else {
-			result.result = Result::InstalledAndCommitted;
-		}
+	if (!in_progress) {
+		return {
+			Result::NoUpdateInProgress,
+			context::MakeError(context::NoUpdateInProgressError, "Cannot commit")};
```
lluiscampos, line 578: similarly here "Cannot commit. No update in progress found"

src/mender-update/standalone/standalone.cpp
```diff
@@ -201,618 +222,435 @@ error::Error RemoveStateData(database::KeyValueDatabase &db) {
...
-
-	// Failure Enter
-	err = script_runner.RunScripts(
-		executor::State::ArtifactFailure, executor::Action::Enter, executor::OnError::Ignore);
-	if (err != error::NoError) {
-		log::Error("Failure during execution of ArtifactFailure Enter script: " + err.String());
-		result.result = Result::FailedAndRollbackFailed;
-		result.err = result.err.FollowedBy(err);
+			Result::NoUpdateInProgress,
+			context::MakeError(context::NoUpdateInProgressError, "Cannot roll back")};
```
lluiscampos, line 623: And here

lluiscampos (commented):
I got pretty exhausted but I manage to _read_ it. What a beast! Although very long, the code is expressive so I am not against merging it.

My fear is that having two different states machines can also be difficult to maintain in the long run, as the flow may diverge or we implement features in daemon and forget here. But yeah, this feature actually calls for an state machine, so I don't have a better idea TBH.

As discussed privately, the `Result` enum class feels a bit messy, as IMO we are loosing the benefits of enum class when we cast it into other types. I'll think about this overnight :thinking: 

See comments below. One thing I suspect is wrong (see `RebootAndRollbackQueryState`) while the rest are mainly small things to discuss.

kacf:
> My fear is that having two different states machines can also be difficult to maintain in the long run, as the flow may diverge or we implement features in daemon and forget here. But yeah, this feature actually calls for an state machine, so I don't have a better idea TBH.

I know, but I don't really have a better idea either. I've made some attempts at unifying the two, both early on in the implementation, and by looking at it now, but it quickly gets messy, because there are some pretty significant differences between the two state machines. The most prominent difference is that we [don't execute the same states](https://docs.mender.io/3.7/artifact-creation/state-scripts#standalone-mode). There are also several other differences, like deployment status related states, transitioning in and out of idle and sync states, as well as state loop states (which have limited usefulness in standalone mode). The data we need to save between each state is also different, and we can't change it because of compatibility with the earlier versions, which complicates going into those states. It certainly wouldn't be impossible to merge the two state machines, and a part of me wants to, but I'm not convinced it would make it easier to maintain, because there would be a lot of conditionals littered everywhere.

src/mender-update/standalone/states.cpp
```diff
@@ -0,0 +1,507 @@
...
+
+	if (err != error::NoError) {
+		return expected::unexpected(err);
+	}
+
+	if (inner_err != error::NoError) {
+		return expected::unexpected(inner_err);
+	}
+
+	return make_shared<events::io::ReaderFromAsyncReader>(loop, reader);
```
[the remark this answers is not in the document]
kacf: Yes, but all error returns from the callback set `inner_err`, which we check before we get here, so it should not happen, or?

I added an `AssertOrReturnUnexpected`, just to be safe.

src/mender-update/standalone/states.cpp
```diff
@@ -0,0 +1,507 @@
...
+			}
+
+			auto exp_reader = resp->MakeBodyAsyncReader();
+			if (!exp_reader) {
+				inner_err = exp_reader.error();
+				return;
+			}
+			reader = exp_reader.value();
+		},
+		[](http::ExpectedIncomingResponsePtr exp_resp) {
```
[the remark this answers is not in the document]
kacf: It will, actually, but long after we have returned. The reader is a shared pointer and is returned from the function, so the body handler will be run once it resumes the loop and reads through the whole body. So it's important not to lambda-capture anything in this handler, since we are not in the same stack frame anymore.

I'll add a comment, since it's indeed quite unintuitive.

src/mender-update/standalone/states.hpp
```diff
@@ -0,0 +1,127 @@
...
+namespace update {
+namespace standalone {
+
+using namespace std;
+
+namespace sm = mender::common::state_machine;
+
+using StateType = sm::State<Context, StateEvent>;
+
+class StateDataSaveState : virtual public StateType {
```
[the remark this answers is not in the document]
kacf: I thought that maybe it was confusing to have the same name as the daemon variant, but OTOH many other things have the same names, so maybe it's actually better to stick to that. I'll change it.

src/mender-update/standalone/states.cpp
```diff
@@ -0,0 +1,507 @@
...
+		poster.PostEvent(StateEvent::Failure);
+		return;
+	}
+
+	if (rollback_support.value()) {
+		poster.PostEvent(StateEvent::NeedsInteraction);
+		return;
+	}
+
+	cout << "Update Module doesn't support rollback. Committing immediately." << endl;
```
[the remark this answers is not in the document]
kacf: It does, actually. This is a big difference between standalone mode and daemon mode. Standalone mode doesn't deal with rebooting at all, so it just commits immediately if there is no chance of rollback. In hindsight, maybe this was a bad decision, since `ArtifactCommit_Leave` script may actually depend on running a new update. But the Golang client also does this, so we won't change this now. And it only affects update modules without rollback, so rootfs-image updates are not affected.

src/mender-update/standalone/states.cpp
```diff
@@ -0,0 +1,507 @@
...
+	auto reboot = ctx.update_module->NeedsReboot();
+	if (!reboot) {
+		log::Error("Could not query for reboot: " + reboot.error().String());
+		UpdateResult(ctx.result_and_error, {Result::Failed, reboot.error()});
+		poster.PostEvent(StateEvent::Failure);
+		return;
+	}
+
+	if (reboot.value() != update_module::RebootAction::No) {
+		UpdateResult(ctx.result_and_error, {Result::RebootRequired, error::NoError});
```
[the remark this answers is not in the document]
kacf: This is actually correct, at least in terms of following Golang client behavior, see [previous comment](https://github.com/mendersoftware/mender/pull/1642#discussion_r1706745507).

src/mender-update/standalone/states.cpp
kacf: Exactly!

tests/src/mender-update/cli/cli_test.cpp
kacf: This was a bug in the standalone mode of the C++ client. Daemon mode [already does this](https://github.com/mendersoftware/mender/blob/4.0.4/src/mender-update/daemon/state_machine/state_machine.cpp#L199). Our docs also [confirm](https://docs.mender.io/3.7/artifact-creation/state-scripts#transitions-and-ordering) this behavior.

src/mender-update/cli/cli.cpp
kacf: No, see [earlier comment](https://github.com/mendersoftware/mender/pull/1642#discussion_r1706745507).

src/mender-update/standalone/standalone.cpp
kacf: Agree! This is one of the places where having a state machine really shines.

src/mender-update/standalone/standalone.cpp
kacf: Indeed, good catch!

src/mender-update/standalone/standalone.cpp
kacf: It is implicitly elaborated already: The error code will provide a string message similar to what you wrote, once it's printed.

src/mender-update/standalone/standalone.cpp
kacf: Same.

src/mender-update/standalone/standalone.cpp
kacf: Same.

src/mender-update/standalone/standalone.cpp
kacf: Lol! I actually proceeded to fix it, but then I realized that `default_paths` comes from the conf, it's just confusingly named. Changing.

src/mender-update/cli/actions.cpp
kacf: Using a switch statement seems infeasible with the new possibilities for starting and stopping, because there are so many combinations, and adding new results need to take existing results into account, hence the bit flags. However, I do believe I have considerably improved the readability now, using a combination of what we talked about yesterday (split some things which were previously sentences into multiple lines), and some new bit flags. Check it out!

kacf:
@lluiscampos: Thanks for the review! I know this was a difficult one, and sorry for trying to rush it back in June.

I have added all fixes and improvements as new commits for easy reviewing, but I will squash them before merging.

kacf:
It's also worth pointing out that there are still some unknowns here with regards to how this will get used in the orchestrator. I'm going to try to adapt the interface there and test a little before I'm calling this settled.

src/mender-update/standalone/states.cpp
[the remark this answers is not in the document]
lluiscampos: Thanks for clarifying this :memo:

src/mender-update/standalone/standalone.cpp
lluiscampos: I am convinced at this point :laughing:

lluiscampos (approved):
Looks good now! Thank you

src/mender-update/cli/actions.cpp
```diff
@@ -161,68 +161,89 @@ static error::Error ResultHandler(standalone::ResultAndError result) {
...
-		log::Error("Unexpected result value: 0x" + ss.str());
+		str += content;
+	};
+
+	string operation_done;
+	string operation_failure;
+
+	// For failure case, include which attempted operation failed.
+	if (contains(r::DownloadFailed)) {
+		operation_failure = "Streaming failed.";
```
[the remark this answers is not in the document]
kacf: True, we don't make a distinction, currently. The question is whether we *want* to make a distinction. Back when we discussed what to call the command (before the `--stop-after` flag was proposed, when `stream` would be its own command), we chose the word "stream" to avoid the misconception that the artifact is being stored on the device first, and then installed. Even if it is temporarily stored in the Update Module File Tree, it is all part of the same overall operation. This is different from apt, where you can say `apt download emacs`, and then install it completely independently afterwards using `dkpg -i emacs.deb`. So it was an attempt at making it clearer that streaming and installing are strongly connected operations, and even if it's temporarily stored, it doesn't really change that.

That being said, I don't really have very strong opinion, but I think I'm leaning towards keeping it as is. Does it make sense?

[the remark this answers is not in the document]
lluiscampos: You convinced me. Let's keep _stream_

kacf:
@lluiscampos: Good thing I tested, it turns out I did need a bit more from the `--stop-after` flag to complete the rootfs-image integration with the orchestrator. The commits starting with "chore: Remove unused function" are new.

src/mender-update/standalone/states.cpp
```diff
@@ -0,0 +1,555 @@
...
+
+			// If there is state data, resave it with `failed` set to false. The rationale
+			// behind this is that if we have already recorded failure for this run, it will be
+			// returned in the error code. That does not mean that we should record error for
+			// the next run, which is independent. An example is rollback, if we are somewhere
+			// in the rollback flow, we are likely to have a failure here, because the *install*
+			// failed. But when we now exit, and then later resume the rollback, the rollback
+			// should return success, not failure.
+			ctx.state_data.failed = false;
+			return SaveStateData(txn, ctx.state_data);
```
lluiscampos, line 542: We could guard this to spare some writing
```suggestion
			if (ctx.state_data.failed) {
				ctx.state_data.failed = false;
				return SaveStateData(txn, ctx.state_data);
			}
```

lluiscampos (approved):
It took many brain cycles... but it makes sense :+1: 

Two suggestions below

src/mender-update/cli/cli.cpp
```diff
@@ -275,24 +283,78 @@ ExpectedActionPtr ParseUpdateArguments(
...
+				if (value.value == "") {
+					return expected::unexpected(conf::MakeError(
+						conf::InvalidOptionsError, "--stop-after needs an argument"));
+				}
+				stop_after.push_back(value.value);
+				continue;
+			} else if (value.option != "") {
+				return expected::unexpected(
+					conf::MakeError(conf::InvalidOptionsError, "No such option: " + value.option));
+			}
```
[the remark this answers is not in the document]
kacf: I think that's a good idea. I'm even about to repeat some of this in the orchestrator. :-P

PR #1645, opened 2024-07-08
fix: Make crypto load and use the openssl.cnf file

danielskinstad:
Libcrypto does not load the openssl.cnf file automatically. This removes the need for manually having to export e.g. `PKCS11_MODULE_PATH` as an env var when loading a pkcs11 interface (if it's already specified in openssl.cnf).

### DIFF
--- a/src/common/crypto/platform/openssl/crypto.cpp
+++ b/src/common/crypto/platform/openssl/crypto.cpp
@@ -32,6 +32,7 @@
 #endif // MENDER_CRYPTO_OPENSSL_LEGACY
 
 #include <openssl/evp.h>
+#include <openssl/conf.h>
 #include <openssl/pem.h>
 #include <openssl/rsa.h>
 
@@ -281,6 +282,11 @@ ExpectedPrivateKey LoadFrom(const Args &args) {
 #endif // ndef MENDER_CRYPTO_OPENSSL_LEGACY
 
 ExpectedPrivateKey PrivateKey::Load(const Args &args) {
+	// Load OpenSSL config
+	if ((CONF_modules_load_file(nullptr, nullptr, 0) != OPENSSL_SUCCESS)) {
+		log::Warning("Failed to load OpenSSL configuration file: " + GetOpenSSLErrorMessage());
+	}
+
 	log::Trace("Loading private key");
 	if (args.ssl_engine != "") {
 		return LoadFromHSMEngine(args);


src/common/crypto/platform/openssl/crypto.cpp
```diff
@@ -281,6 +282,11 @@ ExpectedPrivateKey LoadFrom(const Args &args) {
 #endif // ndef MENDER_CRYPTO_OPENSSL_LEGACY
 
 ExpectedPrivateKey PrivateKey::Load(const Args &args) {
+	// Load OpenSSL config
+	if ((CONF_modules_load_file(NULL, NULL, 0) != 1)) {
+		log::Debug("Failed to load OpenSSL configuration file: " + GetOpenSSLErrorMessage());
+	}
+
```
merlin-northern, line 289: amazing find. thanks a lot. I only have some doubts about the place, Load method of the PrivateKey class does not seem a good place to load the config file.

danielskinstad: I agree, but I didn't find an obvious place to load it other than here, I can look a bit harder though :)

src/common/crypto/platform/openssl/crypto.cpp
```diff
@@ -281,6 +282,11 @@ ExpectedPrivateKey LoadFrom(const Args &args) {
 #endif // ndef MENDER_CRYPTO_OPENSSL_LEGACY
 
 ExpectedPrivateKey PrivateKey::Load(const Args &args) {
+	// Load OpenSSL config
+	if ((CONF_modules_load_file(NULL, NULL, 0) != 1)) {
```
lluiscampos, line 286: Use `nullptr` for NULL
```suggestion
	if ((CONF_modules_load_file(nullptr, nullptr, 0) != 1)) {
```

lluiscampos, line 286: Use `OPENSSL_SUCCESS` instead of hard-coded 1
```suggestion
	if ((CONF_modules_load_file(NULL, NULL, 0) != OPENSSL_SUCCESS)) {
```

src/common/crypto/platform/openssl/crypto.cpp
```diff
@@ -281,6 +282,11 @@ ExpectedPrivateKey LoadFrom(const Args &args) {
...
+		log::Debug("Failed to load OpenSSL configuration file: " + GetOpenSSLErrorMessage());
```
lluiscampos, line 287: Shouldn't it be a warning? `Debug` messages won't be printed by default, and this message is very useful is the client later fails to authenticate for example. Or is it the conf file is not present in most of the environments?

src/common/crypto/platform/openssl/crypto.cpp
lluiscampos: Given the current design, I think `Load` is the best place, actually.

It is the entry point for creating a `PrivateKey` object from existing keys, and it is used from for the rest of the methods dealing with private keys like like `Sign` and `ExtractPublicKey`.

lluiscampos (approved):
Great find! Two nitpicks below, see comments.

As soon as we merge this one please cherry-pick to 4.0.x and ping me to prepare the instructions for testing it with yocto.

src/common/crypto/platform/openssl/crypto.cpp
danielskinstad: It's present, so I agree that warning is better.

danielskinstad:

danielskinstad:

- 4.0.x

danielskinstad:
* 4.0.x

PR #1647, opened 2024-07-18
Proxy basic authentication support

vpodzime:


vpodzime:

src/common/http/http.cpp
```diff
@@ -115,11 +115,24 @@ error::Error BreakDownUrl(const string &url, BrokenDownUrl &address) {
 		address.path = tmp.substr(split_index);
 	}
 
-	if (address.host.find("@") != string::npos) {
-		address = {};
-		return error::Error(
-			make_error_condition(errc::not_supported),
-			"URL Username and password is not supported");
+	auto auth_index = address.host.find("@");
```
alfrunes, line 118: The Golang URL parser allows the username/password to contain `@` and instead uses the [last occurrence](https://cs.opensource.google/go/go/+/refs/tags/go1.22.5:src/net/url/url.go;l=587) of the character as the authority/host separator.
I know that `wget` and `curl` does not allow it, and will confuse the `host:port` section if you include additional `@` characters. In both cases it also works to URL escape `@` using `%40`.

src/common/http/http.cpp
```diff
@@ -115,11 +115,24 @@ error::Error BreakDownUrl(const string &url, BrokenDownUrl &address) {
...
+		auto user_password = address.host.substr(0, auth_index);
+		address.host = address.host.substr(auth_index + 1);
+		auto u_pw_sep_index = user_password.find(":");
+		if (u_pw_sep_index == string::npos) {
+			// no password
+			address.username = std::move(user_password);
+		} else {
+			address.username = user_password.substr(0, u_pw_sep_index);
+			address.password = user_password.substr(u_pw_sep_index + 1);
+		}
```
alfrunes, line 135: Should we unescape the `%` escape sequences for the username/password?

vpodzime: As discussed in the SU, this is tricky and goes against the special handling of `@` in the username/password. Either we treat it as encoded and `@` should never show up there or it is not encoded and we should not treat things like `%20` in any special way. But let's try to be as much backwards compatible as possible with the Go client and do both.

src/common/http/http.cpp
```diff
@@ -188,6 +188,23 @@ string URLEncode(const string &value) {
...
+string URLDecode(const string &value) {
+	stringstream unescaped;
+
+	auto len = value.length();
+	for (size_t i = 0; i < len; i++) {
+		if (value[i] != '%') {
+			unescaped << value[i];
+		} else {
+			unsigned int num;
+			sscanf(value.substr(i + 1, 2).c_str(), "%x", &num);
```
alfrunes, line 200: Should we add a boundary check here (if `%` is one of the last 2 characters)?

vpodzime: Hmm, actually, the function should return `ExpectedString` and error in this case (and others).

alfrunes (approved):

vpodzime:
@mender-test-bot pipeline --fast

oldgiova:
The mender-qa pipeline is broken for a server issue, not introduced by this PR. As per conversation in [this thread](https://northern-tech.slack.com/archives/C2D3HRU9M/p1721917406809819), I'm going to force-merge this PR as it has already been approved.

lluiscampos:
* 4.0.x

PR #1657, opened 2024-08-26
MEN-7159: Progress reader updates the output only when progressing

lluiscampos:
Due to a bug in the logic, the progress was being updated after every block, even if the percentage number was not changing. This made the reading/writing of Artifacts slow and noisy.

### DIFF
--- a/src/mender-update/progress_reader/progress_reader.cpp
+++ b/src/mender-update/progress_reader/progress_reader.cpp
@@ -28,8 +28,9 @@ expected::ExpectedSize Reader::Read(
 	if (exp_read) {
 		bytes_read_ += exp_read.value();
 		int percentage = (bytes_read_ / static_cast<float>(tot_size_)) * 100;
-		if (percentage > last_percentage) {
+		if (percentage > last_percentage_) {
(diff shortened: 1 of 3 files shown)


lluiscampos:

src/mender-update/progress_reader/progress_reader.cpp
```diff
@@ -30,6 +30,7 @@ expected::ExpectedSize Reader::Read(
 		int percentage = (bytes_read_ / static_cast<float>(tot_size_)) * 100;
 		if (percentage > last_percentage) {
 			cerr << "\r" << percentage << "%";
+			last_percentage = percentage;
```
kacf, line 33: Minor & preexisting: `last_percentage` should be called `last_percentage_`.

kacf (approved):
Lol! I guess it still *looked* correct to the human eye!

src/mender-update/progress_reader/progress_reader.cpp
lluiscampos: I am fixing but I won't run a new integration pipeline for it.

lluiscampos:
* 4.0.x

PR #1659, opened 2024-08-27
chore: Don't duplicate `DefaultPaths` in every source file.

kacf:
This simple change saves more than 12K in the final binary.

For future reference, using `readelf -s` to look for duplicate symbols is a useful way to hunt for such space waste. I haven't looked deeply at it, but starting by looking in object files for duplicated `LOCAL` symbols without `WEAK` is a good beginning. I just found this one randomly though, and I cannot dive deeper into this rabbit hole now.


lluiscampos (approved):
Thank you for the tip! :rocket:

PR #1663, opened 2024-09-09
fix: add optional dynamic linking to boost::log

aduskett:
When building boost as a dynamic library, hundreds of `undefined reference to `boost::log` errors occur during linking. As the default behavior currently is to link against boost::log statically, it is best to add a new option that explicitly enables dynamic linking to boost::log. This new option is called MENDER_USE_DYNAMIC_BOOST_LOG.

Changelog: None
Ticket: None

### DIFF
--- a/src/common/cmake/boost.cmake
+++ b/src/common/cmake/boost.cmake
@@ -3,6 +3,10 @@
 find_package(Boost 1.74 COMPONENTS log)
 
 option(MENDER_DOWNLOAD_BOOST "Download Boost if it is not found (Default: OFF)" OFF)
+option(MENDER_USE_DYNAMIC_BOOST_LOG "Link dynamically to boost::log  (Default: OFF)" OFF)
+if(MENDER_USE_DYNAMIC_BOOST_LOG)
+  ADD_DEFINITIONS(-DBOOST_LOG_DYN_LINK)
+endif()
 
 if(NOT MENDER_DOWNLOAD_BOOST AND NOT ${Boost_FOUND})
   message(FATAL_ERROR


aduskett:
I would like to see what failed, but the pipeline URL goes to a 404 :(

lluiscampos:

lluiscampos:
> I would like to see what failed,

You copied the example start, pointing the bot to old or non-existing PR for mender and mender-connect. I started it for you.

> but the pipeline URL goes to a 404 :(

Yeah, this pipeline is internal to the Mender team, sorry.

aduskett:
> > I would like to see what failed,
> 
> You copied the example start, pointing the bot to old or non-existing PR for mender and mender-connect. I started it for you.
> 
> > but the pipeline URL goes to a 404 :(
> 
> Yeah, this pipeline is internal to the Mender team, sorry.

Thank you so much!

danielskinstad:
Hey :smiley:  Thank you for your contribution!

I have some comments/questions.

 As far as I know, and can see from the pre-existing code, it defaults to _dynamically_ linking boost, not statically (unless you set `MENDER_DOWNLOAD_BOOST`).

- ldd of mender-auth after a clean build: `libboost_log.so.1.74.0 => /lib/x86_64-linux-gnu/libboost_log.so.1.74.0`

Secondly, I don't get any `undefined reference` errors when linking boost, and I don't see it any pipelines either. Might it be your setup? Does the same happen with a completely, unmodified, fresh build of mender (and following the build guide in the README)?

kacf:
Indeed, Boost is already linked dynamically. The proposed fix [was removed earlier](https://github.com/mendersoftware/mender/commit/04d54e69fdae513520d053480c8bf5b7fa16e1fc) for this reason. It must be something in the environment.

aduskett:
What this looks like is Buildroot not setting BOOST_LOG_DYN_LINK or BOOST_ALL_DYN_LINKS as found in boosts config.hpp:238

```
#   if defined(BOOST_LOG_DYN_LINK) || defined(BOOST_ALL_DYN_LINKS)
#       define BOOST_LOG_DLL
#   endif
```

Which causes the above issue. Both Fedora and Debian do indeed define BOOST_LOG_DYN_LINK which is why the autobuilders aren't having the problem. Now, as to how to set BOOST_LOG_DYN_LINK is the issue.

kacf:
@aduskett: Can you elaborate on how Debian and Fedora define this? I cannot find this on my system, nor on a fresh one, in fact, it's quite clearly not turned on:

```
/usr/include/boost/config/user.hpp:90:// #define BOOST_ALL_DYN_LINK
```

(`BOOST_LOG_DYN_LINK` doesn't have a define anywhere, as far as I can see)

There are also some other issues with the proposal:

* The PR doesn't appear to have any effect on my system when turning it off. This is probably because the absence of definition is having the "default" effect, which appears to be to link dynamically anyway, on my system. But this means the toggle does not work correctly.
* `add_definitions` has been superseded by other variants, see [here](https://cmake.org/cmake/help/latest/command/add_definitions.html#add-definitions), and should anyway be lowercase to match the rest of the code.

I'm wondering if it isn't better to just do this in autobuild with direct flags instead of having an option (which has some pitfalls, as we have seen). Something like this, for example:

```
cmake -D CMAKE_CXX_FLAGS=-DBOOST_LOG_DYN_LINK -D CMAKE_EXE_LINKER_FLAGS=-DBOOST_LOG_DYN_LINK .
```

Would that work for autobuild?

aduskett:
Hey! So I * did * figure out the problem with buildroot, which is from commit `4349746dac6f55b0b0faca21a7e609fb67d1471b`

```
Author: Fabrice Fontaine <fontaine.fabrice@gmail.com>
Date:   Thu Apr 18 11:13:50 2019 +0200

    package/boost: don't install boost cmake files
    
    Since version 1.70.0, boost installs its own cmake files which are
    utterly broken so don't install them otherwise all cmake packages using
    boost (host-thrift, domoticz, i2pd ...) will fail because
    BOOST_INCLUDE_DIRS will be empty
    
    Fixes:
     - http://autobuild.buildroot.org/results/4ada26bab5c1e6504c7d8c92672326ced1d336df
    
    Signed-off-by: Fabrice Fontaine <fontaine.fabrice@gmail.com>
    Signed-off-by: Thomas Petazzoni <thomas.petazzoni@bootlin.com>
```

Five years ago! The fix is to remove `--no-cmake-config` from boost.mk in buildroot and let boost configure the .cmake file before installing them. Sorry about the junk merge request!

kacf:
No worries, great that you figured it out!

PR #1667, opened 2024-09-25
Fixes for `--stop-before` flag.

kacf:


kacf:
Must be merged together with https://github.com/mendersoftware/meta-mender/pull/2171.

tests/src/mender-update/cli/cli_test.cpp
```diff
@@ -2174,8 +2174,8 @@ exit 0
 			"--datastore",
 			tmpdir.Path(),
 			"install",
-			"--stop-after",
-			"ArtifactCommit",
+			"--stop-before",
+			"ArtifactCommit_Leave",
```
lluiscampos, line 2288: What is the difference between stopping before `ArtifactCommit_Leave` or `Cleanup`? For the successful case, there is no real difference. And yes, for failure it will behave different (it would stop or not), but maybe we could merge these two use cases and stop only before `Cleanup`?

From the API point of view, to allow stopping before a _Leave_ state sounds strange. Note that every other allowed state to stop before are _Enter_ states (or Cleanup)

lluiscampos (commented):
See my comment wrt `ArtifactCommit_Leave` below.

Also, I think you got confused by testing this PR with https://github.com/mendersoftware/meta-mender/pull/2171 - what we need to fix for this PR is `mender-image-tests` use of this new flag.

tests/src/mender-update/cli/cli_test.cpp
```diff
@@ -1425,7 +1492,21 @@ TEST(CliTest, InstallAndThenFailRollBack) {
...
+        case "$1" in
+    NeedsArtifactReboot|SupportsRollback)
+        :
+        ;;
+    *)
+        echo "$1" >> $TEST_DIR/call.log
+        ;;
+esac
+        ;;
+esac
```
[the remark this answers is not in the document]
kacf: Oops, thanks!

tests/src/mender-update/cli/cli_test.cpp
```diff
@@ -3210,7 +3312,14 @@ TEST_P(StandaloneStateScriptTest, AllScriptSuccess) {
...
-echo "$1" >> $TEST_DIR/call.log
+case "$1" in
+    NeedsArtifactReboot|SupportsRollback)
+        :
+        ;;
+    *)
+        echo "$1" >> $TEST_DIR/call.log
+        ;;
+esac
 )";
```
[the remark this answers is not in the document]
kacf: Indeed. This also made me realize that I don't have tests for the last commit. Will fix.

tests/src/mender-update/cli/cli_test.cpp
kacf: There is a big difference actually, but I also didn't realize it until I started looking at this in great detail. `ArtifactCommit_Leave` state scripts are only executed when the commit has definitely happened, and will not roll back (this is documented [here](https://docs.mender.io/3.7/artifact-creation/state-scripts#power-loss)). In other words, when executing `ArtifactCommit_Leave` scripts, we need to be sure that we will not roll back afterwards, hence we need to stop before them, and not before `Cleanup`, at which point it would be too late already. Rolling back after a successful commit, but before the Leave scripts, is possible with the orchestrator because *other* payloads may fail to commit in between.

The state machine is indeed quite complicated in this respect, and I can't say I particularly like this complexity. But it underlines how hard it is to both execute all states correctly, including on powerloss, while simultaneously offering a command line interface to fine control those states. It's not easy! 😃

kacf:
> Also, I think you got confused by testing this PR with [mendersoftware/meta-mender#2171](https://github.com/mendersoftware/meta-mender/pull/2171) - what we need to fix for this PR is `mender-image-tests` use of this new flag.

Well, kind of. I did actually want to test this with that PR just to make sure I hadn't broken anything unintentionally. But you're right it's not directly related.

In any case, it's the orchestrator which needs the symlink change, in its integration tests. I have tested it locally, but the PR for that comes a little bit later.

kacf:
Review comments addressed, and tests added for the last commit!

tests/src/mender-update/cli/cli_test.cpp
lluiscampos: Okey. Got it :memo:

lluiscampos (approved):
It looks good - to the best of my knowledge. Let's see how easy is for me next time I have to come around this code in maintenance mode :nerd_face:

kacf:
Thanks!

PR #1669, opened 2024-10-01
feat: change generated key from RSA to ED25519

danielskinstad:
Ticket: MEN-7534
Changelog: Title


src/common/crypto/platform/openssl/crypto.cpp
```diff
@@ -294,10 +294,9 @@ ExpectedPrivateKey PrivateKey::Load(const Args &args) {
 	return LoadFrom(args);
 }
 
-ExpectedPrivateKey PrivateKey::Generate(const unsigned int bits, const unsigned int exponent) {
-#ifdef MENDER_CRYPTO_OPENSSL_LEGACY
+ExpectedPrivateKey PrivateKey::Generate() {
 	auto pkey_gen_ctx = unique_ptr<EVP_PKEY_CTX, void (*)(EVP_PKEY_CTX *)>(
-		EVP_PKEY_CTX_new_id(EVP_PKEY_RSA, nullptr), pkey_ctx_free_func);
+		EVP_PKEY_CTX_new_id(EVP_PKEY_ED25519, nullptr), pkey_ctx_free_func);
```
lluiscampos, line 299: Previously, we used `EVP_PKEY_CTX_new_id` for OpenSSL legacy, and `EVP_PKEY_CTX_new_from_name` for new versions. From what I read in the man page we should prefer `EVP_PKEY_CTX_new_from_name` - Is there a reason for this API call to not work for ED25519?

src/common/crypto/platform/openssl/crypto.cpp
```diff
@@ -306,75 +305,18 @@ ExpectedPrivateKey PrivateKey::Generate(const unsigned int bits, const unsigned
...
-		return expected::unexpected(MakeError(
-			SetupError,
-			"Failed to generate a private key. Parameters setting failed: "
-				+ GetOpenSSLErrorMessage()));
-	}
-	exponent_bn.release();
-
 	EVP_PKEY *pkey = nullptr;
+#ifdef MENDER_CRYPTO_OPENSSL_LEGACY
+	// https://docs.openssl.org/master/man3/EVP_PKEY_keygen
```
lluiscampos, line 310: What exactly are we referring to from the manual?

src/common/crypto/platform/openssl/crypto.cpp
```diff
@@ -825,11 +767,8 @@ error::Error PrivateKey::SaveToPEM(const string &private_key_path) {
 				+ "): " + GetOpenSSLErrorMessage());
 	}
 
-	// PEM_write_bio_PrivateKey_traditional will use the key-specific PKCS1
-	// format if one is available for that key type, otherwise it will encode
-	// to a PKCS8 key.
```
lluiscampos, line 830: I don't remember the details but this had a reason to be.

Are we removing it now because ED25519 does not have a PKCS1 (tradictional) formatting anyway?

danielskinstad: ED25519 doesn't support the traditional format, so we can't use `PEM_write_bio_PrivateKey_traditional()`

src/common/crypto/platform/openssl/crypto.cpp
danielskinstad: If I remember correctly it was that keygen does the same thing as generate, I guess I should elaborate further in the comment

src/common/crypto/platform/openssl/crypto.cpp
lluiscampos: > I don't see anything in the man page that says that we should prefer `EVP_PKEY_CTX_new_from_name()`, (...)

There is no explicit mention (you are right), it is rather my interpretation,

Based on 1) that the `id` parameter itself are referred as legacy (from the manual: "Legacy NID or EVP_PKEY type (...) These  are  EVP_PKEY_RSA, EVP_PKEY_RSA_PSS, .(.. ) and  are used by legacy methods.) and 2) that we had already int he code the distinction when generating RSA keys, I made a case for that we should "prefer" it.

Also, this is not really related to the change that we want to introduce (replace the key pair type).

TBH I don't have an strong opinion. I expressed here my view and do as you think is best. I trust you on this.

src/common/crypto/platform/openssl/crypto.cpp
lluiscampos: Ok, It makes sense to change. I verified that in production code, the `Save` is only called after a `Generate`, so we know what are we saving.

src/common/crypto/platform/openssl/crypto.cpp
danielskinstad: Ah, didn't know about the `id` being considered legacy. I'll go with `EVP_PKEY_CTX_new_id` for legacy and `EVP_PKEY_CTX_new_from_name` for newer then :+1:

lluiscampos (approved):
The code looks good and the changes tidy and well justified. Good work!

Again, the changelog entry could be more user friendly. Think about: when is this used (only when not provided in the configuration file), does it have consequences for an existing integration (no, the existing key won't be re-generated - it only affects installation in new devices), what is the motivation (more efficient computation)

danielskinstad:

PR #1676, opened 2024-10-16
MEN-7613: Size fixes

kacf:


src/common/json.hpp
```diff
@@ -185,7 +214,20 @@ enum class MissingOk {
...
-	const json::Json &json, const string &key, MissingOk missing_ok);
+	const json::Json &json, const string &key, MissingOk missing_ok) {
+	auto exp_value = json.Get(key);
+	if (!exp_value) {
+		if (missing_ok == MissingOk::Yes
+			&& exp_value.error().code != json::MakeError(json::KeyError, "").code) {
+			return T();
+		} else {
+			auto err = exp_value.error();
+			err.message += ": Could not get `" + key + "` from state data";
```
lluiscampos, line 225: From _json_ data?

src/common/http.hpp
```diff
@@ -125,7 +125,7 @@ string MethodToString(Method method);
 struct BrokenDownUrl {
 	string protocol;
 	string host;
-	int port {-1};
+	uint16_t port;
```
lluiscampos, line 128: If what you are concern here is the sign, isn't better to use `unsigned int` and let the compiler chose the most efficient size?

src/common/http/http.cpp
```diff
@@ -146,7 +146,13 @@ error::Error BreakDownUrl(const string &url, BrokenDownUrl &address, bool with_a
...
 			return error::Error(port.error().code, url + " contains invalid port number");
 		}
-		address.port = port.value();
+		if (port.value() < 0 or port.value() > numeric_limits<decltype(address.port)>::max()) {
+			address = {};
+			return error::Error(
+				make_error_condition(errc::invalid_argument),
+				url + " contains invalid port number");
+		}
+		address.port = static_cast<decltype(address.port)>(port.value());
```
lluiscampos, line 155: Nice. We haven't used `decltype` but we should have.

src/common/json.hpp
```diff
@@ -124,19 +124,43 @@ class Json {
...
-	ExpectedInt64 GetInt() const;
+	ExpectedInt64 GetInt64() const;
 	ExpectedDouble GetDouble() const;
 	ExpectedBool GetBool() const;
 
+	// Defined in cpp file as specialized templates.
 	template <typename T>
-	expected::expected<T, error::Error> Get() const;
+	typename enable_if<
+		not is_integral<T>::value or is_same<T, int64_t>::value,
```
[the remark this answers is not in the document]
kacf: Hehe, yes. But to be fair I never said we won't use them, I said we won't *make* them. But that's exactly what I'm doing here, so I guess I'm still guilty. 😄

But laughing aside, I believe it has really improved the code, especially in this module. It's much less code when the type is also a parameter and can be passed around. It has come a long way since what I knew from before this project (which was pre-C++11).

src/common/http.hpp
kacf: No, it's not the signedness. The range of network port numbers fit exactly into the `uint16_t` type, which means we can use the json getter with range checking that I made above, and get the range checking for free. That would not work with a bigger type.

It's also the fact that when passing it to Boost it also uses `uint16_t`, so we don't need a cast there.

src/common/http/http.cpp
kacf: Yes, it's nice.

But now I see this contradicts my answer in the previous comment. Do I need a cast here now...? Let me recheck.

kacf: Following my other comment above, this is no longer a cast, but I still use `decltype` in the `StringTo` call, which then does range checking based on that type. So I removed the explicit range checking from here, and just rely on the return value.

kacf:
@lluiscampos: I got an approval already, but maybe you want to recheck just the area surrounding the two last threads.

src/common/http.hpp
lluiscampos: Thanks for the clarification.

lluiscampos (approved):
> @lluiscampos: I got an approval already, but maybe you want to recheck just the area surrounding the two last threads.

@kacf It looks good!

PR #1677, opened 2024-10-16
build: Enable some more warnings by default.

kacf:
This builds on #1676.

Originally we wanted to include this list:
* `-Wbad-function-cast`
* `-Wcast-qual`
* `-Wconversion`
* `-Wpacked`
* `-Wpadded`
* `-Wpointer-arith`
* `-Wredundant-decls`
* `-Wswitch-default`

but since it is from C, several changes were made:

* `-Wbad-function-cast` - This does not work with C++.
* `-Wcast-qual` - We should use the C++ specific `-Wold-style-cast` instead, which catches more cases than this one.
* `-Wpadded` - This only affects space usage, and would take some effort to fix. Since the C++ client is no longer being pursued for the embedded side, I chose to omit this one now, although it could be revisited later.
* `-Wswitch-default` - This one is a bit tricky, IMHO it does more harm than good to add this. A typical example of where this is a problem is [here](https://github.com/mendersoftware/mender/blob/4.0.4/src/common/error.cpp#L35-L50). We deliberately avoid using a `default` case here, in order for the compiler to warn us if we add labels to the enum without adding them here. We cannot move the final return into a `default` case, because that silences the warning. We also cannot compile it with `-Wswitch-default` on, because there is no way to tell the compiler "we deliberately omit the default case here". So I removed this from the set.

### DIFF
--- a/tests/src/common/http_resumer/http_resumer_test.cpp
+++ b/tests/src/common/http_resumer/http_resumer_test.cpp
@@ -435,7 +435,7 @@ TEST_P(DownloadResumerTest, Cases) {
 
 	http::ResponseHandler user_body_handler = [&test_case,
 											   &loop](http::ExpectedIncomingResponsePtr exp_resp) {
-		EXPECT_EQ((bool) exp_resp, test_case.success);
+		EXPECT_EQ(exp_resp.has_value(), test_case.success);
 		loop.Stop();
 	};
 
--- a/src/artifact/sha/sha.hpp
+++ b/src/artifact/sha/sha.hpp
@@ -70,7 +70,7 @@ public:
 	string String() const {
 		std::stringstream ss {};
 		for (unsigned int i = 0; i < 32; ++i) {
-			ss << std::hex << std::setw(2) << std::setfill('0') << (int) sha_.at(i);
+			ss << std::hex << std::setw(2) << std::setfill('0') << static_cast<int>(sha_.at(i));
 		}
 		return ss.str();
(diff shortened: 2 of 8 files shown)


tests/src/common/http_resumer/http_resumer_test.cpp
```diff
@@ -435,7 +435,7 @@ TEST_P(DownloadResumerTest, Cases) {
 
 	http::ResponseHandler user_body_handler = [&test_case,
 											   &loop](http::ExpectedIncomingResponsePtr exp_resp) {
-		EXPECT_EQ((bool) exp_resp, test_case.success);
+		EXPECT_EQ(exp_resp.has_value(), test_case.success);
```
vpodzime, line 438: Why not `EXPECT_TRUE()`?

vpodzime (approved):
Looks good to me either way.

tests/src/common/http_resumer/http_resumer_test.cpp
kacf: Because `success` is a boolean variable, so it can be both true and false (a better name might be `is_successful`).

lluiscampos (approved):
Thank you

kacf:
Previous pipeline is [here](https://gitlab.com/Northern.tech/Mender/mender-qa/-/pipelines/1506996188). Only formatting has changed since then, and that pipeline only had known spurious errors. Merging.

PR #1680, opened 2024-10-17
QA-712: Use upper case for Mender Client everywhere

lluiscampos:
Following up of what was decided for the Mender docs content. See:
* https://github.com/mendersoftware/mender-docs/pull/2473


estenberg (approved):
Nice, thank you.

PR #1682, opened 2024-10-21
4.0.x cherry-picks

kacf:


kacf:
The failure is just a mistake on my part when I launched the pipeline, and is not related. Merging.

PR #1690, opened 2024-11-06
MEN-7721: Check integrity of version file

lluiscampos:


src/artifact/v3/version/version.hpp
```diff
@@ -50,9 +53,16 @@ extern const ErrorCategoryClass ErrorCategory;
...
-	const int version;
-	const string format;
+class Version {
+public:
+	int version;
+	string format;
+
+	sha::SHA GetShaSum() const {
+		return shasum_;
+	}
```
kacf, line 63: Since this was originally a struct, it has no constructor, and all members are still public, I would just keep the struct, remove this and use the member directly. The `_` suffix is reserved for private members to make the distinction between struct and class more prominent when reading, so that should also be removed if going down this route.

kacf (approved):
Whoa, nice catch!

One nitpick which I think should be improved, but otherwise looks good, so pre-approving!

src/artifact/v3/header/header.cpp
```diff
@@ -235,6 +235,11 @@ ExpectedHeader Parse(io::Reader &reader, ParserConfig conf) {
 		current_index++;
 	}
 
+	if (tok.type == token::Type::Unrecognized) {
+		return expected::unexpected(parser_error::MakeError(
+			parser_error::Code::ParseError, "Unrecognized error while parsing the header"));
```
vpodzime, line 240: ~error~ token, right?

src/artifact/v3/version/version.hpp
lluiscampos: I used [this other class](https://github.com/mendersoftware/mender/blob/121dbbd87c65344679c287b3f1e249a1372f3c4f/src/artifact/v3/manifest/manifest.hpp#L39) as a reference, which that the exact thing... But to be honest I agree with you that it breaks the convention and we should amend the other class instead. I'll fix

vpodzime: Same here, it's just a value, not something that can/has to be calculated on demand.

vpodzime (commented):
+ *...~and~ as explicit errors* in the first commit message.

Looks good to me otherwise.

src/artifact/v3/header/header.cpp
lluiscampos: Not really. The error is logged somewhere else and then we return "unrecognized" token. See [the code here](https://github.com/mendersoftware/mender/blob/121dbbd87c65344679c287b3f1e249a1372f3c4f/src/artifact/lexer.hpp#L57) and [the test here](https://github.com/mendersoftware/mender/pull/1690/commits/994f658e4ee139f26f52468f06908e7290d74996#diff-1321fc7c2ed177c9698ecd7318ddf2df02706110431451b56b2026072528dc0eR365).

I considered refactoring the lexer code to be able to return the error, but quickly realized that it was too much work for this bug.

src/artifact/v3/version/version.hpp
```diff
@@ -50,9 +53,11 @@ extern const ErrorCategoryClass ErrorCategory;
 
 error::Error MakeError(ErrorCode code, const string &msg);
 
-struct Version {
-	const int version;
-	const string format;
+class Version {
+public:
```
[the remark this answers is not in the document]
lluiscampos: Now, yes :laughing:

lluiscampos:

vpodzime (approved):
Looks good to me.

PR #1691, opened 2024-11-06
MEN-7722: Bugfixes for filename parsing in Artifact manifest

lluiscampos:


src/artifact/v3/manifest/manifest.cpp
```diff
@@ -44,6 +44,7 @@ using ExpectedManifestLine = expected::expected<ManifestLine, error::Error>;
 
 const size_t expected_shasum_length {64};
 const size_t expected_whitespace {2};
+const size_t max_allowed_filename_length {4096};
```
vpodzime, line 47: There are the `PATH_MAX` and `MAXPATHLEN` system macros/constants (the latter defined as the value of the former on my system). You could use one of those. But they are set to 4096 on modern GNU/Linux systems. I'd either tried doing `#include <sys/param.h>` and use one of those if available, or at least added a comment that they are most likely 4096.

vpodzime: Also, should all three of those be `static`?

tests/src/artifact/v3/manifest/manifest_test.cpp
```diff
@@ -212,3 +213,23 @@ TEST(ParserTest, TestParseManifestShortFilenames) {
...
+		+ R"(1)";
+
+	std::stringstream ss {manifest_data};
+
+	mender::common::io::StreamReader sr {ss};
+
+	auto expected_manifest = mender::artifact::v3::manifest::Parse(sr);
+
+	ASSERT_FALSE(expected_manifest);
+	;
```
vpodzime, line 230: Missing smile in this emoji. :wink:

vpodzime (commented):
Looks good to me otherwise.

src/artifact/v3/manifest/manifest.cpp
```diff
@@ -55,9 +55,11 @@ namespace {
 string MaybeStripSuffix(string s, vector<string> suffixes) {
 	auto s_ {s};
 	for (const auto &suffix : suffixes) {
-		if (s.substr(s.size() - suffix.size()) == suffix) {
-			s_.erase(s.size() - suffix.size());
-			return s_;
+		if (s.size() > suffix.size()) {
```
kacf, line 59: I take it this means that if the suffix matches the string in `s` exactly, then it will not be removed? That's probably correct, but just checking.

src/artifact/v3/manifest/manifest.cpp
kacf: I think the header and `PATH_MAX` is POSIX-only though.

tests/src/artifact/v3/manifest/manifest_test.cpp
kacf: Haha!

kacf (approved):

src/artifact/v3/manifest/manifest.cpp
lluiscampos: Yes. The reasoning is that if it matches exactly a suffix it is _not_ a suffix :wink:

src/artifact/v3/manifest/manifest.cpp
vpodzime: > What if for the sake of keeping it simple we just restrict it to 1000? or 100?

:+1:

lluiscampos:

PR #1696, opened 2024-11-20
Fix: Private key file permissions

vpodzime:
I intentionally kept the changes from the last commit separate so that they can be easily dropped (or reverted in the future may that be necessary).

### DIFF
--- a/src/common/crypto/platform/openssl/crypto.cpp
+++ b/src/common/crypto/platform/openssl/crypto.cpp
@@ -16,6 +16,7 @@
 
 #include <common/crypto/platform/openssl/openssl_config.h>
 
+#include <cerrno>
 #include <cstdint>
 #include <string>
 #include <vector>
@@ -41,6 +42,7 @@
 #include <common/error.hpp>
 #include <common/expected.hpp>
 #include <common/common.hpp>
+#include <common/path.hpp>
 
 #include <artifact/sha/sha.hpp>
 
@@ -57,6 +59,7 @@ using namespace std;
 
 namespace error = mender::common::error;
 namespace io = mender::common::io;
+namespace path = mender::common::path;
 
 using EnginePtr = unique_ptr<ENGINE, void (*)(ENGINE *)>;
 #ifndef MENDER_CRYPTO_OPENSSL_LEGACY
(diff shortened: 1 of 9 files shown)


olehermanse (approved):

vpodzime:
@kacf I'm also thinking it might be a good idea to do `umask(077)` in both _mender-auth_ and _mender-update_. AFAICT, all files created by the two should be not group/world-accessible and if there are exceptions to that, it's OK to fix the permissions after the files have been created. What do you think? Of course it doesn't help against attacks using pre-existing files.

src/common/path.hpp
```diff
@@ -43,7 +43,7 @@ enum class Perms {
 	Others_exec,
 };
 
-error::Error Permissions(const string &file_path, vector<Perms> perms);
+error::Error Permissions(const string &file_path, vector<Perms> perms, bool warn_on_change = false);
```
kacf, line 46: Maybe the new argument should be an `enum class` with two members? It's a bit confusing to just see `true` when it's called, especially for this argument which is a bit unusual for this type of function.

src/common/crypto/platform/openssl/crypto.cpp
```diff
@@ -299,6 +299,13 @@ ExpectedPrivateKey PrivateKey::Load(const Args &args) {
 	if (args.ssl_engine != "") {
 		return LoadFromHSMEngine(args);
 	}
+	// Try to make sure the key has the right permissions and warn if there was
+	// a change (MEN-7752), but try to load it in any case (it may not even be a
+	// file on the file system).
+	if (common::StartsWith<string>(args.private_key_path, "/")) {
+		path::Permissions(
```
kacf, line 306: Unchecked error return.

src/common/crypto/platform/openssl/crypto.cpp
```diff
@@ -779,26 +779,18 @@ expected::ExpectedBool VerifySign(
 }
 
 error::Error PrivateKey::SaveToPEM(const string &private_key_path) {
-	auto bio_key = unique_ptr<BIO, void (*)(BIO *)>(NULL, bio_free_func);
+	if (path::FileExists(private_key_path)) {
+		path::FileDelete(private_key_path);
```
kacf, line 783: Unchecked error return.

kacf (approved):
Just some minor nitpicks, looks good overall!

> @kacf I'm also thinking it might be a good idea to do `umask(077)` in both _mender-auth_ and _mender-update_. AFAICT, all files created by the two should be not group/world-accessible and if there are exceptions to that, it's OK to fix the permissions after the files have been created. What do you think? Of course it doesn't help against attacks using pre-existing files.

In principle I agree, but I fear that this may have severe consequences for update modules running under mender-update. These are often updating much less privileged apps, and they may become inaccessible if this is the default mode. It's one of those things we should have started out with, but may be a bit too risky to change now. I'm not categorically against the change though, but be wary of big dragons.

We could make the change specifically for mender-auth, I suppose, because it doesn't interact with any user scripts except identity script, which are unlikely to be affected.

src/common/crypto/platform/openssl/crypto.cpp
vpodzime: Well, we don't have a way to react to the error except for logging it, but I was afraid it could produce noise for non-standard private key URIs (because it's not exactly a file system path, right?). If you think the prefix check is good enough for this to log and error, I'm all for that.

src/common/crypto/platform/openssl/crypto.cpp
vpodzime: Again, there's not much to do and we don't really care if the file was actually deleted or not, all we care is if we successfully create a new file in the next step. If, for example, something deletes the file between the `FileExists()` and `FileDelete()` calls, it's not a problem for us. So perhaps a debug-level log message?

vpodzime:
> We could make the change specifically for mender-auth, I suppose, because it doesn't interact with any user scripts except identity script, which are unlikely to be affected.

Worth considering, but probably not very wise doing here in the release time.

lluiscampos:

lluiscampos:
* 4.0.x

PR #1701, opened 2024-12-05
fix: mender-inventory-provides to read the device_type from mender.conf

MuchoLucho:
Title: None
Ticket: None

the form below:

```

<OPTIONAL LONGER DESCRIPTION>

```

### Description

Thank you!


support/mender-inventory-provides
```diff
@@ -10,6 +10,25 @@ set -e
...
 # file contains a newline, so it's important that this is done last, to avoid corrupting other
 # entries.
-grep '^device_type=' /var/lib/mender/device_type
+
+# Set default value for DeviceTypeFile
+default_device_type_file="/var/lib/mender/device_type"
+
+# Read mender.conf file to find the DeviceTypeFile key if it exists prioritizing /etc over /var/lib
+if grep -q '"DeviceTypeFile"' /etc/mender/mender.conf; then
+    device_type_file=$(grep '"DeviceTypeFile"' /etc/mender/mender.conf | awk -F ': ' '{print $2}' | tr -d '"' | tr -d ',')
```
lluiscampos, line 19: This command is not robust if, for example, you have a (valid!) json string in one line like `"DeviceTypeFile": "/my/file", "SomethingElse": "OtherValue"`. See how we do this in [`rootfs-image` update module](https://github.com/mendersoftware/mender/blob/4.0.5/support/modules/rootfs-image#L62-L68) when no `jq` is present. Here we could have something like:

```
MATCH="[Dd][Ee][Vv][Ii][Cc][Ee][Tt][Yy][Pp][Ee][Ff][Ii][Ll][Ee]"
device_type_file="$(sed -ne '/"'"$MATCH"'" *: *"[^"]*"/ { s/.*"'"$MATCH"'" *: *"\([^"]*\)".*/\1/; p }' "/etc/mender/mender.conf" || true)"
```

support/mender-inventory-provides
```diff
@@ -10,6 +10,25 @@ set -e
...
+elif grep -q '"DeviceTypeFile"' /var/lib/mender/mender.conf; then
+    device_type_file=$(grep '"DeviceTypeFile"' /var/lib/mender/mender.conf | awk -F ': ' '{print $2}' | tr -d '"' | tr -d ',')
```
lluiscampos, line 21: Also, the order should be the opposite: `/var/lib/mender/mender.conf` contents should get precedence over `/etc/mender/mender.conf` contents

lluiscampos (changes requested):
Great fix, thanks!

See comments below :point_down:

support/mender-inventory-provides
```diff
@@ -10,6 +10,25 @@ set -e
 # Also submit device_type, which is in a separate file, not in the database. We don't know if the
 # file contains a newline, so it's important that this is done last, to avoid corrupting other
 # entries.
-grep '^device_type=' /var/lib/mender/device_type
+
+# Set default value for DeviceTypeFile
+default_device_type_file="/var/lib/mender/device_type"
+
+# Read mender.conf file to find the DeviceTypeFile key if it exists prioritizing /etc over /var/lib
+if grep -q '"DeviceTypeFile"' /etc/mender/mender.conf; then
```
TheMeaningfulEngineer, line 18: `/etc/mender/mender.conf`
This can be influenced by env variables and flags

```
mender-update -h
NAME:
   mender-update - manage and start Mender Update
...
...
Environment variables:
   - MENDER_CONF_DIR - configuration (default: /etc/mender).
   - MENDER_DATA_DIR - identity, inventory and update modules (default: /usr/share/mender).
   - MENDER_DATASTORE_DIR - runtime datastore (default: /var/lib/mender).

COMMANDS:
...
GLOBAL OPTIONS:
   --config FILE, -c FILE           Configuration FILE path (default:
                                    /etc/mender/mender.conf)
   --fallback-config FILE, -b FILE  Fallback configuration FILE path (default:
                                    /var/lib/mender/mender.conf)
```

Don't have an idea on top of my head on how to approach this.
Just a note that the the hardcoded `mender.conf` may become the new `/var/lib/mender/device_type` :)

MuchoLucho: I assumed that despite you change the env variables, the /etc/mender/mender.cond and /var/lib/mender/mender.conf will not. I know it doesn't make that much sense, but otherwise we need to also alter the first three paragraphs as part of a parallel PR: https://docs.mender.io/client-installation/configuration#configuration

What do you think @TheMeaningfulEngineer ?

support/mender-inventory-provides
MuchoLucho: Hello @lluiscampos, can you confirm we need to fix this paragraph in docs then?

> Any value set in the main configuration file /etc/mender/mender.conf is chosen, whether or not the setting appears in the fallback file /var/lib/mender/mender.conf

From https://docs.mender.io/client-installation/configuration#configuration

I used to define priority as I understoo /etc/ > /var/lib.

support/mender-inventory-provides
MuchoLucho: It does make sense. Thanks.

support/mender-inventory-provides
MuchoLucho: @TheMeaningfulEngineer what do you think about bff1a4ff09dd646023fe10b3dbf08df939aaff87? 
I don't know how to handle if mender-update was executed with any of the follow ones:

```
--config FILE, -c FILE 
--fallback-config FILE, -b FILE
--data DIR, -d DIR
```

Any advise @lluiscampos?

support/mender-inventory-provides
lluiscampos: You are totally right! (I double checked the code now) Thank you :clap:

support/mender-inventory-provides
lluiscampos: We cannot possibly get the command line options of mender-auth or mender-update from the script. It is correct to use the environment variables (good call on that too!).

In a way the more we align this code with [the rootfs-image](https://github.com/mendersoftware/mender/blob/4.0.5/support/modules/rootfs-image#L44) the better.

support/mender-inventory-provides
```diff
@@ -10,6 +10,31 @@ set -e
...
 # file contains a newline, so it's important that this is done last, to avoid corrupting other
 # entries.
-grep '^device_type=' /var/lib/mender/device_type
+
+# Set default values for configuration directories
+MENDER_CONF_DIR=${MENDER_CONF_DIR:-/etc/mender}
+MENDER_DATASTORE_DIR=${MENDER_DATASTORE_DIR:-/var/lib/mender}
+
+# Set default value for DeviceTypeFile
+default_device_type_file="/var/lib/mender/device_type"
```
lluiscampos, line 19: Please use UPPER_CASE or lower_case consistently for local variables

lluiscampos (changes requested):
Looking good!

Few things:
* Squash your three commits into one.
    * Two of your commits have a typo `Title: None` instead of `Ticket: None` :sweat_smile: 
* Please consider a more user friendly changelog entry, this is a bug after all. Example: "Fix inventory reporting of `device_type` to correctly select the file based on Mender configuration and Mender environment variables"
* See also nitpick below :point_down:

lluiscampos:
@MuchoLucho Can you attend this one? We are going to be releasing very soon and this is a good fix to have. I can take over the work if you want.

support/mender-inventory-provides
MuchoLucho: @lluiscampos can you take a look at it now?

support/mender-inventory-provides
```diff
@@ -10,6 +10,31 @@ set -e
...
+default_device_type_file="$mender_datastore_dir/device_type"
+
+# Poor man's case insensitive match.
+# Read mender.conf file to find the DeviceTypeFile key if it exists prioritizing /etc over /var/lib
+match="[Dd][Ee][Vv][Ii][Cc][Ee][Tt][Yy][Pp][Ee][Ff][Ii][Ll][Ee]"
+if [ -f "$mender_conf_dir/mender.conf" ]; then
+    device_type_file=$(sed -ne '/"'"$match"'" *: *"[^"]*"/ { s/.*"'"$match"'" *: *"\([^"]*\)".*/\1/; p }' "$mender_conf_dir/mender.conf" || true)
+fi
+
+if [ -z "$device_type_file" ] && [ -f "$mender_datastore_dir/mender.conf" ]; then
```
lluiscampos, line 28: If `$device_type_file` variable is defined (iow got a value from /etc/mender/mender.conf) but the device_type file does not exist, we'll not get here for the fallback case where a different file could be specified and could be valid.

I don't know how important this is to handle, though. Is just an observation while experimenting with your fix on my device.

Edit: to make clear my comment - it could be handled with something like:
```
if [ -f "$mender_datastore_dir/mender.conf" ]; then
    if [ -z "$device_type_file" ] || [ ! -f "$device_type_file" ]; then
        ...
    fi
fi
```

lluiscampos (approved):

lluiscampos (approved):

lluiscampos:

lluiscampos:
* 4.0.x

PR #1702, opened 2024-12-05
fix: Allow blocks of zeroes after ArchiveEOF

vpodzime:
The tar format says [1]:
> The end of an archive is marked by at least two consecutive zero-filled records.

Which means that if a file is a valid tar archive, a copy of the file with blocks of zeroes appended to its end is still a valid archive, with the same content.

Thus we cannot consider extra zeroes in a file for which libarchive tells us `ARCHIVE_EOF` as superflous data.

[1] https://en.wikipedia.org/wiki/Tar_(computing)#File_format

Ticket: MEN-7810
Changelog: none

### DIFF
--- a/tests/src/artifact/tar/tar_test.cpp
+++ b/tests/src/artifact/tar/tar_test.cpp
@@ -59,6 +59,10 @@ protected:
     cp ${DIRNAME}/test.tar ${DIRNAME}/test-corrupt.tar
     dd if=/dev/random of=${DIRNAME}/test-corrupt.tar seek=10 count=5 bs=1 conv=notrunc
 
+    # Create a valid tar file with trailing zeroes
+    cp ${DIRNAME}/test-large.tar ${DIRNAME}/test-large-with-zeroes.tar
+    dd if=/dev/zero bs=1K count=3 >> ${DIRNAME}/test-large.tar
+
 		exit 0
 		)";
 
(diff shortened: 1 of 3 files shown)


tests/src/artifact/tar/tar_test.cpp
```diff
@@ -205,9 +209,28 @@ TEST_F(TarTestEnv, TestTarReaderExtraDataAtEOF) {
...
+	std::ifstream fs {tmpdir->Path() + "/test-large-with-zeroes.tar"};
+	mender::common::io::StreamReader sr {fs};
+	mender::tar::Reader tar_reader {sr};
+	mender::tar::ExpectedEntry tar_entry = tar_reader.Next();
+	ASSERT_TRUE(tar_entry);
+
+	io::Discard discard;
+	EXPECT_EQ(io::Copy(discard, tar_entry.value()), error::NoError);
+
+	mender::tar::ExpectedEntry next_tar_entry = tar_reader.Next();
```
lluiscampos, line 229: As long as it is only 0s, this second read should return `error::NoError`, like any other read to a tar that has no trailing zeros.

By looking at the code you wrote, that is what I expect too, so I cannot see the bug there.

vpodzime: It returns `TarEOFError` because there is no other entry. Like with any other tar that has no trailing zeroes. Or am I missing something?

lluiscampos: You are right, my apologies. I took the wrong test as a reference - what we want is this test to behave the same as `TestTarReaderEOF` and that is exactly what it is doing :+1:

lluiscampos (approved):
Looks good! Sorry for the mental fart :sweat_smile:

vpodzime:
@lluiscampos I don't think it makes much sense to run a full pipeline on this. If you agree, merge at will! If not, start the pipeline.

lluiscampos:
* 4.0.x

PR #1678, opened 2024-10-16
MEN-7617: Invalidate cached inventory

lluiscampos:

PR #1714, opened 2024-12-11
fix: Resend the inventory when the device has reauthenticated

jo-lund:
When using mutual TLS authentication the devices are authenticated by the gateway and will be automatically authorized in the server. This reauthentication happens without posting any error codes to the callback when polling for new deployments. Need to instead explcitly check if the device has reauthenticated and clear the inventory cache if that's the case.

Ticket: MEN-7820
Changelog: None

the form below:

```

<OPTIONAL LONGER DESCRIPTION>

```

### Description

Thank you!


vpodzime (approved):
Squash the two commits, please. Looks good to me otherwise although I have to admit I had to study the surrounding code for quite a while. Even though it was me who wrote it.

lluiscampos (changes requested):
Great to receive your first contribution @jo-lund :raised_hands: 

The code looks good, the bug was were I suspected :lady_beetle:  :mag:

This change honors a changelog entry, though. The rule of thumb is to ask "is this change release-able? Does it bring a feature or fixes a bug? Then a changelog entry is due"

You can write some user friendly text or just use `Changelog: Title` to use the title of the commit during the release of the software.

Also, you should be able to run the pipeline yourself. I'll follow-up on this privately.

jo-lund:

danielskinstad (commented):
As Lluis said, you should probably set a changelog here, and you'll also need to rebase your PR on top of master :)

lluiscampos:
@jo-lund After a very long conversation with @vpodzime, we caught two other places where a similar fix is needed.

1) Similar handling from `SubmitInventoryState`. Both on explicit `UnauthorizedError` and on success with `http_client.HasReauthenticated`

This was missing already from my original fix at #1678. If a device would have a longer poll interval than inventory submit interval (as possible use case, a device that only uses troubleshoot + inventory with a poll interval of once a year) then the clearing of Inventory data cached won't be handled.

2) A clear of the Inventory data cached from the D-Bus signal handler. Namely in `HandleReceivedToken`

There is a race condition when a 3rd party tool (a Mender add-on, or any other user software that uses our D-Bus API) does the re-authentication while `mender-update` is waiting for the next interval to expire. Then on next interval (poll interval, or inventory interval - it is irrelevant) the state machine only sees a successful reply and no `http_client.HasReauthenticated` to be detected.

lluiscampos:

lluiscampos (approved):
As we agreed face to face, let's merge this one first because it fixes the main issue reported by the user and we'll follow-up with the other fix:

> @jo-lund After a very long conversation with @vpodzime, we caught two other places where a similar fix is needed.
> 
>     1. Similar handling from `SubmitInventoryState`. Both on explicit `UnauthorizedError` and on success with `http_client.HasReauthenticated`
> 

Discussed f2f, this is not really possible because the inventory state will never attempt sending the data in the first place.

> 
>     2. A clear of the Inventory data cached from the D-Bus signal handler. Namely in `HandleReceivedToken`
> 
> 

This we'll follow-up on a separate PR :+1:

jo-lund:

lluiscampos:
* 4.0.x

PR #1715, opened 2024-12-11
fix: retry polling with separate interval

danielskinstad:


danielskinstad:
I should probably create a test for this as well; I will try to create a test, but you can review the implementation in the mean time

src/mender-update/daemon/states.cpp
```diff
@@ -106,8 +114,26 @@ void IdleState::OnEnter(Context &ctx, sm::EventPoster<StateEvent> &poster) {
...
+						if (err.code != make_error_condition(errc::operation_canceled)) {
+							log::Error("DBus poll timer caused error: " + err.String());
+						}
+					} else {
+						poster.PostEvent(StateEvent::InventoryPollingTriggered);
+					}
+				});
+			}
+#endif
+			void(this);
```
vpodzime, line 136: This should not be needed. There's a private member `poll_timer_` being used so `this` is required to be captured. Did you get a warning somewhere?

src/mender-update/daemon/states.cpp
```diff
@@ -106,8 +114,26 @@ void IdleState::OnEnter(Context &ctx, sm::EventPoster<StateEvent> &poster) {
...
+				poll_timer_.AsyncWait(chrono::seconds(5), [&poster](error::Error err) {
+					if (err != error::NoError) {
+						if (err.code != make_error_condition(errc::operation_canceled)) {
+							log::Error("DBus poll timer caused error: " + err.String());
+						}
+					} else {
+						poster.PostEvent(StateEvent::InventoryPollingTriggered);
+					}
+				});
+			}
```
vpodzime, line 134: Missing `return`?

vpodzime (commented):
Both comments apply to both changed blocks.

Please check the real names of the processes/daemons and make sure they are correct in the changelog. Now it mentions `mender-authd` and `mender-auth`. I'd like to believe they are the same thing. :grin:

src/mender-update/daemon/states.cpp
danielskinstad: Yeah, unused variable warning in the no-dbus test.

vpodzime: I wonder if we should do `this->poll_timer_` instead then. But it's strange, I believe we have this in other places too.

src/mender-update/daemon/states.cpp
danielskinstad: I don't believe it's needed? Since we have entered the `SubmitInventoryState`, I figured "Failed to submit inventory: ..." should probably be printed as well

src/mender-update/daemon/states.cpp
danielskinstad: Nothing from the class is being used here if dbus is compiled out, so the error makes sense, right? It only throws the error if we compile out dbus

vpodzime: Oh! Should the capture be different in such a case then? If not, please use `#else` and add a comment why `void(this)` is needed in that case.

src/mender-update/daemon/states.cpp
vpodzime: Yup, after the second look I agree. :+1: I wasn't quite sure even at the first iteration, hence the `?` :) But yes, we should log an error **and** move to the next state in the state machine (i.e. post a failure event).

src/mender-update/daemon/states.cpp
danielskinstad: Without the dbus the capture of `this` isn't needed, but I didn't want to overcomplicate it, but perhaps I should do:

```
#ifdef MENDER_USE_DBUS
    auto handler = [this, &ctx, &poster](error::Error err) {
         ....
 #else
    auto handler = [&ctx, &poster](error::Error err) {
    ...
#endif
```

lluiscampos (dismissed):
As we discussed in person, we need to extend the fix to
* Use the `retry_poll_interval_seconds` + `retry_poll_count` with an `ExponentialBackoff` object on the two states: poll for updates and send inventory
* The error handling should not be DBus specific but any error that "attempts to communicate with the server" according to [our documented behavior](https://docs.mender.io/client-installation/configuration/configuration-options#retrypollintervalseconds).

danielskinstad:
Reworked it to not use a backoff and do it for all errors that attempt to communicate with the server

src/mender-update/daemon/states.cpp
```diff
@@ -103,15 +103,49 @@ void IdleState::OnEnter(Context &ctx, sm::EventPoster<StateEvent> &poster) {
...
+void SubmitInventoryState::HandlePollingError(Context &ctx, sm::EventPoster<StateEvent> &poster) {
+	auto exp_interval = retry_->backoff.NextInterval();
+	if (!exp_interval) {
+		log::Debug(
+			"Not retrying with backoff, retrying InventoryPollIntervalSeconds: "
+			+ exp_interval.error().String());
+		return;
+	}
+	log::Info(
+		"Retrying deployment polling in "
```
lluiscampos, line 122: Retrying inventory submit in

src/mender-update/daemon/states.cpp
```diff
@@ -103,15 +103,49 @@ void IdleState::OnEnter(Context &ctx, sm::EventPoster<StateEvent> &poster) {
...
+		+ to_string(chrono::duration_cast<chrono::seconds>(*exp_interval).count()) + " seconds");
+
+	retry_->wait_timer.Cancel();
+	retry_->wait_timer.AsyncWait(*exp_interval, [&poster](error::Error err) {
+		if (err != error::NoError) {
+			if (err.code != make_error_condition(errc::operation_canceled)) {
+				log::Error("Retry poll timer caused error: " + err.String());
+			}
+		} else {
+			poster.PostEvent(StateEvent::DeploymentPollingTriggered);
```
lluiscampos, line 132:
```suggestion
			poster.PostEvent(StateEvent::InventoryPollingTriggered);
```

src/mender-update/daemon/states.hpp
```diff
@@ -58,25 +58,34 @@ class IdleState : virtual public StateType {
...
 	void OnEnter(Context &ctx, sm::EventPoster<StateEvent> &poster) override;
 
 private:
-	events::Timer poll_timer_;
+	void HandlePollingError(Context &ctx, sm::EventPoster<StateEvent> &poster);
+	struct Retry {
+		http::ExponentialBackoff backoff;
+		events::Timer wait_timer;
+	};
+	optional<Retry> retry_;
```
lluiscampos, line 72: Is it really optional? I think the state will miss-behave if it is not initialized with the retry

src/mender-update/daemon/states.cpp
```diff
@@ -103,15 +103,49 @@ void IdleState::OnEnter(Context &ctx, sm::EventPoster<StateEvent> &poster) {
...
+}
+
 void SubmitInventoryState::DoSubmitInventory(Context &ctx, sm::EventPoster<StateEvent> &poster) {
 	log::Debug("Submitting inventory");
 
-	auto handler = [&ctx, &poster](error::Error err) {
+	auto handler = [this, &ctx, &poster](error::Error err) {
 		if (err != error::NoError) {
+			// Replace the inventory poll timer with a backoff
+			HandlePollingError(ctx, poster);
```
lluiscampos, line 143: Here (and below) I think it is best if we first print the error message (next line) and then we handle it (where it will log that it will be retried in a short while)

lluiscampos (commented):
The change looks good. Nice work

Besides de comments below, few things:
* The changelog is good, I would only modify the start for a more generic (and relatable to our docs) intro: "On failures attempting to communicate with the server when submitting inventory or polling for deployment ..."
* As we discussed f2f, let's not cherry-pick this one to 4.0.x branch. The fix modifies the behavior of the client, and it can be seen as too intrusive on an stable branch.
* And what happened with the special handling on short poll intervals? With demo setup of poll every 5 seconds we will then wait the retry interval (which is longer). Did we agree on not handling this special case (I don't remember)?

src/mender-update/daemon/states.cpp
```diff
@@ -173,8 +238,10 @@ void PollForDeploymentState::OnEnter(Context &ctx, sm::EventPoster<StateEvent> &
 	auto err = ctx.deployment_client->CheckNewDeployments(
 		ctx.mender_context,
 		ctx.http_client,
-		[&ctx, &poster](mender::update::deployments::CheckUpdatesAPIResponse response) {
+		[this, &ctx, &poster](mender::update::deployments::CheckUpdatesAPIResponse response) {
 			if (!response) {
+				// Replace the update poll timer with a backoff
```
lluiscampos, line 243: One important thing,
Here and above (commenting here because it is easier to see on the following lines),

When the error is `auth::UnauthorizedError`, we shall not trigger the retry mechanism! That must be the exception.

src/mender-update/daemon/states.cpp
danielskinstad: Oops

src/mender-update/daemon/states.hpp
danielskinstad: Yeah, I don't think it should be optional - left over from when I ported the backoff here

lluiscampos (approved):
It looks good to me now :rocket:

danielskinstad:

src/mender-update/daemon/states.cpp
```diff
@@ -103,15 +103,59 @@ void IdleState::OnEnter(Context &ctx, sm::EventPoster<StateEvent> &poster) {
 	log::Debug("Entering Idle state");
 }
 
+SubmitInventoryState::SubmitInventoryState(
+	events::EventLoop &event_loop, int retry_interval_seconds, int retry_count) :
+	retry_(Retry {
```
vpodzime, line 108: No need to create the temporary/intermediate `Retry` instance. `retry_()` in the CIL is a constructor call.

src/mender-update/daemon/states.cpp
```diff
@@ -103,15 +103,59 @@ void IdleState::OnEnter(Context &ctx, sm::EventPoster<StateEvent> &poster) {
...
+}
+
+void SubmitInventoryState::HandlePollingError(Context &ctx, sm::EventPoster<StateEvent> &poster) {
+	// On shorter polling intervals, we need to adjust the backoff to not set
+	// intervals longer than the maximum retry polling interval
+	chrono::milliseconds max_interval =
+		chrono::seconds(ctx.mender_context.GetConfig().retry_poll_interval_seconds);
+	if (max_interval < retry_.backoff.SmallestInterval()) {
+		retry_.backoff.SetSmallestInterval(max_interval);
+		retry_.backoff.SetMaxInterval(max_interval);
```
vpodzime, line 120: This turns the exponential backoff into a fixed one, right? Worth a comment, I believe.

src/mender-update/daemon/states.cpp
```diff
@@ -150,6 +194,45 @@ void SubmitInventoryState::OnEnter(Context &ctx, sm::EventPoster<StateEvent> &po
 	DoSubmitInventory(ctx, poster);
 }
 
+PollForDeploymentState::PollForDeploymentState(
+	events::EventLoop &event_loop, int retry_interval_seconds, int retry_count) :
+	retry_(Retry {
```
vpodzime, line 199: Same here with the constructor.

src/mender-update/daemon/states.hpp
```diff
@@ -58,25 +58,34 @@ class IdleState : virtual public StateType {
...
 	void OnEnter(Context &ctx, sm::EventPoster<StateEvent> &poster) override;
 
 private:
-	events::Timer poll_timer_;
+	void HandlePollingError(Context &ctx, sm::EventPoster<StateEvent> &poster);
+	struct Retry {
+		http::ExponentialBackoff backoff;
+		events::Timer wait_timer;
+	};
+	Retry retry_;
```
vpodzime, line 72: I believe this could be combined into a single statement declaring the `retry_` field as a `struct` without a specific type name. But YMMV.

vpodzime (approved):
Just nitpicks.

src/mender-update/daemon/states.cpp
danielskinstad: Yes, it does. I'll add a short comment :+1:

PR #1719, opened 2024-12-18
fix: Clear the inventory data hash on re-authentication

jo-lund:
This fixes the case where e.g. a third party tool re-authenticates by using the D-Bus API while mender-update is waiting for the next poll interval to expire. The inventory will now be resubmitted when the poll triggers.

Ticket: MEN-7873
Changelog: Title

### DIFF
--- a/src/api/auth.hpp
+++ b/src/api/auth.hpp
@@ -68,6 +68,7 @@ struct AuthData {
 using ExpectedAuthData = expected::expected<AuthData, error::Error>;
 
 using AuthenticatedAction = function<void(ExpectedAuthData)>;
+using ReAuthenticatedAction = function<void()>;
 
 class Authenticator {
 public:
@@ -78,8 +79,14 @@ public:
 	}
 
 	void ExpireToken();
-
 	error::Error WithToken(AuthenticatedAction action);
+
+	// Register a callback to be called on re-authentication. Will overwrite the
+	// stored callback with the new one.
+	void RegisterTokenReceivedCallback(ReAuthenticatedAction action) {
+		action_ = action;
+	}
+
 
 protected:
 	enum class NoTokenAction {
@@ -97,11 +104,11 @@ protected:
 	void HandleReceivedToken(common::ExpectedStringPair ex_auth_dbus_data, NoTokenAction no_token);
 
 	events::EventLoop &loop_;
-
 	bool token_fetch_in_progress_ = false;
 	vector<AuthenticatedAction> pending_actions_;
 	chrono::seconds auth_timeout_;
 	events::Timer auth_timeout_timer_;
+	ReAuthenticatedAction action_ {nullptr};
 };
 
 #ifdef MENDER_USE_DBUS
--- a/src/api/client.hpp
+++ b/src/api/client.hpp
@@ -79,19 +79,10 @@ public:
 		authenticator_.ExpireToken();
 	}
 
-	bool HasReauthenticated() {
-		return reauthenticated_;
-	}
-
-	void SetReauthenticated(bool reauthenticated) {
-		reauthenticated_ = reauthenticated;
-	}
-
 private:
 	events::EventLoop &event_loop_;
 	http::Client http_client_;
 	auth::Authenticator &authenticator_;
-	bool reauthenticated_ {false};
 };
 
 } // namespace api
--- a/src/api/auth/auth.cpp
+++ b/src/api/auth/auth.cpp
@@ -80,7 +80,6 @@ error::Error Authenticator::WithToken(AuthenticatedAction action) {
 	}
 	// else record that token is already being fetched (by GetJwtToken()).
 	token_fetch_in_progress_ = true;
-
 	return error::NoError;
 }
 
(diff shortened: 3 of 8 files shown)


jo-lund:

jo-lund:

src/api/client.hpp
```diff
@@ -79,14 +79,6 @@ class HTTPClient : public Client {
 		authenticator_.ExpireToken();
 	}
 
-	bool HasReauthenticated() {
-		return reauthenticated_;
-	}
-
-	void SetReauthenticated(bool reauthenticated) {
-		reauthenticated_ = reauthenticated;
```
danielskinstad, line 87: You should also remove the private `reauthenticated_` variable, as it's not used anymore

jo-lund: Yes, I've done that locally but haven't submitted it yet. Saw the failing tests.

jo-lund:

src/api/auth.hpp
```diff
@@ -33,6 +33,11 @@
 #include <common/optional.hpp>
 
 namespace mender {
+namespace update {
+namespace inventory {
+class InventoryAPI;
+}
+} // namespace update
```
vpodzime, line 40: This should not be needed. `mender-update/inventory.hpp` should be `#include`d instead.

vpodzime (changes requested):
To me this is not a good way to achieve the desired behavior. It is a working solution, but from a high-level/abstract perspective pulling inventory into the Authenticator is wrong. IMHO, the Authenticator should have zero knowledge (dependency) of (on) things like inventory, deployments checking,..., whatever else. 
So let me suggest that we add a new callback to the Authenticator to call on re-authentication. It's a cleaner design and also more future proof because next time we need to do something else in case of re-authentication we won't have to modify the Authenticator itself.

src/api/auth.hpp
jo-lund: I used a forward declaration here exactly to avoid using include, which can reduce compile time etc.

jo-lund:
> To me this is not a good way to achieve the desired behavior. It is a working solution, but from a high-level/abstract perspective pulling inventory into the Authenticator is wrong. IMHO, the Authenticator should have zero knowledge (dependency) of (on) things like inventory, deployments checking,..., whatever else. So let me suggest that we add a new callback to the Authenticator to call on re-authentication. It's a cleaner design and also more future proof because next time we need to do something else in case of re-authentication we won't have to modify the Authenticator itself.

I agree, but the problem here is that HandleReceivedToken can be called from a dbus signal handler which is decoupled from the state machine itself.

src/api/auth.hpp
vpodzime: Yeah, I got it. A nice habit for sure. But we use `#ifndef-#define` guards that should prevent those issues and we should be consistent.

vpodzime:
> a dbus signal handler which is decoupled from the state machine itself

Yes, I know. That's why I told Lluis immediately when discovered this issue that the solution won't be easy and straightforward. But if we allow passing the re-authentication callback as an argument (to the constructor, for example), it can be a lambda with a capture of the `inventory_client` shared pointer and thus having access to something from the state machine.

jo-lund:
> > a dbus signal handler which is decoupled from the state machine itself
> 
> Yes, I know. That's why I told Lluis immediately when discovered this issue that the solution won't be easy and straightforward. But if we allow passing the re-authentication callback as an argument (to the constructor, for example), it can be a lambda with a capture of the `inventory_client` shared pointer and thus having access to something from the state machine.

Yes, that would be a better design, and it would be nice to have a generic API for this. I was actually looking for a publish/subscribe mechanism in the code when implementing this change.

src/api/auth.hpp
jo-lund: Include guards and forward declarations are not the same and serve different purposes. When you include a file the whole file will be inserted into the current one together with all the includes in that file as well (as long as they have not been included already because they have defined an include guard). Include guards prevent the same header file from being included multiple times and therefore prevent redefinition of the same symbol.

With forward declarations you get just the symbol with incomplete type. This can reduce compile time substantially, e.g. because you don't need to recompile several files when a header changes.

vpodzime:
> I was actually looking for a publish/subscribe mechanism in the code when implementing this change.

:+1: That would be nice. But we only have callbacks and lambdas, hoping for the best regarding memory management and object lifetime. :grinning:

jo-lund:

src/api/auth.hpp
vpodzime: Thanks for elaborating on this. It's a nice thing, but since we don't do it anywhere else, I'd not introduce it here in this one place.

vpodzime (approved):
Nice and clean! :+1:

src/api/auth.hpp
lluiscampos: We actually have some (one?) forward declaration like this one - see [here](https://github.com/mendersoftware/mender/blob/5.0.0/src/common/http.hpp#L50). In that case the purpose is to avoid a circular include and not to optimize compile time.

I personally have no inconvenience either way, I only wanted to pint out that there is a precedent.

vpodzime:
>  ci/gitlab — Pipeline failed on Gitlab 

I think we need to skip some (more) tests with ASAN because it doesn't like the memory pool/allocation used by *libdbus*.

danielskinstad:
> > ci/gitlab — Pipeline failed on Gitlab
> 
> I think we need to skip some (more) tests with ASAN because it doesn't like the memory pool/allocation used by _libdbus_.

I saw several of these issues last week, it might work to just re-run it

PR #1724, opened 2025-01-10
MEN-7900: Fix Mender client getting stuck after failure in sync state

lluiscampos:
From the code point, the issue was that the re-scheduling of new polls for updates or inventory were done from the states _after_ the sync state, so unless the state machine reached that point the new polls would not be scheduled.

Fix by creating two new states, that just do the re-scheduling, between idle and sync. Note that the timer(s) have now been moved to the context object so that it can be accessed from multiple states (namely, update polling and submit inventory states which would need to manipulate the timer for exponential back-off retries.

Changelog: Fix issue where any error in Sync state (triggered for example with an error in Sync_Enter or Sync_Leave state scripts) leaves the client stuck in idle state forever and no new polls for update nor submit of inventory would be attempted again.

TODO:
- [x] Modify [the integration tests](https://github.com/mendersoftware/integration/blob/f67ea8820c5944a696fc8502d8fcd7d90cf41d26/tests/tests/test_state_scripts.py#L167) so that this bug would have been caught (check that i fails with master and passes with this PR)
- [x] Consider duplicating the classes (discussion below)
- [x] Consider refactor the backoff (discussion below)

### DIFF
--- a/src/mender-update/daemon/states.cpp
+++ b/src/mender-update/daemon/states.cpp
@@ -103,24 +103,44 @@ void IdleState::OnEnter(Context &ctx, sm::EventPoster<StateEvent> &poster) {
 	log::Debug("Entering Idle state");
 }
 
-SubmitInventoryState::SubmitInventoryState(
-	events::EventLoop &event_loop, int retry_interval_seconds, int retry_count) :
-	retry_ {
-		http::ExponentialBackoff(chrono::seconds(retry_interval_seconds), retry_count),
-		event_loop} {
+ScheduleNextPollState::ScheduleNextPollState(
+	events::Timer &timer, const string &poll_action, const StateEvent event, int interval) :
+	timer_ {timer},
+	poll_action_ {poll_action},
+	event_ {event},
+	interval_ {interval} {
+}
+
+void ScheduleNextPollState::OnEnter(Context &ctx, sm::EventPoster<StateEvent> &poster) {
+	log::Debug("Scheduling the next " + poll_action_ + " in: " + to_string(interval_) + " seconds");
+	timer_.AsyncWait(chrono::seconds(interval_), [this, &poster](error::Error err) {
+		if (err != error::NoError) {
+			if (err.code != make_error_condition(errc::operation_canceled)) {
+				log::Error("Timer caused error: " + err.String());
+			}
+		} else {
+			poster.PostEvent(event_);
+		}
+	});
+
+	poster.PostEvent(StateEvent::Success);
+}
+
+SubmitInventoryState::SubmitInventoryState(int retry_interval_seconds, int retry_count) :
+	backoff_ {chrono::seconds(retry_interval_seconds), retry_count} {
 }
 
 void SubmitInventoryState::HandlePollingError(Context &ctx, sm::EventPoster<StateEvent> &poster) {
-	// When using short polling inteervals, we should adjust the backoff to ensure
+	// When using short polling intervals, we should adjust the backoff to ensure
 	// that the intervals do not exceed the maximum retry polling interval, which
 	// converts the backoff to a fixed interval.
 	chrono::milliseconds max_interval =
 		chrono::seconds(ctx.mender_context.GetConfig().retry_poll_interval_seconds);
-	if (max_interval < retry_.backoff.SmallestInterval()) {
-		retry_.backoff.SetSmallestInterval(max_interval);
-		retry_.backoff.SetMaxInterval(max_interval);
+	if (max_interval < backoff_.SmallestInterval()) {
+		backoff_.SetSmallestInterval(max_interval);
+		backoff_.SetMaxInterval(max_interval);
 	}
-	auto exp_interval = retry_.backoff.NextInterval();
+	auto exp_interval = backoff_.NextInterval();
 	if (!exp_interval) {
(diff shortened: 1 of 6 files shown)


src/mender-update/daemon/states.cpp
```diff
@@ -103,24 +103,43 @@ void IdleState::OnEnter(Context &ctx, sm::EventPoster<StateEvent> &poster) {
...
-		event_loop} {
+ScheduleNextPollState::ScheduleNextPollState(
+	events::Timer &timer, const StateEvent event, int interval) :
+	timer_ {timer},
+	event_ {event},
+	interval_ {interval} {
+}
+
+void ScheduleNextPollState::OnEnter(Context &ctx, sm::EventPoster<StateEvent> &poster) {
+	log::Debug("Scheduling the next check in: " + to_string(interval_) + " seconds");
```
lluiscampos, line 114: As you can see, I reused the same class `ScheduleNextPollState` for both poll for updates and submit inventory, by passing by reference which of the timers and interval to use.

The only disadvantage that I see is the log messages here being more generic: "next check" instead of "next inventory submission" for example.

src/mender-update/daemon/states.cpp
```diff
@@ -174,45 +193,21 @@ void SubmitInventoryState::DoSubmitInventory(Context &ctx, sm::EventPoster<State
...
-PollForDeploymentState::PollForDeploymentState(
-	events::EventLoop &event_loop, int retry_interval_seconds, int retry_count) :
-	retry_ {
-		http::ExponentialBackoff(chrono::seconds(retry_interval_seconds), retry_count),
-		event_loop} {
+PollForDeploymentState::PollForDeploymentState(int retry_interval_seconds, int retry_count) :
+	backoff_ {chrono::seconds(retry_interval_seconds), retry_count} {
 }
 
 void PollForDeploymentState::HandlePollingError(Context &ctx, sm::EventPoster<StateEvent> &poster) {
```
lluiscampos, line 200: This `HandlePollingError` is still in two places, in here and in `SubmitInventoryState`.

We could move it by having the `backoff` objects also in the context. What do you think? Again the main disadvantage would be the logging, which in this case is info and error level so it is more relevant to get the more context for the (human) user reading the logs.

src/mender-update/daemon/states.hpp
```diff
@@ -56,34 +56,37 @@ class IdleState : virtual public StateType {
...
+class ScheduleNextPollState : virtual public StateType {
+public:
+	ScheduleNextPollState(events::Timer &timer, const StateEvent event, int interval);
+
+	void OnEnter(Context &ctx, sm::EventPoster<StateEvent> &poster) override;
+
+private:
+	events::Timer &timer_;
+	const StateEvent event_;
+	int interval_;
```
lluiscampos, line 68: Note to self,
```suggestion
	const events::Timer &timer_;
	const StateEvent event_;
	const int interval_;
```

src/mender-update/daemon/states.cpp
vpodzime: You could add a `const string &action` argument to the constructor and use the action description here.

lluiscampos: Yes, thank you!

I actually thought about this just as I logged of on Friday... and forgot to comment to myself today :sweat_smile: 

This also partially responds to the other question, although there we might want to keep the retry contained in the class(es) than can actually use it instead of in the shared context.

src/mender-update/daemon/states.cpp
```diff
@@ -174,45 +193,21 @@ void SubmitInventoryState::DoSubmitInventory(Context &ctx, sm::EventPoster<State
...
-			if (err != error::NoError) {
-				if (err.code != make_error_condition(errc::operation_canceled)) {
-					log::Error("Inventory poll timer caused error: " + err.String());
-				}
-			} else {
-				poster.PostEvent(StateEvent::InventoryPollingTriggered);
-			}
-		});
-
-	DoSubmitInventory(ctx, poster);
```
vpodzime, line 195: Where is this happening now?

lluiscampos: in `OnEnter`. The previous `OnEnter` only scheduled the next interval and called `DoSubmitInventory`

src/mender-update/daemon/states.cpp
vpodzime: > This `HandlePollingError` is still in two places, in here and in `SubmitInventoryState`.

I think that's fine. These two cases can potentially differ even more in the future.

src/mender-update/daemon/states.cpp
danielskinstad: Yeah, the action argument would be nice

danielskinstad (approved):
As was pointed out, having the action description as an arg and using that when logging would be nice; other than that it looks good to me :rocket:

src/mender-update/daemon/states.cpp
vpodzime: I see. I was looking for this function call somewhere. :+1:

lluiscampos:
Integration test fix at:
* https://github.com/mendersoftware/integration/pull/2711

lluiscampos:
* 5.0.x

PR #1726, opened 2025-01-17
fix: Clear the inventory data hash on re-authentication

jo-lund:
This fixes the case where e.g. a third party tool re-authenticates by using the D-Bus API while mender-update is waiting for the next poll interval to expire. The inventory will now be resubmitted when the poll triggers.

Ticket: MEN-7873
Changelog: Title

### DIFF
--- a/src/api/auth.hpp
+++ b/src/api/auth.hpp
@@ -68,6 +68,7 @@ struct AuthData {
 using ExpectedAuthData = expected::expected<AuthData, error::Error>;
 
 using AuthenticatedAction = function<void(ExpectedAuthData)>;
+using ReAuthenticatedAction = function<void()>;
 
 class Authenticator {
 public:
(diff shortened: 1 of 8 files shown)


jo-lund:

lluiscampos:

jo-lund:

lluiscampos (approved):
It looks good.

For the next time,
* When doing cherry-picks, it is nice to add the `-x` flag on your `commit` command so that the original commit gets referenced
* When a cherry-pick applies clean (iow there are no conflicts) you don't need to wait for peer approval. I usually just type "approved in #xxx) and that's it. As long as CI is green we are all happy.
* An also, try to avoid the merge commits on PRs - a rebase + force-push is much cleaner wrt how the final Git history will be seen from the base branch (5.0.x in this case) :smiley: 

Anyway, let's get this one in and unblock the :train: - FYI @danielskinstad

PR #1740, opened 2025-01-30
chore: Made permissions of test file more strict

olehermanse:


olehermanse:
@lluiscampos @vpodzime I figured it shouldn't be a problem to test with 0600 here, instead of 0664 or 0644 (?).

lluiscampos (approved):

PR #1741, opened 2025-01-31
fix: Fix wrong parameter order for AuthData

jo-lund:
This has not been used since it is a POC meant to be used for platforms without D-Bus

Ticket: MEN-7966


lluiscampos (approved):
Thank you.

In this case, as we don't considered an actual bug that is part of the production code, we can set `Changelog: None` in the commit :+1:

olehermanse:
@jo-lund This PR includes my commits now - try to fetch and rebase on the latest upstream master to get rid of them.

PR #1749, opened 2025-02-03
fix: Disallow copying and assignment for PrivateKey

jo-lund:
Changelog: None
Ticket: MEN-7965

### DIFF
--- a/src/common/crypto.hpp
+++ b/src/common/crypto.hpp
@@ -44,7 +44,7 @@ struct Args {
 };
 
 class PrivateKey;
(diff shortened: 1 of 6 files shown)


jo-lund:

jo-lund:

vpodzime (approved):
I'm not sure the issue here is really worth the more complicated code because, AFAICT, even the original code only copies **a pointer** to the private key, not the actual private key data. But if we want to avoid even that, this is the right way to do it.

lluiscampos (commented):
Looks good to me, thank you.

PR #1751, opened 2025-02-04
fix: Use the actual length of the DER encoded signature for the buffer

jo-lund:
Will now get the expected length of the DER encoded signature and use that to allocate an array to hold the encoded data.

Ticket: MEN-7979
Changelog: None

### DIFF
--- a/src/common/crypto/platform/openssl/crypto.cpp
+++ b/src/common/crypto/platform/openssl/crypto.cpp
@@ -603,7 +603,6 @@ expected::ExpectedString Sign(const Args &args, const vector<uint8_t> &raw_data)
 	return EncodeBase64(signature);
 }
 
-const size_t mender_decode_buf_size = 256;
 const size_t ecdsa256keySize = 32;
 
 // Try and decode the keys from pure binary, assuming that the points on the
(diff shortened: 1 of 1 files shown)


jo-lund:

src/common/crypto/platform/openssl/crypto.cpp
```diff
@@ -665,19 +664,25 @@ static expected::ExpectedBytes TryASN1EncodeMenderCustomBinaryECFormat(
 	r.release();
 	s.release();
 
-	/* Allocate some array guaranteed to hold the DER-encoded structure */
-	vector<uint8_t> der_encoded_byte_array(mender_decode_buf_size);
-	unsigned char *arr_p = &der_encoded_byte_array[0];
-	int len = i2d_ECDSA_SIG(ecSig.get(), &arr_p);
+	/* Get the expected length in bytes of the DER encoded signature */
+	int len = i2d_ECDSA_SIG(ecSig.get(), NULL);
```
vpodzime, line 668: According to the docs, this allocates a buffer and writes the DER data into it. We work with that buffer and `free()` it in the end.

vpodzime: Oh, scratch that. That's the case if the given pointer points to `NULL`, of course. So that would be an alternative solution, but not better in any way, I think.

src/common/crypto/platform/openssl/crypto.cpp
```diff
@@ -665,19 +664,25 @@ static expected::ExpectedBytes TryASN1EncodeMenderCustomBinaryECFormat(
...
 			SetupError,
 			"Failed to set the signature parts in the ECDSA structure: "
 				+ GetOpenSSLErrorMessage()));
 	}
-	/* Resize to the actual size of the DER-encoded signature */
-	der_encoded_byte_array.resize(len);
 
+	/* Allocate an array guaranteed to hold the DER-encoded structure */
+	vector<uint8_t> der_encoded_byte_array(len);
+	unsigned char *arr_p = &der_encoded_byte_array[0];
```
vpodzime, line 678: You can do `.data()` to get the pointer to the data part. And we should use the same type, I think, so `unsigned char` even for the vector.

vpodzime (approved):
Looks good to me either way.

lluiscampos (changes requested):
The code looks good to me!

If I read it correctly, this one is an actual bug fix - please add a `Changelog` entry to the commit.

src/common/crypto/platform/openssl/crypto.cpp
vpodzime: I think `.data()` is cleaner than getting an address of the first element. They are guaranteed to be the same (I hope!), but it's more C++-ish.

vpodzime:
> If I read it correctly, this one is an actual bug fix - please add a `Changelog` entry to the commit.

AFAICT, it changes nothing. ECDSA signature is 256 bytes so there's no difference. It's just a cleaner way. So no actual bug fixed here. If I read the changelog entry, I'd think there was something functionally broken before. But YMMV.

jo-lund:
I agree with @vpodzime. No bug actually fixed but just a nicer way of doing it.

lluiscampos (approved):
@vpodzime @jo-lund Okey, then let's remove the changelog. Sorry for my misinterpretation.

@jo-lund After amending the commit, you don't need to re-run the integration pipeline - I will force-merge it for you.

PR #1752, opened 2025-02-04
fix: Fix undefined behaviour after a move operation

jo-lund:
This fixes the case where we access an object after it has been moved.

Ticket: MEN-7980
Changelog: None


jo-lund:

lluiscampos (approved):

PR #1753, opened 2025-02-05
chore: Remove ssh configuration from the docker test container

jo-lund:
Ticket: MEN-7972
Changelog: None

### DIFF
--- a/tests/Dockerfile.daemon
+++ b/tests/Dockerfile.daemon
@@ -31,15 +31,7 @@ RUN mender-artifact write bootstrap-artifact \
 FROM ubuntu:22.04
 
 RUN mkdir -p /run/dbus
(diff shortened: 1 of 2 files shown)


tests/Dockerfile.daemon
```diff
@@ -33,14 +33,6 @@ FROM ubuntu:22.04
 RUN mkdir -p /run/dbus
 RUN apt update && apt install -y openssh-server libboost-log1.74.0 liblmdb0 libarchive13 dbus iproute2
```
lluiscampos, line 34: We can remove here the package too
```suggestion
RUN apt update && apt install -y libboost-log1.74.0 liblmdb0 libarchive13 dbus iproute2
```

lluiscampos (approved):
Looks good! See a comment below.

Nit pick: I would mark this commit as `chore:` or even`test:`, IMO we are not fixing anything but removing a hack that we don't need anymore.

jo-lund:

PR #1754, opened 2025-02-06
refac: Added warning when using basic auth

victormlg:


src/common/http/platform/beast/http.cpp
```diff
@@ -458,6 +458,7 @@ static inline error::Error AddProxyAuthHeader(OutgoingRequest &req, BrokenDownUr
 		return ex_encoded_creds.error();
 	}
 	req.SetHeader("Proxy-Authorization", "Basic " + ex_encoded_creds.value());
+	log::Warning("Avoid Basic Auth if possible, and make sure if it's used, it's through HTTPS");
```
lluiscampos, line 461: I suggest we use the exact same terminology than in [Wikipedia](https://en.wikipedia.org/wiki/Basic_access_authentication):

```suggestion
	log::Warning("Avoid Basic access authentication if possible, and make sure if it's used, it's through HTTPS");
```

Or, at least, let's please spell out the full world
```suggestion
	log::Warning("Avoid Basic Authorization if possible, and make sure if it's used, it's through HTTPS");
```

olehermanse: @lluiscampos I agree, but basic auth means basic authentication, not authorization. (The fact that there is `Authorization` in the HTTP header is annoying / confusing).

Also, I don't think we need to capitalize this.

See:
https://developer.mozilla.org/en-US/docs/Web/HTTP/Authentication

```suggestion
	log::Warning("Avoid using basic authentication if possible, and make sure if it's used, it's through HTTPS");
```

lluiscampos: > The fact that there is Authorization in the HTTP header is annoying / confusing

Yeah, I wrote it as an alternative to match 1:1 the HTTP header. I still prefer my first suggestion (and with no capitalization, as suggested on top of it)

src/common/http/platform/beast/http.cpp
```diff
@@ -458,6 +458,8 @@ static inline error::Error AddProxyAuthHeader(OutgoingRequest &req, BrokenDownUr
 		return ex_encoded_creds.error();
 	}
 	req.SetHeader("Proxy-Authorization", "Basic " + ex_encoded_creds.value());
+	log::Warning(
+		"Avoid using basic authentification if possible, and make sure if it's used, it's through HTTPS");
```
lluiscampos, line 462:
```suggestion
		"Avoid using basic authentication if possible, and make sure if it's used, it's through HTTPS");
```

I am not the grammar expert, but I think "authentification" is not a real word in English.

victormlg: My bad, I don't even know how I did that

lluiscampos (approved):

danielskinstad (approved):

PR #1763, opened 2025-02-28
fix: Fix deployment on 32 bit systems when rootfs is larger than 4Gb

jo-lund:

Ticket: MEN-8062
Changelog: Title


jo-lund:

lluiscampos (approved):

PR #1766, opened 2025-03-04
4.0.x: fix: Fix deployment on 32 bit systems when rootfs is larger than 4Gb

jo-lund:

Ticket: MEN-8062
Changelog: Title

Signed-off-by: John Olav Lund <john.olav.lund@northern.tech>
(cherry picked from commit 8ba599b6ec48b7be9f4e5181593ec8c34d78b271)


jo-lund:

PR #1767, opened 2025-03-04
5.0.x: fix: Fix deployment on 32 bit systems when rootfs is larger than 4Gb

jo-lund:

Ticket: MEN-8062
Changelog: Title


lluiscampos:
Testing together:
* https://github.com/mendersoftware/mender/pull/1767
* https://github.com/mendersoftware/mender-qa/pull/780

@ [Pipeline 1700586638](https://gitlab.com/Northern.tech/Mender/mender-qa/-/pipelines/1700586638) with only client acceptance tests

PR #1768, opened 2025-03-04
fix: wget hanging while retrying forever

deligatedgeek:

- [√] Make sure that all commits follow the conventional commit [specification](https://www.github.com/mendersoftware/mendertesting/commitlint/grammar.md) for the Mender project.

the form below:

```
fix: wget effectively hangs when using ipinfo.io in restricted environments


Changelog: Add retry limit to inventory geo script to prevent hanging
Ticket: None
```

- [√] Make sure that all commits are signed with [`git --signoff`](https://git-scm.com/book/en/v2/Git-Tools-Signing-Your-Work). Also note that the signoff author must match the author of the commit.

### Description

In mender-inventory-geo the wget command has a timeout but wget will retry forever.  In restricted environments this effectively causes the script to hang, never sending inventory data.
I have added tries=2 to allow wget to fail so the script can handle it

Thank you!

### DIFF
--- a/support/mender-inventory-geo
+++ b/support/mender-inventory-geo
@@ -40,7 +40,7 @@ ATTR_NAME_LAT="geo-lat"
 ATTR_NAME_LON="geo-lon"
 
 if [ ! -f "${TEMPFILE}" ]; then
-    ipinfo=$(wget --timeout=${TIMEOUT} -q -O /dev/stdout \
+    ipinfo=$(wget --timeout=${TIMEOUT} -q -O /dev/stdout --tries=2 \
                   --header 'Accept: application/json' \
                   https://ipinfo.io)
     if [ $? != 0 ] || [ -z "${ipinfo}" ]; then


lluiscampos (changes requested):
Hi @deligatedgeek and thank you for your contribution. This is indeed a good improvement.

The code looks good :100: . Only there are some formalities to address.
* Sign your commit (you clicked the checkbox in the PR but not actually signed your work). You need to amend your current commit sigh the signature and update the PR.
* You are missing the `:` in your commit title: `fix: wget hanging...`

Also, it won't retry forever, according to the man page the default is 20 times.

deligatedgeek:
I have added the signed off and edited the title.

The default may be 20 times, but that results in 20 x 10 second timeouts, which is over 3 minutes.

A long time when manually replacing "firmware" on devices deployed remotely on customer sites with firmware containing mender-client.  8-)

deligatedgeek:
Created a new PR with a correctly formatted commit.

PR #1772, opened 2025-03-12
fix: Check for control characters in URLDecode

jo-lund:
Ticket: MEN-7963
Changelog: None


jo-lund:

src/common/http/http.cpp
```diff
@@ -202,6 +202,10 @@ expected::ExpectedString URLDecode(const string &value) {
 			}
 			unsigned int num;
 			sscanf(value.substr(i + 1, 2).c_str(), "%x", &num);
+			if (num < 0x20) {
```
vpodzime, line 205: Would be nice to add a comment about why `0x20` is the key value.

vpodzime (approved):
Looks good to me either way.

PR #1782, opened 2025-04-02
feat: Add User-Agent header to HTTP requests

jo-lund:
The Mender client version will now be sent to the server through the User-Agent header.

Tiicket: MEN-1979
Changelog: Title

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -20,4 +20,3 @@ add_subdirectory(client_shared)
 add_subdirectory(mender-update)
 add_subdirectory(mender-auth)
 add_subdirectory(artifact)
-
--- a/src/common/http/platform/beast/http.cpp
+++ b/src/common/http/platform/beast/http.cpp
@@ -24,6 +24,8 @@
 #include <common/common.hpp>
 #include <common/crypto.hpp>
 
+#include <mender-version.h>
+
 namespace mender {
 namespace common {
 namespace http {
@@ -424,6 +426,9 @@ error::Error Client::AsyncCall(
 	req->SetHeader("HOST", header_url);
 
 	log::Trace("Setting HOST address: " + header_url);
+
+	// Add User-Agent header for all requests
+	req->SetHeader("User-Agent", "Mender/" MENDER_VERSION);
 
 	header_handler_ = header_handler;
 	body_handler_ = body_handler;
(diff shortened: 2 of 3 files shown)


jo-lund:

src/CMakeLists.txt
```diff
@@ -9,6 +9,18 @@ if(NOT "${GIT_CMD_ERROR}" STREQUAL "")
   message(FATAL_ERROR "Git command failed:\n" ${GIT_CMD_ERROR})
 endif()
 
+execute_process(
+  COMMAND sh -c "git describe --tags --dirty --abbrev=0"
+  WORKING_DIRECTORY ${PROJECT_SOURCE_DIR}
+  OUTPUT_VARIABLE MENDER_CLIENT_VERSION
```
lluiscampos, line 15: We already had `MENDER_VERSION`, is there a reason for not using the same?

jo-lund: This contained the commit hash when I tested this on master.

lluiscampos: But I think that is what we want. Do you think otherwise?

Besides, if there is an issue with the commit sha we should then fix the former variable, otherwise `mender-update --version` and the version in the User-Agent won't match (which also means that device inventory and device HTTP info to the server won't match neither)

jo-lund: I agree that we should use the current variable for this. However, I would expect an actual version number for the client and not a commit hash in a User-Agent string, but I will change to using the variable we already have. And it's a lot better anyway to only use one variable for this.

lluiscampos: But we get the version number when building from the Git tag, which is what we expect for production. That is the software that we test and release. If the user builds from master or from any fork then they get the Git sha.

src/CMakeLists.txt
```diff
@@ -20,4 +20,3 @@ add_subdirectory(client_shared)
 add_subdirectory(mender-update)
 add_subdirectory(mender-auth)
 add_subdirectory(artifact)
-
```
lluiscampos, line 23: Unrelated change

lluiscampos (approved):

PrimeQH:
捕获
@lluiscampos  Hope this helps to solve this log bug.thanks!!!

lluiscampos:
Hi @PrimeQH,

This is a different code base - is this a fork from `mendersoftware/mender`?

I would like to take a look but please create an issue in our [bug tracker](https://tracker.mender.io/projects/MEN), adding the full output (in text, not in a picture) and steps to reproduce :innocent:

PR #1807, opened 2025-07-17
chore: bump libboost_log to 1.83.0

danielskinstad:
The following errors are present in mendersoftware/mender-client-docker:mender-master

```
mender-auth: error while loading shared libraries: libboost_log.so.1.83.0: cannot open shared object file: No such file or directory
mender-update: error while loading shared libraries: libboost_log.so.1.83.0: cannot open shared object file: No such file or directory
```


vpodzime (approved):
Makes sense

PR #1810, opened 2025-07-22
fix: directory-artifact-gen passthrough arguments mess up

Stanley00:

the form below:

```

<OPTIONAL LONGER DESCRIPTION>

```

### Description
This ensures the `echo` is not the built-in command, since built-in command from `sh` shell don't support -n and just will just make the passthrough arguments messed up


danielskinstad (commented):
I haven’t looked closely at the entire script, but I think making sure the script is fully portable and sh compatible would be the correct fix here (I’m not asking you to do this though).

support/modules-artifact-gen/directory-artifact-gen
```diff
@@ -170,7 +170,7 @@ echo "$dest_dir" > $dest_dir_file
 # Runs in a subshell to allow overriding some parameters passed
 # to mender-artifact
 passthrough_args_modified=" "
-echo -n $passthrough_args | xargs -n 2 printf "%s %s\n" | (while read -r flag arg; do
+\echo -n $passthrough_args | xargs -n 2 printf "%s %s\n" | (while read -r flag arg; do
```
Stanley00, line 173: I read the POSIX spec for `echo` is that it won't accept and flags, so I guess remove `-n` flag here should work. I did tested few manually test and it seems to work. Not sure if we have any test case specific for this one?

New change should look like this instead:
```
echo $passthrough_args | xargs -n 2 printf "%s %s\n" | (while read -r flag arg; do
```

danielskinstad: That seems to work. 

> Not sure if we have any test case specific for this one?

No this is only indirectly tested through things like https://github.com/mendersoftware/integration/blob/90498718bea213d8fcbd0b292cb30a885c67fd9c/tests/tests/test_provides_depends.py#L58C6-L58C7

danielskinstad: Will you modify your PR? If so please squash everything to 1 commit and force-push.

lluiscampos:
Superseded by #1830. Thanks @Stanley00 for your contribution!

PR #1811, opened 2025-07-22
fix: Use /proc/cmdline for root device detection instead of stat in rootfs update module

adbjo:
See topic in forum for background:
https://hub.mender.io/t/mounted-root-does-not-match-boot-loader-environment-error-with-overlayroot-setup/7911


TheYoctoJester:
Hi @adbjo,

Thanks for the submission! Looking at it a bit, I'm not sure if completely replacing the logic by looking at the kernel `cmdline` is a good idea. I agree that in most cases it will be valid, but something (maybe just gut feeling) tells me that it should not be the only source of information.

So for now, I think we should keep the existing logic. What I'm not sure about is, if the logic should be the fallback if `cmdline` evaluation fails, or if `cmdline` evaluation should be the fallback if the existing logic fails.

What's your thoughts on this?

Greetz,
Josef

adbjo:
> Hi @adbjo,
> 
> Thanks for the submission! Looking at it a bit, I'm not sure if completely replacing the logic by looking at the kernel `cmdline` is a good idea. I agree that in most cases it will be valid, but something (maybe just gut feeling) tells me that it should not be the only source of information.
> 
> So for now, I think we should keep the existing logic. What I'm not sure about is, if the logic should be the fallback if `cmdline` evaluation fails, or if `cmdline` evaluation should be the fallback if the existing logic fails.
> 
> What's your thoughts on this?
> 
> Greetz, Josef

Hey Josef, thanks for checking it out and providing great input :)

I have now updated the PR so that the existing logic is the default, I think this minimizes changes to behavior and makes the most sense.

So, the input device will be checked against the parsed `ROOT_DEVICE` as a fallback. I also made it additionally check that the device must be mounted, for some extra verification.

Let me know what you think. Also, a big reason we're doing this is because we have a pro subscription and want to use delta updates feature (we already have an alternative approach that works but `overlayroot` seems very handy so it would be nice if we were able to use it). Does the delta update module inherit logic from this `rootfs` module, or how does that work? Maybe I should ask support about this?

TheYoctoJester:
There was a failure on `vexpress-qemu-flash`, which is the UBI related test suite. Also not sure what triggered it exactly, maybe @danielskinstad or @vpodzime?

adbjo:
> There was a failure on `vexpress-qemu-flash`, which is the UBI related test suite. Also not sure what triggered it exactly, maybe @danielskinstad or @vpodzime?

Fixed it by removing the `mount | grep ...`. This approach doesn't work with UBI and is probably unreliable. `findmnt` might work, but I suspect it has the same problem.

danielskinstad:

danielskinstad (approved):
Looks good to me, @adbjo 

Thanks for taking caring of this one,  @TheYoctoJester :rocket:

adbjo:
> > Btw, I talked to support, and the delta update module will not automatically get the same fix as far as I understood. So maybe this should not be merged until the same fix is ready in the delta update module ?
> 
> 
> 
> The delta Update Module is a completely different implementation, so it doesn't apply anyways.
> 
> 
> 
> > I can imagine having a read-only setup that works with the rootfs update module, but not with delta update would be very frustrating.
> 
> 
> 
> It's actually the other way around. The delta Update Module would not work *without* the root filesystem being read-only. So no worries there. Tests have passed, so we're probably good to go? @danielskinstad
> 

Well, what I meant was that this PR makes the behavior of the rootfs update module diverge from the delta update module. Maybe that's fine, I'll let you guys decide on that :)

adbjo:
I have now tested with the delta update module

> > > Btw, I talked to support, and the delta update module will not automatically get the same fix as far as I understood. So maybe this should not be merged until the same fix is ready in the delta update module ?
> > 
> > 
> > The delta Update Module is a completely different implementation, so it doesn't apply anyways.
> > > I can imagine having a read-only setup that works with the rootfs update module, but not with delta update would be very frustrating.
> > 
> > 
> > It's actually the other way around. The delta Update Module would not work _without_ the root filesystem being read-only. So no worries there. Tests have passed, so we're probably good to go? @danielskinstad
> 
> Well, what I meant was that this PR makes the behavior of the rootfs update module diverge from the delta update module. Maybe that's fine, I'll let you guys decide on that :)
> 
> Potentially at least, I don't actually know if the delta update module needs the same fix, I will try!

I have now tested with the delta update module, it doesn't have the same problem. So ignore my above comments :)

PR #1812, opened 2025-07-22
chore: Use repo instead of standalone deb packages

elkoniu:
Mender supports tool installation using repositories now. This is the correct way of installing the packages instead of downloading fixed version as a *.deb files.

Ticket: QA-1090
Changelog: None

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -14,7 +14,7 @@ variables:
     description: "Version of docker to use in pipelines"
 
   # mender-artifact version for tests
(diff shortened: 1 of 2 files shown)


vpodzime (commented):
Looks good to me, but fails CI. And I would actually split this into two changes:
- fix the URL to use the new repo
- change the tests to use the repo and `apt-get` instead of `wget` and `dpkg -i`

And the latter change is not really that important here at this point so if causes a headache, we can skip it for now.

michalkopczan (approved):

vpodzime (approved):
Looks good to me.

PR #1815, opened 2025-08-15
WIP

danielskinstad:
Ticket: ME-527
Changelog: Add a new state event, `DeploymentAborted`, which is posted and checked whenever we send status updates to the server. Each state that sends status now explicitly checks if the deployment is aborted and handles it accordingly. Previously it was only handled in the commit state, which meant that if you aborted a deployment during e.g. download, the device could still reboot.

### DIFF
--- a/src/mender-update/daemon/state_machine/state_machine.cpp
+++ b/src/mender-update/daemon/state_machine/state_machine.cpp
@@ -139,6 +139,7 @@ StateMachine::StateMachine(Context &ctx, events::EventLoop &event_loop) :
 
 	// Cannot fail due to FailureMode::Ignore.
 	main_states_.AddTransition(send_download_status_state_,             se::Success,                     ss.download_enter_,                      tf::Immediate);
+	main_states_.AddTransition(send_download_status_state_,             se::DeploymentAborted,           ss.download_leave_,                      tf::Immediate);
 
 	main_states_.AddTransition(ss.download_enter_,                      se::Success,                     update_download_state_,                  tf::Immediate);
 	main_states_.AddTransition(ss.download_enter_,                      se::Failure,                     ss.download_error_,                      tf::Immediate);
@@ -165,6 +166,7 @@ StateMachine::StateMachine(Context &ctx, events::EventLoop &event_loop) :
 
 	// Cannot fail due to FailureMode::Ignore.
 	main_states_.AddTransition(send_install_status_state_,              se::Success,                     ss.install_enter_,                       tf::Immediate);
+	main_states_.AddTransition(send_install_status_state_,              se::DeploymentAborted,           clear_artifact_data_state_,              tf::Immediate);
 
 	main_states_.AddTransition(update_install_state_,                   se::Success,                     ss.install_leave_,                       tf::Immediate);
 	main_states_.AddTransition(update_install_state_,                   se::Failure,                     ss.install_error_rollback_,              tf::Immediate);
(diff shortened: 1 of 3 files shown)


danielskinstad:
@mender-test-bot start client pipeline --pr integration/2782

src/mender-update/daemon/state_machine/state_machine.cpp
```diff
@@ -139,6 +139,7 @@ StateMachine::StateMachine(Context &ctx, events::EventLoop &event_loop) :
 
 	// Cannot fail due to FailureMode::Ignore.
```
lluiscampos: We need to update these comments. The important bit is that only on specific abort error the deployment stops, otherwise is ignored.

src/mender-update/daemon/state_machine/state_machine.cpp
```diff
@@ -210,6 +213,7 @@ StateMachine::StateMachine(Context &ctx, events::EventLoop &event_loop) :
 
 	main_states_.AddTransition(send_commit_status_state_,               se::Success,                     ss.commit_enter_,                        tf::Immediate);
 	main_states_.AddTransition(send_commit_status_state_,               se::Failure,                     update_check_rollback_state_,            tf::Immediate);
+	main_states_.AddTransition(send_commit_status_state_,               se::DeploymentAborted,           update_check_rollback_state_,            tf::Immediate);
```
lluiscampos, line 217: And from here on, I would add a comment with the opposite meaning. Once we are committing and after we have commited the generic failure and the abort are treated the same way.

tests/src/mender-update/daemon/state_test.cpp
```diff
@@ -2895,13 +2930,13 @@ vector<StateTransitionsTestCase> GenerateStateTransitionsTestCases() {
...
+					// "rebooting", // Missing because of fail_status_report_status below
 					"failure",
 				},
 			.install_outcome = InstallOutcome::SuccessfulRollback,
 			.fail_status_report_count = 100,
-			.fail_status_report_status = deployments::DeploymentStatus::Installing,
+			.fail_status_report_status = deployments::DeploymentStatus::Rebooting,
 			.fail_status_aborted = true,
 			// When aborting an update, it should react immediately.
 			.long_retry_times = true,
```
lluiscampos, line 2942: What about test `Aborted_update_before_commit` ? I know that you have not changed the logic on this but I think is good to add the test.

lluiscampos (commented):
Good bug fix. Nicely done.

Your changelog entry is quite comprehensive. I would however swap the sentences so that the first that the user reads is the high level description and then the low level detail. Example:

Fix an issue in which when a user aborted a deployment during  e.g. download,
the device could still reboot and install the update to then rollback. It is fixed by ading a new state event, `DeploymentAborted`, which...

(also note that I changed "you" by "user" 😆 )

See comments below too.

tests/src/mender-update/daemon/state_test.cpp
danielskinstad: I wanted to create that, but couldn't find an easy way to do it here (please let me know if you of a way here). You can't just do `deployments::DeploymentStatus::Committing`

lluiscampos: So there was a good reason for not being there in the first place :sweat_smile: I'll have a look but most likely I won't find a solution neither.

lluiscampos (approved):

PR #1821, opened 2025-08-28
fix: add timeout to prevent hanging downloads on connection loss

danielskinstad:
If network connection is lost while downloading an artifact, the client
can hang when calling `beast::http::async_read_some` in `AsyncReadNextBodyPart`.
Implement an AsyncWait timer that triggers an error handler after 5 minutes.

Ticket: MEN-8717
Changelog: Fix an issue where mender-update hangs when network connection is
lost during artifact download. Implement a 5-minute timeout in `AsyncReadNextBodyPart`
to allow the HTTP resumer to resume the download.

### DIFF
--- a/src/common/http/platform/beast/http.cpp
+++ b/src/common/http/platform/beast/http.cpp
@@ -293,7 +293,8 @@ Client::Client(
 	no_proxy_ {client.no_proxy},
 	cancelled_ {make_shared<bool>(true)},
 	resolver_(GetAsioIoContext(event_loop)),
-	body_buffer_(HTTP_BEAST_BUFFER_SIZE) {
+	body_buffer_(HTTP_BEAST_BUFFER_SIZE),
+	read_timeout_timer_(event_loop) {
 }
 
 Client::~Client() {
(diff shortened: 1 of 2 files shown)


lluiscampos (approved):
From what I know about the HTTP code, the fix makes to me.

As discussed privately, I would like to manually test that the  ~`StateScriptTimeoutSeconds`~ `ModuleTimeoutSeconds` timeout, which is configurable, is still able to cancel the whole download even after the new 5 minutes timer has kicked in couple of times.

Pre-approving.

Good commit message by the way!

src/common/http/platform/beast/http.cpp
```diff
@@ -1168,9 +1168,8 @@ void Client::AsyncReadNextBodyPart(
 	response_data_.last_buffer_size_ = smallest;
 
 	auto &cancelled = cancelled_;
-	auto &response_data = response_data_;
 
-	auto async_handler = [this, cancelled, response_data](const error_code &ec, size_t num_read) {
+	auto async_handler = [this, cancelled](const error_code &ec, size_t num_read) {
```
vpodzime: I hope this was not to make sure `response_data` lives for long enough. I'm sure I was looking at exactly this before but didn't dare to touch it. :grin:

src/common/http/platform/beast/http.cpp
```diff
@@ -1169,8 +1170,24 @@ void Client::AsyncReadNextBodyPart(
...
+	// Add a timeout timer to ensure we don't get stuck if we lose connection
+	read_timeout_timer_.AsyncWait(chrono::minutes(5), [this, cancelled](error::Error err) {
+		if (!*cancelled) {
+			if (err != error::NoError) {
+				if (err.code != make_error_condition(errc::operation_canceled)) {
+					log::Error("Read timeout timer caused error: " + err.String());
+					CallErrorHandler(err, request_, body_handler_);
+				}
+			} else {
+				CallErrorHandler(
```
vpodzime, line 1183: Shouldn't we do `DoCancel()` here first? Otherwise, what happens with the read when it later times out (errors) itself?

vpodzime (commented):
Looks good to me otherwise.

src/common/http/platform/beast/http.cpp
danielskinstad: Shouldn't that be covered here? https://github.com/mendersoftware/mender/blob/636953feb2bbf383703eee880b5e3a5fdd765e6f/src/common/http/platform/beast/http.cpp#L633 CallErrorHandler calls DoCancel

src/common/http/platform/beast/http.cpp
danielskinstad: https://github.com/mendersoftware/mender/blob/636953feb2bbf383703eee880b5e3a5fdd765e6f/src/common/http.hpp#L533

Looks like you're right. Perhaps we shouldn't mess with this :sweat_smile:

src/common/http/platform/beast/http.cpp
vpodzime: Oh! Somehow I'd expect the exact opposite...

vpodzime (approved):
Looks good to me.

PR #1826, opened 2025-09-22
Send system type for System devices

danielskinstad:


tests/src/common/CMakeLists.txt
```diff
@@ -28,8 +28,6 @@ add_dependencies(tests json_test)
 add_executable(yaml_test EXCLUDE_FROM_ALL yaml_test.cpp)
 target_link_libraries(yaml_test PUBLIC common_yaml main_test gmock)
 
-# NOTE: MENDER_USE_YAML_CPP=OFF by default. Therefore, in order to test it, we
-# need to do a separate build, with it turned ON.
 add_test(
```
lluiscampos: All this `add_test` as an extra CMake command can be replaced with the usual:
```
gtest_discover_tests(yaml_test ${MENDER_TEST_FLAGS} NO_PRETTY_VALUES)
add_dependencies(tests yaml_test)
```

src/mender-update/context.hpp
```diff
@@ -97,6 +101,13 @@ class MenderContext {
 	// Suffix used for updates that either can't roll back or fail their rollback.
 	static const string broken_artifact_name_suffix;
 
+	// The payload type of mender-orchestrator manifests.
+	// Used to identify artifacts containing a mender-orchestrator manifest.
+	// If a DeviceTier is set to system and the artifact payload type equals
+	// this, we should use the system_type from the topology rather than the
+	// device_type.
```
lluiscampos: This comment is incomplete, the comment from the code at `MenderContext::GetCompatibleType` is much better. To be honest I don't think we need to elaborate on the logic here, I would just write:

```
The payload type of mender-orchestrator manifests
Used to identify artifacts containing a mender-orchestrator manifest and handle accordingly.
See GetCompatibleType()
```

src/mender-update/context/context.cpp
```diff
@@ -43,8 +47,10 @@ namespace json = mender::common::json;
 namespace kv_db = mender::common::key_value_database;
 namespace log = mender::common::log;
 namespace path = mender::common::path;
+namespace yaml = mender::common::yaml;
```
lluiscampos, line 52:
```
#ifdef MENDER_USE_YAML_CPP
namespace yaml = mender::common::yaml
#endif // MENDER_USE_YAML_CPP
```

src/mender-update/context/context.cpp
```diff
@@ -235,6 +275,33 @@ expected::ExpectedString MenderContext::GetDeviceType() {
...
+	if (config_.device_tier == "system") {
+#ifdef MENDER_USE_YAML_CPP
+		if (payload_type == "") {
+			// Deployment polling -> use system_type
+			return GetSystemType();
+		} else if (payload_type == orchestrator_manifest_payload_type) {
+			// Manifest installtion -> use system_type
+			return GetSystemType();
+		} else {
+			// Regular artifact processing (called by orchestrator) -> use device_type
```
lluiscampos: Yes, it is called by orchestartor, but the most important part is that this is being updated in standalone mode, not really who calls it. I think this details would be good to capture in the commit message too.

src/mender-update/context/context.cpp
```diff
@@ -400,20 +467,24 @@ error::Error MenderContext::CommitArtifactData(
...
 
 expected::ExpectedBool MenderContext::MatchesArtifactDepends(const artifact::HeaderView &hdr_view) {
-	auto ex_dev_type = GetDeviceType();
-	if (!ex_dev_type) {
-		return expected::unexpected(ex_dev_type.error());
+	auto ex_compatible_type = GetCompatibleType(hdr_view.type_info.type);
+	if (!ex_compatible_type) {
+		return expected::unexpected(ex_compatible_type.error());
 	}
+	string compatible_type = ex_compatible_type.value();
```
lluiscampos: `string &compatible_type` or `auto &compatible_type`

tests/src/client_shared/config_parser_test.cpp
```diff
@@ -35,6 +35,7 @@ const string complete_config = R"({
   "UpdateLogPath": "UpdateLogPath_value",
   "TenantToken": "TenantToken_value",
   "DaemonLogLevel": "DaemonLogLevel_value",
+  "DeviceTier": "DeviceTier_value",
```
lluiscampos, line 38: Nit pick: the tests from this file belong to the first commit.

tests/src/mender-update/context_test.cpp
```diff
@@ -830,3 +830,146 @@ TEST(ContextArtifactTests, ArtifactMatchesContextErrorsTest) {
...
+	ASSERT_TRUE(ex_compatible_type) << ex_compatible_type.error().String();
+	EXPECT_EQ(*ex_compatible_type, "system_device_type");
+
+	// Test regular artifact processing - should use device_type
+	ex_compatible_type = ctx.GetCompatibleType("some-regular-payload");
+	ASSERT_TRUE(ex_compatible_type) << ex_compatible_type.error().String();
+	EXPECT_EQ(*ex_compatible_type, "regular_device_type");
+
+	unsetenv("MENDER_ORCHESTRATOR_TOPOLOGY_DIR");
+}
```
lluiscampos, line 939: Nice test

src/mender-update/context/context.cpp
danielskinstad: Nice catch

PR #1828, opened 2025-09-25
feat: send tier in authentication request

danielskinstad:
Ticket: MEN-8636
Changelog: Add device tier support to authentication requests. The client now sends a 'tier' parameter in authentication requests, supporting "standard" (default), "micro", and "system" tiers. The tier is configurable via the DeviceTier configuration option, with "standard" as the default.

### DIFF
--- a/src/client_shared/config_parser/config_parser.cpp
+++ b/src/client_shared/config_parser/config_parser.cpp
@@ -31,6 +31,7 @@ using namespace std;
 namespace expected = mender::common::expected;
 namespace json = mender::common::json;
 namespace log = mender::common::log;
+namespace device_tier = mender::common::device_tier;
 
 const ConfigParserErrorCategoryClass ConfigParserErrorCategory;
 
@@ -109,7 +110,13 @@ ExpectedBool MenderConfigFromFile::LoadFile(const string &path) {
 		const json::Json value_json = e_cfg_value.value();
 		const json::ExpectedString e_cfg_string = value_json.GetString();
 		if (e_cfg_string) {
-			this->device_tier = e_cfg_string.value();
+			const string &tier = e_cfg_string.value();
+			if (!device_tier::IsValid(tier)) {
+				auto err = MakeError(
+					ConfigParserErrorCode::ValidationError, "Invalid DeviceTier: " + tier);
+				return expected::unexpected(err);
+			}
+			this->device_tier = tier;
 			applied = true;
 		}
 	}
--- a/src/common/device_tier.hpp
+++ b/src/common/device_tier.hpp
@@ -1,0 +1,34 @@
+// Copyright 2025 Northern.tech AS
+//
+//    Licensed under the Apache License, Version 2.0 (the "License");
+//    you may not use this file except in compliance with the License.
+//    You may obtain a copy of the License at
+//
+//        http://www.apache.org/licenses/LICENSE-2.0
+//
+//    Unless required by applicable law or agreed to in writing, software
+//    distributed under the License is distributed on an "AS IS" BASIS,
+//    WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+//    See the License for the specific language governing permissions and
+//    limitations under the License.
+
+#ifndef MENDER_COMMON_DEVICE_TIER_HPP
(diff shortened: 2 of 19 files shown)


src/client_shared/config_parser/config_parser.cpp
```diff
@@ -109,7 +110,13 @@ ExpectedBool MenderConfigFromFile::LoadFile(const string &path) {
 		const json::Json value_json = e_cfg_value.value();
 		const json::ExpectedString e_cfg_string = value_json.GetString();
 		if (e_cfg_string) {
-			this->device_tier = e_cfg_string.value();
+			string tier = e_cfg_string.value();
```
lluiscampos:
```suggestion
			string &tier = e_cfg_string.value();
```

☝️ I am not sure, can it be a reference? We are copying it later in `this->device_tier = ...`

src/common/device_tier.hpp
```diff
@@ -0,0 +1,36 @@
...
+
+#include <string>
+
+namespace mender {
+namespace common {
+namespace device_tier {
+
+static const std::string standard = "standard";
+static const std::string system = "system";
+static const std::string micro = "micro";
```
lluiscampos: I think this is much better with a `enum class`. It does complicate a bit the code wrt string conversion, but then the further manipulations are safer. If we go this way we would rework `IsValid` and have another method `DeviceTierToString` or similar.

tests/src/mender-auth/api/auth_test.cpp
```diff
@@ -255,3 +259,202 @@ TEST_F(AuthTests, FetchJWTTokenFailTest) {
 
 	ASSERT_EQ(err, error::NoError) << "Unexpected error: " << err.message;
 }
+
+TEST_F(AuthTests, FetchJWTTokenWithMicroTier) {
+	const string JWT_TOKEN = "FOOBARJWTTOKEN";
```
lluiscampos, line 264: Why UPPERCASE? 😆 
```suggestion
	const string jwt_token = "FOOBARJWTTOKEN";
```

tests/src/mender-auth/api/auth_test.cpp
```diff
@@ -255,3 +259,202 @@ TEST_F(AuthTests, FetchJWTTokenFailTest) {
...
+	ASSERT_TRUE(expected_json) << "Failed to parse request body as JSON: "
+							   << expected_json.error().String();
+	auto request_json = expected_json.value();
+
+	auto tier_result = request_json.Get("tier");
+	ASSERT_TRUE(tier_result) << "tier field not found in request";
+	auto tier_string = tier_result.value().GetString();
+	ASSERT_TRUE(tier_string) << "tier field is not a string";
+	EXPECT_EQ(tier_string.value(), device_tier::standard);
+}
```
lluiscampos, line 460: These tests are good. I wonder if we could have parametrized them using `TEST_P` instead.

src/common/device_tier.hpp
danielskinstad: I also experimented a bit with that, but I thought it was over-complicating it. I have no strong opinions here though, so we can go this way

tests/src/mender-auth/api/auth_test.cpp
danielskinstad: I copy-pasted much of it from the others tests, which are also UPPERCASE :sweat_smile:

src/common/device_tier.hpp
lluiscampos: Let's see if @michalkopczan has an opinion :innocent:

vpodzime: I like the strings. But they should only be **declared** here and **defined** in a `.cpp` file. And the names of the variables should start with `k`. See the [update state names](https://github.com/mendersoftware/mender/blob/master/src/mender-update/daemon/context.hpp#L197-L208) for reference.

danielskinstad: Thanks for the feedback. 
Then, unless there are any objections, I'll go with the simple route: prefix them with a k and move the the assignments out of the header

src/client_shared/config_parser/config_parser.cpp
danielskinstad: Indeed, we should be able to do `const string &tier` :+1:

PR #1830, opened 2025-09-30
MEN-8818: Make Update modules generators POSIX compatible

lluiscampos:
By removing the unnecessary `-n` flag. According to POSIX specification "Implementations shall not support any options". Ref:
* https://pubs.opengroup.org/onlinepubs/9699919799/utilities/echo.html


danielskinstad (approved):
Thanks for taking care of this one :beer: 
The commit requires a changelog, but otherwise it looks good to me :rocket:

PR #1911, opened 2026-03-10
fix: Don't retry status update on 413

danielskinstad:
Changelog: Added an explicit check for 413 Request Body Too Large errors when sending deployment logs in order to not go into an unnecessary retry loop when the deployment logs are too big.
Ticket: ME-616

### DIFF
--- a/src/mender-update/daemon/states.cpp
+++ b/src/mender-update/daemon/states.cpp
@@ -700,6 +700,12 @@ void SendStatusUpdateState::DoStatusUpdateHandler(
 			// failure, even if retry is enabled.
 			poster.PostEvent(StateEvent::DeploymentAborted);
 			return;
+		} else if (
+			error.error.code
+			== deployments::MakeError(deployments::RequestBodyTooLargeError, "").code) {
+			// There is no need to retry if the request body is too large
+			poster.PostEvent(StateEvent::Failure);
+			return;
 		}
 
 		switch (mode_) {
--- a/src/mender-update/deployments/deployments.cpp
+++ b/src/mender-update/deployments/deployments.cpp
@@ -68,6 +68,8 @@ string DeploymentsErrorCategoryClass::message(int code) const {
 		return "Deployment was aborted on the server";
 	case TooManyRequestsError:
 		return "Too many requests";
+	case RequestBodyTooLargeError:
+		return "Request body too large";
 	}
 	assert(false);
 	return "Unknown";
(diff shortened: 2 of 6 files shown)


src/mender-update/deployments/deployments.cpp
```diff
@@ -664,6 +666,11 @@ error::Error DeploymentClient::PushLogs(
 
 			if (status == http::StatusNoContent) {
 				api_handler(LogsAPIResponse {status, nullopt, error::NoError});
+			} else if (status == http::StatusRequestBodyTooLarge) {
```
vpodzime: What does this status cause in the header handler?

vpodzime (commented):
Looks good to me otherwise, but we should double-check.

src/mender-update/deployments/deployments.cpp
danielskinstad: AFAICT it shouldn't cause anything special. Anything in particular you're thinking about?

michalkopczan: Similar question: are we sure we get a body from the server in HTTP413? Maybe it should be handled in the header? I did it in HTTP429.
In HTTP429 case, the hM server sends a body actually, but it is not required based on HTTP docs, that's why I added it in the header handler. We should check if here we may have a similar situation.

src/mender-update/daemon/states.cpp
```diff
@@ -700,6 +700,12 @@ void SendStatusUpdateState::DoStatusUpdateHandler(
 			// failure, even if retry is enabled.
 			poster.PostEvent(StateEvent::DeploymentAborted);
 			return;
+		} else if (
+			error.error.code
+			== deployments::MakeError(deployments::RequestBodyTooLargeError, "").code) {
+			// There is no need to retry if the request body is too large
+			poster.PostEvent(StateEvent::Failure);
```
michalkopczan, line 707: What will happen here in the state machine? We can get here while being in multiple different states, right?

vpodzime: Yes, but deployment logs are only uploaded on failure so we are on the "failure path" here already.

src/mender-update/deployments/deployments.cpp
danielskinstad: I couldn't find anything for certain in api docs, but from what I can tell from the mender-server sources and from my manual testing is that it does return a body. However, I am sure we can handle it in the header handler if you'd prefer that.

michalkopczan: Hm, on one hand it doesn't seem necessary. On the other, as we already handle 429 and (i believe) 401 in the header handler, it would be nice to have some consistency - as our http code is hard to understand already.

@vpodzime what do you think?

danielskinstad: I pretty much have it implemented, so unless any of you think it should stay as-is, I'll force-push a version where I handle it in the header handler to be consistent with 429

michalkopczan: In that case, I vote for header handling.

danielskinstad:

michalkopczan (approved):

src/mender-update/deployments/deployments.cpp
vpodzime: Same here.

vpodzime (approved):
Looks good to me. Thanks for the adjustments!

danielskinstad:
* 5.0.x
* 5.1.x

PR #1914, opened 2026-03-12
fix: Handle first timestamp in deployment logs more carefully

vpodzime:
Depending on the system and build configuration, the timestamps in logs can use lower (likely) or higher (unlikely) time resolution than expected (nanoseconds). So in case of a deployment failure with corrupted logs, when using the first timestamp as a replacement in the extra `(THE ORIGINAL LOGS CONTAINED INVALID ENTRIES)` log entry, care must be taken to make sure the extra log entry still has a valid timestamp and that the result is valid JSON.

Ticket: MEN-9427
Changelog: commit

Steps:
- [x] code fix
- [x] new unit tests
- [x] new integration test: https://github.com/mendersoftware/integration/pull/2851

### DIFF
--- a/src/mender-update/deployments/deployments.cpp
+++ b/src/mender-update/deployments/deployments.cpp
@@ -534,7 +534,30 @@ error::Error JsonLogMessagesReader::SanitizeLogs() {
 		if (!clean_logs_) {
 			auto bad_data_msg_tstamp_start =
 				bad_data_msg_.begin() + 15; // len(R"({"timestamp": ")")
+
+			// The actual timestamp from logs can potential have a different
+			// (likely lower) time resolution and thus length than our default.
+			const auto first_tstamp_size = first_tstamp.size();
+			if (first_tstamp_size > default_tstamp_.size()) {
+				// In case the time resolution is higher and the timestamp
+				// longer (unlikely to happen)
+				if (first_tstamp[first_tstamp_size - 1] == 'Z') {
+					first_tstamp[default_tstamp_.size() - 1] = 'Z';
+				}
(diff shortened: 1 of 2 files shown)


vpodzime:
@mender-test-bot start client pipeline --pr mender/1914 --pr integration/2851

src/mender-update/deployments/deployments.cpp
```diff
@@ -534,7 +534,30 @@ error::Error JsonLogMessagesReader::SanitizeLogs() {
 		if (!clean_logs_) {
 			auto bad_data_msg_tstamp_start =
 				bad_data_msg_.begin() + 15; // len(R"({"timestamp": ")")
+
+			// The actual timestamp from logs can potential have a different
+			// (likely lower) time resolution and thus length than our default.
+			const auto first_tstamp_size = first_tstamp.size();
```
michalkopczan, line 540: Does it take care of the situation where the first log message is corrupted, so we replace it with our timestamp? If we have the first timestamp created by us, it may have a different length than the other ones in the file (or does it not matter?)

michalkopczan (approved):

src/mender-update/deployments/deployments.cpp
vpodzime: > or does it not matter?

Exactly. It doesn't matter as long as all timestamps are valid (i.e. using the expected format).

vpodzime: In fact, this is exactly what happens in the setup our integration tests (incl. the new one) run.

vpodzime:
@mender-test-bot  cherry-pick to:
 * 5.0.x
 * 5.1.x

PR #1916, opened 2026-03-26
MEN-9415: Trim large deployment logs

vpodzime:


src/mender-update/deployments/deployments.cpp
```diff
@@ -420,12 +422,65 @@ static ExpectedSize GetLogFileDataSize(const string &path) {
...
+		// make sure we don't work with some stale data from the previous read()
+		buf.resize(is.gcount());
+
+		// XXX: This is not perfect as it can fail to find the next boundary if
+		//      it's split between the end of the current buffer contents and
+		//      the start of the next chunk read from the file. However, with a
+		//      1K buffer, this is very unlikely to happen and even if it does
+		//      happen, the logs will just be trimmed a bit more than
+		//      necessary. For the sake of incomparably simpler code that is
+		//      easy to test.
```
michalkopczan, line 454: makes sense

src/mender-update/deployments.hpp
```diff
@@ -202,11 +204,16 @@ class JsonLogMessagesReader : virtual public io::Reader {
...
 	static const string bad_data_msg_tmpl_;
+	static const string too_much_data_msg_tmpl_;
 	io::Vsize header_rem_ = header_.size();
 	io::Vsize closing_rem_ = closing_.size();
 	vector<uint8_t> bad_data_msg_;
 	io::Vsize bad_data_msg_rem_;
+	vector<uint8_t> too_much_data_msg_;
+	io::Vsize too_much_data_msg_rem_;
 	bool clean_logs_;
+	const int32_t maximum_log_size_ = 1000 * 1000 - 256; // 1MB - some room for request headers etc.
```
michalkopczan, line 215: This is actually configurable in the server. I don't know how to use it here really. Maybe a config option would be good. Seems like a low priority however, but consider creating a follow up task.

tests/src/mender-update/deployments_test.cpp
```diff
@@ -1439,6 +1439,48 @@ not valid json
...
+
+	string expected_data_start =
+		R"d({"messages":[{"timestamp": "2016-03-11T13:03:17.063493443Z", "level": "WARNING", "message": "(THE ORIGINAL LOGS WERE TOO BIG, THIS LOG IS TRUNCATED. The full log can be found on the device)"},{"timestamp": "2016-03-11T13:05:17.063493443Z", "level": "INFO", "message": "excessive log data"})d";
+	string expected_data_end =
+		R"d({"timestamp": "2016-03-11T13:05:17.063493443Z", "level": "INFO", "message": "excessive log data"}]})d";
+
+	deps::JsonLogMessagesReader logs_reader {test_log_file_path};
+	EXPECT_EQ(logs_reader.SanitizeLogs(), error::NoError);
+	auto total_size = logs_reader.TotalDataSize();
+	EXPECT_LT(total_size, 1000 * 1000 - 256);
```
michalkopczan: Suggestion: use the const from JsonLogMsgReader. You can make the test class a friend of the source class, so you'll have access to private data there.
See e.g. mender/src/common/http.hpp "friend class ::InventoryAPITests; for reference

michalkopczan (approved):

src/mender-update/deployments.hpp
vpodzime: Yeah, I was considering that. Let's see how bad it would be if we actually did pass this all the way from the config. If I was running an on-prem Mender deployment, I'd probably like to be able to configure this. But I'd do it as a separate task.

vpodzime:
@mender-test-bot start client pipeline --pr mender/1916 --pr integration/2853

vpodzime:
> ci/gitlabExpected — Waiting for status to be reported

Not sure, what's going on here. Two pipelines ran successfully and even "succeeded" at reporting the status to GH, but it's not showing up here.

vpodzime:
@mender-test-bot  cherry-pick to:
 * 5.0.x
 * 6.0.x

vpodzime:
* 5.1.x

PR #1919, opened 2026-03-27
fix: Fail when device tier is invalid instead of warning

elkoniu:
Changelog: Fail when device tier has invalid value
Ticket: ME-636

### DIFF
--- a/src/client_shared/conf/conf.cpp
+++ b/src/client_shared/conf/conf.cpp
@@ -311,8 +311,8 @@ error::Error MenderConfig::LoadConfigFile_(const string &path, bool required) {
 			return error::NoError;
 		} else {
 			// other errors (parsing errors,...) for default paths should produce warnings
-			log::Warning("Failed to load config from '" + path + "': " + ret.error().message);
-			return error::NoError;
+			log::Error("Failed to load config from '" + path + "': " + ret.error().message);
+			return ret.error();
 		}
 	}
 	// else
--- a/src/client_shared/conf/conf.cpp
+++ b/src/client_shared/conf/conf.cpp
@@ -34,6 +34,7 @@ namespace error = mender::common::error;
 namespace expected = mender::common::expected;
 namespace log = mender::common::log;
 namespace json = mender::common::json;
+namespace config_parser = mender::client_shared::config_parser;
 
 const string kMenderVersion = MENDER_VERSION;
 
@@ -310,6 +311,13 @@ error::Error MenderConfig::LoadConfigFile_(const string &path, bool required) {
 			log::Debug("Failed to load config from '" + path + "': " + ret.error().message);
 			return error::NoError;
 		} else {
+			// Incorrect DeviceTier shall lead to the daemon failure
+			auto device_tier_error = config_parser::MakeError(
+				config_parser::ConfigParserErrorCode::DeviceTierError, "Invalid DeviceTier");
+			if (ret.error().code == device_tier_error.code) {
+				log::Error("Failed to get DeviceTier from '" + path + "': " + ret.error().message);
+				return ret.error();
+			}
 			// other errors (parsing errors,...) for default paths should produce warnings
 			log::Warning("Failed to load config from '" + path + "': " + ret.error().message);
 			return error::NoError;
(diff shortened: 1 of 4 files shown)


danielskinstad:
I'll let Michal review the actual changes, but remember that we've migrated to the new commit linter, so you shouldn't prefix your changelog with the `Changelog:` footer unless you do `Changelog: Title`. Our git-cliff config will ignore random footers, meaning it won't end up in the final changelog.

elkoniu:
@danielskinstad thanks for pointing out this new commit message requirements. I have aligned description and removed `Changelog:` tag.

elkoniu:

michalkopczan:
I know that the app crashes now. With the changes, will it exit with this error instead?

elkoniu:
@michalkopczan we are checking output of this function by:
```
if (error::NoError != err) {
```
Now when explicite error is returned it should fail all the way down.
Just to double check I will run it again and copy paste logs here.

elkoniu:
@michalkopczan just to confirm, before change:
```
May 06 00:36:46 raspberrypi systemd[1]: Starting mender-authd.service - Mender authentication service...
May 06 00:36:46 raspberrypi mender-auth[57672]: record_id=1 severity=warning time="2026-May-06 00:36:46.335096" name="Global" msg="Failed to load config from '/etc/mender/mender.conf': Invalid DeviceTier: wrong"
```
and after the change:
```
May 06 01:34:26 raspberrypi systemd[1]: Starting mender-authd.service - Mender authentication service...
May 06 01:34:26 raspberrypi mender-auth[24307]: record_id=1 severity=error time="2026-May-06 01:34:26.822789" name="Global" msg="Failed to load config from '/etc/mender/mender.conf': Invalid DeviceTier: wrong"
```
Also after the change I cant see `Cannot run dameon mode without ServerURL` so the definition of done criteria are meet for this.

src/client_shared/conf/conf.cpp
```diff
@@ -311,8 +311,8 @@ error::Error MenderConfig::LoadConfigFile_(const string &path, bool required) {
 			return error::NoError;
 		} else {
 			// other errors (parsing errors,...) for default paths should produce warnings
-			log::Warning("Failed to load config from '" + path + "': " + ret.error().message);
-			return error::NoError;
+			log::Error("Failed to load config from '" + path + "': " + ret.error().message);
+			return ret.error();
```
danielskinstad: See the comment above this:
> // other errors (parsing errors,...) for default paths should produce warnings

IOW, this will affect all parsing errors for "default paths", not only DeviceTier. This might have unknown consequences, so I don't think this is the way to go.

Perhaps return an invalid device tier error code, check it here and only then return ret.error()

elkoniu: Thank you for pointing this out, I was not aware that there is another option like injecting config via env vars etc. Will go over the code again to narrow this only to the `DeviceTier` case.

src/mender-auth/cli/actions.cpp
```diff
@@ -141,7 +141,7 @@ error::Error DaemonAction::Execute(context::MenderContext &main_context) {
 	if (none_of(config.servers.cbegin(), config.servers.cend(), [](const string &it) {
 			return it != "";
 		})) {
-		log::Error("Cannot run in daemon mode with no server URL specified");
+		log::Error("Server URL not configured");
```
danielskinstad: This is unrelated, right? AFAICT, the previous statement still holds

danielskinstad (commented):
There are some compilation errors, see CI

src/mender-auth/cli/actions.cpp
elkoniu: True, I was trying to fix this message as pointed in the ticket but when we exit due to DeviceTier error it shall not be displayed anyway.

elkoniu:

michalkopczan:
You have unit test failing, besides that LGTM. Let me know when I should take a look on the fixed stuff

elkoniu:

danielskinstad (approved):

michalkopczan (approved):

michalkopczan:
Force-merging as:
- the tests cover the functionality actually, just not in the unit directly (they test it in the caller function)
- mender-qa is a known issue with `test_update_with_autogenerated_delta`, not related to this PR

elkoniu:
* 5.0.x
* 5.1.x

PR #1930, opened 2026-04-07
chore: Enable PoC for codecov coverage reporting

michalkopczan:
Ticket: QA-1128

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -231,3 +231,13 @@ publish:tests:
       --repo-token ${COVERALLS_TOKEN}
       --no-gcov
       --lcov-file coverage.lcov
+
+publish:tests:codecov:
+  stage: publish
+  image: alpine
+  before_script:
+    - apk add --no-cache bash curl findutils git
+  dependencies:
+    - test:unit
+  script:
+    - curl -s https://codecov.io/bash | bash -s -- -Z
--- a/codecov.yml
+++ b/codecov.yml
@@ -1,22 +1,15 @@
 
 comment:
-  # Only comment after 7 reports uploaded
-  after_n_builds: 7
+  # We have one publish:tests job, so all data is reported to codecov
+  # in one report. We should add a comment in the PR immediately.
+  after_n_builds: 1
 
 codecov:
   notify:
-    # Only notify after 7 reports uploaded
-    after_n_builds: 7
-
-coverage:
-  status:
-    project:
-      acceptance:
-        flags:
-          - acceptance
-      unit:
-        flags:
-          - unit
+  # We have one publish:tests job, so all data is reported to codecov
+  # in one report. We should update the codecov status (meaning whether code
+  # coverage requirements for merging the PR are satisfied) in the PR immediately.
+    after_n_builds: 1
 
 ignore:
   - "vendor/"


michalkopczan:

michalkopczan:

codecov.yml
```diff
@@ -1,19 +1,18 @@
 
 comment:
-  # Only comment after 7 reports uploaded
-  after_n_builds: 7
+  # We have one publish:tests job, so all data is reported to codecov
+  # in one report. We should crate a comment in the PR immediately.
```
vpodzime: *crate* -> *create* -> *add*

or say the same as the other comment bellow

.gitlab-ci.yml
```diff
@@ -231,3 +231,13 @@ publish:tests:
...
+
+publish:tests:codecov:
+  stage: publish
+  image: alpine
+  before_script:
+    - apk add --no-cache bash curl findutils git
+  dependencies:
+    - test:unit
+  script:
+    - bash -c "bash <(curl -s https://codecov.io/bash) -F unittests -Z"
```
vpodzime: This could be `curl | bash -s -- -F unittests -Z`, I believe. No need to run `bash` twice.

.gitlab-ci.yml
```diff
@@ -231,3 +231,13 @@ publish:tests:
       --repo-token ${COVERALLS_TOKEN}
       --no-gcov
       --lcov-file coverage.lcov
+
+publish:tests:codecov:
+  stage: publish
+  image: alpine
+  before_script:
+    - apk add --no-cache bash curl findutils git
```
vpodzime, line 239: When we add this to _mendertesting_ or somewhere to be shared, please add a container image for this to _mender-test-containers_ and use that.

codecov.yml
```diff
@@ -1,19 +1,18 @@
...
 
 coverage:
   status:
     project:
-      acceptance:
-        flags:
-          - acceptance
       unit:
         flags:
           - unit
```
vpodzime: Why keeping `unit` in place?

vpodzime (approved):
Looks good to me as a PoC.

danielskinstad (commented):
Perhaps it would make more sense to commit this directly to mendertesting as a gitlab component and just include it here? That way, if anything needs changing, we only need to update the component in mendertesting rather than having to commit stuff here.

.gitlab-ci.yml
michalkopczan: Good point. Will add this to the follow up task, thanks.

codecov.yml
michalkopczan: The second comment, I now realize, was not clear - the "status" there means whether the code coverage requirements for merging a PR are satisfied.

I updated the second comment, and changed this one to "add". Thanks :)

.gitlab-ci.yml
michalkopczan: You're right. I reused it from our old codecov job for golang client, didn't really go into detail. Changed.

elkoniu:
I like @danielskinstad idea of keeping it in mendertesting and including from there.

codecov.yml
michalkopczan: Removed it, you're right, it's not needed.

michalkopczan:

michalkopczan:
@danielskinstad , @pasinskim , @vpodzime regarding moving to test containers.

The container itself can and should be moved - I'll include it in the follow up task. For now I just wanted a PoC.
@danielskinstad , do you suggest moving the job itself (not the container used for it) to mendertesting? This is more complex, as we may have different needs for C/C++ projects and GO projects. But we could create two jobs, this would be the best approach I guess.

I wonder also about codecov.yml - AFAIK, we need it in each repo separately.

PR #1937, opened 2026-04-14
chore: Add an alternative implementation of KeyValueDatabase

vpodzime:
Probably the most trivial one just using a single blob stored in
persistent storage (a file, for now, on POSIX systems) and using
basic synchronization primitives (`flock()` on the file, for
now). The goal is to be able to build the client without LMDB
and, instead, use a more portable alternative still good enough
for our trivial use-case.

Ticket: MEN-9129
Changelog: none
Signed-off-by: Vratislav Podzimek <vratislav.podzimek+auto-signed@northern.tech>

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -86,6 +86,18 @@ test:unit:
   script:
     - cmake -D COVERAGE=ON .
     - make --jobs=$(nproc --all) --keep-going coverage
+    # Clean things up for another round of coverage data gathering
+    - mv coverage.lcov coverage_base.lcov
+    - make clean
+    - find . \( -name '*.gcda' -o -name '*.gcno' \) -exec rm '{}' \;
+    # Now run key-value DB tests without LMDB (thus testing BlobDB)
+    - cmake -D COVERAGE=ON -D MENDER_USE_LMDB=OFF .
+    - make key_value_database_test
+    - tests/src/common/key_value_database_test
+    - make coverage_no_tests
+    - mv coverage.lcov coverage_no_lmdb.lcov
+    # And combine the coverage data
+    - lcov -a 'coverage_*.lcov' -o coverage.lcov --ignore-errors inconsistent
   artifacts:
     expire_in: 2w
     reports:
--- a/src/common/key_value_database/platform/blobdb/file_blob.cpp
+++ b/src/common/key_value_database/platform/blobdb/file_blob.cpp
@@ -1,0 +1,195 @@
+// Copyright 2026 Northern.tech AS
+//
+//    Licensed under the Apache License, Version 2.0 (the "License");
+//    you may not use this file except in compliance with the License.
+//    You may obtain a copy of the License at
+//
+//        http://www.apache.org/licenses/LICENSE-2.0
+//
+//    Unless required by applicable law or agreed to in writing, software
+//    distributed under the License is distributed on an "AS IS" BASIS,
+//    WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+//    See the License for the specific language governing permissions and
+//    limitations under the License.
+
+#include <common/key_value_database_blobdb.hpp>
+#include <common/key_value_database/platform/blobdb/file_blob.hpp>
+
+// We need these two C headers for open() and close(), respectively, because we
+// need to do flock() and there's no C++ API for that.
+#include <fcntl.h>
+#include <unistd.h>
+#include <sys/file.h> // flock()
+
+#include <cerrno>
+#include <cstdio>
+
+#include <common/path.hpp>
+
+namespace mender {
+namespace common {
+namespace key_value_database {
+
+namespace path = common::path;
+
+error::Error FileBlobdbTransaction::SerializeDB(const DB &db) {
+	// There can be no collision because no two transactions can be serializing
+	// the DB at the same time so no random/special name needed here.
+	string new_file_path = path_or_name_ + ".new";
+	auto ex_ofs = io::OpenOfstream(new_file_path);
+	if (!ex_ofs) {
+		return ex_ofs.error().WithContext("Cannot save DB contents");
+	}
+	auto &ofs = ex_ofs.value();
+
+	auto err = error::NoError;
+	for (const auto &kv : db) {
+		if (kv.first.size() > std::numeric_limits<uint32_t>::max()) {
+			err = error::Error(
+				generic_category().default_error_condition(EOVERFLOW),
+				"Key data too long, cannot be serialized");
+			break;
+		}
+		if (kv.second.size() > std::numeric_limits<uint32_t>::max()) {
+			err = error::Error(
+				generic_category().default_error_condition(EOVERFLOW),
+				"Value data too long, cannot be serialized");
+			break;
+		}
+		uint32_t data_size = static_cast<uint32_t>(kv.first.size());
+		ofs.write(reinterpret_cast<const char *>(&data_size), sizeof(data_size));
+		if (!ofs) {
+			err = error::Error(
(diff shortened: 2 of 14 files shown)


vpodzime:
@michalkopczan the code is ready for reviews, but I still need to figure out how to run the unit tests and combine coverage data (the last commit).

src/common/key_value_database/platform/blobdb/file_blob.hpp
```diff
@@ -0,0 +1,44 @@
...
+#ifndef MENDER_COMMON_FILE_BLOBDB_HPP
+#define MENDER_COMMON_FILE_BLOBDB_HPP
+
+#include <common/key_value_database_blobdb.hpp>
+
+namespace mender {
+namespace common {
+namespace key_value_database {
+
+class FileBlobdbTransaction : virtual public BlobdbTransaction {
```
michalkopczan: Why virtual here? Virtual inheritance is used when you have two base classes that each inherit from one class. See https://www.geeksforgeeks.org/cpp/diamond-problem-in-cpp/

here we have FileBlobdbTransaction->BlobdbTransaction -> Transaction. There's no ambiguity, so no virtual inheritance is needed.

vpodzime: Yup, I took it from elsewhere and checked what it meant. And I came to a conclusion that it doesn't hurt :grin:

vpodzime:
> 69.47% of diff hit (target 81.33%)

As can be seen at https://app.codecov.io/gh/mendersoftware/mender/commit/865fcab4fde8b53b92a06a66952ab96fd6815a89, the issue here is that error reporting is not covered by the tests. But this is nearly impossible to do with our tests setup as it would need to simulate I/O errors at specific times.

src/common/key_value_database_blobdb.hpp
```diff
@@ -0,0 +1,79 @@
+// Copyright 2025 Northern.tech AS
```
michalkopczan: copyright 2026

src/common/key_value_database/platform/blobdb/file_blob.hpp
```diff
@@ -0,0 +1,44 @@
...
+
+private:
+	error::Error SerializeDB(const DB &db) override;
+	ExpectedDB DeserializeDB() override;
+	error::Error LockDB() override;
+	error::Error UnlockDB() override;
+
+	int fd_ = -1;
+
+	friend class KeyValueDatabaseBlobdb;
```
michalkopczan: why friend? You're inheriting from it. There should be no need for friend.

src/common/key_value_database/platform/blobdb/file_blob.cpp
```diff
@@ -0,0 +1,183 @@
...
+	for (const auto &kv : db) {
+		if (kv.first.size() > std::numeric_limits<uint32_t>::max()) {
+			err = error::Error(
+				generic_category().default_error_condition(EOVERFLOW),
+				"Key data too long, cannot be serialized");
+			break;
+		}
+		if (kv.second.size() > std::numeric_limits<uint32_t>::max()) {
+			err = error::Error(
+				generic_category().default_error_condition(EOVERFLOW),
```
michalkopczan, line 55: We use vector<uint8_t> for value, so why would the data be too long? If the size of the vector is outside the 32 bit range, it means we can store this much data, right?
Is this because we're using 32 bit value for the "size" field in the database?

.gitlab-ci.yml
```diff
@@ -86,6 +86,18 @@ test:unit:
   script:
     - cmake -D COVERAGE=ON .
     - make --jobs=$(nproc --all) --keep-going coverage
+    # Clean things up for another round of coverage data gathering
```
michalkopczan, line 89: I understand this is an attempt at testing both lmdb and blobdb and it's a WIP, so not reviewing this now.

src/common/key_value_database/platform/blobdb/file_blob.cpp
```diff
@@ -0,0 +1,183 @@
...
+#include <common/path.hpp>
+
+namespace mender {
+namespace common {
+namespace key_value_database {
+
+namespace path = common::path;
+
+error::Error FileBlobdbTransaction::SerializeDB(const DB &db) {
+	auto ex_ofs = io::OpenOfstream(path_or_name_);
```
danielskinstad: Should we write directly to the database or first write to some tmp file to try to ensure we're not left in an inconsistent state? What happens if there's a powerloss in the middle of writing?

src/common/key_value_database/platform/blobdb/file_blob.cpp
```diff
@@ -0,0 +1,183 @@
...
+			"Failed to open DB at '" + path_or_name_ + "'");
+	}
+
+	int ret;
+	if (write_) {
+		ret = flock(fd_, LOCK_EX);
+	} else {
+		ret = flock(fd_, LOCK_SH);
+	}
+	if (ret != 0) {
```
danielskinstad, line 170: We should close `fd_` here, right?

src/common/key_value_database/platform/blobdb/file_blob.hpp
vpodzime: No, this is inheriting from the transaction class and the DB class needs to access the private functions.

src/common/key_value_database/platform/blobdb/file_blob.cpp
vpodzime: Yes, exactly. We need this to be a fixed number of bytes (so that we can easily read it back). I chose 32 as more than enough in all practically possible cases.

src/common/key_value_database/platform/blobdb/file_blob.cpp
vpodzime: Good point, `rename()` should be atomic.

src/common/key_value_database/platform/blobdb/file_blob.cpp
vpodzime: yup! Good catch!

src/common/key_value_database/platform/blobdb/file_blob.hpp
michalkopczan: Please remove - it's not needed, confusing, and can hide possible errors in case of future changes

src/common/key_value_database/platform/blobdb/file_blob.hpp
vpodzime: Less private stuff now and no friends!

src/common/key_value_database/platform/blobdb/file_blob.hpp
```diff
@@ -0,0 +1,42 @@
...
+namespace mender {
+namespace common {
+namespace key_value_database {
+
+class FileBlobdbTransaction : public BlobdbTransaction {
+public:
+	FileBlobdbTransaction(const string &path_or_name, bool write) :
+		BlobdbTransaction(path_or_name, write) {};
+
+	error::Error SerializeDB(const DB &db) override;
```
michalkopczan, line 29: AFAIK, SerializeDB and DeserializeDB are not used from outside, they could have stayed private. If you think that it makes more sense to have them public - good with me.

michalkopczan (approved):

src/common/key_value_database/platform/blobdb/file_blob.hpp
vpodzime: > If you think that it makes more sense to have them public - good with me.

I think none of them should be public, but that requires refactoring. Ideally, the transaction should care of locking and unlocking, this should be transparent for the code using it. I'm leaving this as-is now and we should revisit this code soon.

PR #1940, opened 2026-04-21
support for building with `libc++`

lhoward:
Various @claude-assisted patches for building with libc++ (as opposed to libstdc++).

### DIFF
--- a/src/artifact/token.hpp
+++ b/src/artifact/token.hpp
@@ -50,7 +50,7 @@ enum class Type {
 	Payload,
 };
 
-const unordered_map<const Type, const string> type_map {
+const unordered_map<Type, string> type_map {
 	{Type::Uninitialized, "Uninitialized"},
 	{Type::EOFToken, "EOF"},
 	{Type::Unrecognized, "Unrecognized"},
--- a/src/artifact/v3/header/token.hpp
+++ b/src/artifact/v3/header/token.hpp
@@ -55,7 +55,7 @@ enum class Type {
 	ArtifactHeaderMetaData,
 };
 
-unordered_map<const Type, string> type_map {
+unordered_map<Type, string> type_map {
 	{Type::Uninitialized, "Uninitialized"},
 	{Type::EOFToken, "EOF"},
 	{Type::Unrecognized, "Unrecognized"},
--- a/src/artifact/v3/scripts/executor.cpp
+++ b/src/artifact/v3/scripts/executor.cpp
@@ -41,7 +41,7 @@ namespace path = mender::common::path;
 
 const int state_script_retry_exit_code {21};
 
-unordered_map<const State, string> state_map {
+unordered_map<State, string> state_map {
 	{State::Idle, "Idle"},
 	{State::Sync, "Sync"},
 	{State::Download, "Download"},
@@ -53,7 +53,7 @@ unordered_map<const State, string> state_map {
 	{State::ArtifactFailure, "ArtifactFailure"},
 };
 
-unordered_map<const Action, string> action_map {
+unordered_map<Action, string> action_map {
 	{Action::Enter, "Enter"},
 	{Action::Leave, "Leave"},
 	{Action::Error, "Error"},
--- a/src/common/error.hpp
+++ b/src/common/error.hpp
@@ -13,6 +13,7 @@
 //    limitations under the License.
 
 #include <functional>
+#include <ostream>
 #include <string>
 #include <system_error>
 #include <type_traits>


vpodzime (commented):
@lhoward thank you for this contribution! A couple notes/questions/suggestions:
- What exactly is this `libc++` the commits are mentioning?
- How could we add a CI job making sure we don't break compatibility with it in the future?
- Shouldn't the extra `include` in the second commit actually go to the `error.hpp` file instead? AFAICT, the `std::ostream` reference is there as well and any code working with that header needs the `ostream` declarations.

lhoward:
Just to answer the first question: https://libcxx.llvm.org (we are using as it is not GPL).

Will reply on others when at desk.

lhoward:
Fixed

vpodzime:
@lhoward could you please rebase this PR? We have slightly changed how CI jobs run on _master_ and I want to make sure this works with that setup.

lhoward:
Rebased.

vpodzime:
> Rebased

Great, thanks! One last thing, for a separate PR which I can open myself, I guess -- how could we add this to our CI? Can we easily do some builds using `libc++` in a, say Debian 13, container? We are doing builds using `clang++`, but that doesn't seem to the same as those have been passing without your changes, @lhoward.

vpodzime:

No functional changes here, no need to run a full pipeline.

vpodzime:
And actually, one more thing -- could you please add `Changelog: none` to the commit messages? These fixes do not qualify for changelog entries (but since they are `fix:`, which I agree with, they would/might end up in changelogs, otherwise).

lhoward:
> > Rebased
> 
> Great, thanks! One last thing, for a separate PR which I can open myself, I guess -- how could we add this to our CI? Can we easily do some builds using `libc++` in a, say Debian 13, container? We are doing builds using `clang++`, but that doesn't seem to the same as those have been passing without your changes, @lhoward.

Will look into it.

lhoward:
Dependencies on Ubuntu are `clang libc++-dev libc++abi-dev`.

```bash
#!/usr/bin/env bash
set -euo pipefail

BUILD_DIR="${BUILD_DIR:-build-libcxx}"

# Wipe stale CMake cache: Boost_* and compiler paths get sticky and a prior
# configure that found system Boost will keep using it.
rm -rf "$BUILD_DIR"

# Unset env vars that would let find_package locate a libstdc++-built Boost.
unset BOOST_ROOT BOOST_INCLUDEDIR BOOST_LIBRARYDIR Boost_DIR

cmake -S . -B "$BUILD_DIR" \
  -DCMAKE_C_COMPILER=clang \
  -DCMAKE_CXX_COMPILER=clang++ \
  -DCMAKE_CXX_FLAGS="-stdlib=libc++" \
  -DCMAKE_EXE_LINKER_FLAGS="-stdlib=libc++ -lc++abi" \
  -DCMAKE_SHARED_LINKER_FLAGS="-stdlib=libc++ -lc++abi" \
  -DSTD_FILESYSTEM_LIB_NAME="" \
  -DMENDER_DOWNLOAD_BOOST=ON \
  -DBoost_NO_SYSTEM_PATHS=ON \
  -DBoost_NO_BOOST_CMAKE=ON \
  "$@"

cmake --build "$BUILD_DIR" -j
```

Note: resulting binaries may still show `libstdc++` because of transitive dependencies but as long as C++ types do not cross ABIs boundaries, this is fine. With Yocto, we do not build anything against the GNU C++ library.

vpodzime:
Thanks! Actually, I don't think we need to go that far. My intention was to make sure *our* code compiles (and links) with `libc++` to make sure it's as much compatible and "correct" as possible. We don't have to worry about Boost and other things, that's not our business, really. So I guess this should be enough for us:

```
cmake -S . -B "$BUILD_DIR" \
  -DCMAKE_C_COMPILER=clang \
  -DCMAKE_CXX_COMPILER=clang++ \
  -DCMAKE_CXX_FLAGS="-stdlib=libc++" \
  -DCMAKE_EXE_LINKER_FLAGS="-stdlib=libc++ -lc++abi" \
  -DCMAKE_SHARED_LINKER_FLAGS="-stdlib=libc++ -lc++abi" \
  -DSTD_FILESYSTEM_LIB_NAME=""
```

lhoward:
You'll get errors if you compile against a system Boost built with libstdc++ (I tried). Anything that crosses the C++ ABI boundary and uses stdlib types needs to use the same stdlib.

vpodzime:
I see. Thanks for the details! I'll create a ticket for such CI job inspired by your suggestion above.

PR #1949, opened 2026-05-07
fix: Fail when device tier is invalid

elkoniu:
DeviceTier is one of the configuration options but it is critical for how device is recognized by the server. When we will fail parsing it out of the configuration we should fail instead to switching to the `standard` device tier.

Ticket: ME-636

Changelog: none

(cherry picked from commit c0f2b20)


elkoniu:
This is a backport of https://github.com/mendersoftware/mender/pull/1919

danielskinstad:
This shouldnt be needed. We don't have device tiers in 5.0.x

elkoniu:
Closing following @danielskinstad feedback about lack of device tiers in 5.0.x

PR #1950, opened 2026-05-07
fix: Make the single-file Update Module compatible with QNX

vpodzime:
The QNX's `sync` command accepts no arguments (it's a global sync) and fails if given some. The easiest way to prevent this from happening is to mask it with a function ignoring arguments and calling the `sync` command without them.

Ticket: MEN-9666
Changelog: none


vpodzime:

I can't see how the changes here could affect any of our tests in CI. No reason to run a full pipeline.

danielskinstad (approved):

michalkopczan (approved):

PR #1955, opened 2026-05-12
chore: Add binary-tarball cmake target

danielskinstad:
Example tarball (compiled with` -DMENDER_USE_DBUS=OFF -DMENDER_EMBED_MENDER_AUTH=ON -DMENDER_USE_LMDB=OFF`):
```
mender-ffaf0abc/
mender-ffaf0abc/usr/
mender-ffaf0abc/usr/bin/
mender-ffaf0abc/usr/bin/mender-update
mender-ffaf0abc/usr/share/
mender-ffaf0abc/usr/share/dbus-1/
mender-ffaf0abc/usr/share/dbus-1/system-services/
mender-ffaf0abc/usr/share/dbus-1/system-services/io.mender.AuthenticationManager.service
mender-ffaf0abc/usr/share/dbus-1/system.d/
mender-ffaf0abc/usr/share/dbus-1/system.d/io.mender.AuthenticationManager.conf
mender-ffaf0abc/usr/share/dbus-1/interfaces/
mender-ffaf0abc/usr/share/dbus-1/interfaces/io.mender.Authentication1.xml
mender-ffaf0abc/usr/share/mender/
mender-ffaf0abc/usr/share/mender/modules/
mender-ffaf0abc/usr/share/mender/modules/v3/
mender-ffaf0abc/usr/share/mender/modules/v3/rootfs-image
mender-ffaf0abc/usr/share/mender/modules/v3/single-file
mender-ffaf0abc/usr/share/mender/modules/v3/directory
mender-ffaf0abc/usr/share/mender/inventory/
mender-ffaf0abc/usr/share/mender/inventory/mender-inventory-os
mender-ffaf0abc/usr/share/mender/inventory/mender-inventory-provides
mender-ffaf0abc/usr/share/mender/inventory/mender-inventory-network
mender-ffaf0abc/usr/share/mender/inventory/mender-inventory-update-modules
mender-ffaf0abc/usr/share/mender/inventory/mender-inventory-intervals
mender-ffaf0abc/usr/share/mender/inventory/mender-inventory-hostinfo
mender-ffaf0abc/usr/share/mender/inventory/mender-inventory-rootfs-type
mender-ffaf0abc/usr/share/mender/inventory/mender-inventory-bootloader-integration
mender-ffaf0abc/usr/share/mender/identity/
mender-ffaf0abc/usr/share/mender/identity/mender-device-identity
mender-ffaf0abc/usr/share/doc/
mender-ffaf0abc/usr/share/doc/mender-auth/
mender-ffaf0abc/usr/share/doc/mender-auth/examples/
mender-ffaf0abc/usr/share/doc/mender-auth/examples/demo.crt
mender-ffaf0abc/etc/
mender-ffaf0abc/etc/mender/
mender-ffaf0abc/etc/mender/mender.conf
mender-ffaf0abc/var/
mender-ffaf0abc/var/lib/
mender-ffaf0abc/var/lib/mender/
mender-ffaf0abc/var/lib/mender/device_type
mender-ffaf0abc/lib/
mender-ffaf0abc/lib/systemd/
mender-ffaf0abc/lib/systemd/system/
mender-ffaf0abc/lib/systemd/system/mender-authd.service
mender-ffaf0abc/lib/systemd/system/mender-updated.service
```

### DIFF
--- a/cmake/dist-package.cmake
+++ b/cmake/dist-package.cmake
@@ -1,0 +1,52 @@
+# Create a tarball containing mender-update/mender-auth binaries and all files required to run mender
+# Example usage:
+#   $ cmake -B build -DMENDER_BUILD_DIST_PACKAGE=on -DMENDER_TENANT_TOKEN=<token> -DMENDER_SERVER_URL=<url> -DMENDER_DEVICE_TYPE=<type>
+#   $ cmake --build build
+# or
+#   $ cmake -B build -DMENDER_TENANT_TOKEN=<token> -DMENDER_SERVER_URL=<url> -DMENDER_DEVICE_TYPE=<type>
+#   $ cmake --build build --target dist-package
+
+include(${CMAKE_CURRENT_LIST_DIR}/version.cmake)
+
+set(MENDER_TENANT_TOKEN "Paste your Tenant token here" CACHE STRING "Tenant token baked into the dist-package mender.conf")
+set(MENDER_SERVER_URL "https://hosted.mender.io/" CACHE STRING "Server URL baked into the dist-package mender.conf")
+# Default device_type: <system>-<processor> lowercased. E.g. linux-x86_64, qnx-aarch64.
+string(TOLOWER "${CMAKE_SYSTEM_NAME}-${CMAKE_SYSTEM_PROCESSOR}" _MENDER_DEVICE_TYPE_DEFAULT)
+set(MENDER_DEVICE_TYPE "${_MENDER_DEVICE_TYPE_DEFAULT}" CACHE STRING "device_type baked into the dist-package /var/lib/mender/device_type")
+option(MENDER_BUILD_DIST_PACKAGE "Build the dist-package tarball as part of the default build" OFF)
+
+function(_setup_dist_package)
+    # Clear potential stale configured files
+    file(REMOVE_RECURSE ${CMAKE_BINARY_DIR}/dist-package)
+    configure_file(${CMAKE_CURRENT_LIST_DIR}/dist-package/mender.conf.in ${CMAKE_BINARY_DIR}/dist-package/mender.conf @ONLY)
+    configure_file(${CMAKE_CURRENT_LIST_DIR}/dist-package/device_type.in ${CMAKE_BINARY_DIR}/dist-package/device_type @ONLY)
+
+    set(STAGE_DIR ${CMAKE_BINARY_DIR}/mender-${MENDER_VERSION})
+
+    set(DIST_DEPS mender-update)
+    if(NOT MENDER_EMBED_MENDER_AUTH)
+        list(APPEND DIST_DEPS mender-auth)
+    endif()
+
+    # If MENDER_BUILD_DIST_PACKAGE is on, ALL is appended to _dp_target_args
+    # ALL means that the target is added to the default build target and always built
+    set(_dp_target_args dist-package)
+    if(MENDER_BUILD_DIST_PACKAGE)
+        list(APPEND _dp_target_args ALL)
+    endif()
+    add_custom_target(${_dp_target_args}
+        COMMAND ${CMAKE_COMMAND} -E rm -rf ${STAGE_DIR}
+        COMMAND ${CMAKE_COMMAND} -E env DESTDIR=${STAGE_DIR} ${CMAKE_COMMAND} --install ${CMAKE_BINARY_DIR} --prefix /usr
+        COMMAND ${CMAKE_COMMAND} -E make_directory ${STAGE_DIR}/etc/mender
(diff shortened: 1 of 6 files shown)


cmake/dist-package.cmake
```diff
@@ -0,0 +1,50 @@
...
+#   $ cmake --build build
+# or
+#   $ cmake -B build -DMENDER_TENANT_TOKEN=<token> -DMENDER_SERVER_URL=<url> -DMENDER_DEVICE_TYPE=<type>
+#   $ cmake --build build --target dist-package
+
+include(${CMAKE_CURRENT_LIST_DIR}/version.cmake)
+
+set(MENDER_TENANT_TOKEN "Paste your Tenant token here" CACHE STRING "Tenant token baked into the dist-package mender.conf")
+set(MENDER_SERVER_URL "https://hosted.mender.io/" CACHE STRING "Server URL baked into the dist-package mender.conf")
+set(MENDER_DEVICE_TYPE "" CACHE STRING "device_type baked into the dist-package /var/lib/mender/device_type")
```
danielskinstad: Note that I haven't set any default here. Should we?

danielskinstad:
I see that make install also added the dbus service file even though I compiled without dbus

cmake/dist-package.cmake
vpodzime: Yeah, I think that we should either create the file with something default and valid or not create the file at all.

vpodzime (commented):
I'd prefer a new target like `install-demo` (or perhaps `install-config`? 🤔) and utilizing it in the `dist-package` target, but I can do this refactoring as part of MEN-9650 if you don't want to spend time on it.

cmake/dist-package.cmake
michalkopczan: What do you mean by valid device_type/tenant token?
Does it make sense to build the tarball if these values are not provided? Maybe we should fail the target if they're not set?

michalkopczan:
LGTM, except the discussion about the default values.
I'll leave approval to Vratislav as this makes the most sense I think.

cmake/dist-package.cmake
danielskinstad: any suggestions for a default value?

danielskinstad:
> I'd prefer a new target like install-demo (or perhaps install-config? 🤔) and utilizing it in the dist-package target, but I can do this refactoring as part of MEN-9650 if you don't want to spend time on it.

What would this entail exactly?  I can do it, I just need to understand what we want/need. However, if you easily know how to do it based on this, then I would say it's easier/faster if you do it, especially since I am off tomorrow. Let me know what you think is best

cmake/dist-package.cmake
vpodzime: > Maybe we should fail the target if they're not set?

No, we want to be able to use the target with user filling in those values when/after installing it.

vpodzime: > any suggestions for a default value?

What about some combination of mender version, target OS and architecture?

vpodzime:
> What would this entail exactly? I can do it, I just need to understand what we want/need. However, if you easily know how to do it based on this, then I would say it's easier/faster if you do it, especially since I am off tomorrow. Let me know what you think is best

When making an APK package, I need all those extra bits the new `dist-package` installs into the DESTDIR, but I don't need the tarball. So basically I need this to be two steps -- install extra bits and create a tarball. But I can do this myself.

danielskinstad:
> So basically I need this to be two steps -- install extra bits and create a tarball. But I can do this myself.

Alright, how do you suggest we move forward? I set a default for device_type, we merge this and you refactor/split/create the targets you need?

vpodzime:
> Alright, how do you suggest we move forward? I set a default for device_type, we merge this and you refactor/split/create the targets you need?

Yes, let's do that. Btw, the commit message mentions a different target name ;)

cmake/dist-package.cmake
danielskinstad: Good idea. We can do e.g. `"${CMAKE_SYSTEM_NAME}-${CMAKE_SYSTEM_PROCESSOR}"`, it will look something like `linux-x86_64`

danielskinstad:
I removed `InventoryPollIntervalSeconds`, `InventoryPollIntervalSeconds` `UpdatePollIntervalSeconds` from the config,  as they are essentially just setting the defaults (except RetryPollIntervalSeconds, which defaults to 0). I originally got the config from examples/mender.conf.production -- do you think we should set it explicitly as I previously did, or should we just set tenant token and server url like I do in the latest force push?

vpodzime:
> just set tenant token and server url

I don't see why we should duplicate the defaults and risk getting out of sync with those. So I'd do as little as possible.

vpodzime (approved):
ACK

vpodzime:

No functional changes here, not running full pipeline.

PR #1962, opened 2026-05-22
QNX inventory scripts

vpodzime:


vpodzime:

No functional changes here, not running full pipeline.

support/mender-inventory-hostinfo
```diff
@@ -8,16 +8,63 @@ set -ue
...
+   # Processor2: 1095749809 Cortex-A76 1500MHz FPU
+   # Processor3: 1095749809 Cortex-A76 1500MHz FPU
+   # Processor4: 1095749809 Cortex-A76 1500MHz FPU
+   pidin info | awk '
+BEGIN {
+    system_info["n_cpus"] = 0
+}
+/^CPU:/ {
+    # Extract CPU and Release
+    split($1, a, ":"); system_info["cpu"] = a[2]
```
danielskinstad: I think it would make sense to align this with linux and call it `cpu_model` rather than `cpu`. We also use `mem_free_mb` on qnx and `mem_free_kB` on linux, but that's probably fine

vpodzime: > We also use `mem_free_mb` on qnx and `mem_free_kB` on linux, but that's probably fine

Yeah, I don't think that's a big deal.

danielskinstad (approved):
LGTM

PR #1970, opened 2026-06-03
ci: Add test build with non standard C++ library

elkoniu:
By default the build is being performed using standard C++ library. We would like to test the build using libc++, which is more strict comparing to the standard library.

Ticket: QA-1626

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -56,6 +56,27 @@ default:
     - k8s
   retry: !reference [.qa-common-default-retry, retry]
 
+# Test the build with libc++ (QA-1626)
+build-with-libc++:
+  stage: test
+  image: registry.gitlab.com/northern.tech/mender/mender-test-containers/base-mender-cpp:master
+  script:
+    - |
+      BUILD_DIR="${BUILD_DIR:-build-libcxx}"
+      rm -rf "$BUILD_DIR"
+      unset BOOST_ROOT BOOST_INCLUDEDIR BOOST_LIBRARYDIR Boost_DIR
+      cmake -S . -B "$BUILD_DIR" \
+        -DCMAKE_C_COMPILER=clang \
+        -DCMAKE_CXX_COMPILER=clang++ \
+        -DCMAKE_CXX_FLAGS="-stdlib=libc++" \
+        -DCMAKE_EXE_LINKER_FLAGS="-stdlib=libc++ -lc++abi" \
+        -DCMAKE_SHARED_LINKER_FLAGS="-stdlib=libc++ -lc++abi" \
+        -DSTD_FILESYSTEM_LIB_NAME="" \
+        -DMENDER_DOWNLOAD_BOOST=ON \
+        -DBoost_NO_SYSTEM_PATHS=ON \
+        -DBoost_NO_BOOST_CMAKE=ON
+      cmake --build "$BUILD_DIR" -j
+
 # Override rules
 renovate:
   needs: []
--- a/cmake/build_mode.cmake
+++ b/cmake/build_mode.cmake
@@ -4,13 +4,13 @@ if ("${CMAKE_BUILD_TYPE}" MATCHES "Rel")
   # The ABI is not guaranteed to be compatible between C++11 and C++17, so in release mode, let's
   # not take that risk, and compile everything under C++17 instead. We still use C++11 in debug mode
   # below.
-  add_compile_options(-std=c++17)
+  add_compile_options($<$<COMPILE_LANGUAGE:CXX>:-std=c++17>)
   add_compile_definitions(MENDER_CXX_STANDARD=17)
   # No need for it in release mode, we compile everything with the same options.
   set(PLATFORM_SPECIFIC_COMPILE_OPTIONS "")
 else()
   # In Debug mode use C++11 by default, so that we catch C++11 violations.
-  add_compile_options(-std=c++11)
+  add_compile_options($<$<COMPILE_LANGUAGE:CXX>:-std=c++11>)
   add_compile_definitions(MENDER_CXX_STANDARD=11)
   # Use this with target_compile_options for platform specific components that need it.
   set(PLATFORM_SPECIFIC_COMPILE_OPTIONS -std=c++17)


elkoniu:
This PR implements CI job containing `libc++` based build mentioned here: https://github.com/mendersoftware/mender/pull/1940#issuecomment-4393802251

.gitlab-ci.yml
```diff
@@ -52,6 +52,29 @@ default:
     - k8s
   retry: !reference [.qa-common-default-retry, retry]
 
+# Test the build with libc++ (QA-1626)
+build-with-libc++:
+  stage: test
+  image: registry.gitlab.com/northern.tech/mender/mender-test-containers/mender-base-ubuntu:master
+  before_script:
+    - apt update && apt install -yyq git pkg-config build-essential libdbus-1-dev libboost-all-dev libssl-dev cmake clang libc++-dev libc++abi-dev
```
vpodzime: Could we add the required bits into the *base-mender-cpp* image and use that instead?

vpodzime (commented):
Looks promising otherwise.

.gitlab-ci.yml
elkoniu: I can do it, but if this would be justified only for `mender` purposes?

That would include:
* git
* pkg-config
* build-essential
* libdbus-1-dev
* ibboost-all-dev
* libssl-dev cmake
* clang
* libc++-dev
* libc++abi-dev

To me it sounds like a candidate for some `mender-base-ubuntu-builder` container which at the end would be used for all `building` related tasks to keep `mender-base-ubuntu` lightweight for all other tasks.

vpodzime: We already have *base-mender-cpp* that is used in other jobs here, I was simply thinking about adding those extra bits there and using it in this new job too.

elkoniu: Sounds like a good plan, I will create a followup PR to update `base-mender-cpp` with new packages.

elkoniu: https://github.com/mendersoftware/mender-test-containers/pull/713

vpodzime: You can use `registry.gitlab.com/northern.tech/mender/mender-test-containers/base-mender-cpp:pr_713` as the container image here to test if it actually works. If it does wen merge that PR and change it to `base-mender-cpp:master` here.

vpodzime:
I see quite a few failures. Perhaps we need a newer version of Boost?

vpodzime:
> I see quite a few failures. Perhaps we need a newer version of Boost?

@lhoward is that the case? The build downloads `boost-1.83.0.tar.gz` and that seems to produce lots of warnings with this setup. Did you use a different version?

lhoward:
> > I see quite a few failures. Perhaps we need a newer version of Boost?
> 
> 
> 
> @lhoward is that the case? The build downloads `boost-1.83.0.tar.gz` and that seems to produce lots of warnings with this setup. Did you use a different version?

I'm not at my desk now but everything built fine today with the latest master-next.

elkoniu:
@lhoward what environment (distro + version) are you using to build this? In this PR I copied your exact recipe and have some build issues + warnings on Debian 13.

lhoward:
I'm building under Yocto wrynose with meta-mender. I can try on Ubuntu tomorrow if that helps?

elkoniu:
> I'm building under Yocto wrynose with meta-mender. I can try on Ubuntu tomorrow if that helps?

I was trying using Debian 13 container and Ubuntu 24 container with the content of this PR and build failed.

lhoward:
You need this patch as `clang` will error if `-std=c++...` is passed to the C compiler.

```diff
diff --git i/cmake/build_mode.cmake w/cmake/build_mode.cmake
index 927963b3..e5d71da0 100644
--- i/cmake/build_mode.cmake
+++ w/cmake/build_mode.cmake
@@ -4,13 +4,13 @@ if ("${CMAKE_BUILD_TYPE}" MATCHES "Rel")
   # The ABI is not guaranteed to be compatible between C++11 and C++17, so in release mode, let's
   # not take that risk, and compile everything under C++17 instead. We still use C++11 in debug mode
   # below.
-  add_compile_options(-std=c++17)
+  add_compile_options($<$<COMPILE_LANGUAGE:CXX>:-std=c++17>)
   add_compile_definitions(MENDER_CXX_STANDARD=17)
   # No need for it in release mode, we compile everything with the same options.
   set(PLATFORM_SPECIFIC_COMPILE_OPTIONS "")
 else()
   # In Debug mode use C++11 by default, so that we catch C++11 violations.
-  add_compile_options(-std=c++11)
+  add_compile_options($<$<COMPILE_LANGUAGE:CXX>:-std=c++11>)
   add_compile_definitions(MENDER_CXX_STANDARD=11)
   # Use this with target_compile_options for platform specific components that need it.
   set(PLATFORM_SPECIFIC_COMPILE_OPTIONS -std=c++17)
```

elkoniu:

cmake/build_mode.cmake
```diff
@@ -4,13 +4,13 @@ if ("${CMAKE_BUILD_TYPE}" MATCHES "Rel")
   # The ABI is not guaranteed to be compatible between C++11 and C++17, so in release mode, let's
   # not take that risk, and compile everything under C++17 instead. We still use C++11 in debug mode
   # below.
-  add_compile_options(-std=c++17)
+  add_compile_options($<$<COMPILE_LANGUAGE:CXX>:-std=c++17>)
```
vpodzime, line 7: This is not removing the flags, it's making sure they are only applied to C++ compilations.

elkoniu: I have rephrased the commit title, please let me knowif now it is ok.

elkoniu:

vpodzime (commented):
The first commit message is also misleading -- both of those libraries are standard C++ libraries. One is _libstdc++_ from gcc, the other one is _libc++_ from LLVM.

elkoniu:

vpodzime (approved):
Thanks.

lhoward:
Thank you and sorry for not alerting you to the other patch earlier.

vpodzime:
> Thank you and sorry for not alerting you to the other patch earlier.

No problem. Thanks for helping with this!

lhoward:
> > Thank you and sorry for not alerting you to the other patch earlier.
> 
> No problem. Thanks for helping with this!

Pleasure. For embedded, minimising even LGPL dependencies can be useful.

PR #1979, opened 2026-06-24
fix: Ensure corrent order of globals initalization

michalkopczan:
Fix a segault caused by global variables: common_global_options and common_description_append referencing another global: DefaultPaths. As the initialization order of globals is unspecified in the standard, it caused referencing uninitialized DefaultPaths on Ubuntu 26.04 when using -flto Link-Time Optimization flag.

Ticket: MEN-9878

### DIFF
--- a/src/client_shared/conf/conf_cli_help.cpp
+++ b/src/client_shared/conf/conf_cli_help.cpp
@@ -41,75 +41,86 @@ const CliOption help_option = {
 	.description = "Show help and exit",
 };
 
-const vector<CliOption> common_global_options = {
-	CliOption {
-		.long_option = "config",
-		.short_option = "c",
-		.description = "Configuration FILE path",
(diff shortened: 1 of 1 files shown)


michalkopczan:

vpodzime (approved):
"Fun" stuff! I'd hope it would be enough to just reorder things, but I guess even if it helped on Ubuntu 26.04 there would be no guarantee it would work everywhere, right?

michalkopczan:
> "Fun" stuff! I'd hope it would be enough to just reorder things, but I guess even if it helped on Ubuntu 26.04 there would be no guarantee it would work everywhere, right?

Right. The order is undefined by the standard explicitly. So if you use a different optimization levels or a different linker, the result may be different. In other words, it's non-deterministic.

michalkopczan:
* 5.1.x
* 5.0.x

PR #1982, opened 2026-06-26
feat: reject non-manifest artifacts on system devices

danielskinstad:
Changelog: On a System Device, mender-update now refuses a deployment whose artifact is not a mender-orchestrator-manifest.
Ticket: MEN-9655

### DIFF
--- a/src/mender-update/daemon/states.cpp
+++ b/src/mender-update/daemon/states.cpp
@@ -15,6 +15,7 @@
 #include <mender-update/daemon/states.hpp>
 
 #include <client_shared/conf.hpp>
+#include <common/device_tier.hpp>
 #include <common/events_io.hpp>
 #include <common/log.hpp>
 #include <common/path.hpp>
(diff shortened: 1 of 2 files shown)


danielskinstad:

rewanrashid-boop (commented):
From the code it looks good to me testing:
1. When tier is system and attempt to install regular artifact 
2. When tier is standard and attempt to install regular artifact
3. When tier is system and attempt to install manifest artifact

Maybe for coverage sake you could have also have a test for when tier is micro and installing a regular/manifest artifact, but i suppose it is the same 🤷. So long as pipeline fully green should be good to merge. Ping when green for approval 🚀 .

vpodzime (commented):
So there's a different tier if _mender-update_ runs as a component interface?

michalkopczan:
> So there's a different tier if _mender-update_ runs as a component interface?

No, there's a separate state machine when running in daemon mode and when running in standalone mode. Daniel did changes only to the daemon state machine. So for standalone, even though the tier is system, normal artifacts can be installed.

Nice, I was wondering if rejecting non-manifest updates in daemon, but allowing them in standalone will be a problem to implement, and it turns out that this was free :D

Daniel , correct me if I'm wrong here.

michalkopczan (approved):

PR #1988, opened 2026-07-02
fix: report download failure when resume give-up occurs mid-stream

TheYoctoJester:
### Summary

Fixes an indefinite hang in `mender-update` when a network outage during
artifact download lasts long enough for the HTTP download resumer to give up.
The daemon stays `active (running)` but wedged: it never reports the deployment
as `Failure` and never returns to the poll loop until it is restarted.

Reported on the Mender Hub:
https://hub.mender.io/t/bug-mender-update-hangs-after-download-resume-gives-up-never-reports-failure-never-returns-to-poll-loop-5-0-3/8299
(reproduced on 5.0.2–5.0.5; the affected code is identical on `master`).

### Root cause

The artifact body is consumed as a stream through `DownloadResumerAsyncReader`.
When the resumer exhausts its retry backoff it logs *"Giving up on resuming the
download"* and delivers the `DownloadResumerError` via
`DownloadResumerClient::CallUserHandler`, which routes it to the **user body
handler**. But the update state machine's body handler deliberately ignores
errors (`src/mender-update/daemon/states.cpp`), expecting them to surface
through the reader instead. The reader's pending read (`last_read_.handler`)
is never completed, so the artifact read blocks forever — the state machine
never posts `Failure`, and (because the daemon's signal handler keeps the
`io_context` busy) the recursive event-loop read never returns.

The existing tests only exercised give-up with the buffered `SetBodyWriter`
model, where the body handler *does* receive the error — so the gap in the
streaming (reader) path went uncovered.

### Fix

In `CallUserHandler`, on any terminal error also complete the body reader's
pending read with the error, in addition to invoking the user body handler.
The buffered `SetBodyWriter` path is unaffected: there the pending read belongs
to an internal `AsyncCopy` that only logs the error, and the user body handler
still delivers the terminal status.

### Testing

- **New unit test** `DownloadResumerTest.ServerDownWhileReadingBodyReportsError`
  streams a download through `MakeBodyAsyncReader`, takes the server down until
  the backoff is exhausted, and asserts the reader receives a
  `DownloadResumerError`. It times out (hangs) without the fix and passes with
  it; the full `http_resumer` suite stays green (37/37).
- **End-to-end** on a `qemux86-64` image (mender 5.0.3) against Hosted Mender:
  cutting the network mid-download now produces
  `Giving up on resuming ... → DownloadCancel → DownloadError scripts →
  ArtifactCleanup → Pushing deployment status: failure`, and the client returns
  to the poll loop instead of wedging.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/src/common/http/http_resumer.cpp
+++ b/src/common/http/http_resumer.cpp
@@ -472,6 +472,23 @@ error::Error DownloadResumerClient::ScheduleNextResumeRequest() {
 void DownloadResumerClient::CallUserHandler(http::ExpectedIncomingResponsePtr exp_resp) {
 	if (!exp_resp) {
 		DoCancel();
+
+		// If a read is pending on the body reader, complete it with the error as well. When the
+		// body is consumed through a DownloadResumerAsyncReader (streaming download, as the
+		// update state machine does), a terminal error such as giving up on resuming after the
+		// retry backoff is exhausted only reaches the user through the user body handler below.
+		// But that handler cannot recover a streaming read, and the update client intentionally
+		// ignores errors there, relying on them surfacing through the reader instead. Without
+		// completing the pending read, that consumer blocks forever and the daemon wedges
+		// without ever reporting a deployment failure. Delivering the error here does not
+		// disturb the buffered SetBodyWriter case, where the pending read belongs to an internal
+		// AsyncCopy that merely logs the error, and the user body handler below still runs.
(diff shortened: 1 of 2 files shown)


vpodzime:
Slightly reworked in https://github.com/mendersoftware/mender/pull/1992.

PR #1995, opened 2026-07-22
fix(auth): Fix segfault when dbus becomes unavailable and available a…

michalkopczan:
…gain

Ticket: MEN-9246

### DIFF
--- a/src/api/auth/auth.cpp
+++ b/src/api/auth/auth.cpp
@@ -55,32 +55,49 @@ void Authenticator::ExpireToken() {
 	}
 }
 
-error::Error Authenticator::WithToken(AuthenticatedAction action) {
-	pending_actions_.push_back(action);
-
+void Authenticator::WithToken(AuthenticatedAction action) {
 	auto err = StartWatchingTokenSignal();
 	if (err != error::NoError) {
-		// Should never fail. We rely on the signal heavily so let's fail
-		// hard if it does.
-		return err;
+		// StartWatchingTokenSignal can fail while D-Bus is unavailable (e.g.
+		// dbus.socket is down). Deliver the failure to THIS action only, via the
+		// event loop, and leave pending_actions_ and token_fetch_in_progress_
+		// untouched: any actions already queued are waiting on an unrelated, in-flight
+		// token fetch and must not be failed with this error. They are delivered when
+		// that fetch resolves, or failed by whichever token-fetch timeout is pending
+		// if it never does resolve.
+		//
+		// This branch is currently never reached while other actions are queued: a
+		// non-empty queue means a fetch is in progress, and the serialized state
+		// machine issues no further WithToken() until that fetch's callback drains the
+		// queue (no WithToken will take place because we are still in the same state, so no new
+		// requests are made). Delivering only this action is therefore defensive, but it keeps the
+		// behavior correct should that ever change.
+		ExpectedAuthData ex_auth_data = expected::unexpected(err);
+		loop_.Post([action, ex_auth_data]() { action(ex_auth_data); });
+		return;
 	}
 
 	if (token_fetch_in_progress_) {
 		// Already waiting for a new token.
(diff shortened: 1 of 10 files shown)


michalkopczan:

michalkopczan:
The 14 lines of missing coverage reported by codecov is superfluous. It assumes that the indentation change is new code which should be covered - which is not the case.

michalkopczan:

src/api/auth/auth.cpp
```diff
@@ -55,32 +55,49 @@ void Authenticator::ExpireToken() {
...
+		// if it never does resolve.
+		//
+		// This branch is currently never reached while other actions are queued: a
+		// non-empty queue means a fetch is in progress, and the serialized state
+		// machine issues no further WithToken() until that fetch's callback drains the
+		// queue (no WithToken will take place because we are still in the same state, so no new
+		// requests are made). Delivering only this action is therefore defensive, but it keeps the
+		// behavior correct should that ever change.
+		ExpectedAuthData ex_auth_data = expected::unexpected(err);
+		loop_.Post([action, ex_auth_data]() { action(ex_auth_data); });
```
vpodzime, line 76: Why not call the `action` here synchronously? I understand we want to guarantee that the `action` will always be called exactly once instead of "maybe not an all in case a synchronous error appears", but this conversion of a synchronous error to an asynchronous one is worth an extra explanation, IMHO.

vpodzime (commented):
Looks OK to me, but the asynchronous handling of some synchronous error is quite a significant change. To me that means two things, at least:
- we should double-check and test other cases affected this, esp. some synchronous errors in `http.AsyncCall()`, and
- it should be properly explained an "justified" in the commit message.

src/api/auth/auth.cpp
michalkopczan: I am looking at it more from the perspective of unifying the error handling. Up until now we had some errors handle synchronously (StartWatchingTokenSignal), while we had others handled asynchronously (GetJwtToken). 

What actually caused the need to handle one synchronously and another one asynchronously? There was no reason for that, and it could (and did, see this task and also https://github.com/mendersoftware/mender/pull/1880/changes/6aa4c776107ae5b038ced51ecdeee32a8f6b9ee5 from MEN-9144) introduce confusion and errors.

Handling all async is cleaner and less prone to error. We could call it here synchronously, but I didn't want to introduce two different timings (after AsyncCall returns and immediately), which could potentially cause problems. As we already call is asynchronously sometimes, I figured it's cleaner to call it asynchronously always.

Is this explanation clear? If yes, I will include it (or something like it) in the commit message.


Re `http.AsyncCall`: this change does not influence it at all, it's just a naming problem: The change is in `class HTTPClient : public mender::api::Client`, and all the places that use it are already checked. Other http client is `class mender::common::http::Client :
	virtual public ClientInterface,
	public events::EventLoopObject,
	virtual public io::Canceller`
And is unrelated. No changes were done there - AsyncCall there still returns an error. It would be great to rework both of them of course, but as there was no need for it now, I didn't.

vpodzime: Makes total sense to me (as much as something can make sense while being on top of insane stuff). Thanks for the extra explanation! Please make sure to include it somehow in the commit message.

michalkopczan: Done, please review

vpodzime (approved):
Looks good to me now, thanks! I'd just mention the name of the class in the commit message too. When I read `AsyncCall()` it makes me think about the `http::Client` class.

michalkopczan:
> Looks good to me now, thanks! I'd just mention the name of the class in the commit message too. When I read `AsyncCall()` it makes me think about the `http::Client` class.

Good point, done.

michalkopczan:

michalkopczan:

michalkopczan:
* 5.1.x

PR #2002, opened 2026-07-29
fix(update-modules): fail the install when the backup fails

LudvigAnderson:
`ret=$?` was read inside `if ! <backup>; then`, where `$?` holds the status of the negated pipeline and is therefore always 0. Both modules exited 0 before applying the payload, so a failed install was reported to the server as successful: the destination was left untouched while the new software version was recorded for it. Capture the status from the command itself instead.

The directory module is the more exposed of the two, since it backs up the whole destination directory, so any unrelated file growing there can exhaust the data partition. Reported from the field with a multi-GB core dump.

Add a pytest suite for the shipped update module scripts, alongside the existing one for the artifact generators, covering both a failing backup and a successful install for each module. Ignore the Python test caches while here, since the new suite is the second pytest directory in the tree.

Changelog: Fixed the directory and single-file Update Modules reporting a deployment as successful when the destination could not be backed up, leaving the destination unchanged.
Ticket: MEN-10021

**Disclosure: heavy reliance on Claude here as I rarely touch this repo, so let me know if this completely off. I did review the code and did make sure the tests failed before adding the code and succeeded after, though.**

### DIFF
--- a/support/modules/directory
+++ b/support/modules/directory
@@ -26,9 +26,10 @@ case "$STATE" in
         test "$dest_dir" = "/" && \
             echo "Error: destination dir is '/', install not supported." && exit 1
         mkdir -p $dest_dir
-        if ! tar -cf ${prev_files_tar} -C ${dest_dir} .
+        ret=0
+        tar -cf ${prev_files_tar} -C ${dest_dir} . || ret=$?
+        if [ "$ret" -ne 0 ]
         then
-            ret=$?
             # Make sure there is no half-backup lying around.
             rm -f ${prev_files_tar}
             exit $ret
--- a/support/modules/tests/test_update_modules.py
+++ b/support/modules/tests/test_update_modules.py
@@ -1,0 +1,174 @@
+# Copyright 2026 Northern.tech AS
+#
+#    Licensed under the Apache License, Version 2.0 (the "License");
+#    you may not use this file except in compliance with the License.
+#    You may obtain a copy of the License at
+#
+#        http://www.apache.org/licenses/LICENSE-2.0
+#
+#    Unless required by applicable law or agreed to in writing, software
+#    distributed under the License is distributed on an "AS IS" BASIS,
+#    WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+#    See the License for the specific language governing permissions and
+#    limitations under the License.
+
+import os
+import pathlib
+import shutil
+import subprocess
+
+import pytest
+
+MODULES_PATH = pathlib.Path(__file__).parent.parent.absolute()
+
+
+def run_module(module, state, files_dir):
+    return subprocess.run(
+        [str(MODULES_PATH / module), state, str(files_dir)],
+        capture_output=True,
+        text=True,
+        check=False,
+    )
+
+
+def snapshot(directory):
+    return {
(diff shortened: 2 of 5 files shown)


LudvigAnderson:

support/modules/directory
```diff
@@ -26,9 +26,9 @@ case "$STATE" in
         test "$dest_dir" = "/" && \
             echo "Error: destination dir is '/', install not supported." && exit 1
         mkdir -p $dest_dir
-        if ! tar -cf ${prev_files_tar} -C ${dest_dir} .
+        tar -cf ${prev_files_tar} -C ${dest_dir} . || ret=$?
+        if [ -n "${ret:-}" ]
```
vpodzime: I'd do `ret=0` before the command and then use `$ret != 0` (or `ret -ne 0`) in the condition as I find that more readable than `-n "${ret:-}"`.

support/modules/tests/test_update_modules.py
```diff
@@ -0,0 +1,174 @@
...
+    fake = bin_dir / name
+    fake.write_text(f'#!/bin/sh\n{body}\nexec {real} "$@"\n')
+    fake.chmod(0o755)
+
+
+def payload_skeleton(tmp_path, dest_dir):
+    files_dir = tmp_path / "payload"
+    (files_dir / "files").mkdir(parents=True)
+    (files_dir / "tmp").mkdir()
+    (files_dir / "files" / "dest_dir").write_text(f"{dest_dir}\n")
```
vpodzime, line 68: Which version of Python do these path constructions and method calls require? I'd personally go with good old `os.path.join()` and `os.makedirs()`, but maybe I'm just old-school. It would be a shame if we couldn't run this in some common environments just because ~if~ of these fancy lines.

vpodzime (commented):
Looks good to me otherwise.

support/modules/directory
LudvigAnderson: Totally agreed, I'm just too crappy at bash to notice that myself :smile:

support/modules/tests/test_update_modules.py
LudvigAnderson: I think that should be no problem :ok_hand:, `pathlib` has been a part of the standard library and f-strings have been around for a long time, Claude says since 3.4 and 3.6 respectively. Our [requirements.txt](https://github.com/mendersoftware/mender/blob/master/support/modules-artifact-gen/tests/requirements.txt) pins `pytest==9.0.3` which states that it requires `Python>=3.10`: https://pypi.org/project/pytest/9.0.3/

So, whatever version is running `pytest==9.0.3` is sure to support this code:)

vpodzime (approved):
Thanks!

vpodzime:
> ci/mender-qaExpected — Waiting for status to be reported

No need to re-run the pipeline after the small change (with no change in behavior).

LudvigAnderson:
> > ci/mender-qaExpected — Waiting for status to be reported
> 
> No need to re-run the pipeline after the small change (with no change in behavior).

@vpodzime Makes sense, but sorry I'm not familiar with the mender client pipeline, is it possible to skip it? :sweat_smile:

vpodzime:
* 5.1.x
* 5.0.x

PR #2004, opened 2026-08-03
ci: configure codecov coverage thresholds

p-targowicz:
Codecov defaults let any coverage decrease pass. This sets an explicit policy: fail the project status on a drop larger than 1%, and report patch coverage at an 80% target without blocking merges on it.

Statuses stay advisory — this does not change branch protection or required checks.

**Merge order: independent.** No dependency on the container or template PRs.

Ticket: [QA-1573](https://northerntech.atlassian.net/browse/QA-1573)

[QA-1573]: https://northerntech.atlassian.net/browse/QA-1573?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/codecov.yml
+++ b/codecov.yml
@@ -11,5 +11,16 @@ codecov:
   # coverage requirements for merging the PR are satisfied) in the PR immediately.
     after_n_builds: 1
 
+coverage:
+  status:
+    project:
+      default:
+        target: auto
+        threshold: 1%
+    patch:
+      default:
+        target: 80%
+        informational: true
+
 ignore:
   - "vendor/"


p-targowicz:
Closing: superseded by the org-wide Codecov Global YAML.

Codecov applies the Global YAML only to repositories that have **no** repo-level
`codecov.yml`. Adding one here would opt this repository out of the org config
entirely, losing `precision`, `range`, `round`, the `comment` layout and the
inherited `project` threshold — and replacing it with a near-identical local
copy. Net effect would be a regression.

The coverage policy now lives in one place:
https://app.codecov.io/account/gh/mendersoftware/yaml

Ticket: QA-1573

PR #2007, opened 2026-08-03
fix(update-modules): fail the install when the backup fails

LudvigAnderson:
`ret=$?` was read inside `if ! <backup>; then`, where `$?` holds the status of the negated pipeline and is therefore always 0. Both modules exited 0 before applying the payload, so a failed install was reported to the server as successful: the destination was left untouched while the new software version was recorded for it. Capture the status from the command itself instead.

The directory module is the more exposed of the two, since it backs up the whole destination directory, so any unrelated file growing there can exhaust the data partition. Reported from the field with a multi-GB core dump.

Add a pytest suite for the shipped update module scripts, alongside the existing one for the artifact generators, covering both a failing backup and a successful install for each module. Ignore the Python test caches while here, since the new suite is the second pytest directory in the tree.

Changelog: Fixed the directory and single-file Update Modules reporting a deployment as successful when the destination could not be backed up, leaving the destination unchanged. Ticket: MEN-10021

(cherry picked from commit 8d6f02c81e4c320fd5e92de6b458d3b8378095a9)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -49,6 +49,10 @@ config.h
 # Test reports
 /reports
 
+# Python test caches
+__pycache__/
+.pytest_cache/
+
 # Misc
 /.clang-format
 build/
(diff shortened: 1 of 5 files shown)


LudvigAnderson:

vpodzime:
> [ci/mender-qa](https://gitlab.com/Northern.tech/Mender/mender-qa/-/pipelines/2726985615)
ci/mender-qa — mender-qa pipeline failed

Unrelated.

PR #2008, opened 2026-08-04
fix(http): time out stalled transactions without killing switched protocols

nickanderson:
A server that accepts a request and then never answers left the response read
outstanding indefinitely, wedging the client silently — alive, no crash, no log
output, recoverable only by restart. Reproduced in a VM on both 5.0.5 and 5.1.0.

The gap was `ReadHeader()`: `AsyncReadNextBodyPart` already had a 5 minute timer,
but nothing bounded the wait for response *headers*.

This arms Beast's stream timeout before each operation of a transaction and
**disarms it at the 101 Switching Protocols handover**. That handover is what the
previous attempt (`d8ddf6da` / `4aa01ee8`) got wrong: it left the timeout armed,
and Beast's `timeout_handler` calls `close()` on expiry regardless of who owns the
socket by then, which killed mender-connect's forwarded WebSocket after five
minutes and got the fix reverted under MEN-9433.

Connect is intentionally not covered — `asio::async_connect` runs on
`lowest_layer()`, the raw socket, which bypasses the stream timeout entirely.

Timeout is `ClientConfig::stream_timeout_seconds`, default 300s. No client
endpoint long-polls, and artifact download is a per-chunk body read (unchanged).

### Tests

Both are red/green verified — each fails when its own half of the fix is removed:

| test | without the fix |
|---|---|
| `HttpTest.TestStalledResponseTimesOut` | hangs to the event-loop watchdog (5006 ms) |
| `HttpForwarderTests.ProtocolSwitchSurvivesIdlePastStreamTimeout` | fails at 1219 ms — MEN-9433 exactly |

The forwarder already had five `ProtocolSwitch*` tests, but all exercise immediate
data flow; none left the connection idle, which is why MEN-9433 escaped.

Full suites pass: `http_test` 47/47, `http_forwarder_test` 13/13.

End-to-end on 5.0.5 with a stalled `deployments/next` and the reporting customer's
state scripts: unpatched logs nothing for 420 s, patched logs
`Error while polling for deployment: The operation timed out` at exactly t+300 s,
runs the SyncError/IdleEnter scripts, and resumes polling.

Ticket: ME-731

PR #2009, opened 2026-08-04
fix(http): time out stalled transactions without killing switched protocols

nickanderson:
Backport of #2008 to 5.0.x. Both 5.0.5 and 5.1.0 were verified affected, so upgrading is
not a workaround.

A server that accepts a request and then never answers left the response read outstanding
indefinitely, wedging the client silently — alive, no crash, no log output, recoverable only
by restart. The gap was `ReadHeader()`: `AsyncReadNextBodyPart` already had a timer, but
nothing bounded the wait for response *headers*.

This arms Beast's stream timeout before each operation of a transaction and **disarms it at
the 101 Switching Protocols handover**. That handover is what the previous attempt
(`d8ddf6da` / `4aa01ee8`) got wrong: it left the timeout armed, and Beast's `timeout_handler`
calls `close()` on expiry regardless of who owns the socket by then, which killed
mender-connect's forwarded WebSocket after five minutes and got the fix reverted under
MEN-9433.

Connect is intentionally not covered — `asio::async_connect` runs on `lowest_layer()`, the
raw socket, which bypasses the stream timeout entirely.

### Why this touches an extra file versus master

`src/common/common.hpp` is included here to bring `common::def_value` over from master; 5.0.x
only has `def_bool`. It is needed for the new `ClientConfig::stream_timeout_seconds` field.

A plain default member initializer would have been smaller but is wrong on this branch:
`cmake/build_mode.cmake` compiles Debug with `-std=c++11` deliberately, and under C++11 a
default member initializer stops `ClientConfig` being an aggregate — which breaks
`http::ClientConfig client_config {"server.localhost.crt"}` in the HTTPS tests. That is the
same reason `def_bool` already exists.

### Tests

Both are red/green verified — each fails when its own half of the fix is removed:

| test | without the fix |
|---|---|
| `HttpTest.TestStalledResponseTimesOut` | hangs to the event-loop watchdog (5006 ms) |
| `HttpForwarderTests.ProtocolSwitchSurvivesIdlePastStreamTimeout` | fails at 1219 ms — MEN-9433 exactly |

The forwarder already had five `ProtocolSwitch*` tests, but all exercise immediate data flow;
none left the connection idle, which is why MEN-9433 escaped.

Built on 5.0.x in **Debug** (the C++11 path, so the aggregate trap above is exercised);
`http_test` and `http_forwarder_test` pass 58/58 via ctest.

End-to-end on 5.0.5 with a stalled `deployments/next` and the reporting customer's state
scripts: unpatched logs nothing for 420 s, patched logs
`Error while polling for deployment: The operation timed out` at exactly t+300 s, runs the
SyncError/IdleEnter scripts, and resumes polling.

Ticket: ME-731

### DIFF
--- a/src/common/common.hpp
+++ b/src/common/common.hpp
@@ -56,6 +56,23 @@ struct def_bool {
 	}
 };
 
+template <typename T, T default_value>
+struct def_value {
+	T value;
+
+	def_value() :
+		value(default_value) {
+	}
+
+	def_value(T init_value) :
+		value(init_value) {
+	}
+
+	operator T() const {
+		return value;
+	}
+};
+
 using StringPair = std::pair<string, string>;
 using ExpectedStringPair = expected::expected<StringPair, error::Error>;
 
--- a/src/common/http.hpp
+++ b/src/common/http.hpp
@@ -396,6 +396,10 @@ private:
 template <typename StreamType>
 class BodyAsyncReader;
 
+// Long enough not to disturb slow but healthy transfers, short enough that a device recovers
+// on its own.
+static constexpr int kDefaultStreamTimeoutSeconds = 300;
+
 // Master object that connections are made from. Configure TLS options on this object before making
 // connections.
 struct ClientConfig {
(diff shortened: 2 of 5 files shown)


nickanderson:

nickanderson:
The failures here are unrelated to the change. Merging through, Daniel said it was ok on the pr to master https://github.com/mendersoftware/mender/pull/2008#issuecomment-5356320707

PR #2013, opened 2026-08-19
fix(http): time out stalled transactions without killing switched protocols

nickanderson:
Backport of #2008 to 5.1.x, as requested in review. The 5.0.x backport is #2009.

A server that accepts a request and then never answers left the response read outstanding
indefinitely, wedging the client silently — alive, no crash, no log output, recoverable only
by restart. The gap was `ReadHeader()`: `AsyncReadNextBodyPart` already had a timer, but
nothing bounded the wait for response *headers*.

This arms Beast's stream timeout before each operation of a transaction and **disarms it at
the 101 Switching Protocols handover**. That handover is what the previous attempt
(`d8ddf6da` / `4aa01ee8`) got wrong: it left the timeout armed, and Beast's `timeout_handler`
calls `close()` on expiry regardless of who owns the socket by then, which killed
mender-connect's forwarded WebSocket after five minutes and got the fix reverted under
MEN-9433.

Connect is intentionally not covered — `asio::async_connect` runs on `lowest_layer()`, the
raw socket, which bypasses the stream timeout entirely.

### Relation to the other two PRs

All four touched files are identical between `master` and `5.1.x`, so this is a clean
`cherry-pick -x` of 51ab5fb6 with a byte-identical diff — no branch adaptation, unlike #2009
(5.0.x needed `common::def_value` ported over as well).

### Tests

Both new tests are red/green verified on master — each fails when its own half of the fix is
removed:

| test | without the fix |
|---|---|
| `HttpTest.TestStalledResponseTimesOut` | hangs to the event-loop watchdog (5006 ms) |
| `HttpForwarderTests.ProtocolSwitchSurvivesIdlePastStreamTimeout` | fails at 1219 ms — MEN-9433 exactly |

Built here on 5.1.x in **Debug** (the C++11 path) in the CI `base-mender-cpp-master` image;
`http_test` and `http_forwarder_test` pass via ctest.

5.1.0 was verified affected by the original hang, so upgrading off 5.0.x is not a workaround.

Ticket: ME-731

### DIFF
--- a/src/common/http.hpp
+++ b/src/common/http.hpp
@@ -412,6 +412,9 @@ template <typename StreamType>
 class BodyAsyncReader;
 
 
+// Long enough not to disturb slow but healthy transfers, short enough that a device recovers
+// on its own.
+static constexpr int kDefaultStreamTimeoutSeconds = 300;
 
 // Master object that connections are made from. Configure TLS options on this object before making
 // connections.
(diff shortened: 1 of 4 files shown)


nickanderson:

vpodzime (commented):
Nice!

nickanderson:
The failures here are unrelated to the change. Merging through, Daniel said it was ok on the pr to master https://github.com/mendersoftware/mender/pull/2008#issuecomment-5356320707

PR #2016, opened 2026-08-28
QNX release build unit test fixes

vpodzime:


vpodzime:

danielskinstad (approved):
LGTM

tests/src/mender-update/update_module/v3/update_module_test.cpp
```diff
@@ -1467,9 +1492,23 @@ struct AliasingAndMaliciousArtifactTestParams {
...
+	// also needs to skip itself via SetUp() below.
+	void SetUp() override {
+		UpdateModuleTests::SetUp();
+		MenderArtifactTest::SetUp();
+	}
+
 	static void SetUpTestSuite() {
+		// Not a real skip (GTEST_SKIP() here would not cascade to the individual TEST_P cases,
+		// see SetUp() above) -- just avoids wastefully running the mender-artifact script below.
+		if (!HasMenderArtifact()) {
```
michalkopczan, line 1508: We effectively have this check both here and in SetUp. Not needed. Although I understand that your intention is to run this check once at the beginning of the whole suite, right?

tests/src/mender-update/context_test.cpp
```diff
@@ -221,8 +225,13 @@ TEST_F(ContextTests, LoadProvidesClosedDB) {
 	err = db.Write("artifact-provides", common::ByteVectorFromString(input_provides_data_str));
 	ASSERT_EQ(err, error::NoError);
 
+#ifdef MENDER_USE_LMDB
```
michalkopczan, line 228: not really related to this tests, so treat is just as a general comment:

I see that both blob and lmdb variants inherit from KeyValueDatabase. If this had a Close() = 0 function, that this ifdef and dynamic casts would not be needed.

michalkopczan (approved):

tests/src/mender-update/update_module/v3/update_module_test.cpp
vpodzime: As it says:
>  just avoids wastefully running the mender-artifact script below.

Which is required for two reasons -- two avoid wasting CPU cycles and to make sure the whole test suite is not marked as failed even if all tests are skipped, leading to exit code 1, because a failure in `SetUpTestSuite()` would cause exactly that.

tests/src/mender-update/context_test.cpp
vpodzime: Yeah, I know. We discussed this before -- the whole KeyValueDatabase API should be refactored/re-thought so that we can use the parent (abstract) class and not worry about the particular implementation (in a child class). Let's finally create a ticket for that! https://northerntech.atlassian.net/browse/MEN-10123

vpodzime:
* 5.1.x

tests/src/mender-update/update_module/v3/update_module_test.cpp
michalkopczan: I meant the double check here and in SetUp.
But now I see that in fact SetUpTestSuite would cause an error in case mender-artifact would not be there. Clear.

tests/src/mender-update/context_test.cpp
michalkopczan: Nice!

PR #2030, opened 2026-09-21
Varlink server

danielskinstad:


src/common/json/platform/nlohmann/nlohmann_json.cpp
```diff
@@ -111,7 +111,47 @@ ExpectedJson Load(io::Reader &reader) {
 }
```
danielskinstad, line 111: With this we should be able to simplify a bunch of stuff like https://github.com/mendersoftware/mender/blob/9ba68f075c1e7ec63f6b2a9ce6d203b230d18edf/src/mender-update/daemon/context.cpp#L263

(i opted to not do a large refactor because this PR is already large enough)

src/common/json/platform/nlohmann/nlohmann_json.cpp
```diff
@@ -111,7 +111,47 @@ ExpectedJson Load(io::Reader &reader) {
...
+
+Json Json::Object(const unordered_map<string, string> &members) {
+	return Json(insensitive_json(members));
+}
+
+Json Json::Array() {
+	return Json(insensitive_json::array());
+}
+
+Json &Json::Append(const Json &value) {
```
vpodzime: This and the others below should do some type checking, right? You can't append an object to a string or add a child item under a string key into an array. Or am I missing something?

src/common/varlink/server.hpp
```diff
@@ -0,0 +1,158 @@
...
+#define MENDER_COMMON_VARLINK_SERVER_HPP
+
+#include <deque>
+#include <map>
+#include <memory>
+#include <set>
+#include <string>
+#include <unordered_map>
+
+#include <boost/asio.hpp>
```
vpodzime: This is a platform-specific `#include`, it should be guarded. And the same applies to all the other ASIO bits below. The header file should be platform-independent or at least "platform-aware".

src/common/varlink/varlink.hpp
```diff
@@ -0,0 +1,95 @@
...
+// those otherwise, for oneway calls, after the terminal frame, and once the
+// connection is gone.
+using Replier = std::function<bool(const Reply &)>;
+// Returns the terminal reply, or nullopt to reply later through the Replier
+// (deferred single reply, or a stream that ends when the handler emits a frame
+// without continues).
+using MethodHandler = std::function<optional<Reply>(const MethodCall &call, const Replier &emit)>;
+
+// Fan-out to every client holding a "more" call open: the varlink equivalent of
+// a D-Bus signal. The handler stores its Replier with Add() and returns nullopt;
```
vpodzime: > The handler stores its Replier with Add() and returns nullopt

This is a very confusing sentence, at least for me.

src/common/varlink/server.hpp
```diff
@@ -0,0 +1,158 @@
...
+
+class Server;
+
+// Frames a peer may leave undrained, or pipeline behind an unfinished call,
+// before we drop it.
+constexpr size_t kMaxQueuedFrames = 1024;
+
+// A single accepted varlink connection. Lives as long as the read/write loop is
+// active. Held by the Server via shared_ptr in a set; removes itself on EOF/error.
+class Connection : public std::enable_shared_from_this<Connection> {
```
vpodzime: Is this class needed in the header file? Shouldn't use of the API provided here solely rely on the `Server` class?

src/common/varlink/server.hpp
```diff
@@ -0,0 +1,158 @@
...
+	// Bounded like write_queue_: a peer that keeps pipelining calls behind one that
+	// never completes (e.g. an open-ended Subscribe) is misbehaving, not streaming.
+	std::deque<std::string> pending_messages_;
+	log::Logger logger_;
+};
+
+using ConnectionPtr = std::shared_ptr<Connection>;
+
+// A hand-rolled varlink server over a Boost.Asio unix stream socket. Messages are
+// JSON objects separated by single NUL bytes.
```
vpodzime: The Boost.Asio mention belongs to the .cpp file, to the particular implementation of this API/class.

src/common/varlink/server.hpp
```diff
@@ -0,0 +1,158 @@
...
+		const std::string &product,
+		const std::string &version,
+		const std::string &url);
+	// Publish a varlink interface's IDL description (returned by
+	// org.varlink.service.GetInterfaceDescription and listed by GetInfo).
+	void AddInterface(const std::string &name, const std::string &description);
+
+	// Binds socket_path with mode 0660 (parent dir 0750 if we have to create it).
+	// A stale socket file from a crashed run is replaced; one that another process
+	// still serves on is refused with TransportError.
```
vpodzime: > one that another process still serves on is refused with TransportError.

? I guess that error is returned here, right?

src/common/platform/varlink/asio/varlink.cpp
```diff
@@ -0,0 +1,572 @@
...
+//
+//    Unless required by applicable law or agreed to in writing, software
+//    distributed under the License is distributed on an "AS IS" BASIS,
+//    WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+//    See the License for the specific language governing permissions and
+//    limitations under the License.
+
+#include <common/varlink/server.hpp>
+
+#include <sys/stat.h>
```
vpodzime: There are definitely boost/asio headers missing here. We should not rely on headers being included in the server.hpp file, IMHO.

src/common/varlink/server.cpp
```diff
@@ -0,0 +1,572 @@
...
+// Serializes a Reply into a NUL-terminated varlink frame.
+static std::string SerializeFrame(const Reply &reply) {
+	json::Json frame = json::Json::Object();
+	if (!reply.error_name.empty()) {
+		frame.Set("error", reply.error_name);
+	}
+	frame.Set("parameters", reply.parameters);
+	// An error reply is always terminal; sd-varlink drops the connection on
+	// error+continues, so never emit that combination.
+	if (reply.continues && reply.error_name.empty()) {
```
vpodzime: This could use `IsTerminal()` from below, right?

src/common/varlink/server.cpp
```diff
@@ -0,0 +1,572 @@
...
+		frame.Set("continues", true);
+	}
+	return frame.Dump(-1) + kMessageSeparator;
+}
+
+// No more frames follow: an error (always terminal) or a reply without continues.
+static bool IsTerminal(const Reply &reply) {
```
vpodzime: `inline`

src/common/platform/varlink/asio/varlink.cpp
```diff
@@ -0,0 +1,572 @@
...
+Connection::Connection(Server &server) :
+	server_ {server},
+	socket_ {server.ctx_},
+	logger_ {"varlink"} {
+}
+
+void Connection::Close() {
+	cancelled_ = true;
+	boost::system::error_code ec;
+	socket_.close(ec);
```
vpodzime, line 119: I guess we should do something with `ec` here? At least log the potential error...?

src/common/platform/varlink/asio/varlink.cpp
```diff
@@ -0,0 +1,572 @@
...
+}
+
+void Connection::ReadFrame() {
+	auto self = shared_from_this();
+	asio::async_read_until(
+		socket_, buf_, kMessageSeparator, [self](const boost::system::error_code &ec, size_t) {
```
vpodzime, line 128: `self` is a lambda-captured shared pointer and the lambda calls this function recursively. Please check it unwinds properly without memory leaks.

src/common/platform/varlink/asio/varlink.cpp
```diff
@@ -0,0 +1,572 @@
...
+
+// Writes the front frame; called when the queue goes from empty to non-empty and
+// after each completed write. front() stays valid: push_back never invalidates
+// deque references and we only pop after this write completes.
+void Connection::DoWrite() {
+	auto self = shared_from_this();
+	asio::async_write(
+		socket_,
+		asio::buffer(write_queue_.front()),
+		[self](const boost::system::error_code &ec, size_t) {
```
vpodzime, line 226: Same story here with the lambda-captured shared pointer and recursion.

src/common/platform/varlink/asio/varlink.cpp
```diff
@@ -0,0 +1,572 @@
...
+	}
+
+	return true;
+}
+
+// Writes the front frame; called when the queue goes from empty to non-empty and
+// after each completed write. front() stays valid: push_back never invalidates
+// deque references and we only pop after this write completes.
+void Connection::DoWrite() {
+	auto self = shared_from_this();
```
vpodzime, line 222: I'd add `assert(!write_queue_.empty())`

src/common/varlink/server.cpp
```diff
@@ -0,0 +1,572 @@
...
+			}
+		});
+}
+
+void Connection::Dispatch(const std::string &message) {
+	auto exp_json = json::Load(message);
+	if (!exp_json) {
+		// Not a varlink frame at all; like sd-varlink, hang up rather than guess a
+		// reply (we cannot even tell whether the call was oneway).
+		logger_.Debug("Dropping varlink peer that sent a malformed frame");
```
vpodzime: I'd qualify this as `Warning`, it shouldn't happen and it always indicates something wrong happening in the system, right?

vpodzime (commented):
Looks good to me otherwise. Many of the comments stem from a slightly wrong organization of the code and a missing split between platform-independent and platform (ASIO) -specific code. You can look at the DBus implementation for how it should be done (I'm not saying I a big fan of that code organization, but that's what we do...)

src/common/json/platform/nlohmann/nlohmann_json.cpp
```diff
@@ -111,7 +111,47 @@ ExpectedJson Load(io::Reader &reader) {
...
+	this->n_json.push_back(value.n_json);
+	return *this;
+}
+
+Json &Json::Set(const string &key, const string &value) {
+	this->n_json[key] = value;
```
michalkopczan: what happens if there' no key? [] will throw? we should guard against that, as we will not handle this exception anywhere

michalkopczan (commented):
Can you explain the general idea behing Server/Client?
I see that the server is listening for connections, so I understand this is the part that listens for something from outside, e.g. mender-auth listening for GetJwtToken requests.

But then the Connection class needs the Server class instance to work. How am I supposed to connect from mender-update (client) to mender-auth(server)? Do both need the Server? If they both need the Server, is it actually a Server?

danielskinstad:
> Can you explain the general idea behing Server/Client? I see that the server is listening for connections, so I understand this is the part that listens for something from outside, e.g. mender-auth listening for GetJwtToken requests.
> 
> But then the Connection class needs the Server class instance to work. How am I supposed to connect from mender-update (client) to mender-auth(server)? Do both need the Server? If they both need the Server, is it actually a Server?

The server creates a `Connection` object for each client it accepts, and uses it to read a client's calls and write replies back.

The idea is that mender-update and mender-auth starts its own varlink server, i.e. its own service on its own socket, and registers the interfaces it exposes. mender-update would have something like `io.mender.Update1` for the update pausing/device-side API, and for mender-auth something like `io.mender.Authentication1` once we get rid of dbus.

For authentication we'd need something like `varlink::Client` in mender-update that connects to the server in mender-auth. It's similar to what we do for dbus

src/common/varlink/server.cpp
danielskinstad: Make sense

src/common/json/platform/nlohmann/nlohmann_json.cpp
danielskinstad: https://json.nlohmann.me/api/basic_json/operator%5B%5D/#notes
`If key is not found in the object, then it is silently added to the object and filled with a null value to make key a valid reference. In case the value was null before, it is converted to an object.`

src/common/json/platform/nlohmann/nlohmann_json.cpp
danielskinstad: > This and the others below should do some type checking, right? 

Good point

src/common/varlink/server.hpp
danielskinstad: I can move it and keep a forward declaration (the server needs to track the connections so it can close them)

src/common/varlink/server.hpp
danielskinstad: You're right

src/common/varlink/server.cpp
danielskinstad: Correct

src/common/platform/varlink/asio/varlink.cpp
danielskinstad: It's fine AFAICT

src/common/platform/varlink/asio/varlink.cpp
danielskinstad: It's fine AFAICT

vpodzime:
> For authentication we'd need something like `varlink::Client` in mender-update that connects to the server in mender-auth. It's similar to what we do for dbus

Which will basically be the code that is already included here in the tests, right?

src/common/json/platform/nlohmann/nlohmann_json.cpp
```diff
@@ -111,7 +112,53 @@ ExpectedJson Load(io::Reader &reader) {
...
+	return Json(insensitive_json(members));
+}
+
+Json Json::Array() {
+	return Json(insensitive_json::array());
+}
+
+Json &Json::Append(const Json &value) {
+	assert(this->n_json.is_null() || this->n_json.is_array());
+	this->n_json.push_back(value.n_json);
```
vpodzime, line 131: What happens in release builds if this is called on a bad type? Will there be an unhandled exception killing the process?

vpodzime: We might need to use `ExpectedJson` and something like `and_then()` or `MaybeSet()` instead of `assert()`.

src/common/platform/varlink.hpp
```diff
@@ -0,0 +1,193 @@
...
+
+Reply ErrorReply(const std::string &error_name, const json::KeyValueMap &parameters = {});
+
+// No more frames follow: an error (always terminal) or a reply without continues.
+inline bool IsTerminal(const Reply &reply) {
+	return !reply.error_name.empty() || !reply.continues;
+}
+
+// Serializes a Reply into a NUL-terminated varlink frame.
+std::string SerializeFrame(const Reply &reply);
```
vpodzime: Is this really needed here? Doesn't look like something code using this API would need to call. I guess it's because platform-specific code needs this while it's actually a platform-independent bit, right? I'd then move this declaration into the platform-specific code instead of having it here. But not a biggie.

src/common/platform/varlink/asio/varlink.cpp
```diff
@@ -0,0 +1,467 @@
...
+//
+//    Unless required by applicable law or agreed to in writing, software
+//    distributed under the License is distributed on an "AS IS" BASIS,
+//    WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+//    See the License for the specific language governing permissions and
+//    limitations under the License.
+
+#include <common/platform/varlink.hpp>
+
+#include <sys/stat.h>
```
vpodzime: Just curious -- what do we need this one for? (perhaps worth adding a comment)

vpodzime (commented):
Very nice, thanks!!! Just a couple small things.

src/common/json/platform/nlohmann/nlohmann_json.cpp
michalkopczan: I agree, assert is not the way to go here.
Look at the already-present Get example:
`ExpectedJson Json::Get(const char *child_key) const {
	if (!this->n_json.is_object()) {
		auto err = MakeError(
			JsonErrorCode::TypeError, "Invalid JSON type to get '" + string(child_key) + "' from");
	}`

We should do similar error handling here.

src/common/json.hpp
```diff
@@ -102,6 +103,32 @@ class Json {
 
 	Json() = default;
 
+	// Builds an object or array incrementally, for output that mixes strings,
+	// numbers and bools (Dump() of a KeyValueMap only takes strings). Use
+	// Dump(-1) for the compact form. Keys dump sorted, not in Set() order.
+	// Append() requires an array, Set() an object; a default-constructed Json
+	// becomes one on first use. Anything else is a programming error (assert).
```
michalkopczan: "a default-constructed Json
	// becomes one on first use. "

A default-constructed Json becomes what on first use? Either an Object or an Array?

src/common/platform/varlink/asio/varlink.cpp
danielskinstad: `umask()`

src/common/platform/varlink.hpp
danielskinstad: Good point. It's not

src/common/platform/varlink.hpp
```diff
@@ -0,0 +1,193 @@
...
+struct Reply {
+	json::Json parameters {json::Json::Object()};
+	bool continues {false};
+	std::string error_name; // non-empty => error reply
+};
+
+Reply ErrorReply(const std::string &error_name, const json::KeyValueMap &parameters = {});
+
+// No more frames follow: an error (always terminal) or a reply without continues.
+inline bool IsTerminal(const Reply &reply) {
```
michalkopczan: Why not make it a member of Reply struct?

src/common/platform/varlink.hpp
```diff
@@ -0,0 +1,193 @@
...
+	return !reply.error_name.empty() || !reply.continues;
+}
+
+// Serializes a Reply into a NUL-terminated varlink frame.
```
michalkopczan: Same here, why not a member?

src/common/json/platform/nlohmann/nlohmann_json.cpp
danielskinstad: Alright, I added something similar to what we do in `Json::Get()`. I don't actually check the return value for `Set()`, but AFAICT that is unnecessary. `Set()` only fails if the json we call it on isn't an object, and the value we set shouldn't matter. We only call it on json objects, so it can't fail

src/common/platform/varlink/asio/varlink.cpp
```diff
@@ -0,0 +1,482 @@
...
+// A single accepted varlink connection. Lives as long as the read/write loop is
+// active. Held by the Server via shared_ptr in a set; removes itself on EOF/error.
+class Connection : public std::enable_shared_from_this<Connection> {
+public:
+	Connection(Server &server);
+
+	void Close();
+
+private:
+	friend class Server;
```
michalkopczan: Why do we need a friend here? Friends are generally something to be avoided, they point at a wrong interface definition.

I see that the Server uses `socket_` and  `ReadFrame` from the Connection class. Consider making ReadFrame public + creating the socket in Server::AsyncAccept and then moving the ownership of it to Connection. This would remove the need for the friend declaration and thus make ownership clearer.

src/common/platform/varlink.hpp
danielskinstad: I've made it a static function and moved it out of the header based on previous feedback

src/common/platform/varlink/asio/varlink.cpp
danielskinstad: > I see that the Server uses socket_ and ReadFrame from the Connection class. Consider making ReadFrame public + creating the socket in Server::AsyncAccept and then moving the ownership of it to Connection. This would remove the need for the friend declaration and thus make ownership clearer.

:+1:

vpodzime (approved):
Looks good to me.

michalkopczan (approved):
LGTM

